PR #47, opened 2024-03-21
chore: do not report an error when reducing resources receives a NotFound error

pmalek:
This PR will reduce the intermittent logs from the operator by not report an error when the resource to be reduced (removed) got deleted before the deletion happens in `reduce` package.


programmer04 (approved):

pmalek:
Ah, the CI pipeline that doesn't fail randomly is indeed something I missed <emotional>

PR #84, opened 2024-04-15
chore(refactor): migrate project from Kubebuilder v3 to v4 layout

programmer04:

Inspired by 

- https://github.com/Kong/gateway-operator-archive/pull/1455

followed the [official manual](https://book.kubebuilder.io/migration/manually_migration_guide_gov3_to_gov4) too. 

This PR does not modify the API surface of KGO since it's currently used as a dependency in the enterprise built. Such efforts can be addressed in a separate PR. 

**Which issue this PR fixes**

Part of https://github.com/Kong/gateway-operator/issues/130

The layout proposed by `kubebuilder` to move `controllers` -> `internal/controller` cannot be implemented because they are pulled by KGO EE, so only renaming to `controller` was performed. 

Known limitation, calling e.g.

```sh
 kubebuilder create api --group batch --version v1 --kind CronJob
```

will put a newly created controller in `internal/controller`. Maybe all controllers that are not pulled explicitly should be put there... It's something to consider and write down as a separate issue. 

The usage of `sed` in `generate.clientsets` is not great but I tested it with bot BSD and GNU `sed` so I think we can live with it.


rainest (approved):
Curious if we could make kubebuilder happier by symlinking its desired names to the actual (public) locations of the packages, but IDK if that causes havoc or how much benefit it'd really provide--I assume we'd still need to move newly-added things, so probably not much.

pmalek:
This will be a breaking change for CG dataplane-controller and we will need to coordinate its update when its KGO dependency is bumped to (probably) 1.3 which will include this change.

https://github.com/Kong/gateway-operator/issues/130

PR #85, opened 2024-04-15
chore: parametrize golangci-lint config path, add renovate annotations, bump KIC

programmer04:

This is part of the improvements done during bumping dependencies and project layout for KGO OSS and EE. This PR contains
- parametrized config path for `golangci-lint` (`Makefile` is reused in KGO EE)
- annotations for renovate mentioned in the original PR https://github.com/Kong/gateway-operator/pull/17#issue-2189063011 as nice to have
- since it's `main` it should depend on `main` KIC, so it's updated manually (we'll do automation for it)


Makefile
```diff
@@ -157,9 +157,10 @@ _build.operator:
 build: generate
 	$(MAKE) build.operator
 
+GOLANGCI_LINT_CONFIG = $(PROJECT_DIR)/.golangci.yaml
```
pmalek: I believe we want `?=` to allow overrides, right?

```suggestion
GOLANGCI_LINT_CONFIG ?= $(PROJECT_DIR)/.golangci.yaml
```

pmalek (approved):

.tools_versions.yaml
```diff
@@ -1,10 +1,20 @@
+# renovate: datasource=github-releases depName=kubernetes/code-generator
```
pmalek, line 1: We still do not have renovate installed in this repo but it's good that you're adding this. 👍

PR #90, opened 2024-04-17
chore(docs): add changing layout of project to CHANGELOG.md

programmer04:

As pointed out in the https://github.com/Kong/gateway-operator/issues/130 such change deserves mentioning in `CHANGELOG.md`.

**Which issue this PR fixes**

Part of https://github.com/Kong/gateway-operator/issues/130

**PR Readiness Checklist**:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,12 @@
 
 ## Unreleased
 
+### Breaking Changes
+
+- Changes project layout to match `kubebuilder` `v4`. Some import paths (due to dir renames) have changed
+  `apis` -> `api` and `controllers` -> `controller`.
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -20,6 +20,12 @@
 
 ## Unreleased
 
+### Breaking Changes
+
+- Changes project layout to match `kubebuilder` `v4`. Some import paths have changed
+  e.g. `apis` -> `api`.
```
pmalek: I believe we should explicitly call out which changes are indeed breaking so `apis` -> `api` and `controllers` -> `controllers` (even though the latter is not really a part of an API that e.g. contains the types that one would normally use in a cluster but rather our controllers' logic.

pmalek (approved):

PR #95, opened 2024-04-17
tests: add helm upgrade tests

pmalek:

This PR adds some basic scaffolding for helm upgrade tests

**Which issue this PR fixes**

Part of https://github.com/Kong/gateway-operator/issues/155

~In order to perform update form pre 1.2 to 1.2+ we need to be able to disable `ControlPlane` and `Gateway` controllers because their respective CRDs (`ControlPlane` and `GatewayConfiguration`) got a breaking change in 1.2 (bump from `v1alpha1` to `v1beta` without a KGO version that would allow a smooth transition).~

~This requires a change in the chart which would allow setting args because only KGO 1.2+ supports configuration via env. https://github.com/Kong/charts/pull/1058 implements that.~

The upgrade from pre 1.2 to 1.2 and up is not done as helm chart (used in this test suite) only contains the CRDs from 1.2 and since helm still doesn't provide CRD management there's no easy way to achieve that (well apart from applying the CRDs manually but we'd need to add manual steps for that).

**PR Readiness Checklist**:


pmalek:
In order to perform update form pre 1.2 to 1.2+ we need to be able to disable `ControlPlane` and `Gateway` controllers because their respective CRDs (`ControlPlane` and `GatewayConfiguration`) got a breaking change in 1.2 (bump from `v1alpha1` to `v1beta` without a KGO version that would allow a smooth transition).

This requires a change in the chart which would allow setting args because only KGO 1.2+ supports configuration via env. https://github.com/Kong/charts/pull/1058 implements that.

test/helpers/setup.go
```diff
@@ -75,14 +75,20 @@ func ParseGoTestFlags(testRunner func(t *testing.T), testSuiteToRun []func(t *te
...
 	// They are explicitly passed to RunTestSuite(...) as an argument.
-	if err := fRun.Value.Set(getFunctionName(testRunner)); err != nil {
+	if err := fRun.Value.Set(getFunctionName(testRunner) + "/" + fRun.Value.String()); err != nil {
 		fmt.Println("testing: unexpected error happened (it should never happen, check the code)", err)
 		os.Exit(1)
 	}
 
+	fmt.Printf("test.run: %s\n", fRun.Value.String())
+
+	// os.Exit(0)
```
programmer04:
```suggestion
```

test/e2e/environment.go
```diff
@@ -101,11 +102,25 @@ func WithOperatorImage(image string) TestEnvOption {
...
+func WithInstallViaKustomize() TestEnvOption {
+	return func(opts *testEnvOptions) {
+		opts.InstallViaKustomize = true
+	}
+}
+
 var loggerOnce sync.Once
 
 // CreateEnvironment creates a new independent testing environment for running isolated e2e test.
+// When running with helm caller is responsible for cleaning up the environment.
```
programmer04:
```suggestion
// When running with Helm, the caller is responsible for cleaning up the environment.
```

programmer04 (commented):
Have you tested it with KGO EE? It seems it should work, but it's better to be sure 😉

test/e2e/test_helm_install_upgrade.go
```diff
@@ -0,0 +1,403 @@
...
+						require.Len(c, gws.Items, 1)
+						gw := &gws.Items[0]
+						assert.True(c, gateway.IsProgrammed(gw))
+						assert.True(c, gateway.AreListenersProgrammed(gw))
+					},
+				},
+			},
+		},
+		{
+			name:             "upgrade from nightly to current",
```
czeslavo, line 172: Does it make sense to test this upgrade path? I believe this is not an upgrade users would perform - maybe it's better to save CI time, WDYT?

test/e2e/test_helm_install_upgrade.go
```diff
@@ -0,0 +1,403 @@
...
+						require.Len(c, gws.Items, 1)
+						gw := &gws.Items[0]
+						assert.True(c, gateway.IsProgrammed(gw))
+						assert.True(c, gateway.AreListenersProgrammed(gw))
+					},
+				},
+			},
+		},
+		{
+			name:             "upgrade from 1.2.3 to current",
```
czeslavo, line 121: We'll probably need a release manual/automated action that will enforce updating test cases to cover the latest patch -> current upgrade path. Can we add this to the release procedure as a manual step for now?

test/e2e/environment.go
```diff
@@ -154,12 +169,21 @@ func CreateEnvironment(t *testing.T, ctx context.Context, opts ...TestEnvOption)
...
 	env, err := builder.Build(ctx)
 	require.NoError(t, err)
+
 	t.Cleanup(func() {
-		cleanupEnvironment(t, context.Background(), env, kustomizeDir.Tests())
+		if opt.InstallViaKustomize {
+			cleanupEnvironment(t, context.Background(), env, kustomizeDir.Tests())
+		} else { //nolint:revive,staticcheck
+			// TODO: using helm for installation, don't clean up because we do not know the release name.
+		}
```
czeslavo:
```suggestion
		}
```
As I understand we decided to not deploy with helm in this helper, but make this responsibility of the upgrade test body itself, so it's no longer a todo here, right?

pmalek:
> Have you tested it with KGO EE? It seems it should work, but it's better to be sure 😉

Good point. I've created https://github.com/Kong/gateway-operator-enterprise/pull/137 to test this

test/e2e/environment.go
```diff
@@ -91,7 +91,8 @@ type TestEnvironment struct {
 type TestEnvOption func(opt *testEnvOptions)
 
 type testEnvOptions struct {
-	Image string
+	Image               string
+	InstallViaKustomize bool
```
czeslavo, line 99: How about leaving here a comment explaining that if we define this to be `false` then no operator will be installed and it's a particular test's responsibility to deploy it in another way (e.g. with Helm)?

pmalek: 👍 Ideally we'd change the whole "framework" that we've got but the usage with helm and kustomize is so different (values vs no values etc) that at this stage I believe there's no point.

Added the comment.

test/e2e/test_helm_install_upgrade.go
pmalek: Tempting to get shorter CI runs 🙃 but this can potentially help us find subtle breaking changes in our most recent commits (that e.g. are not included in the latest tag).

test/e2e/environment.go
pmalek: Good point, 621149c3

pmalek:
> > Have you tested it with KGO EE? It seems it should work, but it's better to be sure 😉
> 
> Good point. I've created [Kong/gateway-operator-enterprise#137](https://github.com/Kong/gateway-operator-enterprise/pull/137) to test this

@programmer04 https://github.com/Kong/gateway-operator-enterprise/pull/137#issuecomment-2096470896

test/e2e/test_helm_install_upgrade.go
pmalek: Added a comment about renovate and https://github.com/Kong/gateway-operator/issues/121. PTAL.

I believe there's no point in adding a manual step to the release procedure since renovate should be enabled soon ™️ 

d153f0ae

czeslavo (approved):

programmer04 (approved):

PR #101, opened 2024-04-19
chore(ci): fix dir exclusion and add gocritic to golangci-lint, fix reported issues

programmer04:

```
make lint
```

reports

```log
WARN [config_reader] The configuration option `run.skip-dirs` is deprecated, please use `issues.exclude-dirs`. 
```

this PR fixes it.

Moreover I spotted that `gocritic` is not enable for this repository, enabling it reported some quite easy-to-fix and nice to have improved issues. See

```go
pkg/utils/kubernetes/metadata.go:75:3: assignOp: replace `name = name + "-"` with `name += "-"` (gocritic)
                name = name + "-"
                ^
test/conformance/suite_test.go:129:2: exitAfterDefer: os.Exit will exit, and `defer cleaner()` will not run (gocritic)
        os.Exit(code)
        ^
pkg/utils/kubernetes/compare/comparators.go:29:16: unlambda: replace `func(a, b corev1.ResourceRequirements) bool {
        return resources.ResourceRequirementsEqual(a, b)
}` with `resources.ResourceRequirementsEqual` (gocritic)
                cmp.Comparer(func(a, b corev1.ResourceRequirements) bool {
                             ^
test/integration/suite.go:189:2: exitAfterDefer: os.Exit will exit, and `defer cleaner()` will not run (gocritic)
        os.Exit(code)
        ^
modules/admission/validator.go:34:2: singleCaseSwitch: should rewrite switch statement to if statement (gocritic)
        switch operation {
        ^
controller/dataplane/owned_deployment.go:310:17: unlambda: replace `func(a, b corev1.ResourceRequirements) bool { return k8sresources.ResourceRequirementsEqual(a, b) }` with `k8sresources.ResourceRequirementsEqual` (gocritic)
                        cmp.Comparer(func(a, b corev1.ResourceRequirements) bool { return k8sresources.ResourceRequirementsEqual(a, b) }),
                                     ^
controller/gateway/controller.go:671:16: unlambda: replace `func(a, b corev1.ResourceRequirements) bool {
        return k8sresources.ResourceRequirementsEqual(a, b)
}` with `k8sresources.ResourceRequirementsEqual` (gocritic)
                cmp.Comparer(func(a, b corev1.ResourceRequirements) bool {
                             ^
controller/controlplane/controller_reconciler_utils.go:129:9: elseif: can replace 'else {if cond {}}' with 'else if cond {}' (gocritic)
        } else {
               ^
controller/controlplane/controller_reconciler_utils.go:211:17: unlambda: replace `func(a, b corev1.ResourceRequirements) bool { return k8sresources.ResourceRequirementsEqual(a, b) }` with `k8sresources.ResourceRequirementsEqual` (gocritic)
                        cmp.Comparer(func(a, b corev1.ResourceRequirements) bool { return k8sresources.ResourceRequirementsEqual(a, b) }),
                                     ^
controller/controlplane/status_conditions.go:13:2: singleCaseSwitch: should rewrite switch statement to if statement (gocritic)
        switch resource := any(resource).(type) {
```

This PR resolves all of them.

### DIFF
--- a/test/conformance/suite_test.go
+++ b/test/conformance/suite_test.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"os"
 	"path"
+	"runtime/debug"
 	"testing"
 	"time"
 
@@ -54,6 +55,14 @@ var (
 // -----------------------------------------------------------------------------
 
 func TestMain(m *testing.M) {
+	var code int
+	defer func() {
+		if r := recover(); r != nil {
+			fmt.Printf("%v stack trace:\n%s\n", r, debug.Stack())
+			code = 1
+		}
+		os.Exit(code)
+	}()
 	ctx, cancel = context.WithCancel(context.Background())
 	defer cancel()
 
@@ -103,7 +112,7 @@ func TestMain(m *testing.M) {
 	exitOnErr(testutils.BuildMTLSCredentials(ctx, clients.K8sClient, &httpc))
 
 	fmt.Println("INFO: environment is ready, starting tests")
-	code := m.Run()
+	code = m.Run()
 	if code != 0 {
 		output, err := env.Cluster().DumpDiagnostics(ctx, "gateway_api_conformance")
 		if err != nil {
@@ -125,8 +134,6 @@ func TestMain(m *testing.M) {
 		fmt.Println("INFO: cleaning up testing cluster and environment")
 		exitOnErr(env.Cleanup(ctx))
 	}
-
-	os.Exit(code)
 }
 
 // -----------------------------------------------------------------------------
--- a/test/integration/suite.go
+++ b/test/integration/suite.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"os"
 	"path"
+	"runtime/debug"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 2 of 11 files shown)


test/conformance/suite_test.go
```diff
@@ -54,6 +54,14 @@ var (
 // -----------------------------------------------------------------------------
 
 func TestMain(m *testing.M) {
+	var code int
+	defer func() {
+		if r := recover(); r != nil {
+			fmt.Printf("ERROR: %v\n", r)
+			code = 1
+		}
```
pmalek: Would `recover` produce a more readable output? I'm wondering if we won't loose information when printing this way.

test/integration/suite.go
```diff
@@ -108,6 +108,14 @@ func TestMain(
...
 	setUpAndRunManager SetUpAndRunManagerFunc,
 ) {
+	var code int
+	defer func() {
+		if r := recover(); r != nil {
+			fmt.Println(r)
+			code = 1
+		}
+		os.Exit(code)
+	}()
```
pmalek: ditto

test/conformance/suite_test.go
programmer04: The rationale for this change was an issue reported by `gocritic` - that due to `os.Exit` deffers won't execute. So let's ensure that `os.Exit` is executed after all defers (as a last defer). We lose stack trace from `panic` called in `exitOnErr` it's not very important I think

pmalek: We can always `defer` only the `os.Exit()` as the last one. IMHO panic stack traces are very helpful in debugging when something goes wrong.

programmer04: > We can always defer only the os.Exit() as the last one. 

Unfortunately, it doesn't, it will exit silently.

But stack trace can be ensured as I did now.

But still whatever we do we can't ensure that defers will execute, because in case of panicking in a different goroutine, it crashes entire program.

pmalek (approved):

test/conformance/suite_test.go
pmalek: Ok got it. Let's merge this as is and we can iterate on printing stack traces.

PR #108, opened 2024-04-22
chore(refactor): get rid of golang.org/x/exp

programmer04:

Functionalities provided by `golang.org/x/exp` are now available in stdlib and `github.com/samber/lo` so this PR removes redundant dependency `golang.org/x/exp` and implements protection from accidentally pulling it again.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,6 @@ require (
 	github.com/kr/pretty v0.3.1
 	github.com/samber/lo v1.39.0
 	github.com/stretchr/testify v1.9.0
-	golang.org/x/exp v0.0.0-20231226003508-02704c960a9b
 	k8s.io/api v0.29.4
 	k8s.io/apimachinery v0.29.4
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -52,6 +51,7 @@ require (
 	go.opentelemetry.io/otel/metric v1.24.0 // indirect
 	go.opentelemetry.io/otel/trace v1.24.0 // indirect
 	go4.org/netipx v0.0.0-20230728184502-ec4c8b891b28 // indirect
+	golang.org/x/exp v0.0.0-20231226003508-02704c960a9b // indirect
```
pmalek, line 54: We still end up depending on this module it seems 🤷 But at least it's not a direct dependency

pmalek (approved):

go.mod
programmer04: Yeah, but it's not our fault at least

PR #202, opened 2024-04-23
chore(ci): update release workflow

pmalek:

- add `mise` installation steps where it manages the tools used in workflows
- ~do not set credentials to checkout KGO as private submodule (from EE): not needed anymore~
  - this is needed because the workflow is reused
- do not require build-push-images for integration tests in release pipeline (those can be run in parallel with the build stage)
- ~do not set `PAT_GITHUB` as env when building the docker image.~
  - this is needed by EE's Dockerfile

### DIFF
--- a/.github/workflows/__build-workflow.yaml
+++ b/.github/workflows/__build-workflow.yaml
@@ -137,7 +137,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Configure Git for private repositories (temporary until making KGO public)
+      - name: Configure Git for private repositories (this is needed by repositories that include this workflow and have other private dependencies)
         run: git config --global url."https://${{ secrets.gh-pat }}@github.com".insteadOf "https://github.com"
 
       - name: Checkout KGO submodule
@@ -273,7 +273,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Configure Git for private repositories (temporary until making KGO public)
+      - name: Configure Git for private repositories (this is needed by repositories that include this workflow and have other private dependencies)
         run: git config --global url."https://${{ secrets.gh-pat }}@github.com".insteadOf "https://github.com"
 
       - name: Checkout KGO submodule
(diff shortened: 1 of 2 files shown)


.github/workflows/__build-workflow.yaml
```diff
@@ -231,7 +228,6 @@ jobs:
             REPO_INFO=https://github.com/${{ github.repository }}.git
             GOPATH=${{ env.GOPATH}}
             GOCACHE=${{ env.GOCACHE}}
-            PAT_GITHUB=${{ secrets.gh-pat }}
```
pmalek: @czeslavo I do not see any references of this env in the repo. There's a secret with that name but Dockerfile(s) do not require this env. Is there an implicit dependency on this somewhere or can we get rid of this?

.github/workflows/__build-workflow.yaml
```diff
@@ -137,9 +137,6 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Configure Git for private repositories (temporary until making KGO public)
-        run: git config --global url."https://${{ secrets.gh-pat }}@github.com".insteadOf "https://github.com"
-
```
programmer04: Do we reuse those pipelines in EE, don't we?
So, it's required for https://github.com/Kong/go-enterprise-license that is private and required by EE build

programmer04 (changes requested):

.github/workflows/__build-workflow.yaml
pmalek: ~For KGO EE this would not be necessary because workflows on Github already contain the token that has access to checkout the repo that they are running for (I'm pretty sure that's the case).~

This would be necessary as you've pointed out for the ee license repo though to check it out. Let me revert that change

.github/workflows/__build-workflow.yaml
pmalek: This seems to be required in EE's Dockerfile. This is complex 🙃

.github/workflows/__build-workflow.yaml
programmer04: I think that comment should be updated, because 
> temporary until making KGO public

is no longer true

pmalek: 4821c04f

programmer04 (approved):

PR #206, opened 2024-04-23
fix(ci): fix release CI to work with release branches

pmalek:

This fixes the release pipeline to make it work with release branches.

Up until now the `.github/workflows/release-bot.yaml` was only triggered on `main` and `.github/workflows/__release-workflow.yaml` was only creating release PRs against `main`. That caused 1.2.2 to be released on `main` which we did not want (due to that branch already including a breaking change which we did not want to include in 1.2.x).

This change makes the release workflow to accept a base branch argument so that we can release from release branches.

Ideally the `VERSION` and manifests would also get an update on `main` but that can be done separately.

This also retracts the 1.2.2 tag since it was not created on a desired SHA. Deleting and re-pushing the tag does not help because golang's proxy caches modules indefinitely (https://sum.golang.org/#faq-retract-version) hence the `retract` clause added in `go.mod`.

### DIFF
--- a/.github/workflows/__release-workflow.yaml
+++ b/.github/workflows/__release-workflow.yaml
@@ -178,6 +178,11 @@ jobs:
           KONG_LICENSE_DATA: ${{ secrets.kong-license-data }}
           KONG_TEST_GATEWAY_OPERATOR_IMAGE_OVERRIDE: ${{ needs.build-push-images.outputs.full_tag }}
 
+  # NOTE: This job's steps are run when:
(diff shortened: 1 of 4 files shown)


programmer04 (approved):

pmalek:
@programmer04 PTAL: I've included the changes 159aa8c5
 to trigger a PR against `main` as well when a release on release branch is created.

programmer04 (approved):

PR #221, opened 2024-04-24
chore(build): use build-platform instead of emulation

pmalek:

Same thing as https://github.com/Kong/kubernetes-ingress-controller/pull/5788/files but for KGO.

Example: https://github.com/Kong/gateway-operator/actions/runs/8817901181


programmer04 (approved):
🐎

PR #222, opened 2024-04-24
fix: conflict updating ControlPlane's finalizers is not an error

pmalek:

This PR prevents the following error when e.g. creating a `Gateway`:

```
{"level":"error","ts":"2024-04-24T09:53:30Z","msg":"Reconciler error","controller":"controlplane","controllerGroup":"gateway-operator.konghq.com","controllerKind":"ControlPlane","ControlPlane":{"name":"gw1-42khv","namespace":"default"},"namespace":"default","name":"gw1-42khv","reconcileID":"ff149a0b-f471-4671-966c-37184e4224ee","error":"failed updating ControlPlane's finalizers : Operation cannot be fulfilled on controlplanes.gateway-operator.konghq.com \"gw1-42khv\": the object has been modified; please apply your changes to the latest version and try again","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.17.2/pkg/internal/controller/controller.go:329\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.17.2/pkg/internal/controller/controller.go:266\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.17.2/pkg/internal/controller/controller.go:227"}
```

which is not an error which user should be aware of so just mask it and retry.

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:

### DIFF
--- a/controller/controlplane/controller.go
+++ b/controller/controlplane/controller.go
@@ -211,11 +211,17 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	crbFinalizerSet := controllerutil.AddFinalizer(cp, string(ControlPlaneFinalizerCleanupClusterRoleBinding))
 	vwcFinalizerSet := controllerutil.AddFinalizer(cp, string(ControlPlaneFinalizerCleanupValidatingWebhookConfiguration))
 	if crFinalizerSet || crbFinalizerSet || vwcFinalizerSet {
-		log.Trace(logger, "Setting finalizers", cp)
+		log.Trace(logger, "setting finalizers", cp)
 		if err := r.Client.Update(ctx, cp); err != nil {
+			if k8serrors.IsConflict(err) {
+				log.Debug(logger, "conflict found when updating ControlPlane, retrying", cp)
+				return ctrl.Result{Requeue: true, RequeueAfter: requeueWithoutBackoff}, nil
+			}
 			return ctrl.Result{}, fmt.Errorf("failed updating ControlPlane's finalizers : %w", err)
 		}
-		return ctrl.Result{}, nil
+		// Requeue to ensure that we do not miss next reconciliation request in case
+		// AddFinalizer calls returned true but the update resulted in a noop.
+		return ctrl.Result{Requeue: true, RequeueAfter: requeueWithoutBackoff}, nil
 	}
 
 	k8sutils.InitReady(cp)


controller/controlplane/controller.go
```diff
@@ -211,11 +211,17 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 		if err := r.Client.Update(ctx, cp); err != nil {
+			if k8serrors.IsConflict(err) {
+				log.Debug(logger, "conflict found when updating ControlPlane, retrying", cp)
+				return ctrl.Result{Requeue: true, RequeueAfter: requeueWithoutBackoff}, nil
+			}
 			return ctrl.Result{}, fmt.Errorf("failed updating ControlPlane's finalizers : %w", err)
 		}
-		return ctrl.Result{}, nil
+		// Requeue to ensure that we do not miss next reconciliation request in case
+		// AddFinalizer calls returned true but the update resulted in a noop.
```
czeslavo, line 223: Can you elaborate on when this could happen? Is it that when finalizers are added in parallel, before a call to `Client.Update`, `Client.Update` will return no error then?

czeslavo (commented):
I think we could add a note the changelog for that.

controller/controlplane/controller.go
pmalek: It's a theoretical situation where the update that's meant to be applied would yield no update (a noop) which would cause no reconciliation (which would continue deploying  `ControlPlance` owned subresources below).

It's a "better safe than sorry" clause here to prevent the controller to "get stuck".

pmalek: We should in theory be guarded against that by https://github.com/kubernetes-sigs/controller-runtime/blob/cb5fcace24fe2b6bcfe94f6b7fe143bd7e8ea2ef/pkg/config/controller.go#L35-L36 which by default is 1 and we [do not change it](https://github.com/Kong/gateway-operator/blob/30d829f9537f9b47d0ca495b086081ea05f76ce0/modules/manager/run.go#L158-L173) but there might also be other, external factors which might cause a change. Hypothetically.

czeslavo: Makes sense, thanks 👍

czeslavo (approved):

PR #224, opened 2024-04-25
chore: move and publish CRD checker

rainest:

We have some features that require CRD availability and _do not_ use Reconcilers. These features should check that their required CRDs are available before attempting to interact with them.

To support this, move the manager's CRD checker utility into the `kubernetes` utils package and publish it.

**Which issue this PR fixes**

Related to https://github.com/Kong/gateway-operator-enterprise/issues/118

**PR Readiness Checklist**:

- [ ] ~the `CHANGELOG.md` release notes have been updated to reflect significant changes~ Nothing user-facing

### DIFF
--- a/pkg/utils/kubernetes/crd.go
+++ b/pkg/utils/kubernetes/crd.go
@@ -1,0 +1,54 @@
+package kubernetes
+
+import (
+	"errors"
+	"fmt"
+	"net/url"
+
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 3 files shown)


pkg/utils/kubernetes/crd.go
pmalek, line 1: Could we add a UT or 2 for that? This PR just moves the code but I believe it's a good opportunity to add UTs while we're at it.

rainest: That was the only purpose of the original `modules/manager/controller_setup_test.go`, so I just moved that entire file as well.

pmalek (approved):
Didn't notice the test being moved 🤦

PR #248, opened 2024-05-08
chore: handle Gateway cleanup errors

pmalek:

This PR makes handling Gateway cleanup errors hide errors that users do not need to see, which in most cases will cause a requeue anyway.

For example:

```
Gateway.gateway.networking.k8s.io \"kong\" is invalid: metadata.finalizers: Forbidden: no new finalizers can be added if the object is being deleted, found new finalizers []string{\"gateway-operator.konghq.com/cleanup-network-policies\"}
```

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:

### DIFF
--- a/controller/gateway/controller_cleanup.go
+++ b/controller/gateway/controller_cleanup.go
@@ -20,7 +20,7 @@ import (
 // Reconciler - Cleanup
 // ----------------------------------------------------------------------------
 
-const requeueWithoutBackoff = time.Millisecond * 200
+const requeueWithoutBackoff = 200 * time.Millisecond
 
 // cleanup determines whether cleanup is needed/underway for a Gateway and
 // performs all necessary cleanup steps. Namely, it cleans up resources
--- a/controller/gateway/controller_cleanup.go
+++ b/controller/gateway/controller_cleanup.go
@@ -135,6 +135,11 @@ func (r *Reconciler) cleanup(
 }
 
 func handleGatewayFinalizerPatchOrUpdateError(err error, gateway *gatewayv1.Gateway, logger logr.Logger) (ctrl.Result, error) {
+	// Short cirtcuit.
+	if err == nil {
+		return ctrl.Result{}, nil
+	}
+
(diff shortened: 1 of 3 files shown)


controller/gateway/controller_cleanup.go
```diff
@@ -129,3 +133,28 @@ func (r *Reconciler) cleanup(
...
+		log.Debug(logger, "failed to delete a finalizer on Gateway, requeueing request", gateway, "cause", cause)
+		return ctrl.Result{
+			Requeue:      true,
+			RequeueAfter: requeueWithoutBackoff,
+		}, nil
+	}
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	return ctrl.Result{}, nil
```
programmer04:
```suggestion
	return ctrl.Result{}, err
```

btw. can we shortcircuit at the top with
`if err == nil`?

controller/gateway/controller_cleanup.go
```diff
@@ -18,6 +20,8 @@ import (
 // Reconciler - Cleanup
 // ----------------------------------------------------------------------------
 
+const requeueWithoutBackoff = time.Millisecond * 200
```
programmer04:
```suggestion
const requeueWithoutBackoff = 200 * time.Millisecond
```

controller/gateway/controller_cleanup.go
pmalek: This way it's consistent that the least indented path is a happy path + the explicit `nil` makes it easy to figure out.

I'm OK with accepting the suggestion if you fill like this would improve the readability on your end. I might be biased.

programmer04: Both have benefits and drawbacks, this can be left as it is. 

Btw. putting `if err == nil` to return at the top?

pmalek: 4d9f6e6b025529ea01f69dc3bb62c6481373236e

👍

programmer04 (approved):

PR #326, opened 2024-06-12
feat: detect when ControlPlane's admission webhook is disabled and ensure relevant resources do not exist

### LINKED ISSUE #308: Allow disabling `ControlPlane`'s admission webhook deployment and management
### Problem statement

https://github.com/Kong/gateway-operator-archive/pull/1545 introduced deploying `ControlPlane`'s (KIC's) admission webhook in an unconditional manner: each `ControlPlane` gets its admission webhook configuration (and related resources) deployed.

This can be an issue for several use cases:

- KIC's `HTTPRoute` admission rules do not allow admission of routes' that contain invalid/unsupported `Group` and/or `Kind` in `spec.rules.backendRefs`: https://github.com/Kong/kubernetes-ingress-controller/blob/a9749c8a2589d3b3e421ffe1b5a6ac4d43359c78/internal/admission/validation/gateway/httproute.go#L163-L171. 
  - [A decision has been made](https://github.com/Kong/kubernetes-ingress-controller/pull/6104#issuecomment-2137828852) to retain that behavior but to allow users to disable admission webhook (this is already possible in KIC itself) to be conformant with Gateway API (which requires the routes to be admitted and their statuses appropriately filled).

- Due to the nature of KGO generating patches and applying them to resources in the cluster, when the environment mutates resources managed by KGO then it might end up in an endless loop as in #239.
  - Allowing users to disable the creation and management of `ControlPlane`'s admission webhook configuration we'll allow users have a working KGO setup until #239's underlying issue (the way KGO calculates patches and the expected state) is resolved.

### Proposed solution

- Detect if `ControlPlane` has `CONTROLLER_ADMISSION_WEBHOOK_LISTEN` overridden with `off` (https://github.com/Kong/kubernetes-ingress-controller/blob/bcdd40a4099b797e274ba72e276a62fc1785053b/internal/manager/setup.go#L201) and if that's the case, disable creation of admission webhook configuration (and other related resources) for `ControlPlane`.
  - This would not introduce a new field to not allow conflicts with the aforementioned env
- Detect change in `CONTROLLER_ADMISSION_WEBHOOK_LISTEN` env when `ControlPlane` is deployed
  - If its value changes from `off` to anything else (that's a valid `<address>:<port>` string) then deploy admission webhook configuration
  - If it changes from anything to `off` then remove admission webhook configuration and any other resources related to it.

### Special notes

Putting tentatively in 1.3

pmalek:

Ensure that relevant resources 

- webhook service
- admissionwebhookconfiguration
- admission webhook certificate service

do not exist when `ControlPlane`'s admission webhook is disabled via `CONTROLLER_ADMISSION_WEBHOOK_LISTEN` env.

**Which issue this PR fixes**

Fixes #308 

**PR Readiness Checklist**:

### DIFF
--- a/controller/controlplane/controller.go
+++ b/controller/controlplane/controller.go
@@ -482,7 +482,7 @@ func (r *Reconciler) ensureWebhookResources(
 	}
 
 	log.Trace(logger, "ensuring admission webhook service", cp)
-	res, admissionWebhookService, err := r.ensureAdmissionWebhookService(ctx, logger, r.Client, cp)
+	res, admissionWebhookService, err := r.ensureAdmissionWebhookService(ctx, r.Client, cp, webhookEnabled)
 	if err != nil {
 		return "", res, fmt.Errorf("failed to ensure admission webhook service: %w", err)
 	}
@@ -496,7 +496,7 @@ func (r *Reconciler) ensureWebhookResources(
 	}
 
 	log.Trace(logger, "ensuring admission webhook certificate", cp)
-	res, admissionWebhookCertificateSecret, err := r.ensureAdmissionWebhookCertificateSecret(ctx, logger, cp, admissionWebhookService)
+	res, admissionWebhookCertificateSecret, err := r.ensureAdmissionWebhookCertificateSecret(ctx, cp, admissionWebhookService, webhookEnabled)
 	if err != nil {
 		return "", res, err
 	}
@@ -510,7 +510,7 @@ func (r *Reconciler) ensureWebhookResources(
 	}
 
 	log.Trace(logger, "ensuring admission webhook configuration", cp)
-	res, err = r.ensureValidatingWebhookConfiguration(ctx, cp, admissionWebhookCertificateSecret, admissionWebhookService)
+	res, err = r.ensureValidatingWebhookConfiguration(ctx, cp, admissionWebhookCertificateSecret, admissionWebhookService, webhookEnabled)
 	if err != nil {
 		return "", res, err
 	}
--- a/controller/controlplane/controller.go
+++ b/controller/controlplane/controller.go
@@ -63,6 +63,9 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 	validatinWebhookConfigurationOwnerPredicate.UpdateFunc = func(e event.UpdateEvent) bool {
 		return r.validatingWebhookConfigurationHasControlPlaneOwner(e.ObjectOld)
 	}
+	validatinWebhookConfigurationOwnerPredicate.DeleteFunc = func(e event.DeleteEvent) bool {
+		return r.validatingWebhookConfigurationHasControlPlaneOwner(e.Object)
+	}
 
 	return ctrl.NewControllerManagedBy(mgr).
 		// watch ControlPlane objects
@@ -285,7 +288,6 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	}
 	changed := controlplane.SetDefaults(
 		&cp.Spec.ControlPlaneOptions,
-		nil,
 		defaultArgs)
 	if changed {
 		log.Debug(logger, "updating ControlPlane resource after defaults are set since resource has changed", cp)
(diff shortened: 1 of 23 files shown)


pkg/utils/kubernetes/resources/deployments_test.go
```diff
@@ -274,3 +277,308 @@ func TestGenerateNewDeploymentForDataPlane(t *testing.T) {
...
+							DNSPolicy:                     corev1.DNSClusterFirst,
+							SchedulerName:                 corev1.DefaultSchedulerName,
+							TerminationGracePeriodSeconds: lo.ToPtr(int64(30)),
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "no webhook cert secret name specified doesn't the webhook volume, volume mount nor port",
```
czeslavo, line 447:
```suggestion
			name: "no webhook cert secret name specified doesn't generate the webhook volume, volume mount nor port",
```
```suggestion
			name: "no webhook cert secret name specified doesn't the webhook volume, volume mount nor port",
```

pkg/utils/kubernetes/resources/deployments.go
```diff
@@ -121,11 +129,17 @@ func GenerateNewDeploymentForControlPlane(params GenerateNewDeploymentForControl
 	return deployment, nil
 }
 
+// GenerateContainerForControlPlaneParams is a parameter struct for GenerateControlPlaneContainer function.
+type GenerateContainerForControlPlaneParams struct {
+	Image                          string
+	AdmissionWebhookCertSecretName string
```
czeslavo: How about using `mo.Option` or just a pointer to signal this is optional?

controller/controlplane/controller_reconciler_utils.go
```diff
@@ -467,6 +468,24 @@ func (r *Reconciler) ensureAdmissionWebhookCertificateSecret(
 	matchingLabels := client.MatchingLabels{
 		consts.SecretUsedByServiceLabel: consts.ControlPlaneServiceKindWebhook,
 	}
+	if !isAdmissionWebhookEnabled(ctx, r.Client, logger, cp) {
```
czeslavo, line 471: I guess this call can be considered quite cheap (given it fetches the pod container from the local cache), but we might calculate it once in `ensureWebhookResources` and pass it along to all functions depending on that. The drawback I can see is the growing number of parameters to be passed. Have you considered it?

controller/controlplane/controller.go
```diff
@@ -489,3 +470,80 @@ func (r *Reconciler) patchStatus(ctx context.Context, logger logr.Logger, update
...
+			log.Debug(logger, "ValidatingWebhookConfiguration has been created/updated", cp)
+		}
+	}
+	if webhookEnabled {
+		return admissionWebhookCertificateSecret.Name, res, nil
+	}
+	return "", res, nil
+}
+
+func isAdmissionWebhookEnabled(ctx context.Context, cl client.Client, logger logr.Logger, cp *operatorv1beta1.ControlPlane) bool {
```
czeslavo, line 530: Have you considered another API for that, e.g. adding an explicit toggle `DisableValidatingWebhook` to the CP specification? Given this is a net new functionality (disabling the webhook), having it exposed properly via strongly typed API would make sense to me. Using the Pod template spec patch to disable the webhook by setting an env var name that I need to know from somewhere seems like not the best UX to me.

czeslavo (commented):
LGTM in general. I'm interested in your opinion on making it depend on the pod spec vs. strong API.

controller/controlplane/controller.go
pmalek: Yes, I have (that is briefly mentioned in #308's Proposed solution section).

Assuming that this field would be enabled by default users would have to disable the webhook both via that field and in the env (unless we provide some sort of admission rules that would reject `ControlPlane`s that have those 2 conflicted)

The problem with that is that we'd allow users to create conflicted configurations:

- disable the dedicated field but forget to disable the env: `ControlPlane` will go into `CrashLoopBackoff` due to missing webhook certs in the container
- do not use the dedicated field (leave it enabled by default) but disable the webhook via env: `ControlPlane` will have its webhook resources created but the webhook won't be functional. This would cause all the admission to fail but [get ignored due to our policy setting](https://github.com/Kong/gateway-operator/blob/3e2ae4b34d78e97a83a9c5128a3f5b66b4008f7c/pkg/utils/kubernetes/resources/validatingwebhookconfig/zz_generated_kic_ge3_1.go#L21-L24).

---

Do you believe we should consider a dedicated field given those challenges?

pkg/utils/kubernetes/resources/deployments.go
pmalek: 👍 I'll go with a pointer just for the sake to not warrant a dependency on `mo` since we do not use that just yet in this project

controller/controlplane/controller_reconciler_utils.go
pmalek: I've considered it and had exactly the same thoughts about that.

Let me try to pass that as an arg b4bec069e94a33b3e903eda5018e689f448fa683

controller/controlplane/controller.go
czeslavo: > disable the dedicated field but forget to disable the env: ControlPlane will go into CrashLoopBackoff due to missing webhook certs in the container

I assume that, by default, `ControlPlane` doesn't have to specify any Pod template spec patches - its basic functionality should be achievable without customization. Am I wrong?

> do not use the dedicated field (leave it enabled by default) but disable the webhook via env: ControlPlane will have its webhook resources created but the webhook won't be functional. This would cause all the admission to fail but [get ignored due to our policy setting](https://github.com/Kong/gateway-operator/blob/3e2ae4b34d78e97a83a9c5128a3f5b66b4008f7c/pkg/utils/kubernetes/resources/validatingwebhookconfig/zz_generated_kic_ge3_1.go#L21-L24).

Yeah, that sounds like a valid argument which is also valid for any other functionality we allow modifying with the Pod template spec patches (users without Kong's university degree can shoot their feet easily by manually modifying arbitrary env variables in ways unexpected by us). Having this exposed via a strongly typed API would allow us to add a validation that would prevent setting this specific env var via patches. 

I don't want to halt this, but I believe that is something worth discussing. I'm worried we're going to end up with APIs that are not easy to use. If we can avoid relying on patches that are quite unintuitive to me (IMO they should only be used by "power users" for use cases that we haven't thought of yet, or haven't implemented yet), I think we should at least consider that.

controller/controlplane/controller_reconciler_utils.go
czeslavo: > Let me try to pass that as an arg https://github.com/Kong/gateway-operator/commit/b4bec069e94a33b3e903eda5018e689f448fa683

Not great, not terrible. I think calling `isAdmissionWebhookEnabled` for readability's and maintainability's sake might make sense. Let's leave it - we'll always have something to improve in terms of performance, and I don't think this will be the most burning issue. 🙂

controller/controlplane/controller.go
pmalek: > I assume that, by default, `ControlPlane` doesn't have to specify any Pod template spec patches - its basic functionality should be achievable without customization. Am I wrong?

That scenario could happen when someone disables the webhook as part of an update.

> I don't want to halt this, but I believe that is something worth discussing. I'm worried we're going to end up with APIs that are not easy to use. If we can avoid relying on patches that are quite unintuitive to me (IMO they should only be used by "power users" for use cases that we haven't thought of yet, or haven't implemented yet), I think we should at least consider that.

💯 I believe that `podTemplateSpec` implementation that we provided was done a bit too far reaching from the get go but now we have to support it (at least until next major version of the API) and use it consistently for all the features that we implement.

That's something I'd consider for v2 API (this is probably not happening for v1): namely, to get rid of the `podTemplateSpec` and provide a limited number for dedicated fields which we have more control over.

controller/controlplane/controller_reconciler_utils.go
pmalek: 51185ab7bd7725da75b455694009951275d5def0

controller/controlplane/controller.go
mlavacca: I think we should be very careful about the addition of such high-level knobs. ControlPlane and DataPlane are low-level APIs and adding high-level abstractions to low-level settings in the same API (or in different APIs) is a bad idea in my opinion, as it opens the doors to many potential problems, such as the one already discussed in this thread.

mlavacca (approved):

controller/controlplane/controller.go
pmalek: 👍 

Obviously that would need to be carefully designed but one potential way to do this could be to remove the `podTemplateSpec` from `GatewayConfiguration` and putting in its place those high level knobs.

Users would still be able to customize their `DataPlane`s and `ControlPlane`s but at the `GatewayConfiguration` level they would have to use those higher level, more managed knobs.

czeslavo: I think that my main concern here is that we add a functionality (possibility to forcibly not enable CP validating webhook) via a very implicit API (you need to know what specific env variable needs to be set to `off` in order to trigger this quite non-trivial side-effect). Reading the API specification, users won't be able to find it easily I'm afraid. 

Another side of the coin is that we primarily need this feature for passing conformance tests. The intention may be to actually _not_ expose this toggle to the users openly, thus "hiding" this somewhat within the patches' infinite possibilities 😄 If that's the case, I'm fine with doing it as it is.

czeslavo (approved):

PR #343, opened 2024-06-18
chore: update shared CI workflows

programmer04:

Include changes that adjust the build process to changes presented in https://github.com/Kong/gateway-operator-enterprise/pull/167

**Which issue this PR fixes**

Part of https://github.com/Kong/gateway-operator-enterprise/issues/66

### DIFF
--- a/.github/workflows/__build-workflow.yaml
+++ b/.github/workflows/__build-workflow.yaml
@@ -39,6 +39,12 @@ on:
         required: false
         type: string
         default: Dockerfile
+      additional-build-contexts:
+        description: >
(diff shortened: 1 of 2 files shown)


.github/workflows/__build-workflow.yaml
```diff
@@ -39,6 +39,10 @@ on:
         required: false
         type: string
         default: Dockerfile
+      additional-build-contexts:
+        description: Additional build context to include in the build (format name=path).
```
czeslavo: Could we leave a comment on why this might be useful for our future selves?

pmalek (approved):

PR #357, opened 2024-06-20
ci: create release branches after release

pmalek:

This PR makes the release pipeline 

- create release branches only for major or minor release **after** the release has been made
- makes the `release-bot.yaml` workflow cherry-pick the commit with release from release branch to `main` (when `latest`, patch release is released)

### DIFF
--- a/.github/workflows/release-bot.yaml
+++ b/.github/workflows/release-bot.yaml
@@ -140,4 +140,4 @@ jobs:
         with:
           branch: main
           title: '[cherry-pick] ${{ needs.semver.outputs.version }} - ${{ github.sha }}'
-          body: 'Cherry picking ${{ needs.semver.outputs.version }} commit (${{ github.sha }}) onto this branch'
+          body: 'Cherry picking ${{ needs.semver.outputs.version }} commit (${{ github.sha }}) onto main'
--- a/.github/workflows/release-bot.yaml
+++ b/.github/workflows/release-bot.yaml
@@ -14,6 +14,7 @@ jobs:
     outputs:
       release_found: ${{ steps.commit_parser.outputs.release_found }}
       release_type: ${{ steps.commit_parser.outputs.release_type }}
+      release_latest: ${{ steps.commit_parser.outputs.release_latest }}
     runs-on: ubuntu-latest
     
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/release-bot.yaml
```diff
@@ -97,3 +104,40 @@ jobs:
...
+    if: ${{ needs.look_for_release.outputs.release_found == 'true' && needs.semver.outputs.patch != '0' && needs.semver.outputs.prerelease == '' && needs.look_for_release.outputs.release_latest == 'true' && github.ref_name != 'main' }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: carloscastrojumo/github-cherry-pick-action@v1.0.1
+        with:
+          branch: main
+          title: '[cherry-pick] {old_title}'
+          body: 'Cherry picking #{old_pull_request_id} onto this branch'
```
czeslavo: `.github/workflows/release-bot.yaml` is triggered on push to main or release branches. As I understand the [action docs](https://github.com/carloscastrojumo/github-cherry-pick-action?tab=readme-ov-file#action-inputs) it might happen that the special strings (`old_title` and `old_pull_request_id`) won't be populated in such a case as the action expects pull-request trigger.

Maybe we could make this a separate workflow to mitigate this? Or just accept that we won't have this relation between PRs included in the body and title?

czeslavo (commented):
Ah, reviewing GH workflows 🫠 Read it twice, beside the comment about special strings possibly not being populated by `carloscastrojumo/github-cherry-pick-action` it looks right to me.

.github/workflows/release-bot.yaml
pmalek: Good catch! Yes, it seems there's no easy way to get that relation in that place. Changed that to just include the version and commit SHA.

PTAL.

.github/workflows/release-bot.yaml
```diff
@@ -97,3 +104,40 @@ jobs:
...
+    if: ${{ needs.look_for_release.outputs.release_found == 'true' && needs.semver.outputs.patch != '0' && needs.semver.outputs.prerelease == '' && needs.look_for_release.outputs.release_latest == 'true' && github.ref_name != 'main' }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: carloscastrojumo/github-cherry-pick-action@v1.0.1
+        with:
+          branch: main
+          title: '[cherry-pick] ${{ needs.semver.outputs.version }} - ${{ github.sha }}'
+          body: 'Cherry picking ${{ needs.semver.outputs.version }} commit (${{ github.sha }}) onto this branch'
```
czeslavo:
```suggestion
          body: 'Cherry picking ${{ needs.semver.outputs.version }} commit (${{ github.sha }}) onto main'
```

czeslavo (approved):

PR #374, opened 2024-06-24
chore: perform go mod download in shared build-workflow

programmer04:

Ensure that Go dependencies are downloaded when `inputs.additional-build-contexts` is specified. It points to a package downloaded by Go toolchain in case of EE build and due to limitations of GH Workflows it has to be done here

### DIFF
--- a/.github/workflows/__build-workflow.yaml
+++ b/.github/workflows/__build-workflow.yaml
@@ -195,6 +195,12 @@ jobs:
         uses: actions/setup-go@v5
         with:
           go-version-file: go.mod
+      
+      # Path for additional-build-contexts may ponits to dependences that are pulled by Go toolchain,
+      # so we need to download them before building the image.
+      - name: Set up Go dependencies for additional build contexts
+        if: ${{ inputs.additional-build-contexts != ''}}
+        run: go mod download
 
       - run: echo "GOPATH=$(go env GOPATH)" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


.github/workflows/__build-workflow.yaml
```diff
@@ -195,6 +195,12 @@ jobs:
         uses: actions/setup-go@v5
         with:
           go-version-file: go.mod
+      
+      # Path for additional-build-contexts may ponits to dependences that are pulled by Go toolchain,
+      # so we need to download them before building the image.
```
pmalek:
```suggestion
      # Path for additional-build-contexts may point to dependences that are pulled by Go toolchain,
      # so we need to download them before building the image.
```

pmalek:
> and due to limitations of GH Workflows it has to be done here

To be specific: due to the fact that Github Actions' jobs are not guaranteed to run on the same  runners

.github/workflows/__build-workflow.yaml
```diff
@@ -195,6 +195,12 @@ jobs:
         uses: actions/setup-go@v5
         with:
           go-version-file: go.mod
+      
+      # Path for additional-build-contexts may point to dependences that are pulled by Go toolchain,
```
pmalek:
```suggestion
      # Path for additional-build-contexts may point to dependencies that are pulled by Go toolchain,
```

pmalek (approved):

PR #387, opened 2024-06-28
feat: set proper User-Agent for each request

programmer04:

Proper `User-Agent` header is now set on outgoing HTTP requests. An example one is
`kong-gateway-operator-oss-v1.4.0`. The naming pattern follows how Docker images are named where the OSS build has `-oss` suffix.

**Which issue this PR fixes**

Part of https://github.com/Kong/gateway-operator-enterprise/issues/111

Take a look at https://github.com/Kong/gateway-operator-enterprise/pull/229 which requires changes presented in this PR.

**PR Readiness Checklist**:

### DIFF
--- a/modules/manager/metadata/metadata.go
+++ b/modules/manager/metadata/metadata.go
@@ -3,7 +3,7 @@ package metadata
 
 import (
 	"fmt"
-	"strings"
+	"runtime"
 )
 
 // -----------------------------------------------------------------------------
@@ -49,10 +49,9 @@ type Info struct {
 
 // UserAgent returns the User-Agent string to use in all HTTP requests made by KGO.
 func (inf Info) UserAgent() string {
-	org := strings.ToLower(inf.Organization)
 	return fmt.Sprintf("%s/%s (%s/%s) (%s)",
-			inf.ProjectName, inf.Release, runtime.GOOS, runtime.GOARCH, inf.Flavor,
-		)
+		inf.ProjectName, inf.Release, runtime.GOOS, runtime.GOARCH, inf.Flavor,
+	)
 }
 
 var (
--- a/modules/manager/metadata/metadata.go
+++ b/modules/manager/metadata/metadata.go
@@ -50,10 +50,9 @@ type Info struct {
 // UserAgent returns the User-Agent string to use in all HTTP requests made by KGO.
 func (inf Info) UserAgent() string {
 	org := strings.ToLower(inf.Organization)
-	if inf.Flavor == OSSFlavor {
-		return fmt.Sprintf("%s-%s-%s/%s", org, inf.ProjectName, OSSFlavor, inf.Release)
-	}
-	return fmt.Sprintf("%s-%s/%s", org, inf.ProjectName, inf.Release)
+	return fmt.Sprintf("%s/%s (%s/%s) (%s)",
+			inf.ProjectName, inf.Release, runtime.GOOS, runtime.GOARCH, inf.Flavor,
+		)
 }
 
 var (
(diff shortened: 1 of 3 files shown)


modules/manager/metadata/metadata.go
```diff
@@ -64,6 +88,6 @@ func Metadata() Info {
 		Commit:       commit,
 		ProjectName:  projectName,
 		Organization: organization,
-		Flavor:       flavor,
+		Flavor:       BuildFlavor(flavor),
```
czeslavo: Could we change the above 
```go
	// Flavor is the flavor of the build.
	flavor = "oss"
```
to
```go
	// Flavor is the flavor of the build.
	flavor = OSSFlavor
```
to avoid duplication of this magic string?

programmer04: 😅 sure I overlooked this, thx

modules/manager/metadata/metadata.go
```diff
@@ -29,7 +44,16 @@ type Info struct {
...
+	Flavor BuildFlavor
+}
+
+// UserAgent returns the User-Agent string to use in all HTTP requests made by KGO.
+func (inf Info) UserAgent() string {
+	org := strings.ToLower(inf.Organization)
+	if inf.Flavor == OSSFlavor {
+		return fmt.Sprintf("%s-%s-%s/%s", org, inf.ProjectName, OSSFlavor, inf.Release)
+	}
+	return fmt.Sprintf("%s-%s/%s", org, inf.ProjectName, inf.Release)
```
pmalek: I'd suggest to include
- OS
- arch

Similarly to what Kubernetes in tree components like controller manager do it:

```
kube-controller-manager/v1.30.0 (linux/arm64) kubernetes/7c48c2b/leader-election
```

We can also put the flavor in the UA regardless to it's value

```suggestion
	return fmt.Sprintf("%s-%s/%s (%s/%s) (%s)",
		org, inf.ProjectName, inf.Release, runtime.GOOS, runtime.GOARCH, inf.Flavor,
	)
```

programmer04: I did it like that to be consistent with [what we do in KIC](https://github.com/Kong/kubernetes-ingress-controller/blob/b996cd5c8a978d8a08d69524e1c2ecd0bbbe1821/internal/manager/metadata/metadata.go#L45-L47). More info is better so  I can adjust in both repos to stay consistent. 

In terms of name, I'm leaning towards being consistent with Docker image naming, because this is the most typical way of running it

pmalek: I see. The problem is that we use `ProjectName` here which is bound to GitHub's repo name (not the docker repo name). 

Do you mean by this that you intend to change your initial change?

programmer04: TBH I don't know how to do it in the best way, because we don't have a variable in `Makefile` that maps one to one. So I came up with that construction of the name. On the other hand, I don't want to hardcode anything here. For now names map (repo and Docker image), but in the future whoever knows...

pmalek: I'd change my initial proposal to

- drop the org as that's already implied that it's a kong project
- keep the project name and not bother about mapping to docker image 1:1. The latter might also change.

WDYT about:

```suggestion
	return fmt.Sprintf("%s/%s (%s/%s) (%s)",
			inf.ProjectName, inf.Release, runtime.GOOS, runtime.GOARCH, inf.Flavor,
		)
```

?

My reasoning is that User Agent is mostly to know what's causing particular requests against the API server. I believe we don't need to stress about the consistency that much (Docker vs GitHub repo).

pmalek (approved):

PR #400, opened 2024-07-08
feat: introduce CRD KongPluginInstallation

### LINKED ISSUE #378: New CRD for custom Kong Plugin installation
## Problem Statement

A new CRD for dealing with Kong Plugins is needed. It needs to have well-thought fields that won't require changes in the future it should be extensible and take into account adding in the foreseeable future.

- https://github.com/Kong/gateway-operator/issues/372

Of course, fields for signature verification are **not in the scope** of this issue. 

Related Kubernetes controller typical boilerplate has to be generated too.

## Proposed Solution

There is an [initial proposition presented in Kong Gateway Plugins distribution](https://docs.google.com/document/d/1-gYEaf_31BR_QXmZXNhy0bOrXYJ4X_cXobQ_vbLDIhE/edit#heading=h.3tn2kjmrw2q5) that can be used as a starting point, inspiration. Naming hasn't been decided yet the proposed options for now are
- `KongPluginInstallation`
- `KongPluginInstance`

## Acceptance Criteria
- [ ] There is a new CRD with well thought fields that serves the purpose and all boilerplate

programmer04:

This PR introduces a new CRD for dealing with Custom Kong Plugins. The name of that CRD is `KongPluginInstallation`. It's a namespace CRD that may be extended in the future. It allows referencing K8s secrets from arbitrary chosen namespaces. Here the minimal set of fields is presented. For status, it follows an approach of GWAPI objects. 

This PR contains only `KongPluginInstallation`, which describes whether an image with a plugin has been successfully fetched and unpacked. The user will learn about the actual status of installation in the status field of CRDs, which will reference the instance of `KongPluginInstallation`.

**Which issue this PR fixes**

Closes #378 

When K8s deals with fetching images it allows to specify `imagePullPolicy` to `IfNotPresent`, `Always`, `Never`. The default behavior is tricky. Thus I decided to not add this field and to fetch images always.

Also, machinery for generating boilerplate was adjusted as part of this PR to use mise installed kustomize.

**PR Readiness Checklist**:


api/v1alpha1/kongplugin_installation.go
```diff
@@ -0,0 +1,85 @@
...
+	SchemeBuilder.Register(&KongPluginInstallation{}, &KongPluginInstallationList{})
+}
+
+//+genclient
+//+k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+//+kubebuilder:object:root=true
+//+kubebuilder:subresource:status
+//+kubebuilder:resource:shortName=kpi,categories=kong;all
+// +kubebuilder:subresource:status
+//+kubebuilder:printcolumn:name="Ready",description="The Resource is ready",type=string,JSONPath=`.status.conditions[?(@.type=='Ready')].status`
```
pmalek: Since `Ready` is explicitly added as a print column I believe it would be worth adding it and its possible reasons that it can be used with to the API (similarly as we've done in KIC in : https://github.com/Kong/kubernetes-ingress-controller/blob/7e567cc4192532561b630d58f6500009e067ba88/pkg/apis/configuration/v1/conditions.go).

This will make it clear what does it mean for `KongPluginInstallation` to be ready.

After initially going through this PR it's not immediately clear to me what that would mean.

WDYT?

api/v1alpha1/kongplugin_installation.go
```diff
@@ -0,0 +1,85 @@
...
+
+// KongPluginInstallation is the Schema
```
pmalek: We might add a bit more elaborate comment to this struct explaining what is its purpose. This will also get copied to CRD ref doc which will help users understand why they might want to use it. 

WDYT?

api/v1alpha1/kongplugin_installation.go
```diff
@@ -0,0 +1,85 @@
...
+// KongPluginInstallationSpec defines the desired state of KongPluginInstallation.
+type KongPluginInstallationSpec struct {
+
+	// Image is OCI image reference for the Kong Plugin.
+	//
+	// +kubebuilder:validation:Required
+	Image string `json:"image"`
+
+	// SecretRef allows specifying secret to be used for OCI registry authentication to pull the image.
+	// When omitted it is assumed that the image is public.
```
pmalek: I believe a link to a page (probably k8s docs if that's relevant) explaining the expected structure of the provided `Secret` might be beneficial. This will not force users to go and read the docs but find this information in here.

api/v1alpha1/kongplugin_installation.go
```diff
@@ -0,0 +1,85 @@
...
+
+// KongPluginInstallationStatus defines the observed state of KongPluginInstallation.
+type KongPluginInstallationStatus struct {
+	// Conditions describe the current conditions of the KongPluginInstallation.
+	//
+	// Known condition types are:
+	//
+	//   - "Accepted"
+	//   - "Fetching"
+	//   - "PluginFetched"
```
pmalek: - IMHO it'd be better (as suggested above) to make those part of the API and propose them in this PR. This would also allow adding comments to condition type and reason consts to let users know when they can expect them.
- `Fetching` doesn't seem like a good candidate for a condition `Type`. Perhaps this was meant to be condition's reason?

api/v1alpha1/kongplugin_installation.go
pmalek, line 1: To keep consistent this file should have a `_types.go` suffix just like all the other files defining API types (e.g. https://github.com/Kong/gateway-operator/blob/3876430e09e61edce58bd8464989e33236bd1872/api/v1beta1/gatewayconfiguration_types.go)

api/v1alpha1/kongplugin_installation.go
```diff
@@ -0,0 +1,129 @@
...
+	// available for use a specific Custom Kong Plugin.
+	//
+	// It is a positive-polarity summary condition, and so should always be
+	// present on the resource with ObservedGeneration set.
+	//
+	// It should be set to Unknown if the controller performs updates to the
+	// status before it has all the information it needs to be able to determine
+	// if the condition is true.
+	//
+	// Possible reasons for this condition to be true are:
```
pmalek:
```suggestion
	// Possible reasons for this condition to be True are:
```

Just to be consistent with other comments below.

api/v1alpha1/kongplugin_installation.go
```diff
@@ -0,0 +1,129 @@
...
+	//
+	// Possible reasons for this condition to be False are:
+	//
+	// * "Invalid"
+	// * "Fetching"
+	//
+	// Possible reasons for this condition to be Unknown are:
+	//
+	// * "Fetching".
+	//
```
pmalek: It seems that some lines are missing here, no?

For instance `Programmed` condition reason is not described in here.

api/v1alpha1/kongplugin_installation.go
```diff
@@ -0,0 +1,129 @@
...
+// KongPluginInstallationStatus.Conditions field.
+type KongPluginInstallationConditionType string
+
+// KongPluginInstallationConditionReason defines the set of reasons that explain why
+// a particular KongPluginInstallation condition type has been raised.
+type KongPluginInstallationConditionReason string
+
+const (
+	// This condition indicates whether the controller has fetched and make
+	// available for use a specific Custom Kong Plugin.
```
pmalek:
```suggestion
	// This condition indicates whether the controller has fetched the plugin image
	// and made it  available for use as a specific Custom Kong Plugin.
```

api/v1alpha1/kongplugin_installation.go
```diff
@@ -0,0 +1,129 @@
...
+	// * "Fetching"
+	//
+	// Possible reasons for this condition to be Unknown are:
+	//
+	// * "Fetching".
+	//
+
+	// Controllers should prefer to use the values of KongPluginInstallationConditionReason
+	// for the corresponding Reason, where appropriate.
+	KongPluginInstallationConditionStatusAccepted KongPluginInstallationConditionType = "Accepted"
```
pmalek: This comment doesn't seem like it belongs to this condition type. It should describe the `Accepted` condition type.

Additionally, to please the linter it would be better to start all those comments with the name of the `const` being described.

api/v1alpha1/kongplugin_installation.go
```diff
@@ -0,0 +1,129 @@
...
+	SchemeBuilder.Register(&KongPluginInstallation{}, &KongPluginInstallationList{})
+}
+
+//+genclient
+//+k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+//+kubebuilder:object:root=true
+//+kubebuilder:subresource:status
+//+kubebuilder:resource:shortName=kpi,categories=kong;all
+//+kubebuilder:subresource:status
+//+kubebuilder:printcolumn:name="Programmed",type=string,JSONPath=`.status.conditions[?(@.type=="Programmed")].status`
```
pmalek: There is no such condition type added to the API. Is this a type and this should be `Accepted`?

I'm wondering if making the "primary" condition for this CRD type `Ready`, wouldn't be a bit better in terms of API semantics.

If the user defines the `KongPluginInstallation`, it cares if it's made "Ready" for him. This CRD type can't really be programmed anywhere ( correct me if I'm wrong ) but it can be "used" by other types, right?

pmalek (commented):
Food for thought on the conditions for this type:

- make `Ready` a primary type for this CRD
- it can have its reason set to `Fetched` when status is `True`
- it can have its reason set to `Failed` when status is `False` and there are problems fetching the image
- it can have its reason set to `Unknown` initially


I'm interested to hear your thoughts on this.

api/v1alpha1/kongplugin_installation.go
programmer04: Sure, to my surprise linter wasn't complain

api/v1alpha1/kongplugin_installation.go
programmer04: I'd prefer `Ready` too. I think so

api/v1alpha1/kongplugin_installation_types.go
```diff
@@ -0,0 +1,129 @@
...
+	// * "Pending"
+	//
+	// Possible reasons for this condition to be "Unknown" are:
+	//
+	// * "Pending".
+	//
+	KongPluginInstallationConditionStatusAccepted KongPluginInstallationConditionType = "Ready"
+
+	// KongPluginInstallationReasonReady is used with the "Ready" condition when
+	// the condition is "True".
```
pmalek: I believe this is already good but I wonder if we can expand this to give readers a bit more insight into the state of the resource they can expect when this reason is used. 

What are the conditions of the resource when this is used? What has happened? What is available to the user? Can we provide this information here?

api/v1alpha1/kongplugin_installation_types.go
```diff
@@ -0,0 +1,129 @@
...
+	// if the condition is true.
+	//
+	// Possible reasons for this condition to be "True" are:
+	//
+	// * "Ready"
+	//
+	// Possible reasons for this condition to be "False" are:
+	//
+	// * "Failed"
+	// * "Pending"
```
pmalek, line 109: I wonder when `Pending` is used with `False`: what can cause this scenario? Wouldn't `Failed` be the only natural failed state?( at least initially, I can see us expanding this to allow other failed reasons in the future )

 How "pending" can be failed?

api/v1alpha1/kongplugin_installation_types.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// KongPluginInstallationStatus defines the observed state of KongPluginInstallation.
+type KongPluginInstallationStatus struct {
+	// Conditions describe the current conditions of this KongPluginInstallation.
+	//
+	//+listType=map
+	//+listMapKey=type
+	//+kubebuilder:validation:MaxItems=8
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
```
pmalek, line 79: Don't we want to use `Ready` with `reason=Pending` and `status=Unknown` as default?

api/v1alpha1/kongplugin_installation_types.go
```diff
@@ -0,0 +1,129 @@
...
+	KongPluginInstallationReasonReady KongPluginInstallationConditionReason = "Ready"
+
+	// KongPluginInstallationReasonFailed is used with the "Ready" condition when
+	// the KongPluginInstallation can't be configured e.g. image can't be fetched.
+	// More details can be obtained from the condition's message.
+	KongPluginInstallationReasonFailed KongPluginInstallationConditionReason = "Failed"
+
+	// KongPluginInstallationReasonPending is used with the "Ready" condition when the requested
+	// controller has started processing the KongPluginInstallation, but it hasn't finished yet.
+	KongPluginInstallationReasonPending KongPluginInstallationConditionReason = "Pending"
```
pmalek: Using `Pending` as a reason when the controller has already picked this for reconciliation is slightly semantically misleading.

WDYT about following Gateway API in this aspect and using that reason when the controller is yet to pick this resource for reconciliation (that means setting this reason with `Unknown` status as default)?

ref: https://github.com/kubernetes-sigs/gateway-api/blob/c1b375bda97d344948a50ebd40814ffcd68b78ac/apis/v1/gateway_types.go#L858-L861

---

OTOH if we want to express the resource state when the controller has picked this up - or has "accepted" it - I suggest we consider adding a separate condition type for this and call it `Accepted` https://github.com/kubernetes-sigs/gateway-api/blob/c1b375bda97d344948a50ebd40814ffcd68b78ac/apis/v1/gateway_types.go#L820-L845.

WDYT?

api/v1alpha1/kongplugin_installation_types.go
rainest: The `Conditions` array doesn't have any default, or [shouldn't, with empty as the implied default unknown](https://maelvls.dev/kubernetes-conditions/#conditions-vs-state-machine). If we want something that does have a default, that should have a dedicated field under `Status`, separate from `Conditions`.

`Conditions` entries should only appear after the controller has touched it, and presumably has something of note to report, even if that's just "I've seen this and haven't done anything with it yet".

api/v1alpha1/kongplugin_installation_types.go
```diff
@@ -0,0 +1,129 @@
...
+//+k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+//+kubebuilder:object:root=true
+//+kubebuilder:subresource:status
+//+kubebuilder:resource:shortName=kpi,categories=kong;all
+//+kubebuilder:subresource:status
+//+kubebuilder:printcolumn:name="Ready",type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].status`
+
+// KongPluginInstallation allows using a custom Kong Plugin distributed as a container image available in a registry.
+// Such a plugin can be associated with GatewayConfiguration or DataPlane to be available for particular Kong Gateway
+// and configured with KongPlugin CRD.
```
rainest, line 38:
```suggestion
// KongPluginInstallation represents a Kong plugin available for installation by a Gateway or DataPlane.
// When a Gateway or DataPlane resource requests a KongPluginInstallation, the operator retrieves
// the plugin indicated by the KongPluginInstallation's spec and configures Volumes, VolumeMounts,
// and environment variables for the Deployment managed by the requesting resource.
```

- I'd expect this is added to the Gateway rather than the GatewayConfiguration--DataPlanes and Gateways are roughly equivalent-level resources for non-GWAPI and GWAPI environments. The original doc has these in a spec field (and we can't add one to Gateway), but doesn't indicate anything more complex than a CSV list of CR names. Do we think we'd ever need more than the names (and maybe namespaces) KongPluginInstallations? Could we stick this in an annotation instead if no? Tag is the only extra info I can think of--if we do allow multiple requesters for the same resource, we'd probably want to let the spec indicate a tag so that different requesters can use different versions an image.
- Only top-level "is thing, controller does action with thing" description here: we can get into the details of registries and images in the associated spec field comments.
- I kinda violate my previous point with the discussion of Volumes and such, but AFAIK this _won't_ have multiple installation implementations on the managed Deployment side--if we did have multiple options, that'd probably warrant an additional spec field describing them (and we'd then go into the details with those spec fields' comments).
- KongPlugin mention probably unnecessary here: there's no tight coupling between the two resources.

api/v1alpha1/kongplugin_installation_types.go
```diff
@@ -0,0 +1,129 @@
...
+type KongPluginInstallationList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []KongPluginInstallation `json:"items"`
+}
+
+// KongPluginInstallationSpec defines the desired state of KongPluginInstallation.
+type KongPluginInstallationSpec struct {
+
+	// The image is an OCI image URL for a packaged Custom Kong Plugin.
```
rainest:
```suggestion
	// The image is an OCI image URL for a packaged custom Kong plugin.
```
https://docs.konghq.com/gateway/latest/kong-plugins/#custom-plugins doesn't capitalize these terms, we shouldn't either.

api/v1alpha1/kongplugin_installation_types.go
```diff
@@ -0,0 +1,129 @@
...
+//+kubebuilder:object:root=true
+
+// KongPluginInstallationList contains a list of KongPluginInstallation.
+type KongPluginInstallationList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []KongPluginInstallation `json:"items"`
+}
+
+// KongPluginInstallationSpec defines the desired state of KongPluginInstallation.
```
rainest:
```suggestion
// KongPluginInstallationSpec provides the information necessary to retrieve and install a Kong custom plugin.
```

api/v1alpha1/kongplugin_installation_types.go
```diff
@@ -0,0 +1,129 @@
...
+type KongPluginInstallationSpec struct {
+
+	// The image is an OCI image URL for a packaged Custom Kong Plugin.
+	//
+	//+kubebuilder:validation:Required
+	Image string `json:"image"`
+
+	// SecretRef allows referring specific Kubernetes Secret to use for OCI registry authentication for pulling an image with Custom Kong Plugin.
+	// The Secret format should follow https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry.
+	// When the field is omitted it is assumed that the image is public and can be fetched without providing credentials.
```
rainest:
```suggestion
	// SecretRef is a reference to a Kubernetes Secret containing credentials necessary to pull the OCI image
	// in Image. It must follow the format in https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry.
	// It is optional. If the image is public, omit this field.
```

api/v1alpha1/kongplugin_installation_types.go
```diff
@@ -0,0 +1,129 @@
...
+	//
+	//+listType=map
+	//+listMapKey=type
+	//+kubebuilder:validation:MaxItems=8
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+}
+
+// KongPluginInstallationConditionType is the type for status conditions on
+// KongPluginInstallation resources. This type should be used with the
+// KongPluginInstallationStatus.Conditions field.
```
rainest:
```suggestion
// The following areKongPluginInstallation specific types for
// https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.30/#condition-v1-meta fields

// KongPluginInstallationConditionType is the type for Conditions in a KongPluginInstallation's
// Status.Conditions array.
```

api/v1alpha1/kongplugin_installation_types.go
```diff
@@ -0,0 +1,129 @@
...
+type KongPluginInstallationConditionType string
+
+// KongPluginInstallationConditionReason defines the set of reasons that explain why
+// a particular KongPluginInstallation condition type has been raised.
```
rainest:
```suggestion
// KongPluginInstallationConditionReason is a reason for the KongPluginInstallation condition's last transition.
```

Borrowing (abbreviated) upstream language this one, since may as well--we're using them for the same purpose, just scoping them.

api/v1alpha1/kongplugin_installation_types.go
```diff
@@ -0,0 +1,129 @@
...
+	// Possible reasons for this condition to be "False" are:
+	//
+	// * "Failed"
+	// * "Pending"
+	//
+	// Possible reasons for this condition to be "Unknown" are:
+	//
+	// * "Pending".
+	//
+	KongPluginInstallationConditionStatusAccepted KongPluginInstallationConditionType = "Ready"
```
rainest:
```suggestion
	KongPluginInstallationConditionStatusAccepted KongPluginInstallationConditionType = "Accepted"
```

I think this was the intent here? `"Ready"` is one of the Reasons.

api/v1alpha1/kongplugin_installation_types.go
```diff
@@ -0,0 +1,129 @@
...
+const (
+	// This condition indicates whether the controller has fetched the plugin image
+	// and made it available for use as a specific Custom Kong Plugin.
+	//
+	// It is a positive-polarity summary condition, and so should always be
+	// present on the resource with ObservedGeneration set.
+	//
+	// It should be set to Unknown if the controller performs updates to the
+	// status before it has all the information it needs to be able to determine
+	// if the condition is true.
```
rainest: Retrieval and installation should probably be separate Condition types, since they can fail or succeed independently. Likely scenarios include:

- The controller cannot and has never retrieved the plugin image, e.g. because the credentials are invalid. Retrieval is `False` and installation is also `False`.
- The controller has retrieved the plugin image successfully, but cannot install it, e.g. because the API server died in the interim between retrieval and the Deployment modification attempt (ignore that the controller won't be able to update status _either_ in this case--there just are scenarios where retrieval is fine, but the work after can't proceed). Retrieval is `True` but installation is `False`.
- The controller has previously retrieved and installed the plugin, but the image server goes offline and the controller can't check it for updates, so retrieval is currently `False` while installation is `True`.

Do we have a rough implementation outline for the binding UX between DataPlanes/Gateways and KongPluginInstallations? This PR should probably include scaffolding for that alongside the CRD, since the binding semantics matter for status information. The design doc doesn't cover that AFAICT.

The main point I can think of is whether the KongPluginInstallation to DataPlane/Gateway relationship is many to one or many to many. The doc implies multiple KongPluginInstallations (it only allows for a single image in a given CR, and I expect we want to support multiple custom plugins), but doesn't indicate if they can then be requested by multiple DataPlane/Gateway resources or only one.

The many to one case is simple, as the installation type condition need only reflect a single modified Deployment, but the many to many case is less simple. If we support many to many, we'd need either:
- Installation is an aggregate condition. All requesters satisfied (`True`) or not all requesters satisfied (`False`). We'd presumably need to indicate _which_ requesters are unsatisfied in `message`, since it's the only dynamic value field.
- Installation either isn't a condition or is an aggregate condition, but `Status` includes an additional field with per-requester information.

I kinda considered a separate `InstallationConditions` array for per-instance, but there's not really any field in `Condition` that'd make sense for storing the instance identifier.

api/v1alpha1/kongplugin_installation_types.go
rainest: `Pending` seems reasonable if the controller has downloaded the plugin image and has a request to install it, but can't install it yet, since the Deployment for the requester doesn't exist yet. You could hit a scenario where the controller can't create the requester's Deployment for whatever reason, and thus can't attempt installation.

On the retrieval side, status is `False` while reason is `Pending` when retrieval hasn't started yet or hasn't completed.

`Failed`, in contrast, means that the controller can attempt retrieval or installation, did so, and definitely did not succeed.

`Unknown` is the more open question for me, since I can't really think of a scenario where we know we need to do something, but don't know if we've tried or if that attempt has succeeded yet. I guess you could use it to differentiate "Pending, not started" and "Pending, in progress, will probably reach Ready soon"? We probably don't need that level of granularity--most downloads should finish or fail within 1-3 seconds.

api/v1alpha1/kongplugin_installation_types.go
rainest: Ditto the separation between installation and retrieval types in my other comment. With that, comments would be something like:

```
	// KongPluginInstallationReasonReady indicates that the controller has added
	// volumes, mounts, and environment configuration to the requester's Deployment.
```
```
	// KongPluginRetrievalReasonReady indicates that the controller has downloaded the plugin
	// and can install it on a DataPlane or Gateway.
```

api/v1alpha1/kongplugin_installation_types.go
```diff
@@ -0,0 +1,129 @@
...
+	//
+	// * "Pending".
+	//
+	KongPluginInstallationConditionStatusAccepted KongPluginInstallationConditionType = "Ready"
+
+	// KongPluginInstallationReasonReady is used with the "Ready" condition when
+	// the condition is "True".
+	KongPluginInstallationReasonReady KongPluginInstallationConditionReason = "Ready"
+
+	// KongPluginInstallationReasonFailed is used with the "Ready" condition when
```
rainest: This ignores the requested type split. As-is this a reason referring to another reason, and should instead refer to a type.

```suggestion
	// KongPluginInstallationReasonFailed is used with the "Accepted" condition type when
```

api/v1alpha1/kongplugin_installation_types.go
```diff
@@ -0,0 +1,129 @@
...
+	// the KongPluginInstallation can't be configured e.g. image can't be fetched.
+	// More details can be obtained from the condition's message.
+	KongPluginInstallationReasonFailed KongPluginInstallationConditionReason = "Failed"
+
+	// KongPluginInstallationReasonPending is used with the "Ready" condition when the requested
```
rainest: Same reason/type relationship mismatch.

```suggestion
	// KongPluginInstallationReasonPending is used with the "Accepted" condition type when the requested
```

CHANGELOG.md
```diff
@@ -30,6 +30,9 @@
 - Proper `User-Agent` header is now set on outgoing HTTP requests.
   [#387](https://github.com/Kong/gateway-operator/pull/387)
 
+- Introduce `KongPluginInstallation` CRD
+  [400](https://github.com/Kong/gateway-operator/pull/400)
```
rainest: This should maybe wait pending the rest of the feature, and then be a single changelog entry that mentions multiple PRs. Could also just leave this in and then update it as the other bits are added.

rainest (changes requested):
Some brevity copyedits and two major bits:

- Retrieval and installation states are independent and should have separate condition types.
- It's unclear whether we'll allow reuse of the same KongPluginInstallation across mulitple DataPlanes. If so, that has major implications for status, since there are now multiple simultaneous installation states on the same CR.

For the second, only allowing one is simpler and maybe good practice (there's no way to accidentally update a bunch of other Deployments by modifying the shared resource). The effort to duplicate them is low and the impact probably is also--even if we don't cache image downloads for the same URL, plugins aren't huge files and re-downloading them doesn't waste much bandwidth.

CHANGELOG.md
programmer04: My typical flow is to add it as fast as possible and updating then

api/v1alpha1/kongplugin_installation_types.go
programmer04: I think that name + namespace should be enough. Because as you mentioned I can't specify new fields in GWAPI `Gateway`. So in the design doc CRDs `GatewayConfiguration` and `DataPlane` are places to add additional fields. It should serve all purposes. And there is one place to influence Gateway's configuration a `GatewayConfiguration` CRD.

The idea with annotations is OK too, but harder to extend in the future. I can't say now whether we would extend something related to this... For me, annotations are something that you use when you can't modify CRD and you want to configure it and there is no other way to achieve it. But it's more I think personal/community opinion/preference

I thought that in case of want a different image (a different version) a separate instance of `KongPluginInstallation` would be configured. 

In terms of docstrings, I think we can adjust them when we sort out the above

The open question is that maybe instead of `KongPluginInstallation` it should be called `KongPluginInstance` as @mlavacca proposed in the design doc.

programmer04:
For https://github.com/Kong/gateway-operator/pull/400#pullrequestreview-2167573317

- single instance is reused by multiple Kong Gateways (I start thinking about lifecycle because probably to Kong Gateway to see an updated version of a plugin needs a restart_
- status reports only whether `ConfigMap` has been successfully created because we can't reason about successful installation, it's just mount, AFAIK only when the plugin is used we can say if it works
- To know if it's mounted I thought about two possibilities
  - extending `KongPluginInstallation` status with a list of Gateways that reference that plugin (it will be managed by Gateway Controller) and had its configuration adjusted and be restarted 
  - extending `Gateway` & `Dataplane` status with a field that says about the status of plugins if they're mounted and Gateway has been restarted

api/v1alpha1/kongplugin_installation_types.go
rainest: Harder to extend in the future, yeah, but unless we can think of something obvious it's YAGNI territory. If we do find that we need something more complex in the binding, worst case we have two methods for binding, simple and more complex.

Sync discussion came down on the side of only allowing one DataPlane or Gateway to bind a KPI, so if we go with GatewayConfiguration, that single-attach relationship would presumably extend to the whole GatewayConfiguration CRD. That may make sense anyway, but AFAIK it's not something we've enforced in the past.

I like having config visibility directly on the Gateway resource itself when possible (though I recognize we've already gone well past that point with everything else in GatewayConfiguration) and having the same interface for both DataPlane and Gateway, but similar fields in a spec is probably close enough for it to not matter that much.

I prefer `KongPluginInstallation` since it's more descriptive (it's a CR that installs plugins, a name that reflects that is good).

rainest:
From sync discussion:

We want to limit a given KPI to a single DataPlane or Gateway, so we can use the simpler condition-based status. We have to use a pull model (push from the KPI to its installation target would create all sorts of fun security problems), so we'll presumably need an additional "you tried to install this on multiple gateways, you can't" condition.

We should persist the original installation in such a case, but won't be able to tell which request was first from existing conditions (they indicate if the plugin is installed, but not where). We should add a status field or some other mechanism to indicate what gateway resource "owns" a given KPI so that bad configuration doesn't disrupt a working environment.

AFAIK OCI URLs indicate image version, so we shouldn't need to poll the URL for updated code--we will know an update occurred because the URL actually changed.

This and other operator features can trigger a Deployment update and new Pod rollout, which may be undesirable. We may want to consider some indicator that allows DataPlane/Gateway resource owners to indicate "hey, even if there's a change to other operator resources that require an update and rollout, wait until I give the okay". The specifics of this would need their own design doc though, so for now it's just something to keep in mind and track any related user requests.

Re:

> status reports only whether ConfigMap has been successfully created because we can't reason about successful installation, it's just mount, AFAIK only when the plugin is used we can say if it works

We can know for certain that we've also successfully added the associated Volume(Mount) and envvars to the Deployment, so we should report as much. As long as we've added that configuration to the Deployment, we can reasonably expect Kubernetes to realize it, or report if it can't through normal channels (Deployment status will indicate if the expected Pods for the latest revision are happy or not).

api/v1alpha1/kongplugin_installation_types.go
programmer04: To conclude, the decision has been made
- name `KongPluginInstallation`
- `KongPluginInstallation` can be configured in `GatewayConfiguration` and `DataPlane`

pmalek:
> We should persist the original installation in such a case, but won't be able to tell which request was first from existing conditions (they indicate if the plugin is installed, but not where). We should add a status field or some other mechanism to indicate what gateway resource "owns" a given KPI so that bad configuration doesn't disrupt a working environment.

Actually: what stands in our way to allow multi use of KPI? I'd imagine that

- KPI is a thing that triggers download and store (in a ConfigMap but that's an implementation detail) action and then it gets a status condition with a CM ref, where the storage happened
- DataPlane or Gateway (through GatewayConfiguration) sees the KPI refs in the spec and mounts those (read only?). Operator after a successful `Deployment` `spec` change (setting mounts) and rollout being complete, sets a status condition (or a dedicated status KPI condition with a KPI ref). The rollout will be triggered as per `Deployment`'s  [`spec.strategy`](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#strategy).

Am I missing something?

rainest:
> KPI is a thing that triggers download and store (in a ConfigMap but that's an implementation detail) action and then it gets a status condition with a CM ref, where the storage happened

This is all set--retrieval side seems fine.

> Am I missing something?

Depends on where the status info ultimately lives. If it's on the KPI CR and there are multiple clients, we need a status spec that can present that info. If it ends up as part of the client (DataPlane/GatewayConfiguration) status, then those client resource statuses can reflect their local state.

api/v1alpha1/kongplugin_installation_types.go
programmer04: It's like that, I've added examples to descriptions.

programmer04:
I've updated the PR description to include that the scope for status in `KongPluginInstallation` is to tell whether the plugin has been successfully fetched. DataPlane/GatewayConfiguration statuses' will reflect each if the plugin has been mounted successfully 
cc: @rainest

CHANGELOG.md
```diff
@@ -29,6 +29,8 @@
 
 - Proper `User-Agent` header is now set on outgoing HTTP requests.
   [#387](https://github.com/Kong/gateway-operator/pull/387)
+- Introduce `KongPluginInstallation` CRD
```
randmonkey: Should we complete the introduction of the usage in this PR? like 
```suggestion
- Introduce `KongPluginInstallation` CRD to install custom plugins.
  [#400](https://github.com/Kong/gateway-operator/pull/400)
```

api/v1alpha1/kongplugin_installation_types.go
```diff
@@ -0,0 +1,133 @@
...
+	//
+	//+kubebuilder:validation:Required
+	Image string `json:"image"`
+
+	// SecretRef is a reference to a Kubernetes Secret containing credentials necessary to pull the OCI image
+	// in Image. It must follow the format in https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry.
+	// It is optional. If the image is public, omit this field.
+	//
+	//+optional
+	SecretRef *corev1.SecretReference `json:"secretRef,omitempty"`
```
randmonkey: naming: I would suggest to name it as `ImagePullSecret` or `ImagePullSecretRef`. The name is clearer and I think there might be possibly other secrets to be used in spec.

rainest (approved):

PR #404, opened 2024-07-12
chore: disable build artifacts upload on GitHub

pmalek:

Disable build report artifacts.

Reference: https://github.com/docker/build-push-action?tab=readme-ov-file#environment-variables

### DIFF
--- a/.github/workflows/__build-workflow.yaml
+++ b/.github/workflows/__build-workflow.yaml
@@ -208,6 +208,9 @@ jobs:
       - name: Build image
         id: build
         uses: docker/build-push-action@v6.3.0
+        env:
+          DOCKER_BUILD_RECORD_UPLOAD: false
+          DOCKER_BUILD_SUMMARY: false
         with:
           context: .
           build-contexts: ${{ inputs.additional-build-contexts }}


.github/workflows/__build-workflow.yaml
```diff
@@ -208,6 +208,9 @@ jobs:
       - name: Build image
         id: build
         uses: docker/build-push-action@v6.3.0
+        env:
+          DOCKER_BUILD_RECORD_UPLOAD: false
+          DOCKER_BUILD_SUMMARY: false
```
saisatishkarra, line 213: I would still keep the `DOCKER_BUILD_SUMMARY=true` to upload summary as GH job summay where the inputs are masked / redacted on github. Ex: https://github.com/Kong/gateway-operator/actions/runs/9885055008.(Refer `Build inputs` on Docker summary for each job)

.github/workflows/__build-workflow.yaml
```diff
@@ -208,6 +208,9 @@ jobs:
       - name: Build image
         id: build
         uses: docker/build-push-action@v6.3.0
+        env:
+          DOCKER_BUILD_RECORD_UPLOAD: false
```
saisatishkarra, line 212: The `DOCKER_BUILD_RECORD_UPLOAD: false` must be set to false as this is leading to non-redacted secrets being exposed in the oci archive summary report as pointed by the researcher

.github/workflows/__build-workflow.yaml
pmalek: Discussed in slack.

saisatishkarra: Agree to disable this as it  will expose the summary which shows that we're using the token which might already be an inclination to folks scavenging across the :github:. However i would have the team enable this summary when the secrets are used to pass the PAT instead of docker args.

saisatishkarra (approved):
gtm

PR #424, opened 2024-07-23
feat: introduce basic reconciliation for KongPluginInstallation CRD

### LINKED ISSUE #379: Controller of Kong Plugin installation CRD is able to fetch OCI image and unpack content to ConfigMap
## Problem Statement

Based on values configured by the user proper 

Provide actual implementation of a controller for Kong Plugin installation. Based on configured values in CRD, it should fetch OCI image with the plugin unpack it to secret, and report useful information about the status in `status` field of the defined object. The format of the expected OCI image is 

```dockerfile
FROM scratch
COPY plugin.lua ./
```

such image goes with the below flow

diagram

[Source](https://docs.google.com/document/d/1-gYEaf_31BR_QXmZXNhy0bOrXYJ4X_cXobQ_vbLDIhE/edit#heading=h.dcuuuiouv5ro)

What is in the blue box is in the scope of this issue.

## Proposed Solution

Istio uses the aforementioned [packaging as an image/blob](https://github.com/istio-ecosystem/wasm-extensions/blob/master/doc/how-to-build-oci-images.md).

This is how Istio implements it. It uses [crane library](https://github.com/google/go-containerregistry/blob/main/cmd/crane/doc/crane.md) to deal with them. [See how it looks in their code]([https://github.com/istio/istio/blob/ef8bfe6fae1952787e05a7a7080511995d613075/pkg/wasm/imagefetcher.go) and maybe reuse. 

## Acceptance Criteria
- [ ] adding configuration for plugin leads to creation of a ConfigMap with a plugin and proper status is set in K8s object for available states

programmer04:

It introduces code for fetching and unpacking the content of container images that contain plugins. The expected format that will be documented is 

```dockerfile
FROM scratch

COPY plugin.lua /plugin.lua
```

also basic reconciliation of `KongPluginInstallation` is introduced. A plugin is unpacked to a respective `ConfigMap` and status conditions are set accordingly. 

**Which issue this PR fixes**

Closes https://github.com/Kong/gateway-operator/issues/379

Tests verify the current state, they will evolve with time. Currently, integration tests don't focus on actual behavior, because it's not there yet. For now, it's better to have such a test than nothing.

It's not the final state, first of all, let's get functionality end to end, and next improve it.

I'm happy to make crucial improvements as part of this PR, for more minor/harder I'd rather create new tasks to do as a part of the Kong plugin distribution epic. 

**PR Readiness Checklist**:


api/v1alpha1/kongplugin_installation_types.go
```diff
@@ -31,7 +31,7 @@ func init() {
 //+kubebuilder:subresource:status
 //+kubebuilder:resource:shortName=kpi,categories=kong;all
 //+kubebuilder:subresource:status
-//+kubebuilder:printcolumn:name="Ready",type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].status`
+//+kubebuilder:printcolumn:name="Accepted",description="The Resource is ready",type=string,JSONPath=`.status.conditions[?(@.type=='Accepted')].status`
```
mlavacca:
```suggestion
//+kubebuilder:printcolumn:name="Accepted",description="The Resource is accepted",type=string,JSONPath=`.status.conditions[?(@.type=='Accepted')].status`
```

controller/kongplugininstallation/controller.go
```diff
@@ -0,0 +1,216 @@
...
+					Name:      kpi.Name,
+					Namespace: kpi.Namespace,
+				},
+			})
+		}
+	}
+	return recs
+}
+
+func setStatusConditionFailed(ctx context.Context, client client.Client, kpi *v1alpha1.KongPluginInstallation, msg string) error {
```
mlavacca: Since the whole package has visibility of this function, can we rename it to make clear it is about the `kongPluginInstallation` resource?

controller/kongplugininstallation/controller.go
```diff
@@ -0,0 +1,216 @@
...
+	_, index, found := lo.FindIndexOf(kpi.Status.Conditions, func(c metav1.Condition) bool {
+		return c.Type == string(v1alpha1.KongPluginInstallationConditionStatusAccepted)
+	})
+	if found {
+		kpi.Status.Conditions[index] = status
+	} else {
+		kpi.Status.Conditions = append(kpi.Status.Conditions, status)
+	}
+	return client.Status().Update(ctx, kpi)
+}
```
mlavacca: I think we may want to make this function available to all our controllers, not just to the KPI one. What do you think about moving it to a utility package?

controller/kongplugininstallation/controller.go
```diff
@@ -0,0 +1,216 @@
...
+		cm.Name = kpi.Status.UnderlyingConfigMapName
+		cm.Data = map[string]string{
+			fmt.Sprintf("%s.lua", kpi.Name): string(plugin),
+		}
+		if err := r.Client.Update(ctx, &cm); err != nil {
+			return ctrl.Result{}, err
+		}
+	default:
+		// It should never happen.
+		return ctrl.Result{}, errors.New("unexpected error happened - more than one ConfigMap found")
```
mlavacca, line 153: We have the reduce package (https://pkg.go.dev/github.com/kong/gateway-operator/pkg/utils/kubernetes/reduce). We may want to create a new configmap-related function to use here. WDYT?

controller/kongplugininstallation/controller.go
```diff
@@ -0,0 +1,216 @@
...
+			predicate.Funcs{
+				DeleteFunc: func(e event.DeleteEvent) bool {
+					return true
+				},
+				CreateFunc: func(e event.CreateEvent) bool {
+					return false
+				},
+				UpdateFunc: func(e event.UpdateEvent) bool {
+					return false
+				},
```
mlavacca: Why don't we want to trigger reconciliation for CM update and creation?

controller/kongplugininstallation/controller.go
```diff
@@ -0,0 +1,216 @@
...
+		if err := ctrl.SetControllerReference(&kpi, &cm, r.Scheme); err != nil {
+			return ctrl.Result{}, err
+		}
+		if err := r.Client.Create(ctx, &cm); err != nil {
+			return ctrl.Result{}, err
+		}
+		kpi.Status.UnderlyingConfigMapName = cm.Name
+	case 1:
+		cm = cms[0]
+		cm.Name = kpi.Status.UnderlyingConfigMapName
```
mlavacca: I think we should check the cm name matches the `UnderlyingConfigMapName` instead of enforcing it. A resource name is an immutable field, so it is a bit odd to enforce it in my opinion.

controller/kongplugininstallation/image/image.go
```diff
@@ -0,0 +1,142 @@
...
+					mut.Unlock()
+				}
+				return nil
+			},
+		},
+	}); err != nil {
+		return nil, fmt.Errorf("can't fetch image: %s, because: %w", imageURL, err)
+	}
+	// How to build such image is described in details,
+	// following manual results in the image with exactly one layer that contains a plugin.
```
mlavacca: I'm not sure I understand this comment 🤔 Can you explain it to me?

controller/kongplugininstallation/image/image.go
```diff
@@ -0,0 +1,142 @@
...
+func CredentialsStoreFromString(s string) (credentials.Store, error) {
+	// TODO: Now we create temporary file, which is not great and should be changed,
+	// but it's the only way to use credentials.NewFileStore(...) which robustly
+	// parses config.json (format used by Docker and Kubernetes).
+	tmpFile, err := os.CreateTemp("", "credentials")
+	if err != nil {
+		return nil, fmt.Errorf("failed to create temporary file: %w", err)
+	}
+	defer tmpFile.Close()
+	defer os.Remove(tmpFile.Name())
```
mlavacca: defer is a stack (LIFO). Ee should defer `remove` first, then `close`

modules/cli/cli.go
```diff
@@ -43,6 +43,7 @@ func New(m metadata.Info) *CLI {
 
 	// controllers for specialized APIs and features
 	flagSet.BoolVar(&cfg.AIGatewayControllerEnabled, "enable-controller-aigateway", false, "Enable the AIGateway controller. (Experimental).")
+	flagSet.BoolVar(&cfg.KongPluginInstallationControllerEnabled, "enable-controller-kongplugininstallation", true, "Enable the KongPluginInstallation controller.")
```
mlavacca: Since the `KongPluginInstallation` is an alpha feature (the API is in alpha), we should default such a flag to `false`

test/integration/utils.go
```diff
@@ -175,3 +176,7 @@ func GetVolumeMountsByVolumeName(volumeMounts []corev1.VolumeMount, name string)
 	}
 	return ret
 }
+
+func GetKongPluginImageRegistryCredentialsForTests() string {
+	return os.Getenv("KONG_PLUGIN_IMAGE_REGISTRY_CREDENTIALS")
```
mlavacca, line 181: Can we assign this env var key to a const?

controller/kongplugininstallation/image/image.go
```diff
@@ -0,0 +1,142 @@
...
+	if credentialsStore != nil {
+		credentialFunc = credentials.Credential(credentialsStore)
+	}
+	httpClient := auth.DefaultClient
+	httpClient.Header = map[string][]string{"User-Agent": {metadata.Metadata().UserAgent()}}
+	registry.Client = &auth.Client{
+		Client:     httpClient.Client,
+		Cache:      auth.NewCache(),
+		Credential: credentialFunc,
+	}
```
czeslavo: `auth.DefaultClient` is a pointer to `auth.Client` so setting its `Header` field is ineffective in our case. We should explicitly set the `Header` field below when instantiating `auth.Client`.
```suggestion
	registry.Client = &auth.Client{
		Client:     auth.DefaultClient.Client,
		Header:     map[string][]string{"User-Agent": {metadata.Metadata().UserAgent()}},
		Cache:      auth.NewCache(),
		Credential: credentialFunc,
	}
```

controller/kongplugininstallation/image/image_test.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+}
+
+func TestFetchPluginContent(t *testing.T) {
+	t.Run("invalid image URL", func(t *testing.T) {
+		_, err := image.FetchPluginContent(context.Background(), "foo bar", nil)
+		require.ErrorContains(t, err, "unexpected format of image url: could not parse reference: foo bar")
+	})
+
+	const registryUrl = "northamerica-northeast1-docker.pkg.dev/k8s-team-playground/"
```
czeslavo, line 66: As I understand, this effectively makes this test an integration test (as we're testing against a real registry, with pre-baked images). I think we need to make sure it's properly documented how to access the registry, how to prepare images for testing, etc.

BTW have you considered mocking the registry in tests so that we can inject different images in tests and have the code responsible for creating them close?

controller/kongplugininstallation/image/image.go
```diff
@@ -0,0 +1,142 @@
...
+			}
+		case errors.Is(err, io.EOF):
+			return nil, fmt.Errorf("file %q not found in the image", kongPluginName)
+		case errors.Is(err, io.ErrUnexpectedEOF):
+			return nil, fmt.Errorf("plugin size exceed %d bytes", sizeLimit_1MiB)
+		default:
+			return nil, fmt.Errorf("unexpected error during looking for plugin: %w", err)
+		}
+	}
+}
```
czeslavo: Can we add a unit test for this function to cover the edge/error cases (no `plugin.lua` in the archive, `plugin.lua` bigger than 1MB, etc.)?

controller/kongplugininstallation/image/image.go
programmer04: I can be clear :)

controller/kongplugininstallation/image/image.go
programmer04: nice catch

test/integration/utils.go
programmer04: I'm consistent with other such cases in this repo - e.g. `suite.go`, `cli.go` Do you suggest making it a global const, exported by package utils? 🤔

controller/kongplugininstallation/image/image.go
programmer04: Great suggestion, it helped spot a bug 😁

controller/kongplugininstallation/controller.go
programmer04: done

controller/kongplugininstallation/controller.go
programmer04: I thought about this, but it is used only in the package `kongplugininstallation` and tied to KPI resource and it can't be done as a generic to be universal, so I think we can move it when the need arises

controller/kongplugininstallation/image/image_test.go
programmer04: Yes, I want to document it, but I don't know how it will change with the implementation, so I'd like to do it as the last step. Official documentation will describe in detail how to build a valid image.

About tests, I had the same thoughts. I considered putting those as a part of integration tests, but those are strictly tied to running the actual controller, so it's not the right place.

Maybe I should create a dedicated target for that kind of test?

Running its registry is something that I considered and here it should work nice, but in integration tests with an actual controller and K8s cluster is much, much easier to use remote registries (especially for public and private ones). Even for running a registry for tests in that file the easiest way would be to use a testcontainers (they have dedicated registry features), but still, they need a Docker as a dependency and more code to maintain...

So basically I'm wondering how to do that in the best way ;p

controller/kongplugininstallation/controller.go
programmer04: You're right. It's left when I tried to use immutable ConfigMap, but they [have limitations](https://github.com/kubernetes/website/issues/42359), which makes them hard to use

controller/kongplugininstallation/controller.go
programmer04: Nice catch. It's leftover from trying the usage of immutable ConfigMap to avoid edition. On creation, it triggers duplicated reconciliation and the creation of owned CM is done only by this controller, so we don't need to enqueue KPI reconciliation. So triggering for Update and Delete I think is enough.

programmer04: There is an open question should we expect additional annotation or label for secrets considered by this controller or just checking a type is enough, because we'll only use ti when it's referenced

controller/kongplugininstallation/image/image_test.go
czeslavo: I can imagine that if we leave this as is and we accidentally remove the remote registry, it will be painful to rebuild it manually without any docs. If there will be docs for that (or a script) for bootstrapping a registry with all the images used in tests, I'm fine with going this way. If not, I think it's better to have some code that will bootstrap it locally (e.g. with [testcontainers](https://github.com/testcontainers/testcontainers-go/blob/main/modules/registry/registry.go) as you mentioned).

Regarding a separate target - I think it's fine to run it as a part of the UT suite as long as we do not need auth credentials to pass.

programmer04: I've created https://github.com/Kong/gateway-operator/issues/465 and added it as a part of the https://github.com/Kong/gateway-operator/issues/371 to do it when the final shape of tests will be reached. SO I'll handle it in the future

config/debug/manager_debug.yaml
```diff
@@ -31,6 +31,10 @@ spec:
             - -zap-time-encoding=iso8601
             - -cluster-ca-secret-namespace=kong-system
             - -zap-log-level=debug
+            - -enable-controller-gateway=true
+            - -enable-controller-controlplane=true
```
mlavacca: nit: These are not really needed as enabled by default

config/dev/manager_dev.yaml
```diff
@@ -24,6 +24,7 @@ spec:
             - -zap-devel=true
             - -enable-controller-gateway=true
             - -enable-controller-controlplane=true
```
mlavacca: We can get rid of these as well

controller/kongplugininstallation/controller.go
mlavacca: What about this one? What do you think?

controller/kongplugininstallation/controller.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+	_, index, found := lo.FindIndexOf(kpi.Status.Conditions, func(c metav1.Condition) bool {
+		return c.Type == string(v1alpha1.KongPluginInstallationConditionStatusAccepted)
+	})
+	if found {
+		kpi.Status.Conditions[index] = status
+	} else {
+		kpi.Status.Conditions = append(kpi.Status.Conditions, status)
+	}
+	return client.Status().Update(ctx, kpi)
```
mlavacca, line 222: I think this update will always cause a change in the resource status as we are updating the `lastTransitionTime`. We should check that there are some differences between the old and the new status, instead of updating it regardless.

controller/kongplugininstallation/controller.go
programmer04: Good question. I think that it can only happen when someone manually creates such ConfigMap and sets its owner to KPI, that more than one ConfigMap will be here available... On the other hand how to decide which one I want in such case 🤔

czeslavo (approved):

controller/kongplugininstallation/controller.go
mlavacca: ok, let's defer this one and implement in case we need it :)

mlavacca (approved):

.github/workflows/tests.yaml
```diff
@@ -250,6 +251,7 @@ jobs:
       run: make test.integration
       env:
         KONG_TEST_DISABLE_CERTMANAGER: "true"
+        KONG_PLUGIN_IMAGE_REGISTRY_CREDENTIALS: ${{ secrets.KONG_PLUGIN_IMAGE_REGISTRY_CREDENTIALS }}
```
pmalek, line 254: @programmer04 Can we document this secret somewhere, e.g. in https://github.com/Kong/team-k8s and explain how to obtain these and how/when to rotate them?

programmer04: Yes, I'm going to document everything, when the tests reach the final state

pmalek: Can we have an issue tracking this?

programmer04: The issue for this has been created https://github.com/Kong/gateway-operator/issues/465 during merge of this pr

Issue #142, opened 2024-01-15
DataPlane: support `PodDisruptionBudget`


pmalek:
### Problem statement

Users might want to specify the allowed disruption budget for their `DataPlane` workloads to configure e.g. how many replicas can be down during an upgrade.

### Proposed

Support enabling a spec field in `DataPlane` API which will deploy and managed `PodDisruptionBudget` targeting the `DataPlane` instances.

### Acceptance criteria

- [ ] As a user I can set a `DataPlane` field and expect a `PodDisruptionBudget` to be created and managed for me accordingly
- [ ] A guide on how to set this exists under https://docs.konghq.com/gateway-operator/

sentinelleader:
will this be implemented in 1.3 release ? 

we recycle pods every week (as we recycle nodes every week) and sometimes we see that the aws karpenter can go aggressive due to lack of pdb and they recommend pdb to avoid too mucch disruption

```
kubectl get events -n c1b32c25-8557-410c-9ea9-a3c2ca174835
LAST SEEN   TYPE      REASON                   OBJECT                                                                       MESSAGE
13m         Normal    SuccessfulCreate         replicaset/dataplane-7ae39c6e-3999-4b72-be5c-f996c01c2c3b-5qfvv-7bcf544c8f   Created pod: dataplane-7ae39c6e-3999-4b72-be5c-f996c01c2c3b-5qfvv-7bcf5pd2p6
11m         Normal    SuccessfulCreate         replicaset/dataplane-7ae39c6e-3999-4b72-be5c-f996c01c2c3b-5qfvv-7bcf544c8f   Created pod: dataplane-7ae39c6e-3999-4b72-be5c-f996c01c2c3b-5qfvv-7bcf5kwl49
11m         Normal    Evicted                  pod/dataplane-7ae39c6e-3999-4b72-be5c-f996c01c2c3b-5qfvv-7bcf57rbg6          Evicted pod
11m         Normal    Killing                  pod/dataplane-7ae39c6e-3999-4b72-be5c-f996c01c2c3b-5qfvv-7bcf57rbg6          Stopping container proxy
9m33s       Warning   FailedPreStopHook        pod/dataplane-7ae39c6e-3999-4b72-be5c-f996c01c2c3b-5qfvv-7bcf57rbg6          PreStopHook failed
[...]

PR #441, opened 2024-07-29
feat: add API for DataPlane managed PDB

pmalek:

Introduce API for DataPlane managed PDB.

**Which issue this PR fixes**

Part of #142.

`PodDisruptionBudgetSpec` has been copied from https://pkg.go.dev/k8s.io/api@v0.30.3/policy/v1#PodDisruptionBudgetSpec with `Selector` removed as that is going to be managed by the operator to match `DataPlane`'s Pods.

**PR Readiness Checklist**:


api/v1beta1/dataplane_types.go
```diff
@@ -64,6 +65,75 @@ type DataPlaneOptions struct {
...
+	PodDisruptionBudget *PodDisruptionBudget `json:"podDisruptionBudget,omitempty"`
+}
+
+// PodDisruptionBudget defines the configuration for the PodDisruptionBudget.
+type PodDisruptionBudget struct {
+	// Enabled indicates whether the PodDisruptionBudget should be created.
+	//
+	// +optional
+	// +kubebuilder:default=false
+	Enabled bool `json:"enabled"`
```
czeslavo: @pmalek Why do we need this flag? As `resources.podDisruptionBudget` is optional, cannot we support enabling it simply by making it non-nil?

pmalek: My reasoning for this was to allow users to keep the spec but disable the pdb (for testing purposes or to temporarily suspend).

That use case might be contrived though.

pmalek: On a side note without the `enabled` field how do we consider the pdb enabled? If any of the spec fields is non nil? Or do we change the `spec` to be a pointer and check whether that's non nil?

czeslavo: > My reasoning for this was to allow users to keep the spec but disable the pdb (for testing purposes or to temporarily suspend).

I'd drop it for simplicity and add only if there's a specific request for that. 

> On a side note without the enabled field how do we consider the pdb enabled? If any of the spec fields is non nil? Or do we change the spec to be a pointer and check whether that's non nil?

`DataPlaneResources.PodDisruptionBudget` is a pointer that can be made explicitly optional. If it is not `nil`, that means we should create a PDB based on the spec. Do I miss something?

### PDB disabled (`resources.podDistruptionBudget` pointer is nil)
```yaml
kind: DataPlane
metadata:
  name: dataplane-with-pdb
spec:
  # Do not fill resources.podDisruptionBudget to not create a PDB for DP.
  # resources:
  #   podDisruptionBudget:
  deployment:
    replicas: 3
```

### PDB enabled with an empty spec (apparently it is allowed, I'm not sure what it effectively means if there's no minAvailable/maxUnavailable set)
```yaml
kind: DataPlane
metadata:
  name: dataplane-with-pdb
spec:
  resources:
    podDisruptionBudget: {}
  deployment:
    replicas: 3
```

pmalek: 👍 

Removed `enabled` field and added CEL validation as per the docs: https://kubernetes.io/docs/tasks/run-application/configure-pdb/#specifying-a-poddisruptionbudget

> You can specify only one of maxUnavailable and minAvailable in a single PodDisruptionBudget.

czeslavo (approved):

Issue #442, opened 2024-07-29
Konnect entities: reorganize API groups, add `konnect` API group


pmalek:
### Problem statement

This task tracks the effort of introducing a new API group: `konnect` which will hold all the Konnect related types.

This will be added in https://github.com/Kong/kubernetes-configuration, next to https://github.com/Kong/kubernetes-configuration/tree/main/api/configuration which holds mostly KIC types.

~This will be a ~
- ~Non breaking change for YAML API consumers (through k8s manifests)~
- ~Breaking change for Go API consumers as we'll move the APIs from under https://github.com/Kong/gateway-operator/tree/main/api to `gateway-operator` dedicated API group (1 level below)~
  - ~Since we do not hold Go API to the same standard of non breaking guarantees this is OK.~

### Acceptance criteria

- [ ] `konnect` API group exists ~under https://github.com/Kong/gateway-operator/tree/main/api~ in https://github.com/Kong/kubernetes-configuration/.
- [ ] ~Cloud Gateways team is made aware of this breaking change in Go APIs. There exists a PR in https://github.com/Kong/dataplane-controller/ (which is currently the only internal KGO Go API consumer) which migrates from current API to the new API structure, produced as a result of this issue.~
  - With introduction of this new types in https://github.com/Kong/kubernetes-configuration this does not require a breaking change in the operator API paths.

PR #443, opened 2024-07-29
feat(api): add Konnect API group

pmalek:

- move `api/v1beta1` and `api/v1alpha1` under `api/gateway-operator/`
- add `api/konnect` with `v1alpha1` placeholder

**Which issue this PR fixes**

Part of #442 

This is a breaking change in the Go API but not in YAML manifests as API  group name does not change.

We do not treat the Go API currently as part of the user contract so we do not label this change as breaking.

**PR Readiness Checklist**:


config/crd/bases/konnect.konghq.com_konnectapiauthconfigurations.yaml
```diff
@@ -0,0 +1,171 @@
+---
+apiVersion: apiextensions.k8s.io/v1
```
randmonkey: Looks like no corresponding type definitions for the CRD. Was it generated?

config/crd/bases/konnect.konghq.com_konnectcontrolplanes.yaml
```diff
@@ -0,0 +1,206 @@
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
```
randmonkey: Ditto this.

randmonkey (changes requested):
Looks like no go types for Konnect CRDs are defined in this PR but we have manifests for two CRDs. If we do not want include konnect CRDs in this PR, I think we should remove the manifests.

config/crd/bases/konnect.konghq.com_konnectapiauthconfigurations.yaml
pmalek: Good catch. 3c98f5c5e980db9a2b23a0bc031b80c49faded45 removes this.

config/crd/bases/konnect.konghq.com_konnectcontrolplanes.yaml
pmalek: Good catch. 3c98f5c5e980db9a2b23a0bc031b80c49faded45 removes this.

randmonkey (approved):

CHANGELOG.md
```diff
@@ -33,6 +33,14 @@
   plugins distributed as container images.
   [400](https://github.com/Kong/gateway-operator/pull/400)
 
+### Changes
```
czeslavo, line 36: ~I think we could be explicit and call these changes breaking as they technically are, right?~
Nevermind, I missed "We do not treat the Go API currently as part of the user contract so we do not label this change as breaking." in the description. 😅 

```suggestion
### Breaking changes
```

czeslavo (approved):

CHANGELOG.md
pmalek: Yes and no. As stated in the PR description we do not consider Go API as part of the user contract. This is not mentioned anywhere though.

The manifests / YAML API is what we keep up to Kubernetes' API versioning standard: https://kubernetes.io/docs/reference/using-api/#api-versioning which this PR does not change.

I can be persuaded to change this to "breaking" though. My priority here is to give users a good hint about what's coming in the upcoming release. Do you feel putting this in `Changes` section does not fit well in that context?

czeslavo: I have no hard opinion on that and leaving it as "changes" works as well. 

I initially thought that if a technically public Go API is changed, we might communicate it as breaking as we never know who relies on that. If we do not want to make the Go API public, we could a) make it internal or b) explicitly state somewhere in the docs we do not guarantee its stability in any way.

mlavacca (changes requested):
I thought we were going to add the new konnect CRDs in kubernetes-configuration rather than here. We agreed on putting off the migration of the existing CRDs to the new repo, but for what concerns the new CRDs, why are we splitting them up between here and kubernetes-config?

pmalek:
> I thought we were going to add the new konnect CRDs in kubernetes-configuration rather than here. We agreed on putting off the migration of the existing CRDs to the new repo, but for what concerns the new CRDs, why are we splitting them up between here and kubernetes-config?

There are 2 "types" of CRDs we're talking about here: 

- Kong configuration CRDs which are already in https://github.com/Kong/kubernetes-configuration/tree/main/api/configuration (`KongConsumer`, `KongConsumerGroup`, `KongPlugin` etc) and those that might still end up there or are being actively worked on (`KongService` or `KongRoute`)
- Top level Konnect entity types like `KonnectControlPlane`. This is the type of CRDs that will land in the newly added API group.

The former make sense in both KGO and KIC context. The latter are only used by the operator and are only used in the Konnect context.

---

From what I understand @mlavacca, you're suggesting to introduce the latter also in https://github.com/Kong/kubernetes-configuration under `konnect` API group, leaving KGO API unchanged. 

I'm wondering if there are any benefits to this approach. I'm curious about your thoughts on this. Given that we'd only use those CRDs in the operator I'm not sure there's much value in putting them outside of this repo. 🤔

CHANGELOG.md
pmalek: @czeslavo PTAL at #445.

mlavacca:
> > I thought we were going to add the new konnect CRDs in kubernetes-configuration rather than here. We agreed on putting off the migration of the existing CRDs to the new repo, but for what concerns the new CRDs, why are we splitting them up between here and kubernetes-config?
> 
> There are 2 "types" of CRDs we're talking about here:
> 
> * Kong configuration CRDs which are already in https://github.com/Kong/kubernetes-configuration/tree/main/api/configuration (`KongConsumer`, `KongConsumerGroup`, `KongPlugin` etc) and those that might still end up there or are being actively worked on (`KongService` or `KongRoute`)
> * Top level Konnect entity types like `KonnectControlPlane`. This is the type of CRDs that will land in the newly added API group.
> 
> The former make sense in both KGO and KIC context. The latter are only used by the operator and are only used in the Konnect context.
> 
> From what I understand @mlavacca, you're suggesting to introduce the latter also in https://github.com/Kong/kubernetes-configuration under `konnect` API group, leaving KGO API unchanged.
> 
> I'm wondering if there are any benefits to this approach. I'm curious about your thoughts on this. Given that we'd only use those CRDs in the operator I'm not sure there's much value in putting them outside of this repo. 🤔

My point is that since we are making the effort to group the CRDs in a unique place, it could make sense to treat this new place as where "Kong CRDs live". No matters if they are shared among KIC or KGO, or just used by KGO, I think that having a centralized place where our APIs live is a good approach, rather than having them scattered across multiple repositories.

Does it make sense to you?

mlavacca:
One additional point to the discussion is there we will inevitably split the konnect CRDs between k8s-conf and kgo if we choose the approach of putting the `KonnectControlPlanes` here: the kubernetes configuration objects will need the `konnectAPIAuthConfiguration` API as it is used in their status, and to avoid circular dependencies we'll have to put such resources in the same repo. We'd end up having the Kong CRDs scattered across multiple repos, AND the same API group scattered across multiple repositories, which is something I think we really need to stay away from.

pmalek:
> One additional point to the discussion is there we will inevitably split the konnect CRDs between k8s-conf and kgo if we choose the approach of putting the `KonnectControlPlanes` here: the kubernetes configuration objects will need the `konnectAPIAuthConfiguration` API as it is used in their status, and to avoid circular dependencies we'll have to put such resources in the same repo. We'd end up having the Kong CRDs scattered across multiple repos, AND the same API group scattered across multiple repositories, which is something I think we really need to stay away from.

I was almost going to nod on this but then I realized that with the approach proposed by @czeslavo in the design doc (https://docs.google.com/document/d/1TxW2ovqPRm-w1td1e_V0CyaET4lpiEBCJ7hiYMZJqng/edit?disco=AAABS5JK2YM) we could drop the idea of having the `KonnectAuthConfiguration` reference in resources living within the CP because the latter would contain the auth config anyway. This way we don't need the `KonnectAuthConfiguration` type information.

Would you agree @mlavacca?

I guess at this point, it's a matter of where we'd **prefer** this CRDs to live. Let's discuss this on today's epic review.

pmalek:
After a discussion with @mlavacca and @czeslavo we've decided that it will be better to place all new Konnect related types and CRDs into https://github.com/Kong/kubernetes-configuration.

With this, https://github.com/Kong/kubernetes-configuration/pull/12 supersedes this PR.

PR #446, opened 2024-07-30
feat(konnect): add generic reconciler scaffolding

### LINKED ISSUE #437: Konnect entities: implement a generic reconciler scaffolding
### Problem statement

Konnect entities #370 will require reconciler(s) to react to changes to CRDs in the cluster.

This issue tracks the effort of introducing a generic reconciler scaffolding which will be used by most (if not all) of the Konnect entity types.

### Acceptance criteria

- [x] A reconciler scaffolding exists and can be used by Konnect related CRD types defined in #370

pmalek:

This PR adds generic reconciler scaffolding.

**Which issue this PR fixes**

Part of #437 

**PR Readiness Checklist**:


controller/konnect/conditions.go
```diff
@@ -0,0 +1,42 @@
+package konnect
+
+// TODO(pmalek): move this to Konnect API directory so that it's part of the API contract.
```
mlavacca, line 3: Can we open an issue in https://github.com/Kong/kubernetes-configuration to keep track of this?

controller/konnect/constraints.go
```diff
@@ -0,0 +1,42 @@
...
+	configurationv1alpha1 "github.com/kong/kubernetes-configuration/api/configuration/v1alpha1"
+)
+
+// SupportedKonnectEntityType is an interface that all Konnect entity types
+// must implement.
+type SupportedKonnectEntityType interface {
+	configurationv1alpha1.KongService |
+		configurationv1alpha1.KongRoute |
+		configurationv1.KongConsumer
+	// TODO: add other types
```
mlavacca, line 20: What about `ConsumerGroup`? Is there anything that prevents us to already add it here?

controller/konnect/constraints.go
```diff
@@ -0,0 +1,42 @@
...
+// Separating this from SupportedKonnectEntityType allows us to use EntityType
+// where client.Object is required, since it embeds client.Object and uses pointer
+// to refer to the SupportedKonnectEntityType.
+type EntityType[
+	T SupportedKonnectEntityType,
+] interface {
+	*T
+
+	// Kubernetes Object methods
+
```
mlavacca:
```suggestion
```

controller/konnect/constraints.go
```diff
@@ -0,0 +1,42 @@
...
+	GetObjectMeta() metav1.Object
+	client.Object
+
+	// Added methods
```
mlavacca: Can we set a more meaningful comment here?

controller/konnect/reconciler_generic.go
```diff
@@ -0,0 +1,263 @@
...
+			ent.GetGeneration(),
+		),
+		ent,
+	)
+	if err := r.Client.Status().Update(ctx, ent); err != nil {
+		if k8serrors.IsConflict(err) {
+			return ctrl.Result{Requeue: true}, nil
+		}
+		return ctrl.Result{}, fmt.Errorf("failed to update status with %s condition: %w", KonnectEntityAPIAuthConfigurationRefValidConditionType, err)
+	}
```
mlavacca: We should return in case the resource has been updated. The update will trigger another reconciliation loop.

controller/konnect/reconciler_generic.go
```diff
@@ -0,0 +1,263 @@
...
+
+		return ctrl.Result{}, nil
+	}
+
+	// TODO: relying on status ID is OK but we need to rethink this because
+	// we're using a cached client so after creating the resource on Konnect it might
+	// happen that we've just created the resource but the status ID is not there yet.
+	//
+	// We should look at the "expectations" for this:
+	// https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/controller_utils.go
```
mlavacca, line 235: I think this is fine. We are returning after setting the ID, and the status update is going to re-trigger another reconciliation loop. In the new loop, we are guaranteed the status has been properly updated, as that specific update is the one that triggered the new loop. (Unless other requests are already in the queue maybe?)

controller/konnect/conditions.go
pmalek: 6955959821e472ccfb5f117c655a851f6fb23006

controller/konnect/constraints.go
pmalek: `GetTypeName()` method is missing.

https://github.com/Kong/kubernetes-configuration/pull/15

controller/konnect/constraints.go
pmalek: Done. PTAL.

controller/konnect/reconciler_generic.go
pmalek: This is now handled slightly better with `updateStatusWithCondition` PTAL

https://github.com/Kong/gateway-operator/pull/446/files#diff-5b6aadc0fe9b4d0ba8cfd31756eef41ffdcd65687cd077d3b8d427294dcc81b5R300-R331

controller/konnect/constraints.go
```diff
@@ -0,0 +1,44 @@
...
+	T SupportedKonnectEntityType,
+] interface {
+	*T
+
+	// Kubernetes Object methods
+	GetObjectMeta() metav1.Object
+	client.Object
+
+	// Additional method which are used in reconciling Konnect entities.
+
```
pmalek:
```suggestion
```

controller/konnect/constraints.go
```diff
@@ -0,0 +1,43 @@
...
+type EntityType[
+	T SupportedKonnectEntityType,
+] interface {
+	*T
+
+	// Kubernetes Object methods
+	GetObjectMeta() metav1.Object
+	client.Object
+
+	// Additional method which are used in reconciling Konnect entities.
```
pmalek:
```suggestion
	// Additional methods which are used in reconciling Konnect entities.
```

controller/konnect/reconciler_generic.go
pmalek: Imagine 2 subsequent reconciliations. The first one creates a Konnect resource because the Konnect ID is empty. The second one kicks in: the status should be set but since we're working with a cached client it might not be there yet. This would cause lines below to attempt creating the entity on Konnect again.

That's what this comment tries to cover.

mlavacca: yes, so this is the case I mentioned with 
> Unless other requests are already in the queue maybe?

?

It should be a harmless problem, as the 2nd creation fails and an error is returned. Bad UX though

controller/konnect/constraints.go
mlavacca: merged. Do you want to bump the `go.mod` and add ConsumerGroup as well or do you prefer doing it in another PR? I'm fine with both

pmalek: Let's do it separately. I also forgot about condition related funcs...

controller/konnect/reconciler_generic.go
pmalek: > Bad UX though

Yes. Additionally this will set the status accordingly, for a brief while, but still.

For this reason I'd prefer to leave the comment in. Does that sound good?

mlavacca (approved):

PR #453, opened 2024-07-31
feat: `DataplaneKonnectExtension` CRD

### LINKED ISSUE #203: Native support for Konnect integration in DataPlane
## Problem Statement

Currently, to deploy `DataPlane`s that will sync with Konnect Control Plane, one has to define a very verbose PodTemplateSpec patch that will define all the required environment variables Kong Gateway needs for that purpose:

```yaml
kind: DataPlane
metadata:
  name: dataplane-example
  namespace: kong
spec:
  deployment:
    podTemplateSpec:
      spec:
        containers:
        - name: proxy
          image: kong/kong-gateway:3.6.1.3
          env:
            - name: KONG_ROLE
              value: data_plane
            - name: KONG_DATABASE
              value: "off"
            - name: KONG_CLUSTER_MTLS
              value: pki
            - name: KONG_CLUSTER_CONTROL_PLANE
              value: YOUR_CP_ID.us.cp0.konghq.com:443
            - name: KONG_CLUSTER_SERVER_NAME
              value: YOUR_CP_ID.us.cp0.konghq.com
            - name: KONG_CLUSTER_TELEMETRY_ENDPOINT
              value: YOUR_CP_ID.us.tp0.konghq.com:443
            - name: KONG_CLUSTER_TELEMETRY_SERVER_NAME
              value: YOUR_CP_ID.us.tp0.konghq.com
            - name: KONG_CLUSTER_CERT
              value: /etc/secrets/kong-cluster-cert/tls.crt
            - name: KONG_CLUSTER_CERT_KEY
              value: /etc/secrets/kong-cluster-cert/tls.key
            - name: KONG_LUA_SSL_TRUSTED_CERTIFICATE
              value: system
            - name: KONG_KONNECT_MODE
              value: "on"
            - name: KONG_VITALS
              value: "off"
          volumeMounts:
            - name: cluster-certificate
              mountPath: /var/cluster-certificate
            - name: kong-cluster-cert
              mountPath: /etc/secrets/kong-cluster-cert/
              readOnly: true
        volumes:
          - name: cluster-certificate
          - name: kong-cluster-cert
            secret:
              secretName: kong-cluster-cert
              defaultMode: 420
``` 

It's not the best user experience as there's a lot of repetition and it's easy to make a mistake.

We should make DataPlane natively integrated with Konnect by extending the CRD with a Konnect-specific spec section.

## Proposed Solution

Extend `DataPlaneOptions` with `Konnect` section as below:

```go
// DataPlaneOptions defines the information specifically needed to
// deploy the DataPlane.
type DataPlaneOptions struct {
        // ...

	// +optional
	Konnect *KonnectOptions `json:"konnect,omitempty"`
}

// DataPlaneKonnectKind defines the kind/role of the DataPlane deployment.
type DataPlaneKonnectKind string

const (
	ManagedByCloudControlPlaneDataPlaneKonnectKind DataPlaneKonnectKind = "managed-by-cloud-control-plane"

	// Note: just for future compatibility if we decide to implement Gateway (KIC + DPs) with Konnect integration as well.
	ManagedByIngressControllerPlaneDataPlaneKonnectKind = "managed-by-ingress-controller"
)

type KonnectOptions struct {
	// Kind defines the kind of the DataPlane deployment, which determines how it will be configured.
	Kind DataPlaneKonnectKind `json:"kind"`

	// ControlPlaneID is the identifier of the Konnect Control Plane.
	ControlPlaneID string `json:"controlPlaneID"`

	// ControlPlaneRegion is the region of the Konnect Control Plane.
	// If not set, 'us' is used as the default region.
	// +optional
	ControlPlaneRegion *string `json:"controlPlaneRegion,omitempty"`

	// ClusterCertificate is a name of the Secret containing the Konnect Control Plane's cluster certificate.
	ClusterCertificate string `json:"clusterCertificate"`
}
```

This will reduce the repetition and allow rigid validation of the input based on the defined schema. 

## Acceptance Criteria
- [ ] As a Konnect-hosted Control Plane user who wants to deploy DataPlanes in Kubernetes using KGO, I'm provided with an API that will allow me to configure those with minimal information provided (CP ID, cluster certificate secret name), so I do not have to know what exact environment variables need to be adjusted in their Deployments

mlavacca:

**Which issue this PR fixes**

Part of #203.

The original issue is about adding a new konnect-related set of fields directly in the `DataPlane` API. I think we should be very cautious about this approach because KGO is an open-source project that aims to be used by Kong and Konnect customers AND community users. Including SaaS-related fields directly in the API would negatively affect the API itself form the community point of view. Hence this new extension CRD that with an approach very similar to the `ControlPlaneMetricsExtension` is intended to be used by the DataPlane controller to customize the DataPlane deployment with konnect-related configuration.

**PR Readiness Checklist**:


api/v1alpha1/dataplane_konnect_extension_types.go
```diff
@@ -0,0 +1,74 @@
...
+	SchemeBuilder.Register(&DataPlaneKonnectExtension{}, &DataPlaneKonnectExtensionList{})
+}
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:categories=kong;all
+// +kubebuilder:subresource:status
+
+// DataPlane is the Schema for the dataplanes API
```
pmalek: Apart from this being a copy paste can we add here a comment explaining how this would work? What does this extension do?

api/v1alpha1/dataplane_konnect_extension_types.go
```diff
@@ -0,0 +1,74 @@
...
+// +kubebuilder:object:root=true
+
+// DataPlaneList contains a list of DataPlane
+type DataPlaneKonnectExtensionList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []DataPlaneKonnectExtension `json:"items"`
+}
+
+// DataPlaneSpec defines the desired state of DataPlane
```
pmalek: Ditto, copy paste.

api/v1alpha1/dataplane_konnect_extension_types.go
```diff
@@ -0,0 +1,74 @@
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   DataPlaneKonnectExtensionSpec   `json:"spec,omitempty"`
+	Status DataPlaneKonnectExtensionStatus `json:"status,omitempty"`
+}
+
+// +kubebuilder:object:root=true
+
+// DataPlaneList contains a list of DataPlane
```
pmalek: Ditto, copy paste.

api/v1beta1/gatewayconfiguration_types.go
```diff
@@ -18,16 +18,19 @@ package v1beta1
...
 
-//+genclient
-//+kubebuilder:object:root=true
-//+kubebuilder:subresource:status
-//+kubebuilder:resource:shortName=kogc,categories=kong;all
+// +genclient
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:shortName=kogc,categories=kong;all
+// +kubebuilder:validation:XValidation:message="Extension not allowed for DataPlane config options",rule="has(self.spec.dataPlaneOptions.extensions) ? has(self.spec.dataPlaneOptions.extensions) && self.spec.dataPlaneOptions.extensions.all(e, e.group == 'gateway-operator.konghq.com' && e.kind == 'DataPlaneKonnectExtension') : true"
```
pmalek: I believe the second `has` is not necessary right? 

```suggestion
// +kubebuilder:validation:XValidation:message="Extension not allowed for DataPlane config options",rule="has(self.spec.dataPlaneOptions.extensions) ?  self.spec.dataPlaneOptions.extensions.all(e, e.group == 'gateway-operator.konghq.com' && e.kind == 'DataPlaneKonnectExtension') : true"
```

api/v1beta1/dataplane_types.go
```diff
@@ -32,6 +34,7 @@ func init() {
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
 // +kubebuilder:resource:shortName=kodp,categories=kong;all
+// +kubebuilder:validation:XValidation:message="Extension not allowed for DataPlane",rule="has(self.spec.extensions) ? has(self.spec.extensions) && self.spec.extensions.all(e, e.group == 'gateway-operator.konghq.com' && e.kind == 'DataPlaneKonnectExtension') : true"
```
pmalek:
```suggestion
// +kubebuilder:validation:XValidation:message="Extension not allowed for DataPlane",rule="has(self.spec.extensions) ? self.spec.extensions.all(e, e.group == 'gateway-operator.konghq.com' && e.kind == 'DataPlaneKonnectExtension') : true"
```

pmalek (commented):
Can we add a sample manifest showing how this can be used?

I believe also that further comment in the PR description about this CRD would be beneficial to clue in potential reviewers.

api/v1alpha1/dataplane_konnect_extension_types.go
mlavacca: I added a more descriptive comment

mlavacca:
> Can we add a sample manifest showing how this can be used?
> 
> I believe also that further comment in the PR description about this CRD would be beneficial to clue in potential reviewers.

Yep, I added an example under `config/samples` and updated the PR description to make it clearer

api/v1alpha1/dataplane_konnect_extension_types.go
```diff
@@ -0,0 +1,81 @@
...
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []DataPlaneKonnectExtension `json:"items"`
+}
+
+// DataPlaneKonnectExtensionSpec defines the desired state of DataPlaneKonnectExtension.
+type DataPlaneKonnectExtensionSpec struct {
+	// ControlPlaneID is the identifier of the Konnect Control Plane.
+	// +kubebuilder:validation:Required
+	ControlPlaneID string `json:"controlPlaneID"`
```
pmalek: Given that we're working on `KonnectControlPlane`, do you think it would make sense to pause for a bit with this approach and rework it, so that it supports different types of CP, e.g. instead of referencing by ID we could also reference by NN, thus having all the resources managed within the k8s cluster. 

WDYT?

mlavacca: I think this is a good point. I'll capture it into the design doc I'm writing 👍

api/v1alpha1/dataplane_konnect_extension_types.go
```diff
@@ -0,0 +1,81 @@
...
+
+	// ControlPlaneRegion is the region of the Konnect Control Plane.
+	// If not set, 'us' is used as the default region.
+	// +optional
```
randmonkey: How do we convert `ControlPlaneRegion` to API server URL here? Also, the `validation` seems to limit the extension of regions (for example, if `au` is available, customers has to update CRD for using `au`.)
And as @pmalek said in previous comment, this can also be retrieved from `KonnectControlPlane` reference, so a `default` may not be required.

mlavacca: The path is built in the following way:
`<CONTROL-PLANE-ID>.<REGION>.tp0|cp0.<URL(tech.konghq.com)>`
- `CONTROL-PLANE-ID` can be either provided or fetched from the CP NamespacedRef, as I just implemented after Patryk's suggestion
- `REGION` has to provided. It could be fetched from the APIAuthConfiguration resource referenced by the CP as it is part of the [ServerURL](https://github.com/Kong/kubernetes-configuration/blob/ac4e44233638b368f3b979c29046519bb2c5f760/config/samples/konnect_controlplane.yaml#L38-L39) section. However, In case the controlplane is referenced via `KonnectID`, we do need to have a way to specify it in the API itself
- `cp0` and `tp0` are hardcoded values, that have to be set depending on the env var to configure
- `URL` has to be specified in this API.

mlavacca: I removed the enum on the region, I agree we may want to have it unbound

CHANGELOG.md
```diff
@@ -53,8 +53,10 @@
 - Added command line flags to configure the certificate generator job's images.
   [#516](https://github.com/Kong/gateway-operator/pull/516)
 - Add `KongPluginBinding` reconciler for Konnect Plugins.
-  [#513](https://github.com/Kong/gateway-operator/pull/513)
-  [#535](https://github.com/Kong/gateway-operator/pull/535)
+  [#513](https://github.com/Kong/gateway-operator/pull/513), [#535](https://github.com/Kong/gateway-operator/pull/535)
```
randmonkey, line 56: Looks like the item 
```
[#535](https://github.com/Kong/gateway-operator/pull/535)
```
is lost in resolving conflicts.

api/v1alpha1/dataplane_konnect_extension_types.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+// DataPlaneKonnectExtensionSpec defines the desired state of DataPlaneKonnectExtension.
+type DataPlaneKonnectExtensionSpec struct {
+	// ControlPlaneRef is a reference to a ControlPlane this DataPlaneKonnectExtension is associated with.
+	// +kubebuilder:validation:Required
+	ControlPlaneRef configurationv1alpha1.ControlPlaneRef `json:"controlPlaneRef"`
+
+	// ControlPlaneRegion is the region of the Konnect Control Plane.
+	// ++kubebuilder:validation:Required
```
randmonkey: One more `+` here?

api/v1beta1/dataplane_types.go
```diff
@@ -32,6 +34,7 @@ func init() {
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
 // +kubebuilder:resource:shortName=kodp,categories=kong;all
+// +kubebuilder:validation:XValidation:message="Extension not allowed for DataPlane",rule="has(self.spec.extensions) ? self.spec.extensions.all(e, e.group == 'gateway-operator.konghq.com' && e.kind == 'DataPlaneKonnectExtension') : true"
```
randmonkey, line 37: Should the message be `Extension other then Konnect not allowed for DataPlane` here?

api/v1beta1/gatewayconfiguration_types.go
```diff
@@ -18,16 +18,19 @@ package v1beta1
...
 
-//+genclient
-//+kubebuilder:object:root=true
-//+kubebuilder:subresource:status
-//+kubebuilder:resource:shortName=kogc,categories=kong;all
+// +genclient
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:shortName=kogc,categories=kong;all
+// +kubebuilder:validation:XValidation:message="Extension not allowed for DataPlane config options",rule="has(self.spec.dataPlaneOptions.extensions) ? self.spec.dataPlaneOptions.extensions.all(e, e.group == 'gateway-operator.konghq.com' && e.kind == 'DataPlaneKonnectExtension') : true"
```
randmonkey, line 33: ditto here.

randmonkey (commented):
Generally LGTM, only some minor comments.

api/v1alpha1/dataplane_konnect_extension_types.go
```diff
@@ -0,0 +1,86 @@
...
+	ControlPlaneRegion string `json:"controlPlaneRegion"`
+
+	// ServerURL is the URL of the Konnect server.
+	// +kubebuilder:validation:Required
```
czeslavo: It might be worth adding an example value via `+kubebuilder:example:=` marker to make it clear what we're expecting here.

api/v1alpha1/dataplane_konnect_extension_types.go
```diff
@@ -0,0 +1,86 @@
...
+
+// DataPlaneKonnectExtensionSpec defines the desired state of DataPlaneKonnectExtension.
+type DataPlaneKonnectExtensionSpec struct {
+	// ControlPlaneRef is a reference to a ControlPlane this DataPlaneKonnectExtension is associated with.
+	// +kubebuilder:validation:Required
+	ControlPlaneRef configurationv1alpha1.ControlPlaneRef `json:"controlPlaneRef"`
+
+	// ControlPlaneRegion is the region of the Konnect Control Plane.
+	// ++kubebuilder:validation:Required
+	ControlPlaneRegion string `json:"controlPlaneRegion"`
```
czeslavo, line 65: Ditto - `+kubebuilder:example:=` could make sense here.

api/v1alpha1/dataplane_konnect_extension_types.go
czeslavo: Shouldn't we make this optional and add a CEL rule depending on the type of the CP reference? This should only be required if the CP is referred by an ID, no? I understand we can infer the region from the `KonnectGatewayControlPlane` URL.

api/v1beta1/dataplane_types.go
mlavacca: I'd keep this as we can add other extensions in the future and don't bother ourselves with changing this error message.

randmonkey (approved):

api/v1alpha1/dataplane_konnect_extension_types.go
mlavacca: We could infer the region from the `APIAuth` referenced by the `KonnectGatewayControlPlane`. This means we have to fetch the ControlPlane, then fetch the referenced APIAuth, and parse the serverURL to get the region. I think it's better to mark it as required here.

api/v1alpha1/dataplane_konnect_extension_types.go
mlavacca: I've updated this field to be a hostname. I added validation and an example: cc2851d28ed110bec1d3928a566ff1f734c6e909

api/v1alpha1/dataplane_konnect_extension_types.go
czeslavo: I think that fetching `KonnectGatewayControlPlane` should be enough as it has the `ServerURL` in status. Anyway, it's not that critical - we can always loosen this requirement later 👍

czeslavo (approved):

api/v1alpha1/dataplane_konnect_extension_types.go
```diff
@@ -0,0 +1,98 @@
...
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []DataPlaneKonnectExtension `json:"items"`
+}
+
+// DataPlaneKonnectExtensionSpec defines the desired state of DataPlaneKonnectExtension.
+type DataPlaneKonnectExtensionSpec struct {
+	// ControlPlaneRef is a reference to a ControlPlane this DataPlaneKonnectExtension is associated with.
+	// +kubebuilder:validation:Required
+	ControlPlaneRef configurationv1alpha1.ControlPlaneRef `json:"controlPlaneRef"`
```
pmalek, line 59: Just a note based on a sync with @mlavacca: when we get to implement the konnectNamespacedRef we should make the CP ref immutable to prevent going from 1 type of CP ref to another (unless that gets implemented)

PR #454, opened 2024-07-31
fix: requeue instead of error when finalizer removal yields a conflict

pmalek:

Observed in one of the CI test runs https://github.com/Kong/gateway-operator/actions/runs/10148426698/job/28061112791:

```
2024-07-29T16:57:20Z	ERROR	Reconciler error	{"controller": "gateway", "controllerGroup": "gateway.networking.k8s.io", "controllerKind": "Gateway", "Gateway": {"name":"f8518cc4-d6c5-4365-bcfa-5105d5d5b009","namespace":"cc7f49ee-b420-4e22-ad0d-7f8c3e0b7894"}, "namespace": "cc7f49ee-b420-4e22-ad0d-7f8c3e0b7894", "name": "f8518cc4-d6c5-4365-bcfa-5105d5d5b009", "reconcileID": "f141fe3c-f044-43a1-adf0-a9fe2b4aae1a", "error": "failed updating Gateway's finalizers: Operation cannot be fulfilled on gateways.gateway.networking.k8s.io \"f8518cc4-d6c5-4365-bcfa-5105d5d5b009\": the object has been modified; please apply your changes to the latest version and try again"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.18.4/pkg/internal/controller/controller.go:324
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.18.4/pkg/internal/controller/controller.go:261
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
	/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.18.4/pkg/internal/controller/controller.go:222
```

This should be retried without an error so that users do not need to be confused about a false positive error being reported in the logs (and also controller-runtime metrics).

**PR Readiness Checklist**:

### DIFF
--- a/controller/gateway/controller.go
+++ b/controller/gateway/controller.go
@@ -116,9 +116,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	}
 
 	log.Trace(logger, "managing cleanup for gateway resource", gateway)
-	cleanupUnderway, result, err := r.cleanup(ctx, logger, &gateway)
-	if cleanupUnderway {
+	shouldReturnEarly, result, err := r.cleanup(ctx, logger, &gateway)
+	if err != nil || result.Requeue {
 		return result, err
+	} else if shouldReturnEarly {
+		return ctrl.Result{}, nil
 	}
 
 	log.Trace(logger, "managing the gateway resource finalizers", gateway)
--- a/controller/gateway/controller.go
+++ b/controller/gateway/controller.go
@@ -116,8 +116,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	}
 
 	log.Trace(logger, "managing cleanup for gateway resource", gateway)
-	shouldReturnEarly, result, err := r.cleanup(ctx, logger, &gateway)
-	if err != nil || result.Requeue {
+	if shouldReturnEarly, result, err := r.cleanup(ctx, logger, &gateway); err != nil || result.Requeue {
 		return result, err
 	} else if shouldReturnEarly {
 		return ctrl.Result{}, nil
--- a/controller/gateway/controller_cleanup.go
+++ b/controller/gateway/controller_cleanup.go
@@ -31,7 +31,7 @@ func (r *Reconciler) cleanup(
 	logger logr.Logger,
 	gateway *gatewayv1.Gateway,
 ) (
-	bool, // whether or not cleanup is being performed
+	bool, // Whether the caller should return early from the reconciliation loop.
 	ctrl.Result,
 	error,
 ) {
@@ -42,7 +42,7 @@ func (r *Reconciler) cleanup(
 
 	if gateway.DeletionTimestamp.After(time.Now()) {
 		log.Debug(logger, "gateway deletion still under grace period", gateway)
-		return true, ctrl.Result{
+		return false, ctrl.Result{
 			Requeue:      true,
 			RequeueAfter: time.Until(gateway.DeletionTimestamp.Time),
 		}, nil
@@ -54,15 +54,16 @@ func (r *Reconciler) cleanup(
 	// for cleaning up owned controlplanes when they disappeared.
 	controlplanes, err := gatewayutils.ListControlPlanesForGateway(ctx, r.Client, gateway)
 	if err != nil {
-		return true, ctrl.Result{}, err
+		return false, ctrl.Result{}, err
 	}
 	if len(controlplanes) > 0 {
 		deletions, err := r.ensureOwnedControlPlanesDeleted(ctx, gateway)
 		if err != nil {
-			return true, ctrl.Result{}, err
+			return false, ctrl.Result{}, err
 		}
 		if deletions {
 			log.Debug(logger, "deleted owned controlplanes", gateway)
+			// Return early from reconciliation, deletion will trigger a new reconcile.
 			return true, ctrl.Result{}, err
 		}
 	} else {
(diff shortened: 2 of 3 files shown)


controller/gateway/controller.go
```diff
@@ -116,9 +116,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	}
 
 	log.Trace(logger, "managing cleanup for gateway resource", gateway)
-	cleanupUnderway, result, err := r.cleanup(ctx, logger, &gateway)
-	if cleanupUnderway {
+	shouldReturnEarly, result, err := r.cleanup(ctx, logger, &gateway)
+	if err != nil || result.Requeue {
```
randmonkey: Is `else if` required here?
Should this be clearer?
```golang
shouldReturnEarly, result, err := r.cleanup(ctx, logger, &gateway)
	if err != nil || result.Requeue {
		return result, err
	}  
        if shouldReturnEarly {
		return ctrl.Result{}, nil
	}
```

randmonkey (commented):
In many branches of `cleanup` it returns `false` on `shouldReturnEarly`  but in the previous code the cases returns `true` on `cleanupUnderway`. This leeds to behavior change in `Reconcile` (whether to return early in reconcile). Is this intended?

controller/gateway/controller.go
pmalek: It's not required. If you feel that this is more readable then let's go ahead with it.

pmalek:
> In many branches of `cleanup` it returns `false` on `shouldReturnEarly` but in the previous code the cases returns `true` on `cleanupUnderway`. This leeds to behavior change in `Reconcile` (whether to return early in reconcile). Is this intended?

The behavior doesn't change because in those code paths the `err` is not `nil` (or requeue is set) so we return anyway and those will trigger another reconciliation.

The `bool` return value is required when `err == nil` or `requeue` is not set.

Does that make sense?

randmonkey:
> > In many branches of `cleanup` it returns `false` on `shouldReturnEarly` but in the previous code the cases returns `true` on `cleanupUnderway`. This leeds to behavior change in `Reconcile` (whether to return early in reconcile). Is this intended?
> 
> The behavior doesn't change because in those code paths the `err` is not `nil` (or requeue is set) so we return anyway and those will trigger another reconciliation.
> 
> The `bool` return value is required when `err == nil` or `requeue` is not set.
> 
> Does that make sense?

Got that. Indeed it does not change the behavior, I think it could make more sense to explain the meanings and affects on reconcile loops of the return values.

controller/gateway/controller_cleanup.go
```diff
@@ -54,68 +54,75 @@ func (r *Reconciler) cleanup(
...
 	if len(controlplanes) > 0 {
 		deletions, err := r.ensureOwnedControlPlanesDeleted(ctx, gateway)
 		if err != nil {
-			return true, ctrl.Result{}, err
+			return false, ctrl.Result{}, err
 		}
 		if deletions {
 			log.Debug(logger, "deleted owned controlplanes", gateway)
+			// Return early from reconciliation, deletion will trigger a new reconcile.
 			return true, ctrl.Result{}, err
```
mlavacca: I know this was not part of your changes, but since we are here, can you return `nil` instead of the error? err here is always nil and it would increase the code's readability.

controller/gateway/controller_cleanup.go
```diff
@@ -140,13 +149,14 @@ func handleGatewayFinalizerPatchOrUpdateError(err error, gateway *gatewayv1.Gate
 		return ctrl.Result{}, nil
 	}
 
-	// If the Gateway is not found, then requeue without an error.
-	if k8serrors.IsNotFound(err) {
+	// If the Gateway is not found or there's a conflict patching, then requeue without an error.
+	if k8serrors.IsNotFound(err) || k8serrors.IsConflict(err) {
```
mlavacca, line 163: I'm not sure we want to silently return in case of a conflict. It can be useful to understand what's going on from time to time. WDYT?

pmalek: What value do you think, the conflict error would give to users?

Since the controller is the one that manages these entities, at the end of the day, it is what causes the conflict to be there in the first place. By logging those errors we're letting the user know that "hey, the controller that you're using observed a conflict. You don't need to act on it, but it's an error."

controller/gateway/controller_cleanup.go
pmalek:
```suggestion
			return true, ctrl.Result{}, nil
```

controller/gateway/controller_cleanup.go
mlavacca: Rather than being useful for users, I find this kind of error useful when debugging. But I see your point, let's keep it as it is 👍

mlavacca (approved):

controller/gateway/controller.go
pmalek: Changed this, PTAL.

pmalek:
> > > In many branches of `cleanup` it returns `false` on `shouldReturnEarly` but in the previous code the cases returns `true` on `cleanupUnderway`. This leeds to behavior change in `Reconcile` (whether to return early in reconcile). Is this intended?
> > 
> > 
> > The behavior doesn't change because in those code paths the `err` is not `nil` (or requeue is set) so we return anyway and those will trigger another reconciliation.
> > The `bool` return value is required when `err == nil` or `requeue` is not set.
> > Does that make sense?
> 
> Got that. Indeed it does not change the behavior, I think it could make more sense to explain the meanings and affects on reconcile loops of the return values.

@randmonkey PTAL. I've added more comments in `cleanup`'s godoc explaining this.

PR #456, opened 2024-07-31
feat(konnect): add KonnectAPIAuthConfiguration reconciler

### LINKED ISSUE #431: Konnect entities: KonnectAuthConfiguration reconciler
### Problem statement

Implement `KonnectAuthConfiguration` reconciler.

For details see https://docs.google.com/document/d/1TxW2ovqPRm-w1td1e_V0CyaET4lpiEBCJ7hiYMZJqng/edit#heading=h.xeiruffwggdz

### Proposed solution

- Kong Configuration CRDs (like `Consumer`, Kong's `Service` etc) will have to contain references to `KonnectAuthConfiguration` hence its definition has to live outside of KGO to prevent import cycles (between KIC and KGO, as well as https://github.com/Kong/kubernetes-configuration and KGO). For this reason this CRD and its types has to live in https://github.com/Kong/kubernetes-configuration

- Introduction of this reconciler should be accompanied by introduction of a toggle for konnect related reconcilers which will allow disabling/enabling them when needed.

### Acceptance criteria

- [x] KGO can be configured with a CRD which will contain (and later on be able to reference) Konnect API token for communication with Konnect
  - [x] Done in #456
  - [x] Indication whether the token is valid (using [`/v3/organizations/me` endpoint](https://github.com/Kong/platform-api/blob/096e50e30d864985714c41158511e52ec0a5c88c/build/complete/go-sdk/public.yaml#L3525-L3541) which works for both user and service PATs in Konnect.
- [x] As a User I can use `secretRef` to reference secret holding my Konnect token as an alternative to explicit token inside the `KonnectAPIAuthConfiguration`
  - [x] #459 implements that.

pmalek:

Adds basic reconciler for https://github.com/Kong/kubernetes-configuration/pull/12

Requires: https://github.com/Kong/kubernetes-configuration/pull/18

**Which issue this PR fixes**

Part of #431 

This uses code from https://github.com/Kong/kubernetes-configuration/pull/12

**PR Readiness Checklist**:


controller/konnect/reconciler_konnectapiauth.go
```diff
@@ -0,0 +1,159 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/kong/gateway-operator/controller/pkg/log"
+	k8sutils "github.com/kong/gateway-operator/pkg/utils/kubernetes"
+)
+
+// KonnectAPIAuthConfigurationReconciler reconciles a KonnectAPIAuthConfiguration object.
+type KonnectAPIAuthConfigurationReconciler struct {
```
randmonkey, line 21: Would it stutter to use `konnect.Konnect***` as some linters may suggest?

pmalek: On one hand yes, that's true. On the other hand though we'd ideally want this to have a "konnect" prefix the same way as the CRDs (e.g. added in https://github.com/Kong/kubernetes-configuration/pull/12). 

So just for consistency I'd keep this as is even though it stutters.

---

Listing those CRDs is then unambiguous using e.g. `kubectl` using short resource names: so we end up with 

```
kubectl get konnectapiauthconfigurations
# equivalent of
kubectl get konnectapiauthconfigurations.konnect.konghq.com
```

and not

```
kubectl get apiauthconfigurations
# equivalent of
kubectl get apiauthconfigurations.konnect.konghq.com
```

---

For the full context:

I believe this can be overriden using kubebuilder code markers like so:

```
// +kubebuilder:resource:scope=Namespaced,path=konnectapiauthconfigurations,singular=konnectapiauthconfiguration
```

controller/konnect/conditions.go
```diff
@@ -0,0 +1,14 @@
...
+const (
+	// KonnectAPIAuthConfigurationValidConditionType is the type of the condition
+	// that indicates whether the APIAuth configuration is valid.
+	KonnectAPIAuthConfigurationValidConditionType = "Valid"
+
+	// KonnectAPIAuthConfigurationReasonValid is the reason used with the Valid
+	// condition type indiciating that the APIAuth configuration is valid.
+	KonnectAPIAuthConfigurationReasonValid = "Valid"
+	// KonnectAPIAuthConfigurationReasonValid is the reason used with the Valid
+	// condition type indiciating that the APIAuth configuration is invalid.
```
mlavacca:
```suggestion
	// condition type indicating that the APIAuth configuration is invalid.
	// It is intended to be used when the condition status is false.
```

controller/konnect/conditions.go
```diff
@@ -0,0 +1,14 @@
+package konnect
+
+const (
+	// KonnectAPIAuthConfigurationValidConditionType is the type of the condition
+	// that indicates whether the APIAuth configuration is valid.
+	KonnectAPIAuthConfigurationValidConditionType = "Valid"
+
+	// KonnectAPIAuthConfigurationReasonValid is the reason used with the Valid
+	// condition type indiciating that the APIAuth configuration is valid.
```
mlavacca:
```suggestion
	// condition type indicating that the APIAuth configuration is valid.
	// It is intended to be used when the condition status is false.
```

controller/konnect/sdk_factory.go
mlavacca, line 1: I like this `sdkFactory` approach. We could leverage it in https://github.com/Kong/gateway-operator/issues/437 as well. What do you think about creating a small PR just with it and rebasing this one and #437 before getting them merged? Would it make sense to you?

modules/cli/cli.go
```diff
@@ -44,6 +44,9 @@ func New(m metadata.Info) *CLI {
 	// controllers for specialized APIs and features
 	flagSet.BoolVar(&cfg.AIGatewayControllerEnabled, "enable-controller-aigateway", false, "Enable the AIGateway controller. (Experimental).")
 
+	// controllers for Konnect APIs
+	flagSet.BoolVar(&cfg.KonnectControllersEnabled, "enable-controller-konnect", false, "Enable the KonnectAPIAuthConfiguration controller.")
```
mlavacca: I think we should enable/disable all the konnect-related controllers based on this flag, not only the `konnectAPIAuthConfiguration` one. What do you think?

pmalek: Argh, this is what the flag does (in https://github.com/Kong/gateway-operator/pull/456/files#diff-21545a58e07f7be7b539e19a03651f2f7fed85ad4cbd1fc72d15cdb765888706R277-R288) but I forgot to change the description. 

269dc5a3379c3996e5844eca78873f1c82514707 does just that.

controller/konnect/sdk_factory.go
pmalek: That's fine by me. I don't see the benefit given how small it is now but here it is: https://github.com/Kong/gateway-operator/pull/458. PTAL

config/rbac/role/role.yaml
```diff
@@ -603,6 +603,24 @@ rules:
...
   - update
+- apiGroups:
+  - konnect.konghq.com
+  resources:
+  - konnectapiauthconfigurations
+  verbs:
+  - get
+  - list
+  - patch
+  - update
```
tao12345666333, line 614: Do we need the update permission for this resource?

pmalek: In order for the reconciler to be able to work with KonnectAPIAuthConfiguration resources, yes.

programmer04 (approved):

PR #458, opened 2024-08-01
chore: add Konnect SDK factory

pmalek:

Add Konnect SDK factory.

### DIFF
--- a/controller/konnect/sdkfactory.go
+++ b/controller/konnect/sdkfactory.go
@@ -1,0 +1,36 @@
+package konnect
+
+import (
+	sdkkonnectgo "github.com/Kong/sdk-konnect-go"
+	sdkkonnectgocomp "github.com/Kong/sdk-konnect-go/models/components"
+)
+
+// SDKToken is a token used to authenticate with the Konnect SDK.
+type SDKToken string
+
+// SDKFactory is a factory for creating Konnect SDKs.
+type SDKFactory interface {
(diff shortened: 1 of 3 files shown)


controller/konnect/sdkfactory.go
```diff
@@ -0,0 +1,36 @@
...
+type SDKToken string
+
+// SDKFactory is a factory for creating Konnect SDKs.
+type SDKFactory interface {
+	NewKonnectSDK(serverURL string, token SDKToken) *sdkkonnectgo.SDK
+}
+
+type sdkFactory struct {
+	ServerURL string
+	Token     SDKToken
```
mlavacca: What's the point of having these fields in the factory, since they are passed to the `NewKonnectSDK` function?

pmalek: Whoops, you're right.

controller/konnect/sdkfactory.go
```diff
@@ -0,0 +1,36 @@
...
+}
```
pmalek:
```suggestion
type sdkFactory struct {}
```

controller/konnect/sdkfactory.go
pmalek: Removed in 6466abce2253ee90f78c945d96d82ca53dd05745.

mlavacca (approved):

PR #459, opened 2024-08-01
feat(konnect): support Secrets in KonnectAPIAuthConfiguration

### LINKED ISSUE #431: Konnect entities: KonnectAuthConfiguration reconciler

pmalek:

Add support for tokens in Secrets in `KonnectAPIAuthConfiguration` reconciler.

This allows the following:

```
kind: KonnectAPIAuthConfiguration
metadata:
  name: konnect-api-auth
  namespace: default
spec:
  type: secretRef
  secretRef:
    name: konnect-api-auth-secret
  serverURL: eu.api.konghq.com
---
kind: Secret
apiVersion: v1
metadata:
  name: konnect-api-auth-secret
  namespace: default
  labels:
    konghq.com/credential: konnect
stringData:
  token: kpat_xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
```

The `konghq.com/credential=konnect` is required to efficiently watch for changes in Secrets.

~Depends on #456  and https://github.com/Kong/kubernetes-configuration/pull/16~

**Which issue this PR fixes**

Part of #431 

**PR Readiness Checklist**:

### DIFF
--- a/config/samples/konnect_apiauth_configuration.yaml
+++ b/config/samples/konnect_apiauth_configuration.yaml
@@ -1,0 +1,30 @@
+kind: KonnectAPIAuthConfiguration
+apiVersion: konnect.konghq.com/v1alpha1
+metadata:
+  name: konnect-api-auth-1
+  namespace: default
+spec:
+  type: token
+  token: kpat_xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
+  serverURL: eu.api.konghq.com
+---
+kind: KonnectAPIAuthConfiguration
+apiVersion: konnect.konghq.com/v1alpha1
+metadata:
+  name: konnect-api-auth-2
+  namespace: default
+spec:
+  type: secretRef
+  secretRef:
+    name: konnect-api-auth-secret
+  serverURL: eu.api.konghq.com
+---
+kind: Secret
+apiVersion: v1
+metadata:
+  name: konnect-api-auth-secret
+  namespace: default
+  labels:
+    konghq.com/credential: konnect
+stringData:
+  token: kpat_xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
--- a/config/samples/konnect_apiauth_configuration.yaml
+++ b/config/samples/konnect_apiauth_configuration.yaml
@@ -27,6 +27,8 @@ metadata:
   name: konnect-api-auth-secret
   namespace: default
   labels:
+    # NOTE: this label is required on Konnect credential secrets to make
+    # Secret watch efficient in the operator.
     konghq.com/credential: konnect
 stringData:
   token: kpat_xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
--- a/config/samples/konnect_apiauth_configuration.yaml
+++ b/config/samples/konnect_apiauth_configuration.yaml
@@ -6,6 +6,7 @@ metadata:
 spec:
   type: token
   token: kpat_xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
+  # For complete list of available API URLs see: https://docs.konghq.com/konnect/network/
   serverURL: eu.api.konghq.com
 ---
 kind: KonnectAPIAuthConfiguration
@@ -17,6 +18,7 @@ spec:
   type: secretRef
   secretRef:
     name: konnect-api-auth-secret
+  # For complete list of available API URLs see: https://docs.konghq.com/konnect/network/
   serverURL: eu.api.konghq.com
 ---
 kind: Secret
--- a/controller/konnect/reconciler_konnectapiauth.go
+++ b/controller/konnect/reconciler_konnectapiauth.go
@@ -6,10 +6,15 @@ import (
 	"time"
 
 	sdkkonnectgoops "github.com/Kong/sdk-konnect-go/models/operations"
+	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	"github.com/kong/gateway-operator/controller/pkg/log"
 	k8sutils "github.com/kong/gateway-operator/pkg/utils/kubernetes"
@@ -23,6 +28,17 @@ type KonnectAPIAuthConfigurationReconciler struct {
 	DevelopmentMode bool
 	Client          client.Client
 }
+
+const (
+	// SecretTokenKey is the key used to store the token in the Secret.
+	SecretTokenKey = "token"
+	// SecretCredentialLabel is the label used to identify Secrets holding
+	// KonnectAPIAuthConfiguration tokens.
+	SecretCredentialLabel = "konghq.com/credential" //nolint:gosec
+	// SecretCredentialLabelValueKonnect is the value of the label used to
+	// identify Secrets holding KonnectAPIAuthConfiguration tokens.
+	SecretCredentialLabelValueKonnect = "konnect"
+)
 
 // NewKonnectAPIAuthConfigurationReconciler creates a new KonnectAPIAuthConfigurationReconciler.
 func NewKonnectAPIAuthConfigurationReconciler(
@@ -39,8 +55,26 @@ func NewKonnectAPIAuthConfigurationReconciler(
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *KonnectAPIAuthConfigurationReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	secretLabelPredicate, err := predicate.LabelSelectorPredicate(
+		metav1.LabelSelector{
+			MatchLabels: map[string]string{
+				SecretCredentialLabel: SecretCredentialLabelValueKonnect,
+			},
+		},
+	)
+	if err != nil {
+		return fmt.Errorf("failed to create Secret label selector predicate: %w", err)
+	}
+
 	b := ctrl.NewControllerManagedBy(mgr).
 		For(&konnectv1alpha1.KonnectAPIAuthConfiguration{}).
+		Watches(
+			&corev1.Secret{},
+			handler.EnqueueRequestsFromMapFunc(
+				listKonnectAPIAuthConfigurationsReferencingSecret(mgr.GetClient()),
+			),
+			builder.WithPredicates(secretLabelPredicate),
+		).
 		Named("KonnectAPIAuthConfiguration")
 
 	return b.Complete(r)
@@ -81,9 +115,30 @@ func (r *KonnectAPIAuthConfigurationReconciler) Reconcile(
 		return ctrl.Result{}, nil
 	}
 
+	token, err := getTokenFromKonnectAPIAuthConfiguration(ctx, r.Client, &apiAuth)
+	if err != nil {
+		k8sutils.SetCondition(
+			k8sutils.NewConditionWithGeneration(
+				KonnectEntityAPIAuthConfigurationValidConditionType,
+				metav1.ConditionFalse,
+				KonnectEntityAPIAuthConfigurationReasonInvalid,
+				err.Error(),
+				apiAuth.GetGeneration(),
+			),
+			&apiAuth,
+		)
+		if err := r.Client.Status().Update(ctx, &apiAuth); err != nil {
+			if k8serrors.IsConflict(err) {
+				return ctrl.Result{Requeue: true}, nil
+			}
+			return ctrl.Result{}, fmt.Errorf("failed to update status of %s: %w", entityTypeName, err)
+		}
+		return ctrl.Result{}, err
+	}
+
 	sdk := r.SDKFactory.NewKonnectSDK(
 		"https://"+apiAuth.Spec.ServerURL,
-		SDKToken(apiAuth.Spec.Token),
+		SDKToken(token),
 	)
 
 	// TODO(pmalek): check if api auth config has a valid status condition
@@ -101,6 +156,7 @@ func (r *KonnectAPIAuthConfigurationReconciler) Reconcile(
 			cond.Status != metav1.ConditionFalse ||
 			cond.Reason != KonnectEntityAPIAuthConfigurationReasonInvalid ||
 			cond.ObservedGeneration != apiAuth.GetGeneration() ||
+			cond.Message != err.Error() ||
 			apiAuth.Status.OrganizationID != "" ||
 			apiAuth.Status.ServerURL != apiAuth.Spec.ServerURL {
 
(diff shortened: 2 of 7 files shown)


config/samples/konnect_apiauth_configuration.yaml
```diff
@@ -0,0 +1,30 @@
+kind: KonnectAPIAuthConfiguration
+apiVersion: konnect.konghq.com/v1alpha1
+metadata:
+  name: konnect-api-auth-1
+  namespace: default
+spec:
+  type: token
+  token: kpat_xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
+  serverURL: eu.api.konghq.com
```
randmonkey, line 10: Could we add some comments about the environment using the `serverURL`?

controller/konnect/reconciler_konnectapiauth.go
```diff
@@ -24,6 +29,17 @@ type KonnectAPIAuthConfigurationReconciler struct {
 	Client          client.Client
 }
 
+const (
+	// SecretTokenKey is the key used to store the token in the Secret.
+	SecretTokenKey = "token"
+	// SecretCredentialLabel is the label used to identify Secrets holding
+	// KonnectAPIAuthConfiguration tokens.
+	SecretCredentialLabel = "konghq.com/credential" //nolint:gosec
```
randmonkey, line 37: Just for curiosity: why do we need `gosec` here? The linter just fails because we defined a constant containing "credential"?

config/samples/konnect_apiauth_configuration.yaml
```diff
@@ -0,0 +1,30 @@
...
+    name: konnect-api-auth-secret
+  serverURL: eu.api.konghq.com
+---
+kind: Secret
+apiVersion: v1
+metadata:
+  name: konnect-api-auth-secret
+  namespace: default
+  labels:
+    konghq.com/credential: konnect
```
randmonkey, line 32: I think a comment should be here to tell the users that the label is required to use it for Konnect API auth. The definition is in another repo.

controller/konnect/reconciler_konnectapiauth_watch.go
```diff
@@ -0,0 +1,73 @@
...
+				operatorerrors.ErrUnexpectedObject,
+				"failed to run map funcs",
+				"expected", "Secret", "found", reflect.TypeOf(obj),
+			)
+			return nil
+		}
+
+		var konnectAPIAuthConfigList konnectv1alpha1.KonnectAPIAuthConfigurationList
+		if err := cl.List(ctx, &konnectAPIAuthConfigList); err != nil {
+			log.FromContext(ctx).Error(
```
czeslavo:
```suggestion
			logger.Error(
```

controller/konnect/reconciler_konnectapiauth.go
```diff
@@ -149,3 +216,38 @@ func (r *KonnectAPIAuthConfigurationReconciler) Reconcile(
...
+		var secret corev1.Secret
+		nn := types.NamespacedName{
+			Namespace: apiAuth.Spec.SecretRef.Namespace,
+			Name:      apiAuth.Spec.SecretRef.Name,
+		}
+		if nn.Namespace == "" {
+			nn.Namespace = apiAuth.Namespace
+		}
+
+		if err := cl.Get(ctx, nn, &secret); err != nil {
```
czeslavo:
```suggestion
		nn := types.NamespacedName{
			Namespace: apiAuth.Spec.SecretRef.Namespace,
			Name:      apiAuth.Spec.SecretRef.Name,
		}
		if nn.Namespace == "" {
			nn.Namespace = apiAuth.Namespace
		}

		var secret corev1.Secret
		if err := cl.Get(ctx, nn, &secret); err != nil {
```
Just to keep the var as close as possible to its first usage.

controller/konnect/reconciler_konnectapiauth.go
pmalek:
```
controller/konnect/reconciler_konnectapiauth.go:37:2: G101: Potential hardcoded credentials (gosec)
```

go.mod
```diff
@@ -17,7 +17,7 @@ require (
 	github.com/cloudflare/cfssl v1.6.5
 	github.com/go-logr/logr v1.4.2
 	github.com/google/uuid v1.6.0
-	github.com/kong/kubernetes-configuration v0.0.0-20240801170722-d6586edd1b81
+	github.com/kong/kubernetes-configuration v0.0.0-20240802130201-3233f9f923a7
```
czeslavo: Could we start tagging `github.com/kong/kubernetes-configuration` so we do not use main refs here?

controller/konnect/reconciler_konnectapiauth.go
```diff
@@ -39,8 +55,26 @@ func NewKonnectAPIAuthConfigurationReconciler(
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *KonnectAPIAuthConfigurationReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	secretLabelPredicate, err := predicate.LabelSelectorPredicate(
+		metav1.LabelSelector{
+			MatchLabels: map[string]string{
+				SecretCredentialLabel: SecretCredentialLabelValueKonnect,
+			},
```
mlavacca, line 62: What happens if such a label is removed for any reason? the `updatePredicate` is not triggered, as the new object does not have the label. We may circumvent this by explicitly setting the `updatePredicate` to check the old object has the label.

Also, I can't find the place where you add such a label to the referenced secret. Did I miss it?

mlavacca (commented):
Could you add a CHANGELOG entry?

config/samples/konnect_apiauth_configuration.yaml
pmalek, line 10:
```suggestion
  # For complete list of available API URLs see: https://docs.konghq.com/konnect/network/
  serverURL: eu.api.konghq.com
```

config/samples/konnect_apiauth_configuration.yaml
```diff
@@ -0,0 +1,30 @@
...
+kind: KonnectAPIAuthConfiguration
+apiVersion: konnect.konghq.com/v1alpha1
+metadata:
+  name: konnect-api-auth-2
+  namespace: default
+spec:
+  type: secretRef
+  secretRef:
+    name: konnect-api-auth-secret
+  serverURL: eu.api.konghq.com
```
pmalek, line 22:
```suggestion
  # For complete list of available API URLs see: https://docs.konghq.com/konnect/network/
  serverURL: eu.api.konghq.com
```

config/samples/konnect_apiauth_configuration.yaml
pmalek: PTAL : https://github.com/Kong/gateway-operator/pull/459/commits/04466b36d58708e58d705e663e788dbd308f5aa0

config/samples/konnect_apiauth_configuration.yaml
```diff
@@ -0,0 +1,32 @@
...
+  # For complete list of available API URLs see: https://docs.konghq.com/konnect/network/
+  serverURL: eu.api.konghq.com
+---
+kind: Secret
+apiVersion: v1
+metadata:
+  name: konnect-api-auth-secret
+  namespace: default
+  labels:
+    konghq.com/credential: konnect
```
pmalek, line 32:
```suggestion
    # NOTE: this label is required on Konnect credential secrets to make
    # Secret watch efficient in the operator.
    konghq.com/credential: konnect
```

config/samples/konnect_apiauth_configuration.yaml
pmalek: 7daf42f340a9e733517e73d68f2a08e60f8e0a58

go.mod
pmalek: 👍 8b1abef6aaf4d942a878996145442fa3d57f9d15

controller/konnect/reconciler_konnectapiauth.go
pmalek: It's defined and used by the user, e.g. in the sample manifest: https://github.com/Kong/gateway-operator/pull/459/files#diff-d1e630fb0525f89bb052779443f73fb4c0159f01ce4b1dec343edd2577b883d2R32

---

On one hand we want to try to trigger a reconciliation of referencing KonnectAPIAuthConfigurations, when the label is removed. OTOH when it's removed it's not a credential anymore (or at least not explicitly marked as such).

Nevertheless, at that point we would return an error stating that user is using the Secret without the label (in `getTokenFromKonnectAPIAuthConfiguration`). 

I'm curious for your thoughts on this. Does that sound like good UX to you?

mlavacca: Then, the user in order to set a secret as the token source has to do both the actions:
- set the proper field in the `KonnectAPIAuthConfiguration` resource, and
- label the secret.

If one of the above is not performed, the secret is not taken into account as a token container.

Is that correct?

randmonkey (approved):

controller/konnect/reconciler_konnectapiauth.go
pmalek: Yes. This way we efficiently watch the `Secrets` (via the label selector).

Returning an error will enforce this so that users have the best possible experience and we don't support the use case (no label) with suboptimal UX. WDYT?

randmonkey (changes requested):
Looks like some existing code still importing Konnect related types from `configurationv1alpha1`. They should be also updated after `Kong/kubernetes-configuration` is bumped.

controller/konnect/reconciler_konnectapiauth.go
mlavacca: I think that requiring these two actions to the user is not the best UX and can be confusing. We could attach such a label from the controller logic, instead of asking the user to do it.

In the first reconciliation loop, we fetch the secret and attach a label, then we ensure such a label is always there. We have to be careful though, as the problem I raised in the first comment of this thread has to be avoided.

WDYT?

randmonkey (approved):

controller/konnect/reconciler_konnectapiauth.go
pmalek: I see your point but I believe that this will only complicate the code and make the operator more difficult to be used in gitops setups (this label will need to be ignored as it will not be present in user applied manifests).

We already do use this label in KIC (this change was introduced in 3.0 https://docs.konghq.com/kubernetes-ingress-controller/latest/guides/migrate/credential-kongcredtype-label/ to improve the performance and make the credential secrets explicitly marked as such) and I believe KGO should use the same approach.

czeslavo (approved):

controller/konnect/reconciler_konnectapiauth.go
mlavacca: > I see your point but I believe that this will only complicate the code and make the operator more difficult to be used in gitops setups (this label will need to be ignored as it will not be present in user applied manifests).

Ok, it makes sense.

I still find this approach a bit odd though, and I wonder: do we really need such an optimization? I understand there can be a massive amount of secrets in the cluster, but they do not likely change often. I can see the point of such an improvement for highly mutable objects that may have frequent status updates (pods for example), but secrets do not have status at all and they should be pretty stable objects. Can you help me understand why we really need it?

pmalek: This has been raised quite some time ago by one of KIC users in https://github.com/Kong/kubernetes-ingress-controller/issues/2431.

Do note that this was about webhook validation which can only use label and not fields (as it was done prior to solving that issue).

What we're discussing here is filtering at the controller-runtime predicates level but if we ever need to implement any additional logic elsewhere (say in the aforementioned webhook) we'll hit a similar issue: how to efficiently list if we don't have the label to select on.

> I understand there can be a massive amount of secrets in the cluster, but they do not likely change often. I can see the point of such an improvement for highly mutable objects that may have frequent status updates (pods for example), but secrets do not have status at all and they should be pretty stable objects.

That might be an assumption which is not always true or more importantly: it's not really relevant when we e.g. list `Secret`s that can be used as credentials.

I'm happy to discuss this on a Zoom / Slack 👍

mlavacca: > Do note that this was about webhook validation which can only use label and not fields (as it was done prior to solving that issue).

I see the root problem, and how it was solved, but I really think this is not a good UX. Having a webhook on secrets is a very bad idea at first instance: if the secret is malformed for a consumer we should reflect this in the consumer status in my opinion. But I know that we really LOVE webhooks and we want to have them all over our code.

But this is a different use-case: we are doing it just for optimization purpose, and as I said above I don't see how this is problematic. 

> That might be an assumption which is not always true

Can you give me an example of frequently changing secrets? I can't find any 🤔 

> we e.g. list Secrets that can be used as credentials.

We have the secret reference in the `KonnectAPIAuthConfiguration`, we don't need to list them.

pmalek: > if the secret is malformed for a consumer we should reflect this in the consumer status in my opinion

We already had this discussion w.r.t to a different feature that had rejecting at webhook level invalid configuration. This came straight from our users who want to prevent invalid configuration to get into the cluster.

We can always evangelize best practices internally and externally ( so that e.g. users use objects status for monitoring instead of trying to block invalid configuration at the webhook level). It doesn't seem that we do that (enough), since users are calling for the latter. Do we have an alternative for them? Do we understand their use case and have a solution for that?

Having said that: ideally that would be configurable when people don't want this level of validation. Here, IMHO configurability would only make this simple feature, a complicated one.

> But I know that we really LOVE webhooks and we want to have them all over our code.

That's not a feeling shared by all involved 🙃 

> Can you give me an example of frequently changing secrets? I can't find one 🤔

It doesn't really how to be the use case of frequently changing secrets. Let's try to make a bit less contrived example: a user or an internal PM asks for Konnect token rotation operator. There's 1 requirement: all Konnect tokens embedded either in `KonnectAPIAuthConfiguration` or in `Secret`s has to be checked 7 days ahead of expiration and rotated. Token in `Secret`s should be rotated **regardless** if they are being used or not.

For this use case we'd need to list all `Secret`s in the cluster. Having the label we can easily filter only those that we care about.

---

Additionally, without the label, `listKonnectAPIAuthConfigurationsReferencingSecret` is being called for each `Secret`. Each create or update on a `Secret` in a cluster will cause 

- a list operation of `KonnectAPIAuthConfiguration`s
- check if any of the items in the returned listm, references the `Secret`

mlavacca: > It doesn't really how to be the use case of frequently changing secrets. Let's try to make a bit less contrived example: a user or an internal PM asks for Konnect token rotation operator. There's 1 requirement: all Konnect tokens embedded either in KonnectAPIAuthConfiguration or in Secrets has to be checked 7 days ahead of expiration and rotated. Token in Secrets should be rotated regardless if they are being used or not.

My impression is that we are speculating on something that could happen in the future, and providing a solution before having an actual request or issue. But if you strongly believe this is a good decision, feel free to merge the PR. Approving it.

mlavacca (approved):

controller/konnect/reconciler_konnectapiauth.go
pmalek: > My impression is that we are speculating on something that could happen in the future, and providing a solution before having an actual request or issue.

This way we save ourselves from a breaking change in the future.

I believe this is the right direction and this approach has a valuable trait: we can  relax the requirement without causing a breaking change.

pmalek:
Needed to rebase on `main` and resolve conflicts. PTAL.

PR #462, opened 2024-08-02
feat(konnect): add `KonnectControlPlane` reconciler

### LINKED ISSUE #432: Konnect entities: KonnectControlPlane CRD
### Problem statement

Introduce `KonnectControlPlane`.

For details see https://docs.google.com/document/d/1TxW2ovqPRm-w1td1e_V0CyaET4lpiEBCJ7hiYMZJqng/edit#heading=h.5hsnm3xa1yie

### Proposed solution

- Introduce `KonnectControlPlane` CRD in new API group called: `konnect`.

### Acceptance criteria

- [x] KGO can create, delete and update Konnect's ControlPlanes.

pmalek:

Adds `KonnectControlPlane` reconciler.

Example manifest:

```
kind: KonnectControlPlane
metadata:
  name: test1
  namespace: default
spec:
  name: test1
  labels:
    app: test1
    key1: test1
  konnect:
    authRef:
      name: konnect-api-auth-dev-1
---
kind: KonnectAPIAuthConfiguration
metadata:
  name: konnect-api-auth-dev-1
  namespace: default
spec:
  type: token
  token: kpat_xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
  serverURL: eu.api.konghq.com
```

~Depends on https://github.com/Kong/kubernetes-configuration/pull/20~

**Which issue this PR fixes**

Part of #432 

**PR Readiness Checklist**:

### DIFF
--- a/controller/konnect/ops.go
+++ b/controller/konnect/ops.go
@@ -3,6 +3,8 @@ package konnect
 import (
 	"context"
 	"fmt"
+	"io"
+	"net/http"
 	"time"
 
 	sdkkonnectgo "github.com/Kong/sdk-konnect-go"
@@ -13,17 +15,26 @@ import (
 
 	"github.com/kong/gateway-operator/controller/pkg/log"
 	k8sutils "github.com/kong/gateway-operator/pkg/utils/kubernetes"
+
+	konnectv1alpha1 "github.com/kong/kubernetes-configuration/api/konnect/v1alpha1"
 )
 
-// Op is the Konnect operation type.
+// Response is the interface for the response from the Konnect API.
+type Response interface {
+	GetContentType() string
+	GetStatusCode() int
+	GetRawResponse() *http.Response
+}
+
+// Op is the type for the operation type of a Konnect entity.
 type Op string
 
 const (
-	// CreateOp is the Konnect create operation.
+	// CreateOp is the operation type for creating a Konnect entity.
 	CreateOp Op = "create"
-	// UpdateOp is the Konnect update operation.
+	// UpdateOp is the operation type for updating a Konnect entity.
 	UpdateOp Op = "update"
-	// DeleteOp is the Konnect delete operation.
+	// DeleteOp is the operation type for deleting a Konnect entity.
 	DeleteOp Op = "delete"
 )
 
@@ -34,9 +45,12 @@ func Create[
 ](ctx context.Context, sdk *sdkkonnectgo.SDK, logger logr.Logger, cl client.Client, e *T) (*T, error) {
 	defer logOpComplete[T, TEnt](logger, time.Now(), CreateOp, e)
 
-	switch ent := any(e).(type) { //nolint:gocritic
-	// ---------------------------------------------------------------------
-	// TODO: add other Konnect types
+	switch ent := any(e).(type) {
+	case *konnectv1alpha1.KonnectControlPlane:
+		return e, createControlPlane(ctx, sdk, logger, ent)
+
+		// ---------------------------------------------------------------------
+		// TODO: add other Konnect types
 
 	default:
 		return nil, fmt.Errorf("unsupported entity type %T", ent)
@@ -50,9 +64,12 @@ func Delete[
 ](ctx context.Context, sdk *sdkkonnectgo.SDK, logger logr.Logger, cl client.Client, e *T) error {
 	defer logOpComplete[T, TEnt](logger, time.Now(), DeleteOp, e)
 
-	switch ent := any(e).(type) { //nolint:gocritic
-	// ---------------------------------------------------------------------
-	// TODO: add other Konnect types
+	switch ent := any(e).(type) {
+	case *konnectv1alpha1.KonnectControlPlane:
+		return deleteControlPlane(ctx, sdk, logger, ent)
+
+		// ---------------------------------------------------------------------
+		// TODO: add other Konnect types
 
 	default:
 		return fmt.Errorf("unsupported entity type %T", ent)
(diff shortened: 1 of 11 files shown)


controller/konnect/ops_controlplane.go
```diff
@@ -0,0 +1,165 @@
...
+
+	resp, err := sdk.ControlPlanes.DeleteControlPlane(ctx, id)
+	if errHandled := handleResp[konnectv1alpha1.KonnectControlPlane](err, resp, DeleteOp); errHandled != nil {
+		var sdkError *sdkerrors.SDKError
+		if errors.As(errHandled, &sdkError) && sdkError.StatusCode == 404 {
+			logger.Info("entity not found in Konnect, skipping delete",
+				"op", DeleteOp, "type", cp.GetTypeName(), "id", id,
+			)
+			return nil
+		}
```
czeslavo: Couldn't we just check `resp.GetStatusCode() == 404` here instead?

controller/konnect/ops_controlplane.go
```diff
@@ -0,0 +1,165 @@
...
+		Description: cp.Spec.Description,
+		AuthType:    (*components.UpdateControlPlaneRequestAuthType)(cp.Spec.AuthType),
+		ProxyUrls:   cp.Spec.ProxyUrls,
+		Labels:      cp.Spec.Labels,
+	}
+
+	resp, err := sdk.ControlPlanes.UpdateControlPlane(ctx, id, req)
+	var sdkError *sdkerrors.NotFoundError
+	if errors.As(err, &sdkError) {
+		logger.Info("entity not found in Konnect, trying to recreate",
```
czeslavo, line 110: We do not check if the status code is 404 yet we're proceeding as if it was it.

controller/konnect/ops_controlplane.go
```diff
@@ -0,0 +1,165 @@
...
+		k8sutils.SetCondition(
+			k8sutils.NewConditionWithGeneration(
+				KonnectEntityProgrammedConditionType,
+				metav1.ConditionTrue,
+				KonnectEntityProgrammedReason,
+				"",
+				cp.GetGeneration(),
+			),
+			cp,
+		)
```
czeslavo: `createControlPlane` already sets the condition.
```suggestion
```

controller/konnect/ops.go
```diff
@@ -111,3 +131,27 @@ func logOpComplete[
 		"konnect_id", e.GetKonnectStatus().GetKonnectID(),
 	)
 }
+
+// handleResp checks the response from the Konnect API and returns an error if
```
czeslavo: I have a question regarding the SDK. Methods like `CreateControlPlane` always return a response and an error. Why do we need to handle the raw response body again on our own? Shouldn't the SDK do it for us and return a properly parsed error already?

controller/konnect/ops_controlplane.go
pmalek: This type of error is only returned if the entity wasn't found and the error code returned from Konnect was 404: https://github.com/Kong/sdk-konnect-go/blob/ec4101008c80e02e4c241a21f73ddea248bed6d0/controlplanes.go#L707-L715.

Do you believe checking the status code here as well makes sense?

controller/konnect/ops_controlplane.go
```diff
@@ -0,0 +1,165 @@
...
+	ctx context.Context,
+	sdk *sdkkonnectgo.SDK,
+	logger logr.Logger, //nolint:unparam
+	cp *konnectv1alpha1.KonnectControlPlane,
+) error {
+	resp, err := sdk.ControlPlanes.CreateControlPlane(ctx, cp.Spec.CreateControlPlaneRequest)
+	// TODO: handle already exists
+	// Can't adopt it as it will cause conflicts between the controller
+	// that created that entity and already manages it, hm
+	// TODO: implement entity adoption https://github.com/Kong/gateway-operator/issues/460
```
mlavacca, line 29: For the plugin API the SDK provides an `upsert` function. It'd be ideal to have the same function for other APIs as well

controller/konnect/ops_controlplane.go
pmalek: I was just about to make this change but then we'd to take `err` **and** `resp` into account.

I've changed this to explicitly check for `NotFoundError` instead which I believe is better than checking generic `SDKError`.

mlavacca (approved):

controller/konnect/ops.go
pmalek: We could leverage what the sdk offers us. `handleResp` is somewhat of a syntactic sugar to generically use it for all error handling. Do you believe this could be replaced with something more readable, more succinct?

I'm curious what we can improve here.

controller/konnect/ops_controlplane.go
czeslavo: Ah, I missed the type was `sdkerrors.NotFoundError`, not `sdkerrors.SDKError` (as in `deleteControlPlane` above). Nevermind.

controller/konnect/ops.go
czeslavo: We could probably skip `handleResp` as the SDK already reads the body and maps it to specific error types. That means that there should be no case where an SDK method returns `err == nil` and a non-successful `resp.GetStatusCode()`, making the `handleResp` redundant and noop (as it will always return on the first `if err != nil`). Please correct me if I'm wrong.

pmalek: Agreed. I've changed `handleResp` to basically be only a uniform error wrapper. HTTP body was also already read at that point ( SDK functions read the body to get the error details). 

PTAL 02f5c354dde7c6fa39e5484a0cbcb01358726859

czeslavo (approved):

controller/konnect/ops_controlplane.go
pmalek: 👍 

Most if not all entity types have the support upsert in API. For those that do not have it we can submit requests.

PR #474, opened 2024-08-12
chore: change format of image for KongPluginInstallation

programmer04:

Rework a little bit what has been introduced in 

- https://github.com/Kong/gateway-operator/pull/424

It's simple to deal with a plugin in the format

```dockerfile
FROM scratch

COPY plugin /
```

where `plugin` is a dir that contains `handler.lua` and `schema.lua` which are copied to the root of an image (the final image contains only those two files). Tests are adjusted accordingly. 

**Which issue this PR fixes**

Follow up for https://github.com/Kong/gateway-operator/issues/379

Everything mentioned in the respective section of the PR https://github.com/Kong/gateway-operator/pull/424 remains in force. 

**PR Readiness Checklist**:

### DIFF
--- a/controller/kongplugininstallation/image/image.go
+++ b/controller/kongplugininstallation/image/image.go
@@ -148,7 +148,7 @@ func (sl sizeLimitBytes) String() string {
 	return fmt.Sprintf("%.2f MiB", float64(sl)/(1024*1024))
 }
 
-func extractKongPluginFromLayer(r io.Reader) (map[string]string, error) {
+func extractKongPluginFromLayer(r io.Reader) (PluginFiles, error) {
 	// Search for the files walking through the archive.
 	// The size of a plugin is limited to the size of a ConfigMap in Kubernetes.
 	const sizeLimit_1MiB sizeLimitBytes = 1024 * 1024
--- a/controller/kongplugininstallation/image/image.go
+++ b/controller/kongplugininstallation/image/image.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
 	"sync"
 
 	"github.com/google/go-containerregistry/pkg/name"
@@ -23,8 +24,9 @@ import (
 	"github.com/kong/gateway-operator/modules/manager/metadata"
 )
 
-// FetchPluginContent fetches the content of the plugin from the image URL. When authentication is not needed pass nil.
-func FetchPluginContent(ctx context.Context, imageURL string, credentialsStore credentials.Store) ([]byte, error) {
+// FetchPlugin fetches the content of the plugin from the image URL. When authentication is not needed pass nil.
+// It returns a map with two entries - handler.lua and schema.lua, which are the content of the plugin.
+func FetchPlugin(ctx context.Context, imageURL string, credentialsStore credentials.Store) (map[string]string, error) {
 	ref, err := name.ParseReference(imageURL)
 	if err != nil {
 		return nil, fmt.Errorf("unexpected format of image url: %w", err)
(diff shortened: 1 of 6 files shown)


controller/kongplugininstallation/image/image.go
```diff
@@ -23,8 +24,9 @@ import (
 	"github.com/kong/gateway-operator/modules/manager/metadata"
 )
 
-// FetchPluginContent fetches the content of the plugin from the image URL. When authentication is not needed pass nil.
-func FetchPluginContent(ctx context.Context, imageURL string, credentialsStore credentials.Store) ([]byte, error) {
+// FetchPlugin fetches the content of the plugin from the image URL. When authentication is not needed pass nil.
+// It returns a map with two entries - handler.lua and schema.lua, which are the content of the plugin.
+func FetchPlugin(ctx context.Context, imageURL string, credentialsStore credentials.Store) (map[string]string, error) {
```
czeslavo: Given that we know what the required files are upfront, why not define a struct instead of the `map[string]string`? E.g.

```go
type PluginFiles struct {
  Schema  []byte
  Handler []byte
}
```

programmer04: I thought about this, but it seems to me like a more hustle in the great scheme. It's only used in one place where I need the map to pass it to K8s ConfigMap, there are more ifs to set respective fields, harder (more code) to validate whether each field is filled, etc. Introducing something like https://github.com/mitchellh/mapstructure seems to be overkill too.

This is why I did like that, but yeah if you insist I can change it 😉

controller/kongplugininstallation/image/image.go
```diff
@@ -117,34 +119,56 @@ func (sl sizeLimitBytes) String() string {
...
 	}
+
+	var missingFiles []string
+	for _, f := range []string{kongPluginHandler, kongPluginSchema} {
+		if _, ok := pluginFiles[f]; !ok {
+			missingFiles = append(missingFiles, f)
+		}
+	}
+	if len(missingFiles) > 0 {
+		return nil, fmt.Errorf("not found in the image required files: %s", strings.Join(missingFiles, ", "))
```
czeslavo:
```suggestion
		return nil, fmt.Errorf("required files not found in the image: %s", strings.Join(missingFiles, ", "))
```

controller/kongplugininstallation/image/image.go
```diff
@@ -117,34 +119,56 @@ func (sl sizeLimitBytes) String() string {
...
+	}
```
czeslavo: For the readability's sake, could we move this validation to a separate function, e.g., ensureRequiredPluginFiles?

controller/kongplugininstallation/image/image.go
```diff
@@ -117,34 +119,56 @@ func (sl sizeLimitBytes) String() string {
 	return fmt.Sprintf("%.2f MiB", float64(sl)/(1024*1024))
 }
 
-func extractKongPluginFromLayer(r io.Reader) ([]byte, error) {
+func extractKongPluginFromLayer(r io.Reader) (map[string]string, error) {
+	// The target file name expected in image with a custom Kong plugin.
```
czeslavo:
```suggestion
	// The target files' names expected in an image with a custom Kong plugin.
```

controller/kongplugininstallation/image/image.go
```diff
@@ -117,34 +119,56 @@ func (sl sizeLimitBytes) String() string {
...
+	const (
+		kongPluginHandler = "handler.lua"
+		kongPluginSchema  = "schema.lua"
+	)
+	// Search for the file walking through the archive.
+	// Size of plugin is limited to size of a ConfigMap in Kubernetes.
```
czeslavo:
```suggestion
	// Search for the files walking through the archive.
	//The size of a plugin is limited to the size of a ConfigMap in Kubernetes.
```

controller/kongplugininstallation/image/image.go
czeslavo: My assumption was that `FetchPlugin` is able to ensure that the struct it would return is valid, so downstream doesn't have to do that the second time - it'd only use the struct's fields and map them to the file names in the CM's config.

I'm fine with keeping using `map[string]string`, but how about at least a type definition for it and a godoc to make it clear what maps to what exactly? :) 

```go
// PluginFiles maps a plugin's file names to their content.
// It's expected that each plugin consists of `schema.lua` and `handler.lua` files.
type PluginFiles map[string]string
```

controller/kongplugininstallation/image/image.go
```diff
@@ -117,34 +119,56 @@ func (sl sizeLimitBytes) String() string {
...
+	const sizeLimit_1MiB sizeLimitBytes = 1024 * 1024
```
czeslavo, line 154: This const can be put in a single `const` block above.

controller/kongplugininstallation/image/image.go
```diff
@@ -117,34 +119,56 @@ func (sl sizeLimitBytes) String() string {
...
+		if err != nil {
+			return nil, fmt.Errorf("unexpected error during looking for plugin: %w", err)
+		}
+
+		switch fileName := filepath.Base(h.Name); fileName {
+		case kongPluginHandler, kongPluginSchema:
+			file := make([]byte, h.Size)
+			if _, err := io.ReadFull(tr, file); err != nil {
+				if errors.Is(err, io.ErrUnexpectedEOF) {
+					return nil, fmt.Errorf("plugin size exceed %s", sizeLimit_1MiB)
```
czeslavo:
```suggestion
					return nil, fmt.Errorf("plugin size limit of %s exceeded", sizeLimit_1MiB)
```

controller/kongplugininstallation/image/image.go
programmer04: Now it'll be different

controller/kongplugininstallation/image/image.go
programmer04: Yup, good idea, I'm doing it

controller/kongplugininstallation/image/image.go
```diff
@@ -120,13 +149,8 @@ func (sl sizeLimitBytes) String() string {
 }
 
 func extractKongPluginFromLayer(r io.Reader) (map[string]string, error) {
```
czeslavo:
```suggestion
func extractKongPluginFromLayer(r io.Reader) (PluginFiles, error) {
```

czeslavo (approved):

Issue #465, opened 2024-08-07
Streamline and automatize/document testing strategy for `KongPluginInstallation` CRDs


programmer04:
## Problem Statement

For instance, PR https://github.com/Kong/gateway-operator/pull/424 introduces some tests that require remote image registries with properly built images to execute tests. The test setup should be easily reproducible/modifiable in the future to make it easy to maintain. 

## Proposed Solution

See e.g. https://github.com/Kong/gateway-operator/pull/424#discussion_r1705307390

## Additional Information

Reference also official documentation to get more info about actual format

## Acceptance Criteria
- [ ] Test suite/set-up for `KongPluginInstallation` is easy to follow, modify and reproduce

PR #476, opened 2024-08-12
feat: allow assigning KongPluginInstallation to GatewayConfiguration and DataPlane

### LINKED ISSUE #380: Allow configuring additional plugins for Dataplane - Kong Gateway
## Problem Statement

The relationship between an instance of CRD introduced in 
- https://github.com/Kong/gateway-operator/issues/378

and instance of Dataplane needs to be established. Extend with an additional field that matches the name of CRD, eg.

- `GatewayConfiguration` -> `spec.dataPlaneOptions.pluginsToInstall`
- `DataPlane` -> `spec.pluginsToInstall`

[src](https://docs.google.com/document/d/1-gYEaf_31BR_QXmZXNhy0bOrXYJ4X_cXobQ_vbLDIhE/edit#heading=h.3tn2kjmrw2q5)

Kubernetes ConfigMap has to be mounted as a file in the proper paths of Kong Gateway, also environment variables have to be properly configured to make Kong Plugin available for Kong Gateway to use and be configured with KIC's CRD `KongPlugin`. 

diagram

What is in the blue box is in the scope of this issue.

In scope are only LUA plugins without additional dependencies. To add them put plugins in a dedicated directory accessible by Kong Gateway and provide it via environment variable `KONG_LUA_PACKAGE_PATH` (it allows configuration like e.g. `"/opt/?.lua;;"`). All plugins must be in that directory.
The name of the plugin needs to be added to the variable KONG_PLUGINS (e.g. `KONG_PLUGINS=”bundled,myplugin”` where ”bundled” enables Kong default, built-in plugins.
> References: [Gateway docs](https://docs.konghq.com/gateway/latest/plugin-development/distribution/#via-a-dockerfile-or-docker-run-install-and-load), [KIC docs](https://docs.konghq.com/kubernetes-ingress-controller/latest/plugins/custom/)

## Acceptance Criteria
- [ ] Plugin can be configured and used
- [ ] statuses about attachment are in place
- [ ] referenced plugins can't be removed by mistake

programmer04:

Kong custom plugin built from `Dockerfile`

```Dockerfile
FROM scratch

COPY plugin /
```

where `plugin` is a directory that contains `handler.lua` and `schema.lua` as described in [the official docs](https://docs.konghq.com/kubernetes-ingress-controller/latest/plugins/custom/#create-a-custom-plugin). 

Image with the aforementioned Kong custom plugin is available publicly under
- `northamerica-northeast1-docker.pkg.dev/k8s-team-playground/plugin-example/myheader` -> adds header `myheader: roar`
- `northamerica-northeast1-docker.pkg.dev/k8s-team-playground/plugin-example/myheader-2` (it's also mirrored in `plugin-example-private` requires auth) -> adds header `newheader: amazing`

Those are used in `config/samples/gateway-kongplugininstallation-httproute.yaml` and submitted as part of this PR and in the integration test.
 
 Functionality is implemented entirely. For each instance of Kong Gateway deployment, a dedicated `ConfigMap` with a custom plugin is created (based on source `ConfigMap` made as a result of the successful reconciliation of  `KongPluginInstallation`) and its lifecycle is tied to Kong Gateway. Such an approach although a little bit wasteful, because there are many duplications of `ConfigMap`s (but it's guaranteed that its size won't exceed 1MiB), greatly simplifies implementation. This is due to the limitation of Kubernetes - it's impossible to mount a `ConfigMap` from another namespace to a `Deployment`. Furthermore, it reduces the blast radius of some mistakes, bugs, user actions, etc. to only one Dataplane - each one has its own `ConfigMap`s and reconciles them on its own.

**Which issue this PR fixes**

Closes https://github.com/Kong/gateway-operator/issues/380

Those "magic" images used in tests and examples will be documented/reworked to make them easy to maintain in the separate issue #465

Maybe more testing would be useful, e.g. with the usage of newly created `envtests` or more integration ones, I'm open to suggestions that I'm going to address in a separate PR. 

**PR Readiness Checklist**:


api/v1beta1/dataplane_types.go
```diff
@@ -80,6 +80,10 @@ type DataPlaneOptions struct {
 	// +kubebuilder:validation:MinItems=0
 	// +kubebuilder:validation:MaxItems=1
 	Extensions []v1alpha1.ExtensionRef `json:"extensions,omitempty"`
+	// PluginsToInstall is a list of KongPluginInstallation resources that
+	// will be installed and available in the DataPlane.
+	// +optional
+	PluginsToInstall []NamespacedName `json:"pluginsToInstall,omitempty"`
```
pmalek, line 86: Just calling this out explicitly ( I might be missing something in here): we're aware of using plugins from other namespaces without any grant system in place and we're not concerned about this?

controller/dataplane/watch.go
```diff
@@ -27,5 +37,64 @@ func DataPlaneWatchBuilder(mgr ctrl.Manager) *builder.Builder {
...
+		// They may trigger reconciliation of DataPlane resources.
+		Watches(
+			&corev1.ConfigMap{},
+			handler.EnqueueRequestsFromMapFunc(listDataPlanesReferencingKongPluginInstallation(mgr.GetClient())),
+		)
+}
+
+func listDataPlanesReferencingKongPluginInstallation(
+	c client.Client,
+) handler.TypedMapFunc[client.Object, reconcile.Request] {
```
pmalek: Can't we use the actual type `corev1.ConfigMap` in here as type parameter just like in the example in the PR introducing `TypedMapFunc` https://github.com/kubernetes-sigs/controller-runtime/pull/2783/files#diff-6dea9cae12878bf42782202d4cf31b7336732b7ec9856b1c2fd35c305ab0c081R72-R83 ?

controller/dataplane/watch.go
```diff
@@ -27,5 +37,64 @@ func DataPlaneWatchBuilder(mgr ctrl.Manager) *builder.Builder {
...
+		}
+		// Find all DataPlane resources reference KongPluginInstallation
+		// that maps to the ConfigMap enqueued for reconciliation.
+		kpiToFind := kpiCM.Annotations[consts.AnnotationKongPluginInstallationMappedKongPluginInstallation]
+		if kpiToFind == "" {
+			return nil
+		}
+
+		var dataPlaneList operatorv1beta1.DataPlaneList
+		if err := c.List(ctx, &dataPlaneList); err != nil {
```
pmalek:
```suggestion
		if err := c.List(ctx, &dataPlaneList); client.IgnoreNotFound(err) != nil {
```

controller/dataplane/controller_reconciler_utils.go
```diff
@@ -78,6 +84,169 @@ func (r *Reconciler) ensureDataPlaneAddressesStatus(
...
+) (customPlugin, ctrl.Result, error) {
+	kpi, ready, err := verifyKPIReadinessForDataPlane(ctx, logger, c, dataplane, kpiNN)
+	if err != nil {
+		return customPlugin{}, ctrl.Result{}, err
+	}
+	if !ready {
+		return customPlugin{}, ctrl.Result{Requeue: true}, nil
+	}
+
+	log.Trace(logger, "Fetch underlying ConfigMap for KongPluginInstallation", kpi)
```
pmalek: Might log the name of the ConfigMap here.

pkg/consts/kongplugininstallation.go
```diff
@@ -8,4 +8,8 @@ const (
 	// AnnotationKongPluginInstallationName is the annotation key used to store the name of the KongPluginInstallation
```
czeslavo:
```suggestion
	// AnnotationKongPluginInstallationMappedKongPluginInstallation is the annotation key used to store the name of the KongPluginInstallation
```

controller/dataplane/custom_plugin.go
```diff
@@ -0,0 +1,87 @@
...
+		}
+	}
+
+	var (
+		kpisNames        = make([]string, 0, len(customPlugins)+1)
+		kpisGenerations  = make([]string, 0, len(customPlugins))
+		kpisVolumeMounts = make([]corev1.VolumeMount, 0, len(customPlugins))
+		kpisVolumes      = make([]corev1.Volume, 0, len(customPlugins))
+	)
+	kpisNames = append(kpisNames, "bundled") // To preserve plugins available in the default image.
```
czeslavo: How about extracting `bundled` as a constant and documenting this, explaining it is a special value Kong accepts in its `KONG_PLUGINS` to use the default builtin plugins bundle?

controller/dataplane/custom_plugin.go
```diff
@@ -0,0 +1,87 @@
...
+			// It's safe to perform a delete operation on a nil map.
+			delete(
+				d.Spec.Template.Annotations,
+				consts.AnnotationKongPluginInstallationGenerationInternal,
+			)
+		}
+	}
+
+	var (
+		kpisNames        = make([]string, 0, len(customPlugins)+1)
```
czeslavo:
```suggestion
		kpisNames        = make([]string, 0, len(customPlugins)+1) // +1 for the bundled plugins.
```

controller/dataplane/custom_plugin.go
```diff
@@ -0,0 +1,87 @@
...
+
+	return func(deployment *appsv1.Deployment) {
+		if deployment.Spec.Template.Annotations == nil {
+			deployment.Spec.Template.Annotations = make(map[string]string)
+		}
+		deployment.Spec.Template.Annotations[consts.AnnotationKongPluginInstallationGenerationInternal] = strings.Join(kpisGenerations, ",")
+		deployment.Spec.Template.Spec.Containers[0].Env = append(
+			deployment.Spec.Template.Spec.Containers[0].Env,
+			corev1.EnvVar{
+				Name:  "KONG_PLUGINS",
```
czeslavo: Please consider extracting Kong's well-known env vars to constants (I can see it's already used in 4 different places).

controller/dataplane/controller_reconciler_utils.go
```diff
@@ -78,6 +84,169 @@ func (r *Reconciler) ensureDataPlaneAddressesStatus(
...
+	case 0:
+		log.Trace(logger, "Create new ConfigMap for KongPluginInstallation", kpi)
+		cm.GenerateName = dataplane.Name + "-"
+		cm.Namespace = dataplane.Namespace
+		k8sutils.SetOwnerForObject(&cm, dataplane)
+		k8sresources.LabelObjectAsDataPlaneManaged(&cm)
+		k8sresources.AnnotateConfigMapWithKongPluginInstallation(&cm, kpi)
+		cm.Data = underlyingCM.Data
+		if err := c.Create(ctx, &cm); err != nil {
+			return customPlugin{}, ctrl.Result{}, fmt.Errorf("could not create new mapped: %w", err)
```
pmalek: This error message seems a bit vague: a new mapped? How about something like:

```suggestion
			return customPlugin{}, ctrl.Result{}, fmt.Errorf("could not create new ConfigMap for KongPluginInstallation: %w", err)
```

controller/dataplane/controller_reconciler_utils.go
```diff
@@ -78,6 +84,169 @@ func (r *Reconciler) ensureDataPlaneAddressesStatus(
...
+		}
+	case 1:
+		log.Trace(logger, "Update existing ConfigMap for KongPluginInstallation", kpi)
```
pmalek: WDYT about adding the cm name to the log entry?

controller/dataplane/watch.go
```diff
@@ -27,5 +37,64 @@ func DataPlaneWatchBuilder(mgr ctrl.Manager) *builder.Builder {
...
+		ctx context.Context, obj client.Object,
+	) []reconcile.Request {
+		logger := ctrllog.FromContext(ctx)
+
+		kpiCM, ok := obj.(*corev1.ConfigMap)
+		if !ok {
+			logger.Error(
+				fmt.Errorf("unexpected object type"),
+				"Namespace watch predicate received unexpected object type",
+				"expected", "*operatorv1alpha1.KongPluginInstallation", "found", reflect.TypeOf(obj),
```
czeslavo:
```suggestion
				"expected", "*corev1.ConfigMap", "found", reflect.TypeOf(obj),
```

controller/dataplane/controller_reconciler_utils.go
```diff
@@ -78,6 +84,169 @@ func (r *Reconciler) ensureDataPlaneAddressesStatus(
...
+		cm = mappedConfigMapForKPI[0]
+		cm.Data = underlyingCM.Data
+		if err := c.Update(ctx, &cm); err != nil {
```
pmalek: If `k8serrors.IsConflict(err)` I'd suggest to not return an error but a `ctrl.Result{Requeue: true}` to not bother users with an error that most likely stems from the operator synchronization (because of using a cached client). WDYT?

controller/dataplane/watch.go
czeslavo: Or, alternatively `fmt.Sprintf("%T", kpiCM)`.

controller/dataplane/watch.go
```diff
@@ -27,5 +37,64 @@ func DataPlaneWatchBuilder(mgr ctrl.Manager) *builder.Builder {
...
+			)
+			return nil
+		}
+		// Find all DataPlane resources reference KongPluginInstallation
```
czeslavo:
```suggestion
		// Find all DataPlane resources referencing KongPluginInstallation
```

controller/dataplane/watch.go
```diff
@@ -27,5 +37,64 @@ func DataPlaneWatchBuilder(mgr ctrl.Manager) *builder.Builder {
...
+		// that maps to the ConfigMap enqueued for reconciliation.
+		kpiToFind := kpiCM.Annotations[consts.AnnotationKongPluginInstallationMappedKongPluginInstallation]
+		if kpiToFind == "" {
+			return nil
+		}
+
+		var dataPlaneList operatorv1beta1.DataPlaneList
+		if err := c.List(ctx, &dataPlaneList); err != nil {
+			logger.Error(err, "Failed to list gateways in watch", "KongPluginInstallation", kpiToFind)
```
czeslavo:
```suggestion
			logger.Error(err, "Failed to list DataPlanes in watch", "KongPluginInstallation", kpiToFind)
```

controller/dataplane/controller_reconciler_utils.go
```diff
@@ -78,6 +84,169 @@ func (r *Reconciler) ensureDataPlaneAddressesStatus(
 	return false, nil
 }
 
+// ensureKonPluginInstallation ensures that the KongPluginInstallation resources
+// referenced by the DataPlane are resolved and DataPlane is configured to use them.
+// During resolving for each DataPlane based on each instance of KongPluginInstallation
+// ConfigMap is created and mounted. The DataPlane manages its lifecycle.
+func ensureKongPluginInstallation(
```
pmalek: Can we change the name of this `func` (and the godoc on it) to mention that it returns a `Deployment` generation opt rather than an actual function that would ensure the KPI for a `DataPlane`? I believe this might be misleading to future readers.

Altenatively we could split the opt code from code that does prepares the `ConfigMap`s and deletes the old ones.

WDYT?

pkg/utils/test/clients.go
```diff
@@ -42,6 +44,11 @@ func NewK8sClients(env environments.Environment) (K8sClients, error) {
 	if err != nil {
 		return clients, err
 	}
+	clients.KICClient, err = kicclient.NewForConfig(env.Cluster().Config())
```
czeslavo: Please consider switching to `kubernetes-configuration` client instead for manipulating `KongPlugin`s.

controller/dataplane/custom_plugin.go
pmalek, line 1: To be consistent with other files that contain code responsible for `Deployment` opt I'd suggest to rename this to `owned_custom_plugins.go`

pkg/utils/test/clients.go
programmer04: Very good idea, I'm doing it

controller/dataplane/custom_plugin.go
programmer04: 🆗

controller/dataplane/controller_reconciler_utils.go
programmer04: Very good suggestion, btw. we should probably go through our codebase to find some spots like that, because even when you look at logs of our automated tests we can see some transitive errors that may not should bother users

controller/dataplane/custom_plugin.go
programmer04: I'm proposing an alternative approach, I hope you'll like it

controller/dataplane/custom_plugin.go
programmer04: I'm proposing an alternative approach, I hope you'll like it

controller/dataplane/watch.go
programmer04: I tried to do it and failed using `Watches(...)`. I think it can be done only in the way I did based on what you provided me, please check

controller/dataplane/controller_reconciler_utils.go
programmer04: I reworked it like that, please tell WDYT

internal/utils/dataplane/config.go
```diff
@@ -75,3 +84,26 @@ func FillDataPlaneProxyContainerEnvs(existing []corev1.EnvVar, podTemplateSpec *
 	}
 	sort.Sort(k8sutils.SortableEnvVars(container.Env))
 }
+
+// ConfigureKongPluginRelatedEnvVars returns the environment variables
+// needed for configuring the Kong Gateway with the provided Kong Plugin
+// names. If kongPluginNames is nil or empty, nil is returned.
```
czeslavo:
```suggestion
// names. If kongPluginNames is nil or empty, nil is returned. Kong will use bundled
// plugins by default if we do not override `KONG_PLUGINS`.
```
Just to make it clear we do not make a mistake here by returning nil.

pkg/consts/kongplugininstallation.go
programmer04: I'm hoping you think about something like I have know 😉

api/v1beta1/dataplane_types.go
programmer04: This is a valid point, I have to sort out this. And I'll answer here and update the PR description. 

❗ For now, I'm disabling automerge of this PR, and let's leave this conversation open. 
🟢  But I think that the actual core review can be continued

czeslavo (commented):
Apart from Patryk's concern about the namespaced Plugin references in the DP CRD, it looks good to me now. 👍

api/v1beta1/dataplane_types.go
programmer04: As concluded we're leaving it as it is for now

czeslavo (approved):

controller/dataplane/watch.go
pmalek: 👍 

Just an FYI: builder's `Watches()` can't be easily done now given that go doesn't allow type parameters on methods: https://github.com/golang/go/issues/49085

I've created this to track the issue (but knowing that it can't be easily done now). https://github.com/kubernetes-sigs/controller-runtime/issues/2955.

So your approach with using source seems to be the only one that's possible.

controller/dataplane/watch.go
```diff
@@ -27,5 +36,58 @@ func DataPlaneWatchBuilder(mgr ctrl.Manager) *builder.Builder {
...
+		// Find all DataPlane resources referencing KongPluginInstallation
+		// that maps to the ConfigMap enqueued for reconciliation.
+		kpiToFind := kpiCM.Annotations[consts.AnnotationMappedToKongPluginInstallation]
+		if kpiToFind == "" {
+			return nil
+		}
+
+		var dataPlaneList operatorv1beta1.DataPlaneList
+		if err := c.List(ctx, &dataPlaneList); client.IgnoreNotFound(err) != nil {
+			logger.Error(err, "Failed to list DataPlanes in watch", "KongPluginInstallation", kpiToFind)
```
pmalek, line 70: This could also use an index similarly as we're doing it in Konnect entities e.g. https://github.com/Kong/gateway-operator/blob/4731ea64685befe828bbdcdc648d6f5e11660271/controller/konnect/index_kongpluginbinding.go right? Then there's no need for filtering below and listing is faster.

WDYT?

controller/dataplane/owned_custom_plugins.go
```diff
@@ -0,0 +1,81 @@
...
+type customPlugin struct {
+	// Name of the KongPluginInstallation resource.
+	Name string
+	// ConfigMapNN is the namespace/name of the ConfigMap that contains the plugin.
+	ConfigMapNN types.NamespacedName
+	// Generation is the generation of the KongPluginInstallation that contains the plugin.
+	Generation int64
+}
+
+func withCustomPlugins(customPlugins []customPlugin) k8sresources.DeploymentOpt {
```
pmalek: WDYT about changing the signature here so that it accepts a variadic list instead of a slice? This way we don't need to provide `nil` when no custom Plugins are to be configured.

config/samples/gateway-kongplugininstallation-httproute.yaml
```diff
@@ -0,0 +1,170 @@
...
+  name: kong-custom-plugin
+plugin: additional-custom-plugin
+---
+kind: KongPluginInstallation
+apiVersion: gateway-operator.konghq.com/v1alpha1
+metadata:
+  name: additional-custom-plugin
+  namespace: additional
+spec:
+  image: northamerica-northeast1-docker.pkg.dev/k8s-team-playground/plugin-example/myheader
```
pmalek, line 157: Do those need to be named the same? If not I'd make them different in the sample so that users now which is being referred to in the `pluginsToInstall` list above. WDYT?

pmalek (commented):
I'm still going through this one but these are the minor nits I found so far.

config/samples/gateway-kongplugininstallation-httproute.yaml
programmer04: Do you mean if the field `plugin: additional-custom-plugin` of `KongPlugin` has to match `name: additional-custom-plugin` field of `KongPluginInstallation`? Yet it has to, it's the same mechanism when you specify a plugin that is bundled in Kong Gateway, you have to use its name in `KongPlugin`

controller/dataplane/controller.go
```diff
@@ -176,6 +176,17 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	deploymentOpts := []k8sresources.DeploymentOpt{
 		labelSelectorFromDataPlaneStatusSelectorDeploymentOpt(dataplane),
 	}
+
+	log.Trace(logger, "ensuring generation of deployment configuration for KongPluginInstallations configured for DataPlane", dataplane)
+	kpisForDeployment, reconcileResult, err := ensureMappedConfigMapToKongPluginInstallationForDataPlane(ctx, logger, r.Client, dataplane)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("cannot ensure KongPluginInstallation for DataPlane: %w", err)
+	}
+	if reconcileResult.Requeue {
```
pmalek: I can see that for now `populateDedicatedConfigMapForKongPluginInstallation` nor `ensureMappedConfigMapToKongPluginInstallationForDataPlane` set `.RequeueAfter` but I'd consider adding a check for that here to be future proof and also requeue on that. WDYT?

controller/dataplane/controller.go
```diff
@@ -210,10 +220,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{}, nil
 	}
 
-	if res, err := ensureDataPlaneReadyStatus(ctx, r.Client, logger, dataplane, dataplane.Generation); err != nil {
+	if reconcileResult, err := ensureDataPlaneReadyStatus(ctx, r.Client, logger, dataplane, dataplane.Generation); err != nil {
 		return ctrl.Result{}, err
-	} else if res.Requeue {
-		return res, nil
+	} else if reconcileResult.Requeue {
```
pmalek: Ditto here.

controller/dataplane/controller_reconciler_utils.go
```diff
@@ -77,6 +84,170 @@ func (r *Reconciler) ensureDataPlaneAddressesStatus(
...
+	for _, kpiNN := range dataplane.Spec.PluginsToInstall {
+		kpiNN := types.NamespacedName(kpiNN)
+		if kpiNN.Namespace == "" {
+			kpiNN.Namespace = dataplane.Namespace
+		}
+
+		cp, reconciliationResult, err := populateDedicatedConfigMapForKongPluginInstallation(
+			ctx, logger, c, configMapsOwned, kpiNN, dataplane,
+		)
+		if err != nil || reconciliationResult.Requeue {
```
pmalek: Ditto about `.RequeueAfter`.

We might also consider adding a `func` which would simply return `true` if either of those are set.

controller/dataplane/controller_reconciler_utils.go
```diff
@@ -77,6 +84,170 @@ func (r *Reconciler) ensureDataPlaneAddressesStatus(
...
+		return customPlugin{}, ctrl.Result{}, fmt.Errorf("unexpected error happened - more than one ConfigMap found: %s", names)
+	}
+	return customPlugin{
+		Name:        kpi.Name,
+		ConfigMapNN: client.ObjectKeyFromObject(&cm),
+		Generation:  kpi.Generation,
+	}, ctrl.Result{}, nil
+}
+
+// verifyKPIReadinessForDataPlane Reports in DataPlane status when KPI does not exist or it
```
pmalek:
```suggestion
// verifyKPIReadinessForDataPlane reports in DataPlane status when KPI does not exist or it
```

pmalek: Nit: reports feels vague here. Reading this godoc I don't know what are going to be the side effects of this `func`. Looking at the source code it seems that `DataPlane`'s conditions are updated according to the state of KPI. Can we mention that in the comment?

controller/dataplane/controller_reconciler_utils.go
```diff
@@ -77,6 +84,170 @@ func (r *Reconciler) ensureDataPlaneAddressesStatus(
...
+	case 1:
+		cm = mappedConfigMapForKPI[0]
+		log.Trace(logger, fmt.Sprintf("Update existing ConfigMap %s for KongPluginInstallation", client.ObjectKeyFromObject(&cm)), kpi)
+		cm.Data = underlyingCM.Data
+		if err := c.Update(ctx, &cm); err != nil {
+			if k8serrors.IsConflict(err) {
+				return customPlugin{}, ctrl.Result{Requeue: true}, nil
+			}
+			return customPlugin{}, ctrl.Result{}, fmt.Errorf("could not update mapped: %w", err)
+		}
```
pmalek: While this seems OK I'd suggest considering following the pattern that we already have, that is comparing relevant fields and only if those are not as expected, performing an update. This could potentially save some unnecessary updates.

Not critical though to be included in this PR.

programmer04: I think that the most important thing that we should only track here is the content of CM, so I've done it

controller/dataplane/controller.go
programmer04: or maybe instead of returning `ctrl.Result` just return `requeue bool` and ensure returning configured `ctrl.Result{}` only in the body of `Reconcile`. I changed it like I described for the scope of this PR. For `ensureDataPlaneReadyStatus` I can adjust in a separate PR. WDYT?

pmalek: I'm ok with that as well but we already use `.RequeueAfter` in konnect entities reconciler: https://github.com/Kong/gateway-operator/blob/56f5cc1bfdfddfd92421b7efc84b5b78c40b55f5/controller/konnect/ops/ops.go#L141-L159.

That's not a big deal to use different mechanism between the 2 but let's just make sure to account for that field as well here.

controller/dataplane/watch.go
programmer04: I wonder if it would be a big gain for `DataPlane`s. I think that compared to something like consumers, we'd have only a handful of `DataPlane`s. WDYT? Worth to do? I don't have any numbers to backup my thesis 🙃

controller/dataplane/controller_reconciler_utils.go
programmer04: I reworded it entirely, PTAL

controller/dataplane/controller.go
programmer04: So here for Konnect it's handled OK. For the aforementioned, I think that approach with booleans is simpler, so I'll adjust it everywhere but in a separate PR

controller/dataplane/watch.go
pmalek: It might be indeed lower gain than for the `Consumer`s but I believe it's still worth to add this index.

Can be done in a separate PR since this one is pretty big already.

controller/dataplane/controller_reconciler_utils.go
```diff
@@ -77,6 +84,175 @@ func (r *Reconciler) ensureDataPlaneAddressesStatus(
...
+	}
+	configMapsToRetain := make(map[types.NamespacedName]struct{}, len(configMapsOwned))
+
+	for _, kpiNN := range dataplane.Spec.PluginsToInstall {
+		kpiNN := types.NamespacedName(kpiNN)
+		if kpiNN.Namespace == "" {
+			kpiNN.Namespace = dataplane.Namespace
+		}
+
+		cp, requeue, err := populateDedicatedConfigMapForKongPluginInstallation(
```
pmalek: This is OK but I'd suggest to be cautious about redefining the named return variables.

pmalek (approved):

controller/dataplane/controller_reconciler_utils.go
programmer04: Improved, tbh I don't like naked returns, but named returns for me are quite useful as a documentation 😉

controller/dataplane/watch.go
programmer04: Ok, I'll address it separately and check for others opportunities

Issue #447, opened 2024-07-30
Konnect entities: add Konnect mocks for integration testing


pmalek:
### Problem statement

#370 introduces support for managing of Konnect entities in KGO.

Konnect SDK - https://github.com/Kong/sdk-konnect-go/ - instances are created in runtime due to tokens being stored in Kubernetes resources ( `KonnectAPIAuthConfiguration` or Secrets ).

This issue tracks the effort of introducing mocks for Konnect APIs so that we can test Konnect integration in operator's integration test suite.

### Relevant

https://github.com/Kong/gateway-operator/pull/446/files#diff-5b6aadc0fe9b4d0ba8cfd31756eef41ffdcd65687cd077d3b8d427294dcc81b5R157-R164

### Acceptance criteria

- [ ] Konnect mocks exists and are used in integration tests.

PR #507, opened 2024-08-20
refactor(konnect): add SDK interface types and theirs mocks

pmalek:

This PR will allow UTs for entity focused functions when mocks for those are introduced.

This PR does reorganize the `konnect` package a bit to solve the import cycles between `konnect` and newly added `ops` package.

Mocks are generated using [`mockery`](https://github.com/vektra/mockery) and `generate.mocks` makefile target. The config that configures which interfaces to generate mocks for is `.mockery.yaml`.

The tests that are added in this PR are in 
- `controller/konnect/ops/ops_controlplane_test.go`
- `controller/konnect/ops/ops_kongservice_test.go`

More will follow after this PR gets reviewed and merged.

**Which issue this PR fixes**

Part of #447


.golangci.yaml
```diff
@@ -62,6 +62,13 @@ linters-settings:
 
       - pkg: github.com/kong/gateway-operator/internal/types
         alias: gwtypes
+
+      - pkg: github.com/Kong/sdk-konnect-go/models/components
+        alias: sdkkonnectgocomp
+      - pkg: github.com/Kong/sdk-konnect-go/models/operations
+        alias: sdkkonnectgoops
+      - pkg: github.com/Kong/sdk-konnect-go/models/sdkerrors
+        alias: sdkkonnectgoerrs
```
programmer04: I think this `go` in the middle is redundant. WDYT e.g. `sdkkonnectops` instead of `sdkkonnectgoops`? The same for other packages

programmer04 (commented):
Overall LGTM

.golangci.yaml
randmonkey: Changed the aliases.

.mockery.yaml
```diff
@@ -0,0 +1,15 @@
...
+filename: "{{ trimSuffix .InterfaceFile \".go\" | base | lower }}_mock_test.go"
+dir: "{{ .InterfaceDir }}"
+mockname: "Mock{{ .InterfaceName }}"
+outpkg: "{{ .PackageName }}"
+
+packages:
+  github.com/kong/gateway-operator/controller/konnect/ops:
+    interfaces:
+      ControlPlaneSDK:
+      ServicesSDK:
```
czeslavo, line 15: Nit: we might as well already generate mocks for other interfaces that are added in this PR (`ConsumerGroupSDK`, `ConsumersSDK`, `RoutesSDK`).

czeslavo (approved):

.mockery.yaml
randmonkey: I'll add mocks of all used SDKs in the following PR to refactor the usage of SDK.

PR #516, opened 2024-08-22
feat(cli): configurable certgen images

### LINKED ISSUE #515: Make certgen job images configurable
## Problem Statement

We have an environment where we don't allow to access external Docker repositories. The kong-gateway-operator-controller-manager creates Jobs with external images hence it fails to create them.

## Proposed Solution

Add two command line arguments to make this configurable by the user. To keep backward compatibility the default values for the arguments should be the same as the hard coded values were.

```
  -webhook-certificate-config-base-image string
    	The base image for the certgen Jobs. (default "registry.k8s.io/ingress-nginx/kube-webhook-certgen:v1.3.0")
  -webhook-certificate-config-shell-image string
    	The shell image for the certgen Jobs. (default "busybox")
```

## Additional Information

I already have a PR for this, I just wanted to create the issue first so I can reference it. Also if my PR is not accepted then someone else still can find this and implement it better.

## Acceptance Criteria
- The image fields in the gateway-operator-admission job pods should be the ones the user set in the command line.

akunszt:

Add two new command line options to the manager.

```
  -webhook-certificate-config-base-image string
    	The base image for the certgen Jobs. (default "registry.k8s.io/ingress-nginx/kube-webhook-certgen:v1.3.0")
  -webhook-certificate-config-shell-image string
    	The shell image for the certgen Jobs. (default "busybox")
```

Those are optional. If you omit them the defaults will be used which are the previous hard coded values.

By using these command line arguments the user can deploy kong-gateway-operator to a cluster where external Docker repositories are not reachable or prohibited to use them.

**Which issue this PR fixes**

Fixes #515.

I don't think this is a significant change and as this is my first PR I am not sure if I have to touch CHANGELOG.md or not.

**PR Readiness Checklist**:

### DIFF
--- a/modules/cli/cli.go
+++ b/modules/cli/cli.go
@@ -52,6 +52,8 @@ func New(m metadata.Info) *CLI {
 
 	// webhook and validation options
 	flagSet.BoolVar(&deferCfg.ValidatingWebhookEnabled, "enable-validating-webhook", true, "Enable the validating webhook.")
+	flagSet.StringVar(&deferCfg.WebhookCertificateConfigBaseImage, "webhook-certificate-config-base-image", consts.WebhookCertificateConfigBaseImage, "The base image for the certgen Jobs.")
+	flagSet.StringVar(&deferCfg.WebhookCertificateConfigShellImage, "webhook-certificate-config-shell-image", "busybox", "The shell image for the certgen Jobs.")
 
 	flagSet.BoolVar(&deferCfg.Version, "version", false, "Print version information.")
 
@@ -88,10 +90,12 @@ type CLI struct {
 }
 
 type flagsForFurtherEvaluation struct {
-	DisableLeaderElection    bool
-	ClusterCASecretNamespace string
-	ValidatingWebhookEnabled bool
-	Version                  bool
+	DisableLeaderElection              bool
+	ClusterCASecretNamespace           string
+	ValidatingWebhookEnabled           bool
+	Version                            bool
+	WebhookCertificateConfigBaseImage  string
+	WebhookCertificateConfigShellImage string
 }
 
 const (
@@ -167,6 +171,9 @@ func (c *CLI) Parse(arguments []string) manager.Config {
 		anonymousReportsEnabled = false
 	}
 
+	webhookCertificateConfigBaseImage := c.deferFlagValues.WebhookCertificateConfigBaseImage
+	webhookCertificateConfigShellImage := c.deferFlagValues.WebhookCertificateConfigShellImage
+
 	if c.deferFlagValues.Version {
 		type Version struct {
 			Release string `json:"release"`
@@ -218,6 +225,8 @@ func (c *CLI) Parse(arguments []string) manager.Config {
 	c.cfg.WebhookPort = manager.DefaultConfig().WebhookPort
 	c.cfg.LeaderElectionNamespace = controllerNamespace
 	c.cfg.AnonymousReports = anonymousReportsEnabled
+	c.cfg.WebhookCertificateConfigBaseImage = webhookCertificateConfigBaseImage
+	c.cfg.WebhookCertificateConfigShellImage = webhookCertificateConfigShellImage
 
 	return *c.cfg
 }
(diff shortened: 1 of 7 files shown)


modules/cli/cli.go
```diff
@@ -167,6 +171,9 @@ func (c *CLI) Parse(arguments []string) manager.Config {
 		anonymousReportsEnabled = false
 	}
 
+	webhookCertificateConfigBaseImage := c.deferFlagValues.WebhookCertificateConfigBaseImage
+	webhookCertificateConfigShellImage := c.deferFlagValues.WebhookCertificateConfigShellImage
```
pmalek: You shouldn't really need this here. Those 2 flags can simply be used as is.

This func is for flags that require more logic than just binding to a variable.

pmalek (commented):
The proposal looks reasonable. I made a comment about those flags not requiring to be set in the `Parse` func.

As for the changelog: yes, that would most certainly be a nice change to place it in there.

Thanks!

modules/cli/cli.go
akunszt: Thanks for the quick response. I modified the code as you requested.

akunszt:
I added a changelog entry. I hope I did it right, if not then please tell me how to fix it or feel free to correct it.

modules/cli/cli.go
```diff
@@ -52,6 +52,8 @@ func New(m metadata.Info) *CLI {
 
 	// webhook and validation options
 	flagSet.BoolVar(&deferCfg.ValidatingWebhookEnabled, "enable-validating-webhook", true, "Enable the validating webhook.")
+	flagSet.StringVar(&deferCfg.WebhookCertificateConfigBaseImage, "webhook-certificate-config-base-image", consts.WebhookCertificateConfigBaseImage, "The base image for the certgen Jobs.")
+	flagSet.StringVar(&deferCfg.WebhookCertificateConfigShellImage, "webhook-certificate-config-shell-image", "busybox", "The shell image for the certgen Jobs.")
```
czeslavo: Please consider extracting `busybox` to a constant, similarly to `consts.WebhookCertificateConfigBaseImage`, and using it in UTs.

modules/cli/cli.go
```diff
@@ -88,10 +90,12 @@ type CLI struct {
...
-	DisableLeaderElection    bool
-	ClusterCASecretNamespace string
-	ValidatingWebhookEnabled bool
-	Version                  bool
+	DisableLeaderElection              bool
+	ClusterCASecretNamespace           string
+	ValidatingWebhookEnabled           bool
+	Version                            bool
+	WebhookCertificateConfigBaseImage  string
+	WebhookCertificateConfigShellImage string
```
pmalek: This shouldn't be necessary as previously mentioned. Just add it to the `Config` struct: https://github.com/Kong/gateway-operator/blob/fa31122c75a4ff06710f7f71ce666c771bfc34f4/modules/manager/run.go#L61

modules/cli/cli.go
pmalek: Those 2 flags should be bound to the `Config` struct instead of the defer config as there's nothing to evaluate here.

pmalek (changes requested):
Looking good.

Left a couple of comments to move the flags to `Config`.

Please add a handful of UTs for those flags in https://github.com/Kong/gateway-operator/blob/a0614c5216d8ed3f83d6186365fa2b0d805443ea/modules/cli/cli_test.go

modules/cli/cli.go
akunszt: Got it. I just misunderstood the `flagsForFurtherEvaluation`. I thought that those are the flags which are still evaluated to keep, modify or drop by the team. But it is clear that further evaluation means that flags which needs some extra logic and cannot be assigned directly to the configuration.

I will create unit tests and then push the changes.

pmalek:
The Build container image failed because we're using a GH variable in https://github.com/Kong/gateway-operator/blob/771df50f9841309149e6361e1445b9bf4538e37d/.github/workflows/build.yaml#L29-L31 and those are [not accessible to PRs from forks](https://github.com/orgs/community/discussions/44322).

That is not required for the PR to get merged though at this moment (we should change that).

pmalek (approved):

pmalek:
Thanks @akunszt 🙇

PR #537, opened 2024-08-29
tests: add Konnect entities integration test

czeslavo:

Bootstraps an integration test against a Konnect dev environment.

**Which issue this PR fixes**

Part of https://github.com/Kong/gateway-operator/issues/481.

### DIFF
--- a/config/dev/manager_dev.yaml
+++ b/config/dev/manager_dev.yaml
@@ -24,6 +24,7 @@ spec:
             - -zap-devel=true
             - -enable-controller-kongplugininstallation=true
             - -enable-validating-webhook=true
+            - -enable-controller-konnect=true
           name: manager
           env:
           - name: GATEWAY_OPERATOR_DEVELOPMENT_MODE
--- a/config/dev/manager_dev.yaml
+++ b/config/dev/manager_dev.yaml
@@ -22,9 +22,9 @@ spec:
             - -cluster-ca-secret-namespace=kong-system
             - -zap-log-level=debug
             - -zap-devel=true
-            - -enable-controller-kongplugininstallation=true
-            - -enable-validating-webhook=true
-            - -enable-controller-konnect=true
+            - -enable-controller-kongplugininstallation
+            - -enable-validating-webhook
+            - -enable-controller-konnect
           name: manager
           env:
           - name: GATEWAY_OPERATOR_DEVELOPMENT_MODE
--- a/test/integration/test_konnect_entities.go
+++ b/test/integration/test_konnect_entities.go
@@ -23,8 +23,18 @@ import (
 )
 
 func TestKonnectEntities(t *testing.T) {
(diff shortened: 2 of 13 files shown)


config/dev/manager_dev.yaml
```diff
@@ -24,6 +24,7 @@ spec:
             - -zap-devel=true
             - -enable-controller-kongplugininstallation=true
             - -enable-validating-webhook=true
+            - -enable-controller-konnect=true
```
programmer04: Please add this to 
- `config/debug/manager_debug.yaml` (btw. this `=true` next to each flag seems redundant)
- [Makefile](https://github.com/Kong/gateway-operator/blob/ebb8b70cd7548f15cbf165898c6510e75afb79ac/Makefile#L471)

somewhere else? (idk)

test/integration/test_konnect_entities.go
```diff
@@ -0,0 +1,124 @@
...
+	testutils "github.com/kong/gateway-operator/pkg/utils/test"
+	"github.com/kong/gateway-operator/test"
+	"github.com/kong/gateway-operator/test/helpers"
+
+	configurationv1alpha1 "github.com/kong/kubernetes-configuration/api/configuration/v1alpha1"
+	konnectv1alpha1 "github.com/kong/kubernetes-configuration/api/konnect/v1alpha1"
+)
+
+func TestKonnectEntities(t *testing.T) {
+	ns, _ := helpers.SetupTestEnv(t, GetCtx(), GetEnv())
```
programmer04, line 33: Why don't we need a `cleaner` to be created here and all later-created objects added to it?

test/integration/test_konnect_entities.go
```diff
@@ -0,0 +1,124 @@
...
+
+	testID := uuid.NewString()[:8]
```
programmer04, line 38: Why the length of `8`? It's not obvious

programmer04 (changes requested):

test/integration/test_konnect_entities.go
czeslavo: A cleaner is created underneath anyway, and a whole namespace is deleted eventually. I couldn't use a cleaner to delete objects because it handles deletes in FIFO order and that won't work in this case: `KonnectAPIAuthConfiguration` shouldn't be deleted before any other object as that is required for others to complete their finalizer which is deleting a reflecting entity in Konnect. That's why I'm only cleaning up a `KonnectControlPlane` and waiting for its deletion synchronously to ensure it was successfully deleted along with its children, and the `KonnectAPIAuthConfiguration` is implicitly deleted with the namespace.

test/integration/test_konnect_entities.go
czeslavo: Added a comment (tldr: Konnect does not like too long names).

programmer04 (approved):
🚀

PR #550, opened 2024-09-04
Konnect entities: `KongPlugin` reconciler

### LINKED ISSUE #530: Konnect entities: KongPlugin reconciler
## Problem Statement

https://github.com/Kong/gateway-operator/pull/513 introduced the reconciler for the new `KongPluginBinding` CRD. Since we want to allow the legacy approach based on annotations as well, we need to introduce a new `KongPlugin` reconciler that watches for plugin annotations on entities and creates the underlying corresponding `KongPluginBinding`.

## Acceptance Criteria

- [ ] The plugin annotations of an object ar reconciled, and a corresponding `KongPluginBinding` is created

mlavacca:

**Which issue this PR fixes**

Fixes #530 

**PR Readiness Checklist**:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,10 @@
 - Add `KongService`, `KongRoute`, `KongConsumer`, and `KongConsumerGroup` watchers
   in the `KongPluginBinding` reconciler.
   [#571](https://github.com/Kong/gateway-operator/pull/571)
+- Annotating `KongService`s with the `konghq.com/plugins` annotation results in
+  the creation of a managed `KongPluginBInding` resource, which is taken by the
+  `KongPluginBinding` reconciler to create the corresponding plugin object in Konnect.
+  [#550](https://github.com/Kong/gateway-operator/pull/550)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,7 +67,7 @@
   in the `KongPluginBinding` reconciler.
   [#571](https://github.com/Kong/gateway-operator/pull/571)
 - Annotating `KongService`s with the `konghq.com/plugins` annotation results in
-  the creation of a managed `KongPluginBInding` resource, which is taken by the
+  the creation of a managed `KongPluginBinding` resource, which is taken by the
   `KongPluginBinding` reconciler to create the corresponding plugin object in Konnect.
   [#550](https://github.com/Kong/gateway-operator/pull/550)
 
--- a/config/samples/konnect-kongservice-plugin-annotated.yaml
+++ b/config/samples/konnect-kongservice-plugin-annotated.yaml
@@ -1,0 +1,49 @@
+kind: KonnectAPIAuthConfiguration
+apiVersion: konnect.konghq.com/v1alpha1
+metadata:
+  name: demo-auth
+  namespace: default
+spec:
+  type: token
+  token: kpat_XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
+  serverURL: eu.api.konghq.tech 
+---
+kind: KonnectGatewayControlPlane
+apiVersion: konnect.konghq.com/v1alpha1
+metadata:
+  name: demo-cp
+  namespace: default
+spec:
+  name: demo-cp
+  labels:
+    app: demo-cp
+    key1: demo-cp
+  konnect:
+    authRef:
+      name: demo-auth
+      # namespace not required if APIAuthConfiguration is in the same namespace
+---
+apiVersion: configuration.konghq.com/v1
+kind: KongPlugin
+metadata:
+  name: rate-limit-5-min
+  namespace: default
+config:
+  minute: 5
+  policy: local
+plugin: rate-limiting
+---
+kind: KongService
+apiVersion: configuration.konghq.com/v1alpha1
+metadata:
+  name: service-1
+  namespace: default
+  annotations:
+    konghq.com/plugins: rate-limit-5-min
+spec:
+  name: service-1
+  host: example.com
+  controlPlaneRef:
+    type: konnectNamespacedRef
+    konnectNamespacedRef:
+      name: demo-cp
--- a/controller/konnect/reconciler_kongplugin.go
+++ b/controller/konnect/reconciler_kongplugin.go
@@ -22,8 +22,8 @@ import (
 
 // KongPluginReconciler reconciles a KongPlugin object.
 type KongPluginReconciler struct {
-	DevelopmentMode bool
-	Client          client.Client
+	developmentMode bool
+	client          client.Client
 }
 
 // NewKongPluginReconciler creates a new KongPluginReconciler.
@@ -32,8 +32,8 @@ func NewKongPluginReconciler(
 	client client.Client,
 ) *KongPluginReconciler {
 	return &KongPluginReconciler{
-		DevelopmentMode: developmentMode,
-		Client:          client,
+		developmentMode: developmentMode,
+		client:          client,
 	}
 }
 
@@ -57,12 +57,12 @@ func (r *KongPluginReconciler) SetupWithManager(_ context.Context, mgr ctrl.Mana
 func (r *KongPluginReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	var (
 		entityTypeName = "KongPlugin"
-		logger         = log.GetLogger(ctx, entityTypeName, r.DevelopmentMode)
+		logger         = log.GetLogger(ctx, entityTypeName, r.developmentMode)
 	)
 
 	// Fetch the KongPlugin instance
 	var kongPlugin configurationv1.KongPlugin
-	if err := r.Client.Get(ctx, req.NamespacedName, &kongPlugin); err != nil {
+	if err := r.client.Get(ctx, req.NamespacedName, &kongPlugin); err != nil {
 		if k8serrors.IsNotFound(err) {
 			return ctrl.Result{}, nil
 		}
@@ -75,7 +75,7 @@ func (r *KongPluginReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 	// TODO(mlavacca): use indexers instead of listing all KongPluginBindings
 	pluginBindings := []configurationv1alpha1.KongPluginBinding{}
 	referencingBindingList := configurationv1alpha1.KongPluginBindingList{}
-	err := r.Client.List(ctx, &referencingBindingList, client.InNamespace(kongPlugin.Namespace))
+	err := r.client.List(ctx, &referencingBindingList, client.InNamespace(kongPlugin.Namespace))
 	if err != nil {
 		return ctrl.Result{}, err
 	}
@@ -103,7 +103,7 @@ func (r *KongPluginReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 	// Get all the KongServices referenced by the KongPluginBindings
 	// TODO(mlavacca): use indexers instead of listing all KongServices
 	kongServiceList := configurationv1alpha1.KongServiceList{}
-	err = r.Client.List(ctx, &kongServiceList, client.InNamespace(kongPlugin.Namespace))
+	err = r.client.List(ctx, &kongServiceList, client.InNamespace(kongPlugin.Namespace))
 	if err != nil {
 		return ctrl.Result{}, err
 	}
@@ -118,7 +118,7 @@ func (r *KongPluginReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 		// we need to remove the finalizer.
 		if len(pluginBindingsByServiceName[kongService.Name]) == 0 {
 			if controllerutil.RemoveFinalizer(&kongService, consts.CleanupPluginBindingFinalizer) {
-				if err = r.Client.Update(ctx, &kongService); err != nil {
+				if err = r.client.Update(ctx, &kongService); err != nil {
 					if k8serrors.IsConflict(err) {
 						return ctrl.Result{Requeue: true}, nil
 					}
@@ -132,14 +132,14 @@ func (r *KongPluginReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 		// if the KongService is marked for deletion, we need to delete all the PluginBindings that reference it.
 		if !kongService.DeletionTimestamp.IsZero() {
 			for _, pb := range pluginBindingsByServiceName[kongService.Name] {
-				if err := r.Client.Delete(ctx, &pb); err != nil {
+				if err := r.client.Delete(ctx, &pb); err != nil {
 					return ctrl.Result{}, err
 				}
 				log.Debug(logger, "KongPluginBinding deleted", pb)
 			}
 			// Once all the KongPluginBindings that use the KongService have been deleted, remove the finalizer.
 			controllerutil.RemoveFinalizer(&kongService, consts.CleanupPluginBindingFinalizer)
-			if err = r.Client.Update(ctx, &kongService); err != nil {
+			if err = r.client.Update(ctx, &kongService); err != nil {
 				return ctrl.Result{}, err
 			}
 			log.Debug(logger, "KongService finalizer removed", kongService)
@@ -205,14 +205,14 @@ func (r *KongPluginReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 						},
 					}
 					k8sutils.SetOwnerForObject(&kongPluginBinding, &kongPlugin)
-					if err = r.Client.Create(ctx, &kongPluginBinding); err != nil {
+					if err = r.client.Create(ctx, &kongPluginBinding); err != nil {
 						return ctrl.Result{}, err
 					}
 					log.Debug(logger, "Managed KongPluginBinding created", kongPluginBinding)
 				}
 				// add a finalizer to the KongService that means the associated kongPluginBindings need to be cleaned up
 				if controllerutil.AddFinalizer(&kongService, consts.CleanupPluginBindingFinalizer) {
-					if err = r.Client.Update(ctx, &kongService); err != nil {
+					if err = r.client.Update(ctx, &kongService); err != nil {
 						return ctrl.Result{}, err
 					}
 					log.Debug(logger, "KongService finalizer added", kongService)
@@ -223,7 +223,7 @@ func (r *KongPluginReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 
 		// iterate over all the KongPluginBindings to be deleted and delete them.
 		for _, pb := range pluginBindingsToDelete {
-			if err = r.Client.Delete(ctx, &pb); err != nil {
+			if err = r.client.Delete(ctx, &pb); err != nil {
 				if k8serrors.IsNotFound(err) {
 					continue
 				}
@@ -239,7 +239,7 @@ func (r *KongPluginReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 	// deleted or do not reference it anymore.
 	if pluginReferenceFound {
 		if controllerutil.AddFinalizer(&kongPlugin, consts.PluginInUseFinalizer) {
-			if err = r.Client.Update(ctx, &kongPlugin); err != nil {
+			if err = r.client.Update(ctx, &kongPlugin); err != nil {
 				return ctrl.Result{}, err
 			}
 			log.Debug(logger, "KongPlugin finalizer added", kongPlugin)
@@ -247,7 +247,7 @@ func (r *KongPluginReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 		}
 	} else {
 		if controllerutil.RemoveFinalizer(&kongPlugin, consts.PluginInUseFinalizer) {
-			if err = r.Client.Update(ctx, &kongPlugin); err != nil {
+			if err = r.client.Update(ctx, &kongPlugin); err != nil {
 				return ctrl.Result{}, err
 			}
 			log.Debug(logger, "KongPlugin finalizer removed", kongPlugin)
--- a/controller/konnect/reconciler_kongplugin.go
+++ b/controller/konnect/reconciler_kongplugin.go
@@ -63,10 +63,7 @@ func (r *KongPluginReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 	// Fetch the KongPlugin instance
 	var kongPlugin configurationv1.KongPlugin
 	if err := r.client.Get(ctx, req.NamespacedName, &kongPlugin); err != nil {
-		if k8serrors.IsNotFound(err) {
-			return ctrl.Result{}, nil
-		}
-		return ctrl.Result{}, err
+		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
 	log.Debug(logger, "reconciling", kongPlugin)
--- a/controller/konnect/reconciler_kongplugin.go
+++ b/controller/konnect/reconciler_kongplugin.go
@@ -101,6 +101,7 @@ func (r *KongPluginReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 
 	// Get all the KongServices referenced by the KongPluginBindings
 	// TODO(mlavacca): use indexers instead of listing all KongServices
+	// https://github.com/Kong/gateway-operator/issues/596
 	kongServiceList := configurationv1alpha1.KongServiceList{}
 	err = r.client.List(
 		ctx,
--- a/controller/konnect/reconciler_kongplugin.go
+++ b/controller/konnect/reconciler_kongplugin.go
@@ -87,6 +87,7 @@ func (r *KongPluginReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 
 	// TODO(mlavacca): So far we are supporting only KongService targets here. We need to implement
 	// the same logic for KongRoute, KongConsumer, and KongConsumerGroup as well.
+	// https://github.com/Kong/gateway-operator/issues/583
 
 	// Group the PluginBindings by KongService name
 	pluginBindingsByServiceName := map[string][]configurationv1alpha1.KongPluginBinding{}
--- a/controller/konnect/reconciler_kongplugin.go
+++ b/controller/konnect/reconciler_kongplugin.go
@@ -38,7 +38,7 @@ func NewKongPluginReconciler(
 }
 
 // SetupWithManager sets up the controller with the Manager.
-func (r *KongPluginReconciler) SetupWithManager(mgr ctrl.Manager) error {
+func (r *KongPluginReconciler) SetupWithManager(_ context.Context, mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		Named("KongPlugin").
 		For(&configurationv1.KongPlugin{}).
--- a/controller/konnect/reconciler_kongplugin.go
+++ b/controller/konnect/reconciler_kongplugin.go
@@ -189,7 +189,7 @@ func (r *KongPluginReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 				}
 			}
 
-			// iterate over all the KondService annotations
+			// iterate over all the KongService annotations
 			for _, pluginName := range pluginSlice {
 				if pluginName != kongPlugin.Name {
 					continue
--- a/controller/konnect/reconciler_kongplugin.go
+++ b/controller/konnect/reconciler_kongplugin.go
@@ -54,6 +54,8 @@ func (r *KongPluginReconciler) SetupWithManager(_ context.Context, mgr ctrl.Mana
 }
 
 // Reconcile reconciles a KongPlugin object.
+// The purpose of this reconciler is to handle annotations on Kong entities objects that reference KongPlugin objects.
+// As a result of such annotations, KongPluginBinding objects are created and managed by the controller.
 func (r *KongPluginReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	var (
 		entityTypeName = "KongPlugin"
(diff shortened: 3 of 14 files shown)


CHANGELOG.md
```diff
@@ -57,6 +57,9 @@
 - The `DataPlaneKonnectExtension` CRD has been introduced. Such a CRD can be attached
   to a `DataPlane` via the extensions field to have a konnect-flavored `DataPlane`.
   [#453](https://github.com/Kong/gateway-operator/pull/453)
+- Labeling `KongService`s with the `konghq.com/plugins` label results in the creation
+  of a KongPluginBInding resources, that gets reconciled and created in Konnect.
```
```suggestion
  of a KongPluginBinding resource, that gets reconciled and created in Konnect.
```

modules/manager/controller_setup.go
```diff
@@ -382,6 +384,13 @@ func SetupControllers(mgr manager.Manager, c *Config) (map[string]ControllerDef,
 			}
 			controllers[name] = controller
 		}
+		controllers[KongPluginControllerName] = ControllerDef{
+			Enabled: c.KonnectControllersEnabled,
+			Controller: konnect.NewKongPluginReconciler(
+				c.DevelopmentMode,
+				mgr.GetClient(),
+			),
+		}
```
czeslavo: Please add the reconciler to the map defined above: `konnectControllers`. It was added to prevent runtime overrides in case of duplicate map keys.

test/integration/test_kongplugin.go
```diff
@@ -0,0 +1,202 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	"github.com/kong/gateway-operator/pkg/consts"
+	"github.com/kong/gateway-operator/test/helpers"
+
+	configurationv1 "github.com/kong/kubernetes-configuration/api/configuration/v1"
+	configurationv1alpha1 "github.com/kong/kubernetes-configuration/api/configuration/v1alpha1"
+)
+
+func TestKongPlugins(t *testing.T) {
```
pmalek, line 28: WDYT about leaving a comment about the fact that this doesn't really sync anything to Konnect but is only testing the creation/deletion of KongPluginBindings?

modules/manager/controller_setup.go
pmalek: Please update to the scaffolding used above to prevent duplication of controllers.

controller/konnect/reconciler_kongplugin.go
```diff
@@ -0,0 +1,256 @@
...
+		).
+		Watches(
+			&configurationv1alpha1.KongService{},
+			handler.EnqueueRequestsFromMapFunc(r.mapKongServices),
+		).
+		Complete(r)
+}
+
+// Reconcile reconciles a KongPlugin object.
+func (r *KongPluginReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
czeslavo, line 59: Could we try splitting the `Reconcile` method into smaller functions? Now it's hard to quickly grasp what are the actions and their order from the reader perspective.

CHANGELOG.md
pmalek: > ... creation of a KongPluginBInding resources, that gets reconciled and created in Konnect

I'd suggest rewording this as the binding object is not what's created in Konnect but rather a plugin instance is created, binding to objects as per the binding spec. WDYT?

test/integration/test_kongplugin.go
```diff
@@ -0,0 +1,202 @@
...
+		return len(kongPluginBindingList.Items) == 0
+	}, 15*time.Second, time.Second)
+	require.Eventually(t, func() bool {
+		err := GetClients().MgrClient.Get(GetCtx(), client.ObjectKeyFromObject(proxyCachekongPlugin), proxyCachekongPlugin)
+		return k8serrors.IsNotFound(err)
+	}, 15*time.Second, time.Second)
+
+	t.Log("delete the kongservice and check it gets collected, as the kongPluginBinding finalizer should have been removed")
+	err = GetClients().MgrClient.Delete(GetCtx(), kongService)
+	require.NoError(t, err)
```
pmalek:
```suggestion
	require.NoError(t, GetClients().MgrClient.Delete(GetCtx(), kongService))
```

test/integration/test_kongplugin.go
```diff
@@ -0,0 +1,202 @@
...
+			Name:      proxyCachekongPluginName,
+			Namespace: ns.Name,
+		},
+		PluginName: "proxy-cache",
+		Config: apiextensionsv1.JSON{
+			Raw: []byte(`{"response_code": [200], "request_method": ["GET", "HEAD"], "content_type": ["text/plain; charset=utf-8"], "cache_ttl": 300, "strategy": "memory"}`),
+		},
+	}
+	err = GetClients().MgrClient.Create(GetCtx(), proxyCachekongPlugin)
+	require.NoError(t, err)
```
pmalek:
```suggestion
	require.NoError(t, GetClients().MgrClient.Create(GetCtx(), proxyCachekongPlugin))
```

test/integration/test_kongplugin.go
```diff
@@ -0,0 +1,202 @@
...
+		err := GetClients().MgrClient.List(GetCtx(), kongPluginBindingList, client.InNamespace(ns.Name))
+		if err != nil {
+			return false
+		}
+		if len(kongPluginBindingList.Items) != 1 {
+			return false
+		}
+		managedKPB = &kongPluginBindingList.Items[0]
+		return true
+	}, 30*time.Second, time.Second)
```
pmalek: Can we extract those into `const`s and reuse them across this test?

I'd also suggest decreasing the tick time from 1s to something like 250ms or so. There's a bunch of them in this test and that (tick period) is time it takes for the first predicate execution, hence they add up to the minimal time execution ( glancing at this test, it seems there are already 8 invocations like this which makes this test take at the very least 8s).

WDYT?

test/integration/test_kongplugin.go
```diff
@@ -0,0 +1,202 @@
...
+
+	t.Log("delete managed kongPluginBinding, then check it gets recreated")
+	err = GetClients().MgrClient.Delete(GetCtx(), managedKPB)
+	require.NoError(t, err)
```
pmalek:
```suggestion
	require.NoError(t, GetClients().MgrClient.Delete(GetCtx(), managedKPB))
```

test/integration/test_kongplugin.go
```diff
@@ -0,0 +1,202 @@
...
+		},
+		Spec: configurationv1alpha1.KongServiceSpec{
+			KongServiceAPISpec: configurationv1alpha1.KongServiceAPISpec{
+				Name: lo.ToPtr(ksName),
+				URL:  lo.ToPtr("http://example.com"),
+			},
+		},
+	}
+	err = GetClients().MgrClient.Create(GetCtx(), kongService)
+	require.NoError(t, err)
```
pmalek:
```suggestion
	require.NoError(t, GetClients().MgrClient.Create(GetCtx(), kongService))
```

test/integration/test_kongplugin.go
pmalek:
```suggestion
	require.NoError(t, GetClients().MgrClient.Create(GetCtx(), proxyCachekongPlugin))
```

test/integration/test_kongplugin.go
```diff
@@ -0,0 +1,202 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      rateLimitingkongPluginName,
+			Namespace: ns.Name,
+		},
+		PluginName: "rate-limiting",
+		Config: apiextensionsv1.JSON{
+			Raw: []byte(`{"minute": 5, "policy": "local"}`),
+		},
+	}
+	err := GetClients().MgrClient.Create(GetCtx(), rateLimitingkongPlugin)
```
pmalek: Nit: You might consider putting `GetClients().MgrClient` into its own variable.

test/integration/test_kongplugin.go
```diff
@@ -0,0 +1,202 @@
...
+			return false
+		}
+		if len(kongPluginBindingList.Items) != 1 {
+			return false
+		}
+		managedKPB = &kongPluginBindingList.Items[0]
+		return true
+	}, 30*time.Second, time.Second)
+
+	t.Log("delete managed kongPluginBinding, then check it gets recreated")
```
pmalek: Nit: add a name of PKB to the log entry.

test/integration/test_kongplugin.go
```diff
@@ -0,0 +1,202 @@
...
+		if err != nil {
+			return false
+		}
+		if len(kongPluginBindingList.Items) > 0 {
+			return false
+		}
+		return true
+	}, 15*time.Second, time.Second)
+
+	t.Log("deploying an unmanaged KongPluginBinding resource")
```
pmalek: Log the KPB name.

test/integration/test_kongplugin.go
```diff
@@ -0,0 +1,202 @@
...
+			return false
+		}
+		return len(kongPluginBindingList.Items) == 0
+	}, 15*time.Second, time.Second)
+	require.Eventually(t, func() bool {
+		err := GetClients().MgrClient.Get(GetCtx(), client.ObjectKeyFromObject(proxyCachekongPlugin), proxyCachekongPlugin)
+		return k8serrors.IsNotFound(err)
+	}, 15*time.Second, time.Second)
+
+	t.Log("delete the kongservice and check it gets collected, as the kongPluginBinding finalizer should have been removed")
```
pmalek: Log the service name.

controller/konnect/reconciler_kongplugin.go
```diff
@@ -0,0 +1,256 @@
...
+	for _, kongService := range kongServiceList.Items {
+		var pluginSlice []string
+
+		// in case no kongpluginbindings are referencing the kongservice, but it has the finalizer,
+		// we need to remove the finalizer.
+		if len(pluginBindingsByServiceName[kongService.Name]) == 0 {
+			if controllerutil.RemoveFinalizer(&kongService, consts.CleanupPluginBindingFinalizer) {
+				if err = r.Client.Update(ctx, &kongService); err != nil {
+					return ctrl.Result{}, err
+				}
```
pmalek: You might consider using `.IsConflict(err)` to explicitly handle the conflict here. Users shouldn't not be concerned about these types of errors.

Example: https://github.com/Kong/gateway-operator/blob/dcb56adfdf81438a191b4a1533f565a08c6fb704/controller/konnect/reconciler_generic.go#L151-L156

controller/konnect/reconciler_kongplugin.go
```diff
@@ -0,0 +1,256 @@
...
+		return ctrl.Result{}, err
+	}
+	for _, pluginBinding := range referencingBindingList.Items {
+		if pluginBinding.Spec.PluginReference.Name == kongPlugin.Name {
+			pluginBindings = append(pluginBindings, pluginBinding)
+		}
+	}
+
+	// TODO(mlavacca): So far we are supporting only KongService targets here. We need to implement
+	// the same logic for KongRoute, KongConsumer, and KongConsumerGroup as well.
```
pmalek, line 88: Can we have an issue for this linked here?

controller/konnect/reconciler_kongplugin.go
```diff
@@ -0,0 +1,256 @@
...
+	k8sutils "github.com/kong/gateway-operator/pkg/utils/kubernetes"
+
+	configurationv1 "github.com/kong/kubernetes-configuration/api/configuration/v1"
+	configurationv1alpha1 "github.com/kong/kubernetes-configuration/api/configuration/v1alpha1"
+)
+
+// KongPluginReconciler reconciles a KongPlugin object.
+type KongPluginReconciler struct {
+	DevelopmentMode bool
+	Client          client.Client
```
pmalek: These can be unexported given that we have a contructor that exposes those fields.

controller/konnect/watch_kongplugin.go
```diff
@@ -0,0 +1,62 @@
...
+func (r *KongPluginReconciler) mapKongServices(ctx context.Context, obj client.Object) []reconcile.Request {
+	logger := log.GetLogger(ctx, "KongPlugin", r.DevelopmentMode)
+	kongService, ok := obj.(*configurationv1alpha1.KongService)
+	if !ok {
+		log.Error(logger, errors.New("cannot cast object to KongService"), "KongService mapping handler", obj)
+		return []ctrl.Request{}
+	}
+
+	requests := []ctrl.Request{}
+	if plugins, ok := kongService.Annotations["konghq.com/plugins"]; ok {
```
pmalek: I'd consider putting the `konghq.com/plugins` into a named const.

pkg/consts/konnect.go
```diff
@@ -0,0 +1,11 @@
+package consts
+
+const (
+	// CleanupPluginBindingFinalizer is the finalizer attached to entities that are properly
+	// referenced by KongPluginBindings, that should be cleaned up when the KongService
+	// gets deleted.
```
pmalek: 1. I believe this comment should be a bit more explicit. What does "properly referenced" mean in here?
2. "that should be cleaned up when the KongService gets deleted":
  - I believe we can use a generic term like `resource` instead of KongService to prevent us forgetting to update this comment when other types of objects are supported.
  - What should clean this finalizer? I believe this is worth to be mentioned, otherwise users might think that they are somehow expected to clean this up.

pmalek (commented):
Looks very good 👍 Left a couple of comments to make this a bit clearer to follow. Let me know what you think about those.

CHANGELOG.md
mlavacca: Yep I agree, reworded.

modules/manager/controller_setup.go
mlavacca: 👍 see 0efc11a3d37c9300f515aa0bc07f70e213b34de4

test/integration/test_kongplugin.go
mlavacca: sure, please see fa311fe77fd5ea8c5eebe6a836333f10a343f8db

modules/manager/controller_setup.go
mlavacca: done in https://github.com/Kong/gateway-operator/commit/0efc11a3d37c9300f515aa0bc07f70e213b34de4

test/integration/test_kongplugin.go
mlavacca: yep, see db6de2b045694792c3f6f30193512c6fcb54765a

test/integration/test_kongplugin.go
mlavacca: yes, it makes sense. I've updated the tick time to 500 ms as 250 looked a bit too tight to me. See c5362d004ca9626c50b0b973215bf1ed7f0a4747

test/integration/test_kongplugin.go
mlavacca: 6ac46b88b163193a309eaaab096317cb6e743a28

test/integration/test_kongplugin.go
mlavacca: 6ac46b88b163193a309eaaab096317cb6e743a28

test/integration/test_kongplugin.go
mlavacca: 6ac46b88b163193a309eaaab096317cb6e743a28

controller/konnect/reconciler_kongplugin.go
mlavacca: it makes sense, here it is 375457f4017e5873c284f2652e217f0576535f2b

controller/konnect/reconciler_kongplugin.go
mlavacca: sure thing: 5692bc7ae19b5064b92c983ab21877e434a206a3

controller/konnect/watch_kongplugin.go
mlavacca: this is a tricky one. I'd love to have all our annotations gathered in a common place that can be consumed by KIC, KGO, or even 3rd party tools (see https://github.com/kubernetes-sigs/ingress2gateway/blob/main/pkg/i2gw/providers/kong/consts.go#L25-L31). I think that kubernetes-configuration would be a great place to gather them, WDYT?

pmalek: That could be a long term goal but as an interim step I'd suggest to just use 1 const in this repo. We can then use an alias for easy migration and then migrate to a const from another repo (like kubernetes-configuration). WDYT?

pkg/consts/konnect.go
mlavacca: yep, I agree. here it is c4a737641026a56f76d2fc0d75a84d0c41187ce8

controller/konnect/reconciler_kongplugin.go
mlavacca: 2172e05b25fa599ca8d3cc034536b74e0248702b

controller/konnect/watch_kongplugin.go
mlavacca: makes sense 033f86f350150c99a9439da6b31843223cf673fe

controller/konnect/reconciler_kongplugin.go
```diff
@@ -0,0 +1,278 @@
...
+			handler.EnqueueRequestsFromMapFunc(r.mapKongPluginBindings),
+		).
+		Watches(
+			&configurationv1alpha1.KongService{},
+			handler.EnqueueRequestsFromMapFunc(r.mapKongServices),
+		).
+		Complete(r)
+}
+
+// Reconcile reconciles a KongPlugin object.
```
czeslavo, line 56: I think it would be helpful for future readers to add a comment explaining briefly what is the responsibility of this reconciler - that might be not obvious it's meant to create `KongPluginBindings` rather than create plugins in Konnect.

controller/konnect/reconciler_kongplugin.go
```diff
@@ -0,0 +1,278 @@
...
+							return true
+						}
+						return false
+					}) {
+					// then mark it for deletion, as the plugin is not referenced by the KongService anymore
+					pluginBindingsToDelete = append(pluginBindingsToDelete, pb)
+				}
+			}
+
+			// iterate over all the KondService annotations
```
czeslavo:
```suggestion
			// iterate over all the KongService annotations
```

test/integration/test_kongplugin.go
```diff
@@ -0,0 +1,205 @@
...
+	configurationv1 "github.com/kong/kubernetes-configuration/api/configuration/v1"
+	configurationv1alpha1 "github.com/kong/kubernetes-configuration/api/configuration/v1alpha1"
+)
+
+// TestKongPlugins is an integration that aims at testing the KongPluginBinding resources
+// are properly created and cleaned up when Konnect entities get annotated with KongPlugin names.
+// It also tests that unmanaged KongPluginBindings are properly handled when managed KongPluginBindings
+// are in play.
+// NOTE: this test does not test the Konnect integration. No resource created here will be pushed to Konnect.
+func TestKongPlugins(t *testing.T) {
```
czeslavo, line 28: Loose suggestion: this could probably be an envtest to save some CI/local cycles.

controller/konnect/reconciler_kongplugin.go
```diff
@@ -0,0 +1,278 @@
...
+		client.InNamespace(kongPlugin.Namespace),
+	)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	for _, pluginBinding := range referencingBindingList.Items {
+		if pluginBinding.Spec.PluginReference.Name == kongPlugin.Name {
+			pluginBindings = append(pluginBindings, pluginBinding)
+		}
+	}
```
czeslavo: How about extracting this to a separate method `getPluginBindingsUsingKongPlugin`?

controller/konnect/reconciler_kongplugin.go
```diff
@@ -0,0 +1,278 @@
...
+	// Group the PluginBindings by KongService name
+	pluginBindingsByServiceName := map[string][]configurationv1alpha1.KongPluginBinding{}
+	for _, pluginBinding := range pluginBindings {
+		if pluginBinding.Spec.Targets.ServiceReference == nil ||
+			pluginBinding.Spec.Targets.ServiceReference.Group != configurationv1alpha1.GroupVersion.Group ||
+			pluginBinding.Spec.Targets.ServiceReference.Kind != "KongService" {
+			continue
+		}
+		pluginBindingsByServiceName[pluginBinding.Spec.Targets.ServiceReference.Name] = append(pluginBindingsByServiceName[pluginBinding.Spec.Targets.ServiceReference.Name], pluginBinding)
+	}
```
czeslavo: How about extracting this to a separate `groupPluginBindingsByKongServiceName` function?

controller/konnect/reconciler_kongplugin.go
```diff
@@ -0,0 +1,278 @@
...
+
+	pluginBindingsToDelete := []configurationv1alpha1.KongPluginBinding{}
+	// pluginReferenceFound represents whether the plugin is referenced by any KongService
+	var pluginReferenceFound bool
+	for _, kongService := range kongServiceList.Items {
+		var pluginSlice []string
+
+		// in case no kongpluginbindings are referencing the kongservice, but it has the finalizer,
+		// we need to remove the finalizer.
```
czeslavo, line 122: How will this work if we only list indexed Services (the ones referenced by KongPluginBindings) according to TODO? Won't we need a separate reconciler for `KongService`s to handle their finalizers?

controller/konnect/reconciler_kongplugin.go
```diff
@@ -0,0 +1,278 @@
...
+		if pluginBinding.Spec.Targets.ServiceReference == nil ||
+			pluginBinding.Spec.Targets.ServiceReference.Group != configurationv1alpha1.GroupVersion.Group ||
+			pluginBinding.Spec.Targets.ServiceReference.Kind != "KongService" {
+			continue
+		}
+		pluginBindingsByServiceName[pluginBinding.Spec.Targets.ServiceReference.Name] = append(pluginBindingsByServiceName[pluginBinding.Spec.Targets.ServiceReference.Name], pluginBinding)
+	}
+
+	// Get all the KongServices referenced by the KongPluginBindings
+	// TODO(mlavacca): use indexers instead of listing all KongServices
```
czeslavo, line 103: Can we have an issue tracking this?

controller/konnect/reconciler_kongplugin.go
```diff
@@ -0,0 +1,278 @@
...
+		client.InNamespace(kongPlugin.Namespace),
+	)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	pluginBindingsToDelete := []configurationv1alpha1.KongPluginBinding{}
+	// pluginReferenceFound represents whether the plugin is referenced by any KongService
+	var pluginReferenceFound bool
+	for _, kongService := range kongServiceList.Items {
```
czeslavo, line 118: I'm afraid that if we leave this code as is and attempt to handle other entities (routes, consumers, groups) in this reconciler, it will get very hard to maintain considering the volume of code in a single function we have right now. Please try to extract the `Service`-specific code to separate method(s) to avoid that.

CHANGELOG.md
```diff
@@ -64,6 +64,10 @@
 - Add `KongService`, `KongRoute`, `KongConsumer`, and `KongConsumerGroup` watchers
   in the `KongPluginBinding` reconciler.
   [#571](https://github.com/Kong/gateway-operator/pull/571)
+- Annotating `KongService`s with the `konghq.com/plugins` annotation results in
+  the creation of a managed `KongPluginBInding` resource, which is taken by the
```
```suggestion
  the creation of a managed `KongPluginBinding` resource, which is taken by the
```

controller/konnect/reconciler_kongplugin.go
```diff
@@ -0,0 +1,278 @@
...
+								Name: kongPlugin.Name,
+							},
+						},
+					}
+					k8sutils.SetOwnerForObject(&kongPluginBinding, &kongPlugin)
+					if err = r.client.Create(ctx, &kongPluginBinding); err != nil {
+						return ctrl.Result{}, err
+					}
+					log.Debug(logger, "Managed KongPluginBinding created", kongPluginBinding)
+				}
```
czeslavo, line 224: How about extracting this to a separate function `createKongPluginBindingForKongPlugin`? It could accept a struct with objects to be used as targets (in our case only a Service for now).

controller/konnect/reconciler_kongplugin.go
```diff
@@ -0,0 +1,278 @@
...
+		}
+	} else {
+		if controllerutil.RemoveFinalizer(&kongPlugin, consts.PluginInUseFinalizer) {
+			if err = r.client.Update(ctx, &kongPlugin); err != nil {
+				return ctrl.Result{}, err
+			}
+			log.Debug(logger, "KongPlugin finalizer removed", kongPlugin)
+			return ctrl.Result{}, nil
+		}
+	}
```
czeslavo, line 268: How about extracting this to a separate method, e.g. `handleKongPluginFinalizer`?

controller/konnect/reconciler_kongplugin.go
```diff
@@ -0,0 +1,278 @@
...
+		logger         = log.GetLogger(ctx, entityTypeName, r.developmentMode)
+	)
+
+	// Fetch the KongPlugin instance
+	var kongPlugin configurationv1.KongPlugin
+	if err := r.client.Get(ctx, req.NamespacedName, &kongPlugin); err != nil {
+		if k8serrors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
```
pmalek: You can simplify this by using:

```suggestion
		return ctrl.Result{}, client.IgnoreNotFound(err)
```

controller/konnect/reconciler_kongplugin.go
```diff
@@ -0,0 +1,278 @@
...
+	log.Debug(logger, "reconciling", kongPlugin)
+
+	// Get the pluginBindings that use this KongPlugin
+	pluginBindings := []configurationv1alpha1.KongPluginBinding{}
+	referencingBindingList := configurationv1alpha1.KongPluginBindingList{}
+	err := r.client.List(
+		ctx,
+		&referencingBindingList,
+		client.MatchingFields{
+			IndexFieldKongPluginBindingKongPluginReference: kongPlugin.Name,
```
pmalek: According to the `ExtractValue` func in here: https://github.com/Kong/gateway-operator/blob/4731ea64685befe828bbdcdc648d6f5e11660271/controller/konnect/index_kongpluginbinding.go#L60-L70

it extract the value from the plugin binding objects as `<ns>/<name>`. Doesn't that mean that this won't match and we need to prepend the `kongPlugin.Name` with `kongPlugin.Namespace + "/"`?

controller/konnect/reconciler_kongplugin.go
```diff
@@ -0,0 +1,278 @@
...
+		},
+		client.InNamespace(kongPlugin.Namespace),
+	)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	for _, pluginBinding := range referencingBindingList.Items {
+		if pluginBinding.Spec.PluginReference.Name == kongPlugin.Name {
```
pmalek: Do we actually need to do this? Doesn't using the index above guarantee us that we have only the requested plugin bindings (with matching ref name)?

test/integration/test_kongplugin.go
pmalek: Even though this is a good idea I believe we can continue with this integration test and add envtest based test in a separate PR (due to being pressed for time to release 1.4 on time).

We can leave a note here that this can be converted to envtest based test though. WDYT?

pmalek (commented):
I believe this is a good starting point.

Ideally we'd split the `Reconcile()` func into smaller ones but that can also be done when support for other entities is added after this gets merged.

1 thing I'd like to see in here though is a sample manifest in `config/samples`.

controller/konnect/reconciler_kongplugin.go
mlavacca: yep, the integration test was broken indeed :). I added this indexing in a rush and I overlooked this aspect. Fixed in 285eb9c353c58691f0a4477a73ed27eefa694738

controller/konnect/reconciler_kongplugin.go
mlavacca: yep, removed in 285eb9c353c58691f0a4477a73ed27eefa694738

controller/konnect/reconciler_kongplugin.go
mlavacca: Yes, In the tracking issue of KongService indexing I've added this detail: https://github.com/Kong/gateway-operator/issues/596

mlavacca:
Regarding the comments related to splitting up the logic here into smaller pieces, my point of view is that this is the first PR that poses the basis to iterate on it. We already have multiple issues in place (https://github.com/Kong/gateway-operator/issues/583, https://github.com/Kong/gateway-operator/issues/596) that address the improvements on this side, and looking for a way now, in the very first iteration loop to optimize code readability and modularity as much as possible is out of scope IMO. Once we'll add the missing pieces (other Kong entities, entities indexing, etc.), the context will be broad enough to improve the code modularity. As of now, I think we should consider it as a starting point, as @pmalek already pointed out.

config/samples/konnect-kongservice-plugin-annotated.yaml
pmalek, line 1: 👍

controller/konnect/reconciler_kongplugin.go
czeslavo: As we agreed, we're going to improve that in the upcoming PRs.

controller/konnect/reconciler_kongplugin.go
czeslavo: As we agreed, we're going to improve that in the upcoming PRs.

controller/konnect/reconciler_kongplugin.go
czeslavo: As we agreed, we're going to improve that in the upcoming PRs.

controller/konnect/reconciler_kongplugin.go
czeslavo: As we agreed, we're going to improve that in the upcoming PRs.

controller/konnect/reconciler_kongplugin.go
czeslavo: As we agreed, we're going to improve that in the upcoming PRs.

czeslavo (approved):

pmalek (approved):

pmalek:
I've resolved conflicts after merging #593.

Should be able to be merged now.

PR #558, opened 2024-09-06
test: add remaining konnect entities integration

### LINKED ISSUE #481: Konnect entities: add integration tests against an actual Konnect org
### Problem statement

Konnect entities currently lack tests. This issue tracks the effort of adding those.

### Acceptance criteria

- [ ] Integration tests against an actual Konnect dev cluster exist

czeslavo:

Adds remaining Konnect entities integration tests and fixes issues found in those.

**Which issue this PR fixes**

Closes https://github.com/Kong/gateway-operator/issues/481.

### DIFF
--- a/controller/konnect/ops/ops_kongpluginbinding.go
+++ b/controller/konnect/ops/ops_kongpluginbinding.go
@@ -262,8 +262,11 @@ func kongPluginBindingToSDKPluginInput(
 	}
 
 	pluginConfig := map[string]any{}
-	if err := json.Unmarshal(plugin.Config.Raw, &pluginConfig); err != nil {
-		return nil, err
+	if rawConfig := plugin.Config.Raw; rawConfig != nil {
+		// If the config is empty (a valid case), there's no need to unmarshal (as it would fail).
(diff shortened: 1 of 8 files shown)


controller/konnect/ops/ops_kongpluginbinding.go
```diff
@@ -262,8 +262,11 @@ func kongPluginBindingToSDKPluginInput(
 	}
 
 	pluginConfig := map[string]any{}
-	if err := json.Unmarshal(plugin.Config.Raw, &pluginConfig); err != nil {
-		return nil, err
+	if rawConfig := plugin.Config.Raw; rawConfig != nil {
```
randmonkey, line 265: I think it worth a detailed comment to explain that empty `plugin.Config.Raw` is valid but it would make the unmarshal fail so we need to skip the unmarshal when it is empty.

czeslavo: Changed the comment to mention it's a valid case.

randmonkey (approved):

PR #562, opened 2024-09-09
chore: remove deprecated exportloopref from golangci-lint config

pmalek:

Remove `exportloopref` from golangci-lint config to prevent the following deprecation notice when running `make lint`:

```
WARN The linter 'exportloopref' is deprecated (since v1.60.2) due to: Since Go1.22 (loopvar) this linter is no longer relevant. Replaced by copyloopvar.
```


programmer04 (approved):
🏅

PR #588, opened 2024-09-12
chore: group dependabot k8s libs bumps

pmalek:

Group k8s libs bumps in dependabot config as per : https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups


programmer04 (approved):
🆒

PR #594, opened 2024-09-12
fix: conformance tests run properly locally

### LINKED ISSUE #444: Conformance tests fail locally
## Current Behavior

The conformance tests are part of the CI and run at every PR. They work just fine [in the CI,](https://github.com/Kong/gateway-operator/actions/runs/10148932923/job/28062714376) but when we try to run them locally, a bunch are failing.

I personally hit this issue, and @programmer04 confirmed the same behavior on his machine.

## Expected Behavior

Conformance tests must be successful when run locally

## Steps To Reproduce

`make test.conformance`

## Operator Version

`v1.3.0`, and main as well.

mlavacca:

This PR has two fixes:

- generation is always enforced when initializing and updating a status. the status needs to have the proper generation always set.
- looser timeouts have been set to avoid flakiness

**Which issue this PR fixes**

Fixes #444 

**PR Readiness Checklist**:


controller/gateway/controller_reconciler_utils.go
```diff
@@ -555,11 +555,18 @@ func supportedRoutesByProtocol() map[gatewayv1.ProtocolType]map[gatewayv1.Kind]s
 // It also sets the listeners Programmed condition by setting the underlying
 // Listener Programmed status to false.
 func (g *gatewayConditionsAndListenersAwareT) initProgrammedAndListenersStatus() {
-	k8sutils.InitProgrammed(g)
+	k8sutils.SetCondition(
```
pmalek, line 558: I can see that this change pretty much only sets the generation in addition to what has been done previously but what I'm concerned about is that we seem to always set this condition regardless if that condition has already been set on `Gateway` or not.

https://github.com/Kong/gateway-operator/blob/dcb56adfdf81438a191b4a1533f565a08c6fb704/controller/gateway/controller.go#L169-L186

Shouldn't that be checking if `Programmed` has been set and only initialize it if it's not there yet? Just like we do it for listeners in https://github.com/Kong/gateway-operator/blob/508daaf373bf6ff7b9b06b81b9ded9e976516d28/controller/gateway/controller_reconciler_utils.go#L560-L562 ?

mlavacca: It needs to be like this, as the programmed condition is initialized as false and set to true only in case all the other conditions are true. This means that when you have a programmed Gateway and the ControlPlane gets deleted, the programmed condition is set to false again thanks to this initialization a priori. Once the ControlPlane will be provisioned and ready again, the programmed condition will transition to true, as the Gateway Controller will get to the bottom of the reconciliation loop. 

Does it make sense to you?

pmalek: > This means that when you have a programmed Gateway and the ControlPlane gets deleted, the programmed condition is set to false again thanks to this initialization a priori

Ok I just checked the logic (e.g. in https://github.com/Kong/gateway-operator/blob/dcb56adfdf81438a191b4a1533f565a08c6fb704/controller/gateway/controller.go#L526-L538 and now it makes sense to me). I thought that `Programmed` would be set in those places as well but as long as we set it to `False` here and not perform the update in happy path before the bottom of reconciliation loop it sounds good.

pmalek (approved):

PR #602, opened 2024-09-13
fix: fix removing plugin-in-use finalizer

pmalek:

This PR fixes removing the `plugin-in-use` finalizer when `KongPluginBinding` gets removed.

We need to check if the `KongPluginBinding` that we're trying to remove is already removed and not return. When we return, the reconciler does not get to the `KongPlugin` finalizers update and `plugin-in-use` finalizer does not get removed.

This PR also removes `test/integration/test_kongplugin.go` and adds envtest based tests in exchange.


programmer04 (approved):

czeslavo (approved):

pmalek:
This won't work in the integration test because we're not creating the Konnect ControlPlane (and  Plugin and Service do not get programmed).

Im working on envtest based test for this.

pmalek:
@czeslavo @programmer04 PTAL now. I've migrated from integration to envtest based test.

test/envtest/main_test.go
```diff
@@ -0,0 +1,42 @@
...
+		EncodeLevel:    zapcore.LowercaseLevelEncoder,
+		EncodeTime:     zapcore.RFC3339TimeEncoder,
+		EncodeDuration: zapcore.SecondsDurationEncoder,
+		EncodeCaller:   zapcore.ShortCallerEncoder,
+	})
+
+	core := zapcore.NewCore(encoder, zapcore.AddSync(output), zap.DebugLevel)
+
+	return zap.New(core)
+}
```
czeslavo:
```suggestion
```
Appears to be unused.

test/envtest/kongpluginbinding_managed_test.go
```diff
@@ -0,0 +1,240 @@
...
+	require.NoError(t, err)
+	t.Cleanup(func() { w.Stop() })
+	kongPluginBinding := watchFor(t, ctx, w, watch.Added,
+		func(kpb *configurationv1alpha1.KongPluginBinding) bool {
+			return kpb.Spec.Targets.ServiceReference != nil &&
+				kpb.Spec.Targets.ServiceReference.Name == kongService.Name &&
+				kpb.Spec.PluginReference.Name == rateLimitingkongPlugin.Name
+		},
+		"KongPluginBinding wasn't created",
+	)
```
czeslavo: Question out of curiosity: was there any specific reason not to rely on the usual way we handle eventual conditions (`require.Eventually`) but using `Watch`?

pmalek: I wanted to test this out as it feels more natural (might need some refactoring love though): we get notified as soon as something happens, not based on a time period.

czeslavo: I understand it can be better in terms of the time spent on waiting as we receive an event as soon as it happens instead of using a fixed interval. I'm just wondering if we'd like to switch to this approach to not have a mix of both in the repo.

pmalek: I'm happy to hear feedback on this approach.

test/envtest/watch.go
```diff
@@ -0,0 +1,48 @@
...
+	eventType watch.EventType,
+	predicate func(T) bool,
+	failMsg string,
+) T {
+	t.Helper()
+
+	var ret T
+	for found := false; !found; {
+		select {
+		case <-ctx.Done():
```
czeslavo, line 62: It might make sense to set a timeout per `watchFor` call as waiting for the test context timeout might be too long for a single condition.

test/envtest/watch.go
```diff
@@ -0,0 +1,48 @@
...
+
+// watchFor watches for an event of type eventType using the provided watch.Interface.
+// It returns when either context is done - and then it marks the test as failed -
+// or when the event has been received and predicate returned true.
+func watchFor[
+	T metav1.Object,
+](
+	t *testing.T,
+	ctx context.Context,
+	w watch.Interface,
```
czeslavo, line 49: Do you think we could make this helper also run the watch itself instead of relying on the caller to pass it down? This way we could take care of cleaning the watch up in this function, making the caller code leaner.

pmalek: Yes, but the problem is that we first need to set up the watch and then we can perform an action that the effect of which we'll be catching with the watch.

When we encapsulate the watch within the `watchFor` we can't do it. I've tried adding a `preactions ... func(*testing.T, context.Context)` argument to allow optional actions which would be called between the watch set up and the for loop but I'm not convinced about this. 

LMKWYT.

test/envtest/watch.go
pmalek: Added the timeout PTAL.

test/envtest/watch.go
czeslavo: Ah, I see. It makes sense to me now. `preactions` doesn't sound intuitive to me either. 

Possibly, we could have a separate wrapper that creates a watch and sets up its cleanup to reduce boilerplate as much as we can. So instead of 
```go
	w, err := clientWithWatch.Watch(ctx, &configurationv1alpha1.KongPluginBindingList{}, client.InNamespace(ns.Name))
	require.NoError(t, err)
	t.Cleanup(func() { w.Stop() })
```
we'd have
```go
	w := setupWatch(ctx, t, clientWithWatch, &configurationv1alpha1.KongPluginBindingList{}, client.InNamespace(ns.Name))
```

pmalek: Makes sense 👍 c375eadd65916df85a3ffeb35ff775b9103f8997

Let me know if that looks reasonable.

czeslavo (approved):

test/envtest/deploy_resources.go
```diff
@@ -34,6 +36,33 @@ func deployKonnectAPIAuthConfiguration(
...
+func deployKonnectAPIAuthConfigurationWithID(
+	t *testing.T,
+	ctx context.Context,
+	cl client.Client,
+) *konnectv1alpha1.KonnectAPIAuthConfiguration {
+	t.Helper()
+
+	apiAuth := deployKonnectAPIAuthConfiguration(t, ctx, cl)
+
 	apiAuth.Status.Conditions = []metav1.Condition{
```
czeslavo, line 57: It's already done in `deployKonnectAPIAuthConfiguration`, isn't it? The comment also says about populating the Konnect status, but I cannot see it happening.

pmalek: Good catch! Thanks. bbcffcbc44674042511ddced9fd67c2b192f072b should fix this.

czeslavo (approved):

PR #614, opened 2024-09-18
chore(refactor): extract and make generic checking `ReferenceGrant` for `Secret`

programmer04:

Move code related to checking `ReferenceGrant` for `Secret` to a dedicated package and make it reusable. It will be used not only by `Gateway`, but also by `KongPluginInstallation` (see actual implementation https://github.com/Kong/gateway-operator/pull/615), so make it generic too.

**Which issue this PR fixes**

Prerequisite for https://github.com/Kong/gateway-operator/issues/610

Unit test is moved and adjusted to check whether it works properly for other Kind. We maintain no integration test because conformance tests cover this feature for `Gateway` and it proves that changes from this PR work as expected.

### DIFF
--- a/controller/pkg/secrets/ref/ref.go
+++ b/controller/pkg/secrets/ref/ref.go
@@ -32,7 +32,7 @@ func ReferenceGrantForSecretFrom(group gatewayv1.Group, kind gatewayv1.Kind) pre
 	)
 }
 
-// IsReferenceGrantForObj checks if the ReferenceGrant in .Spec.From specifies Group and Kind of the object.
+// IsReferenceGrantForObj checks if ReferenceGrant's from clause matches the provided object's Group, Kind and namespace.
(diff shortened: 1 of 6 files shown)


controller/pkg/secrets/ref/ref.go
```diff
@@ -0,0 +1,125 @@
...
+				if from.Kind == kind && from.Group == group {
+					return true
+				}
+			}
+			return false
+		},
+	)
+}
+
+// IsReferenceGrantForObj checks if the ReferenceGrant in .Spec.From specifies Group and Kind of the object.
```
pmalek:
```suggestion
// IsReferenceGrantForObj checks if ReferenceGrant's from clause matches the provided object's Group, Kind and namespace.
```

pmalek (approved):
Just 1 minor nit, otherwise 👍

PR #625, opened 2024-09-19
feat(konnect): add support for basic auth credentials for consumers

### LINKED ISSUE #617: Konnect entities: Credentials - Basic Auth
### Problem statement

Introduce `KongCredentialBasicAuth`.

API details: https://kong-platform-api.netlify.app/konnect/control-planes-config/v2/openapi.yaml/#tag/Basic-auth-credentials

### Proposed solution

- Introduce `KongCredentialBasicAuth` CRD in https://github.com/Kong/kubernetes-configuration
  - This CRD will contain a `KongConsumer` reference. 
  - ControlPlane reference will be inferred from `KongConsumer` reference
  - Only references that refer to Konnect will result in reconciliation against Konnect.

### Acceptance criteria

- [x] KGO can create, delete and update DataPlane's client certificates in Konnect based on the CRD spec.

pmalek:

This PR adds a first reconciler and scaffolding around it for consumer credentials.

Initial support comes with basic auth credentials.

Related k-cfg PR: https://github.com/Kong/kubernetes-configuration/pull/83

~Blocked until above gets merged.~

**Which issue this PR fixes**

Fixes #617

**PR Readiness Checklist**:


CHANGELOG.md
```diff
@@ -70,6 +70,8 @@
   the creation of a managed `KongPluginBinding` resource, which is taken by the
   `KongPluginBinding` reconciler to create the corresponding plugin object in Konnect.
   [#550](https://github.com/Kong/gateway-operator/pull/550)
+- Add support for `KognConsumer` credentials:
```
```suggestion
- Add support for `KongConsumer` credentials:
```

controller/konnect/ops/ops_credentialbasicauth.go
```diff
@@ -0,0 +1,183 @@
...
+		cred,
+	)
+
+	return nil
+}
+
+// updateRoute updates the Konnect Route entity.
+// It is assumed that provided KongRoute has Konnect ID set in status.
+// It returns an error if the KongRoute does not have a ControlPlaneRef or
+// if the operation fails.
```
czeslavo:
```suggestion
// updateCredentialBasicAuth updates the Konnect BasicAuth entity.
// It is assumed that the provided BasicAuth has Konnect ID set in status.
// It returns an error if the BasicAuth does not have a ControlPlaneRef or
// if the operation fails.
```

controller/konnect/ops/ops_credentialbasicauth.go
```diff
@@ -0,0 +1,183 @@
...
+		),
+		cred,
+	)
+
+	return nil
+}
+
+// deleteRoute deletes a KongRoute in Konnect.
+// It is assumed that provided KongRoute has Konnect ID set in status.
+// It returns an error if the operation fails.
```
czeslavo:
```suggestion
// deleteCredentialBasicAuth deletes a BasicAuth credential in Konnect.
// It is assumed that the provided BasicAuth has Konnect ID set in status.
// It returns an error if the operation fails.
```

controller/konnect/reconciler_generic.go
```diff
@@ -534,7 +588,8 @@ func getServiceRef[T constraints.SupportedKonnectEntityType, TEnt constraints.En
 		*configurationv1beta1.KongConsumerGroup,
 		*konnectv1alpha1.KonnectGatewayControlPlane,
 		*configurationv1alpha1.KongPluginBinding,
-		*configurationv1alpha1.KongUpstream:
+		*configurationv1alpha1.KongUpstream,
+		*configurationv1alpha1.CredentialBasicAuth:
```
czeslavo: How about changing this to work similarly to the just added `getConsumerRef`: only keep explicit cases for the types that support Service references, and return `mo.None[configurationv1alpha1.ServiceRef]()` on the default case?

controller/konnect/watch_credentialbasicauth.go
```diff
@@ -0,0 +1,209 @@
...
+//   reference from the object
+// - lists have their items stored in Items field, not returned via a method
+
+// CredentialBasicAuthReconciliationWatchOptions returns the watch options for
+// the CredentialBasicAuth.
+func CredentialBasicAuthReconciliationWatchOptions(
+	cl client.Client,
+) []func(*ctrl.Builder) *ctrl.Builder {
+	return []func(*ctrl.Builder) *ctrl.Builder{
+		func(b *ctrl.Builder) *ctrl.Builder {
```
czeslavo, line 37: Shouldn't we enqueue credentials requests also on the owning `KongConsumer` changes?

controller/konnect/reconciler_generic.go
pmalek: Initially I thought that `panic` would be better to always remember to add the implementation for a newly added type but perhaps that's not necessary.

Additionally this will require less changes for types that this shouldn't concern.

PTAL 3489f1d6551d55460fa946f98e006ab09274d936

controller/konnect/watch_credentialbasicauth.go
pmalek: I'm curious if that's actually necessary 🤔 

The only thing that would be relevant for the credentials here would be the consumer ID or its CP ID which currently cannot change because we disallow migrations between CPs.

Do you see any particular reason that we should enqueue credentials on consumer changes?

czeslavo: When the `KongConsumer` is deleted, it won't automatically trigger events for the credentials. In the case of `cascade=foreground` deletion that will make the process wait until the credential gets reconciled again on its fixed updated interval. For the `background` it will also make the final deletion of a credential postponed until the fixed update reconciliation is triggered.

controller/konnect/reconciler_generic.go
```diff
@@ -169,6 +170,26 @@ func (r *KonnectEntityReconciler[T, TEnt]) Reconcile(
 	} else if res.Requeue {
 		return res, nil
 	}
+	// If a type has a KongConsumer ref, handle it.
+	res, err = handleKongConsumerRef(ctx, r.Client, ent)
+	if err != nil {
+		if !errors.As(err, &ReferencedKongConsumerIsBeingDeleted{}) {
```
czeslavo: We should also handle the consumer's `NotFound` here to make the cascading deletion of the credential work as expected. Otherwise, after removing a `KongConsumer`, a credential gets stuck on `Reconciler error - {"controller": "CredentialBasicAuth", "controllerGroup": "configuration.konghq.com", "controllerKind": "CredentialBasicAuth", "CredentialBasicAuth": {"name":"basic-auth-1","namespace":"default"}, "namespace": "default", "name": "basic-auth-1", "reconcileID": "3934eb68-137c-4298-805a-370f3fdd0054", "error": "can't get the referenced KongConsumer default/consumer1: KongConsumer.configuration.konghq.com \"consumer1\" not found"}`.

Another aspect is that the `Consumer` can be removed from Konnect without removing its children in advance (the credentials will be deleted in cascade) - that's a different behavior than in the `Route` - `Service` relation where you cannot delete a service without dropping the routes depending on it first. That makes the deletion of entities owned by a removed/being removed consumer not necessary.

I experimented a bit and prepared a patch that seems to work correctly:

```diff
diff --git a/controller/konnect/errors.go b/controller/konnect/errors.go
index 0f1c86c6..bc9442e6 100644
--- a/controller/konnect/errors.go
+++ b/controller/konnect/errors.go
@@ -46,3 +46,14 @@ type ReferencedKongConsumerIsBeingDeleted struct {
 func (e ReferencedKongConsumerIsBeingDeleted) Error() string {
        return fmt.Sprintf("referenced Kong Consumer %s is being deleted", e.Reference)
 }
+
+// ReferencedKongConsumerDoesNotExist is an error type that is returned when
+type ReferencedKongConsumerDoesNotExist struct {
+       Reference types.NamespacedName
+       Err       error
+}
+
+// Error implements the error interface.
+func (e ReferencedKongConsumerDoesNotExist) Error() string {
+       return fmt.Sprintf("referenced Kong Consumer %s does not exist: %v", e.Reference, e.Err)
+}
diff --git a/controller/konnect/reconciler_generic.go b/controller/konnect/reconciler_generic.go
index faf71354..12b5dad3 100644
--- a/controller/konnect/reconciler_generic.go
+++ b/controller/konnect/reconciler_generic.go
@@ -173,20 +173,33 @@ func (r *KonnectEntityReconciler[T, TEnt]) Reconcile(
        // If a type has a KongConsumer ref, handle it.
        res, err = handleKongConsumerRef(ctx, r.Client, ent)
        if err != nil {
-               if !errors.As(err, &ReferencedKongConsumerIsBeingDeleted{}) {
-                       return ctrl.Result{}, err
-               }
-
-               // If the referenced KongConsumer is being deleted (has a non zero deletion timestamp)
-               // then we remove the entity if it has not been deleted yet (deletion timestamp is zero).
-               // We do this because Konnect blocks deletion of entities like Credentials
-               // if they contain entities like Routes.
-               if ent.GetDeletionTimestamp().IsZero() {
-                       if err := r.Client.Delete(ctx, ent); err != nil {
-                               return ctrl.Result{}, fmt.Errorf("failed to delete %s: %w", client.ObjectKeyFromObject(ent), err)
+               consumerGoneOrBeingDeleted := errors.As(err, &ReferencedKongConsumerDoesNotExist{}) || errors.As(err, &ReferencedKongConsumerIsBeingDeleted{})
+               if consumerGoneOrBeingDeleted {
+                       // If the referenced KongConsumer is not found or is being deleted and the object is being deleted, remove the
+                       // finalizer and let the deletion proceed without trying to delete the entity from Konnect as the KongConsumer
+                       // deletion will (or already has - in case of the consumer being gone) take care of it on the Konnect side.
+                       if !ent.GetDeletionTimestamp().IsZero() {
+                               if controllerutil.RemoveFinalizer(ent, KonnectCleanupFinalizer) {
+                                       if err := r.Client.Update(ctx, ent); err != nil {
+                                               if k8serrors.IsConflict(err) {
+                                                       return ctrl.Result{Requeue: true}, nil
+                                               }
+                                               return ctrl.Result{}, fmt.Errorf("failed to remove finalizer %s: %w", KonnectCleanupFinalizer, err)
+                                       }
+                                       log.Debug(logger, "konnect finalizer removed as the owning KongConsumer is being deleted or is already gone", ent)
+                               }
+                       } else {
+                               // If the entity for unknown reason is not being deleted yet (it should be if the owning KongConsumer is gone
+                               // or is being deleted), we should delete it.
+                               if err := r.Client.Delete(ctx, ent); err != nil {
+                                       return ctrl.Result{}, fmt.Errorf("failed to delete %s: %w", client.ObjectKeyFromObject(ent), err)
+                               }
+                               log.Debug(logger, "entity deleted as the owning KongConsumer is being deleted or is already gone", ent)
                        }
                        return ctrl.Result{}, nil
                }
+
+               return ctrl.Result{}, err
        } else if res.Requeue {
                return res, nil
        }
@@ -776,6 +789,9 @@ func handleKongConsumerRef[T constraints.SupportedKonnectEntityType, TEnt constr
                ); errStatus != nil || res.Requeue {
                        return res, errStatus
                }
+               if k8serrors.IsNotFound(err) {
+                       return ctrl.Result{}, ReferencedKongConsumerDoesNotExist{Reference: nn, Err: err}
+               }
 
                return ctrl.Result{}, fmt.Errorf("can't get the referenced KongConsumer %s: %w", nn, err)
        }
```

controller/konnect/watch_credentialbasicauth.go
pmalek: I somehow forgot about the deletion. You're right.

4ce45b527d31e19f2e4f7c07698bd373ace5a7b9 should handle that (also added checking this to the test)

controller/konnect/reconciler_generic.go
pmalek: After having a chat with Greg and after receiving confirmation from Konnect eng that Consumer deletion will not be blocked when Consumer has credentials assigned to it I've refactored it to 2253c07fdb073d2b8ad8764386a7eb04039762d5

PTAL.

controller/konnect/errors.go
```diff
@@ -46,3 +46,14 @@ type ReferencedKongConsumerIsBeingDeleted struct {
 func (e ReferencedKongConsumerIsBeingDeleted) Error() string {
 	return fmt.Sprintf("referenced Kong Consumer %s is being deleted", e.Reference)
 }
+
+// ReferencedKongConsumerDoesNotExist is an error type that is returned when
```
czeslavo:
```suggestion
// ReferencedKongConsumerDoesNotExist is an error type that is returned when the referenced KongConsumer does not exist.
```

controller/konnect/reconciler_generic.go
```diff
@@ -777,15 +795,21 @@ func handleKongConsumerRef[T constraints.SupportedKonnectEntityType, TEnt constr
...
 	if delTimestamp := consumer.GetDeletionTimestamp(); !delTimestamp.IsZero() {
-		return ctrl.Result{}, ReferencedKongConsumerIsBeingDeleted{
-			Reference: nn,
-		}
+		return ctrl.Result{
+				Requeue:      true,
+				RequeueAfter: time.Until(delTimestamp.Time),
+			}, ReferencedKongConsumerIsBeingDeleted{
+				Reference: nn,
+			}
```
czeslavo: How about not changing the result here - it's kind of an antipattern to return a value and an error at once. We can decide on requeueing on the caller side.

pmalek: 👍 d56733225fa4af9c780bc8d747bd01cb02adc769

PTAL.

czeslavo (approved):

PR #627, opened 2024-09-20
feat(konnect):  KongTarget reconciler

### LINKED ISSUE #576: Konnect entities: KongTarget
### Problem statement

Introduce `KongTarget`.

### Proposed solution

- Introduce `KongTarget` CRD in https://github.com/Kong/kubernetes-configuration
  - This CRD will contain a ControlPlane reference. Only references that refer to Konnect will result in reconciliation against Konnect.
  - This CRD will contain an Upstream ref to reference which upstream this should be bound to (required via: https://github.com/Kong/sdk-konnect-go/blob/39a27951b473927fce9c56c13445d917fa2cec0c/models/operations/createtargetwithupstream.go#L13-L14)

### Acceptance criteria

- [x] KGO can create, delete and update `Target`s in Konnect based on the CRD spec.

randmonkey:
Add `KongTarget` reconciler to reconcile `KongTarget`s.
**Which issue this PR fixes**

Major part of #576 

**PR Readiness Checklist**:


controller/konnect/reconciler_upstreamref.go
```diff
@@ -0,0 +1,185 @@
...
+	switch e := any(e).(type) {
+	case *configurationv1alpha1.KongTarget:
+		// Since upstreamRef is required for KongTarget, we directly return spec.UpstreamRef here.
+		return mo.Some(e.Spec.UpstreamRef)
+	default:
+		return mo.None[configurationv1alpha1.TargetRef]()
+	}
+}
+
+// handleKongUpstreamRef handles KongUpstram reference if the entity referenced some KongUpstream.
```
pmalek:
```suggestion
// handleKongUpstreamRef handles KongUpstream reference if the entity references a KongUpstream.
```

controller/konnect/reconciler_upstreamref_test.go
```diff
@@ -0,0 +1,366 @@
...
+)
+
+type handleUpstreamRefTestCase[T constraints.SupportedKonnectEntityType, TEnt constraints.EntityType[T]] struct {
+	name                string
+	ent                 TEnt
+	objects             []client.Object
+	expectResult        ctrl.Result
+	expectError         bool
+	expectErrorContains string
+	// REVIEW: add assertions on modified ent?
```
pmalek: I think we can check things like owner ref and conditions 👍

controller/konnect/reconciler_upstreamref.go
```diff
@@ -0,0 +1,185 @@
...
+		conditions.KongUpstreamRefReasonValid,
+		fmt.Sprintf("Referenced KongUpstream %s programmed", nn),
+	); errStatus != nil || res.Requeue {
+		return res, errStatus
+	}
+
+	cpRef, ok := getControlPlaneRef(kongUpstream).Get()
+	// REVIEW: the logic will cause KongTarget referencing to a KongUpstream that is not controlled by Konnect entity reconcilers
+	// fall into endless reconcile backoff (so do KongRoute).
+	// In such case, should we just ignore the KongTarget and treat it as not controlled?
```
pmalek: That's a valid point. Let's create an issue for this and solve this for those entities that have this problem.

Sadly https://github.com/kubernetes-sigs/controller-runtime/issues/617 - which would allow returning error and not causing a requeue - is not getting implemented anytime soon as it seems from the comments.

controller/konnect/reconciler_upstreamref.go
```diff
@@ -0,0 +1,185 @@
...
+			conditions.ControlPlaneRefValidConditionType,
+			metav1.ConditionFalse,
+			conditions.ControlPlaneRefReasonInvalid,
+			err.Error(),
+		); errStatus != nil || res.Requeue {
+			return res, errStatus
+		}
+		if k8serrors.IsNotFound(err) {
+			// REVIEW: `getCPForRef` generates a new error but does not wrap the original error from client.Get so this is never reached.
+			// Should we change that?
```
pmalek: Good catch! Yes, let's change that.

controller/konnect/reconciler_upstreamref.go
randmonkey: Done.

controller/konnect/reconciler_upstreamref_test.go
randmonkey: Added assertions.

controller/konnect/reconciler_upstreamref.go
randmonkey: Created #629  and added comments here.

controller/konnect/reconciler_upstreamref.go
randmonkey: done.

controller/konnect/reconciler_generic.go
```diff
@@ -211,6 +211,27 @@ func (r *KonnectEntityReconciler[T, TEnt]) Reconcile(
...
+		// If the referenced KongUpstream is being deleted (has a non zero deletion timestamp)
+		// then we remove the entity if it has not been deleted yet (deletion timestamp is zero).
+		// We do this because Konnect blocks deletion of entities like Services/Upstreams
+		// if they contain entities like Routes/Targets.
+		if ent.GetDeletionTimestamp().IsZero() {
+			if err := r.Client.Delete(ctx, ent); err != nil {
+				return ctrl.Result{}, fmt.Errorf("failed to delete %s: %w", client.ObjectKeyFromObject(ent), err)
+			}
+			return ctrl.Result{}, nil
+		}
```
pmalek: I've just checked and Konnect allows removing upstreams that contain targets.

Given this, I suggest we implement the same handling as it has been done for consumers and their credentials: https://github.com/Kong/gateway-operator/blob/ca99be3217aab7c788b27d925669aa8250414d09/controller/konnect/reconciler_generic.go#L134-L152

pmalek (commented):
Looks solid 👍 

Please add a sample manifest in `config/samples/` for Upstream with Target so that users know how this can be used (and also so that it's tested via: https://github.com/Kong/gateway-operator/blob/83369ea831df4870f26a9e23434dd70eb1749cda/.github/workflows/tests.yaml#L69-L96)

pmalek:
When rebasing please use the functions to set programmed condition on entities introduced in https://github.com/Kong/gateway-operator/pull/633.

controller/konnect/reconciler_generic.go
randmonkey: Would it run cascade delete on targets if we call API to delete an upstream? If not, it would bring orphaned targets and we cannot delete them.

randmonkey: Updated.

randmonkey:
> When rebasing please use the functions to set programmed condition on entities introduced in #633.

Done and added sample `KongTarget`.

controller/konnect/reconciler_generic.go
pmalek: > Would it run cascade delete on targets if we call API to delete an upstream? If not, it would bring orphaned targets and we cannot delete them.

I can see we're (correctly) setting the owner reference in https://github.com/Kong/gateway-operator/pull/627/files#diff-bb95aba82e21609e440b7242dc888eda4d9b80d8639c6f336d223035ee199d09R99 (as we do for other refs) so this should cascade delete the target.

In any case an envtest based test for this would be nice to add in a separate  PR ( similarly as we have it for other types).

pmalek (approved):

PR #633, opened 2024-09-23
refactor(konnect): refactor setting programmed condition

pmalek:

Reduce code duplication by adding `SetKonnectEntityProgrammedConditionFalse` and `SetKonnectEntityProgrammedCondition` which are used to set entity's programmed condition.

### DIFF
--- a/controller/konnect/ops/conditions.go
+++ b/controller/konnect/ops/conditions.go
@@ -1,0 +1,61 @@
+package ops
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/kong/gateway-operator/controller/konnect/conditions"
+	"github.com/kong/gateway-operator/pkg/consts"
+	k8sutils "github.com/kong/gateway-operator/pkg/utils/kubernetes"
+)
+
+type entityType interface {
+	SetConditions([]metav1.Condition)
+	GetConditions() []metav1.Condition
+	GetGeneration() int64
+}
+
+// SetKonnectEntityProgrammedCondition sets the KonnectEntityProgrammed condition to true
+// on the provided object.
+func SetKonnectEntityProgrammedCondition(
+	obj entityType,
+) {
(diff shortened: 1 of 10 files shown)


controller/konnect/ops/conditions.go
```diff
@@ -0,0 +1,61 @@
...
+		k8sutils.NewConditionWithGeneration(
+			conditions.KonnectEntityProgrammedConditionType,
+			status,
+			reason,
+			msg,
+			obj.GetGeneration(),
+		),
+		obj,
+	)
+}
```
czeslavo, line 61: I believe starting with a lowercase is enough to signal the method is private. If you think it needs better isolation, please consider moving it to a separate package.
```suggestion
func SetKonnectEntityProgrammedCondition(
	obj entityType,
) {
	setKonnectEntityProgrammedConditon(
		obj,
		metav1.ConditionTrue,
		conditions.KonnectEntityProgrammedReasonProgrammed,
		"",
	)
}

// SetKonnectEntityProgrammedConditionFalse sets the KonnectEntityProgrammed condition
// to false on the provided object.
func SetKonnectEntityProgrammedConditionFalse(
	obj entityType,
	reason consts.ConditionReason,
	msg string,
) {
	setKonnectEntityProgrammedConditon(
		obj,
		metav1.ConditionFalse,
		reason,
		msg,
	)
}

func setKonnectEntityProgrammedConditon(
	obj entityType,
	status metav1.ConditionStatus,
	reason consts.ConditionReason,
	msg string,
) {
	k8sutils.SetCondition(
		k8sutils.NewConditionWithGeneration(
			conditions.KonnectEntityProgrammedConditionType,
			status,
			reason,
			msg,
			obj.GetGeneration(),
		),
		obj,
	)
}
```

pmalek: Agreed. Applied the suggestion.

czeslavo (approved):

PR #644, opened 2024-09-25
feat(konnect): make KongPlugin reconciler create KongPluginBindings for plugins attached to KongRoutes and KongServices

### LINKED ISSUE #525: Konnect entities: allow KongRoute to be targeted by a KongPluginBinding
## Problem Statement

#513 introduced the `KongPluginBinding` reconciler, but only the ServiceRef is taken in consideration. We need to allow the targeting of `KongRoute` as well.

## Acceptance Criteria

- [x] The field `.spec.targets.RouteRef` of `KongPluginBinding`s is properly handled by the `KongPLuginBinding` reconciler.

pmalek:

This PR introduces mechanism of creating `KongPluginBinding`s for both `KongRoute`s and `KongService`s based on annotation on those.

A mechanism to clean unused `KongPluginBinding`s has also been introduced.

Creating a cartesian product of possible relations (if e.g. the same plugin is used on `KongService` and `KongRoute` create 1 binding with both set as targets) is based on https://github.com/Kong/kubernetes-ingress-controller/blob/ee797b4e84bd176526af32ab6db54f16ee9c245b/internal/util/relations.go with 1 small change:

For `KongConsumer`, `KongService` and `KongRoute` relations it creates a single entry (and thus a single `KongPluginBinding`) where KIC (and the linked implementation) would create 2: 

- 1 for `KongConsumer` and `KongService`
- 1 for `KongConsumer` and `KongRoute`

**Which issue this PR fixes**

Fixes #525 

- `envtest` based test will follow in a separate PR to not make this PR any bigger.

- This introduces a small gap where `KongPluginBinding` (and thus Konnect plugin entity) can be deleted and recreated when target set of a plugin has changed, for instance when `KongService` and `KongRoute` are annotated with the same plugin and then `KongRoute` gets that annotation removed the reconciler will remove the `KongPluginBinding` as the target set doesn't match and then recreate a new one, having only the service as target.
  
  Implementing an improvement for this would require adding a mechanism which could map the plugin relations (combinations of resources having the `konghq.com/plugins` annotation set) to existing `KongPluginBinding`s.

**PR Readiness Checklist**:

### DIFF
--- a/controller/konnect/index_kongroute.go
+++ b/controller/konnect/index_kongroute.go
@@ -14,8 +14,6 @@ import (
 const (
 	// IndexFieldKongRouteOnReferencedPluginNames is the index field for KongRoute -> KongPlugin.
 	IndexFieldKongRouteOnReferencedPluginNames = "kongRouteKongPluginRef"
-	// IndexFieldKongRouteOnServiceReference is the index field for KongRoute -> Service.
-	IndexFieldKongRouteOnServiceReference = "kongRouteServiceRef"
 )
 
 // IndexOptionsForKongRoute returns required Index options for KongRoute reconciler.
--- a/controller/konnect/index_kongroute.go
+++ b/controller/konnect/index_kongroute.go
@@ -1,0 +1,46 @@
+package konnect
+
+import (
+	"strings"
+
+	"github.com/samber/lo"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/kong/gateway-operator/pkg/consts"
+
+	configurationv1alpha1 "github.com/kong/kubernetes-configuration/api/configuration/v1alpha1"
+)
+
+const (
+	// IndexFieldKongRouteOnReferencedPluginNames is the index field for KongRoute -> KongPlugin.
+	IndexFieldKongRouteOnReferencedPluginNames = "kongRouteKongPluginRef"
+	// IndexFieldKongRouteOnServiceReference is the index field for KongRoute -> Service.
+	IndexFieldKongRouteOnServiceReference = "kongRouteServiceRef"
+)
+
+// IndexOptionsForKongRoute returns required Index options for KongRoute reconciler.
+func IndexOptionsForKongRoute() []ReconciliationIndexOption {
+	return []ReconciliationIndexOption{
+		{
+			IndexObject:  &configurationv1alpha1.KongRoute{},
+			IndexField:   IndexFieldKongRouteOnReferencedPluginNames,
+			ExtractValue: kongRouteUsesPlugins,
+		},
+	}
+}
+
+func kongRouteUsesPlugins(object client.Object) []string {
+	route, ok := object.(*configurationv1alpha1.KongRoute)
+	if !ok {
+		return nil
+	}
+	ann, ok := route.Annotations[consts.PluginsAnnotationKey]
+	if !ok {
+		return nil
+	}
+
+	namespace := route.GetNamespace()
+	return lo.Map(strings.Split(ann, ","), func(p string, _ int) string {
+		return namespace + "/" + p
+	})
+}
--- a/controller/konnect/index_kongroute.go
+++ b/controller/konnect/index_kongroute.go
@@ -1,12 +1,9 @@
 package konnect
 
 import (
-	"strings"
-
-	"github.com/samber/lo"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	"github.com/kong/gateway-operator/pkg/consts"
+	"github.com/kong/gateway-operator/pkg/annotations"
 
 	configurationv1alpha1 "github.com/kong/kubernetes-configuration/api/configuration/v1alpha1"
 )
@@ -32,13 +29,5 @@ func kongRouteUsesPlugins(object client.Object) []string {
 	if !ok {
 		return nil
 	}
-	ann, ok := route.Annotations[consts.PluginsAnnotationKey]
-	if !ok {
-		return nil
-	}
-
-	namespace := route.GetNamespace()
-	return lo.Map(strings.Split(ann, ","), func(p string, _ int) string {
-		return namespace + "/" + p
-	})
+	return annotations.ExtractPlugins(route)
 }
--- a/controller/konnect/index_kongservice.go
+++ b/controller/konnect/index_kongservice.go
@@ -1,14 +1,11 @@
 package konnect
 
 import (
-	"strings"
-
-	"github.com/samber/lo"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	"github.com/kong/gateway-operator/pkg/consts"
-
 	configurationv1alpha1 "github.com/kong/kubernetes-configuration/api/configuration/v1alpha1"
+
+	"github.com/kong/gateway-operator/pkg/annotations"
 )
 
 const (
@@ -32,13 +29,6 @@ func kongServiceUsesPlugins(object client.Object) []string {
 	if !ok {
 		return nil
 	}
-	ann, ok := svc.Annotations[consts.PluginsAnnotationKey]
-	if !ok {
-		return nil
-	}
 
-	namespace := svc.GetNamespace()
-	return lo.Map(strings.Split(ann, ","), func(p string, _ int) string {
-		return namespace + "/" + p
-	})
+	return annotations.ExtractPlugins(svc)
 }
--- a/controller/konnect/index_kongservice.go
+++ b/controller/konnect/index_kongservice.go
@@ -3,9 +3,9 @@ package konnect
 import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	configurationv1alpha1 "github.com/kong/kubernetes-configuration/api/configuration/v1alpha1"
-
 	"github.com/kong/gateway-operator/pkg/annotations"
+
+	configurationv1alpha1 "github.com/kong/kubernetes-configuration/api/configuration/v1alpha1"
 )
 
 const (
--- a/controller/konnect/index_kongservice.go
+++ b/controller/konnect/index_kongservice.go
@@ -1,0 +1,44 @@
+package konnect
+
+import (
+	"strings"
+
+	"github.com/samber/lo"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/kong/gateway-operator/pkg/consts"
+
+	configurationv1alpha1 "github.com/kong/kubernetes-configuration/api/configuration/v1alpha1"
+)
+
+const (
+	// IndexFieldKongServiceOnReferencedPluginNames is the index field for KongService -> KongPlugin.
+	IndexFieldKongServiceOnReferencedPluginNames = "kongServiceKongPluginRef"
+)
+
+// IndexOptionsForKongService returns required Index options for KongService reconciler.
+func IndexOptionsForKongService() []ReconciliationIndexOption {
+	return []ReconciliationIndexOption{
+		{
+			IndexObject:  &configurationv1alpha1.KongService{},
+			IndexField:   IndexFieldKongServiceOnReferencedPluginNames,
+			ExtractValue: kongServiceUsesPlugins,
+		},
+	}
+}
+
+func kongServiceUsesPlugins(object client.Object) []string {
+	svc, ok := object.(*configurationv1alpha1.KongService)
+	if !ok {
+		return nil
+	}
+	ann, ok := svc.Annotations[consts.PluginsAnnotationKey]
+	if !ok {
+		return nil
+	}
+
+	namespace := svc.GetNamespace()
+	return lo.Map(strings.Split(ann, ","), func(p string, _ int) string {
+		return namespace + "/" + p
+	})
+}
--- a/controller/konnect/kongpluginbuilder.go
+++ b/controller/konnect/kongpluginbuilder.go
@@ -1,7 +1,12 @@
 package konnect
 
 import (
+	"fmt"
+
 	configurationv1alpha1 "github.com/kong/kubernetes-configuration/api/configuration/v1alpha1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 // KongPluginBindingBuilder helps to build KongPluginBinding objects.
@@ -66,6 +71,17 @@ func (b *KongPluginBindingBuilder) WithRouteTarget(routeName string) *KongPlugin
 		Name:  routeName,
 	}
 	return b
+}
+
+func (b *KongPluginBindingBuilder) WithOwnerReference(owner client.Object, scheme *runtime.Scheme) (*KongPluginBindingBuilder, error) {
+	opts := []controllerutil.OwnerReferenceOption{
+		controllerutil.WithBlockOwnerDeletion(true),
+	}
+	if err := controllerutil.SetOwnerReference(owner, b.binding, scheme, opts...); err != nil {
+		return nil, fmt.Errorf("failed to set owner reference: %w", err)
+	}
+
+	return b, nil
 }
 
 // Build returns the KongPluginBinding.
--- a/controller/konnect/kongpluginbuilder.go
+++ b/controller/konnect/kongpluginbuilder.go
@@ -1,0 +1,75 @@
+package konnect
+
+import (
+	configurationv1alpha1 "github.com/kong/kubernetes-configuration/api/configuration/v1alpha1"
+)
+
+// KongPluginBindingBuilder helps to build KongPluginBinding objects.
+type KongPluginBindingBuilder struct {
+	binding *configurationv1alpha1.KongPluginBinding
+}
+
+// NewKongPluginBindingBuilder creates a new KongPluginBindingBuilder.
+func NewKongPluginBindingBuilder() *KongPluginBindingBuilder {
+	return &KongPluginBindingBuilder{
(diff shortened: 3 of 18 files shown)


controller/konnect/index_kongroute.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/kong/gateway-operator/pkg/consts"
+
+	configurationv1alpha1 "github.com/kong/kubernetes-configuration/api/configuration/v1alpha1"
+)
+
+const (
+	// IndexFieldKongRouteOnReferencedPluginNames is the index field for KongRoute -> KongPlugin.
+	IndexFieldKongRouteOnReferencedPluginNames = "kongRouteKongPluginRef"
+	// IndexFieldKongRouteOnServiceReference is the index field for KongRoute -> Service.
+	IndexFieldKongRouteOnServiceReference = "kongRouteServiceRef"
```
czeslavo: It's unused - was the intention to add this index?

controller/konnect/index_kongroute.go
```diff
@@ -0,0 +1,46 @@
...
+	if !ok {
+		return nil
+	}
+	ann, ok := route.Annotations[consts.PluginsAnnotationKey]
+	if !ok {
+		return nil
+	}
+
+	namespace := route.GetNamespace()
+	return lo.Map(strings.Split(ann, ","), func(p string, _ int) string {
```
czeslavo: Please trim whitespace in case of something like `konghq.com/plugins: plugin1, plugin2`.

controller/konnect/index_kongservice.go
```diff
@@ -0,0 +1,44 @@
...
+	if !ok {
+		return nil
+	}
+	ann, ok := svc.Annotations[consts.PluginsAnnotationKey]
+	if !ok {
+		return nil
+	}
+
+	namespace := svc.GetNamespace()
+	return lo.Map(strings.Split(ann, ","), func(p string, _ int) string {
```
czeslavo: Ditto.

controller/konnect/kongpluginbuilder.go
```diff
@@ -0,0 +1,75 @@
...
+	b.binding.Spec.Targets.RouteReference = &configurationv1alpha1.TargetRefWithGroupKind{
+		Group: configurationv1alpha1.GroupVersion.Group,
+		Kind:  "KongRoute",
+		Name:  routeName,
+	}
+	return b
+}
+
+// Build returns the KongPluginBinding.
+func (b *KongPluginBindingBuilder) Build() *configurationv1alpha1.KongPluginBinding {
```
czeslavo, line 90: Please consider validating here whether all required fields were set to avoid invalid KPBs in the result.

controller/konnect/kongpluginbuilder.go
```diff
@@ -0,0 +1,75 @@
...
+func (b *KongPluginBindingBuilder) WithPluginRef(pluginName string) *KongPluginBindingBuilder {
+	b.binding.Spec.PluginReference.Name = pluginName
+	return b
+}
+
+// WithControlPlaneRef sets the control plane reference of the KongPluginBinding.
+func (b *KongPluginBindingBuilder) WithControlPlaneRef(ref *configurationv1alpha1.ControlPlaneRef) *KongPluginBindingBuilder {
+	// TODO: Cross check this with other types of ControlPlaneRefs
+	// used by Route, Consumer and/or ConsumerGroups that also bind this plugin
+	// in this KongPluginBinding spec.
```
czeslavo: Can we create an issue for this?

controller/konnect/ops/ops_kongpluginbinding.go
```diff
@@ -241,7 +253,14 @@ func kongPluginWithTargetsToKongPluginInput(
...
 				ID: lo.ToPtr(t.GetKonnectStatus().ID),
 			}
-		// TODO(mlavacca): add support for KongRoute
+		case *configurationv1alpha1.KongRoute:
+			id := t.GetKonnectID()
+			if id == "" {
+				return nil, fmt.Errorf("KongRoute %s is not configured in Konnect yet", client.ObjectKeyFromObject(t))
+			}
+			pluginInput.Route = &sdkkonnectcomp.PluginRoute{
+				ID: lo.ToPtr(t.GetKonnectStatus().ID),
```
czeslavo:
```suggestion
				ID: lo.ToPtr(id),
```
Same could be applied to `KongService` code above.

controller/konnect/watch_kongplugin.go
```diff
@@ -8,43 +8,67 @@ import (
...
-						Name:      kp.Name,
-					},
-				})
-			}
-		}
+	ann, ok := kongService.Annotations[consts.PluginsAnnotationKey]
+	if !ok {
+		return nil
+	}
+	plugins := strings.Split(ann, ",")
```
czeslavo: Trim whitespace

controller/konnect/reconciler_kongplugin.go
```diff
@@ -88,130 +103,124 @@ func (r *KongPluginReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
-	pluginBindingsToDelete := []configurationv1alpha1.KongPluginBinding{}
-	// pluginReferenceFound represents whether the plugin is referenced by any KongService
+	kongRouteList := configurationv1alpha1.KongRouteList{}
+	err = clientWithNamespace.List(ctx, &kongRouteList,
+		client.MatchingFields{
+			IndexFieldKongRouteOnReferencedPluginNames: kongPlugin.Namespace + "/" + kongPlugin.Name,
+		},
+	)
+	if err != nil {
+		return ctrl.Result{}, err
```
czeslavo:
```suggestion
		return ctrl.Result{}, fmt.Errorf("failed listing KongRoutes referencing %s KongPlugin: %w", client.ObjectKeyFromObject(kongPlugin), err)
```

controller/konnect/reconciler_kongplugin.go
```diff
@@ -88,130 +103,124 @@ func (r *KongPluginReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
-		ctx,
-		&kongServiceList,
-		client.InNamespace(kongPlugin.Namespace),
+	err = clientWithNamespace.List(ctx, &kongServiceList,
+		client.MatchingFields{
+			IndexFieldKongServiceOnReferencedPluginNames: kongPlugin.Namespace + "/" + kongPlugin.Name,
+		},
 	)
 	if err != nil {
 		return ctrl.Result{}, err
```
czeslavo:
```suggestion
		return ctrl.Result{}, fmt.Errorf("failed listing KongServices referencing %s KongPlugin: %w", client.ObjectKeyFromObject(kongPlugin), err)
```

controller/konnect/reconciler_kongplugin.go
```diff
@@ -88,130 +103,124 @@ func (r *KongPluginReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+		),
+		// TODO consumers and consumer groups
+	}
+	grouped, err := foreignRelations.GroupByControlPlane(ctx, r.client)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// pluginReferenceFound represents whether the plugin is referenced by any resource.
 	var pluginReferenceFound bool
```
czeslavo: Please move this declaration just before the `for cpNN, relations := range groupedCombinations` loop (to keep it as close to its usage as possible).

controller/konnect/reconciler_kongplugin.go
```diff
@@ -88,130 +103,124 @@ func (r *KongPluginReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
-	for _, kongService := range kongServiceList.Items {
-		if !kongService.DeletionTimestamp.IsZero() {
-			continue
-		}
+	groupedCombinations := grouped.GetCombinations()
 
-		var pluginSlice []string
+	// Delete the KongPluginBindings that are not used anymore.
+	if err := deleteUnusedKongPluginBindings(ctx, logger, clientWithNamespace, &kongPlugin, groupedCombinations, kongPluginBindingList.Items); err != nil {
+		return ctrl.Result{}, err
```
czeslavo:
```suggestion
		return ctrl.Result{}, fmt.Errorf("failed deleting unused KongPluginBindings for %s KongPlugin: %w", client.ObjectKeyFromObject(kongPlugin), err)
```

controller/konnect/reconciler_kongplugin.go
```diff
@@ -242,31 +251,189 @@ func (r *KongPluginReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+			ownerRef.Name == kongPlugin.Name &&
+			ownerRef.UID == kongPlugin.UID
+	}
 }
 
-func mapKongPluginBindingsByTargetTypeAndRef(bindings []configurationv1alpha1.KongPluginBinding) kongPluginBindingMapping {
-	ret := kongPluginBindingMapping{
-		byServiceName: map[string][]configurationv1alpha1.KongPluginBinding{},
-		byRouteName:   map[string][]configurationv1alpha1.KongPluginBinding{},
+func deleteUnusedKongPluginBindings(
```
czeslavo, line 287: This function contains quite complex logic that ought to be tested. It's a good candidate for unit tests using a fake cache client I believe.

controller/konnect/reconciler_kongplugin.go
```diff
@@ -242,31 +251,189 @@ func (r *KongPluginReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+				break
+			}
+
+			if !combinationFound {
+				pluginBindingsToDelete[client.ObjectKeyFromObject(&pb)] = pb
+				continue
+			}
+
+			s, serviceExists := serviceMap[serviceRef.Name]
+			r, routeExists := routeMap[routeRef.Name]
```
czeslavo: I believe that instead of listing all the services and routes in the namespace, we could just `Get` them one by one here and achieve the same effect without a need to convert lists to maps, etc. We could encapsulate `Get`s so it's more or less equivalent of the current code:

```go
s, serviceExists := get[*configurationv1alpha1.KongService](ctx, clientWithNamespace, types.NamespacedName{Namespace: pb.Namespace, Name: serviceRef.Name})
...
func get[T client.Object](ctx context.Context, c client.Client, key types.NamespacedName) (T, bool) {
	var obj T
	err := c.Get(ctx, key, obj)
	return obj, err == nil
}
```
WDYT?

czeslavo (commented):
Looks good! Left some suggestions, please tell WDYT.

controller/konnect/index_kongroute.go
pmalek: Must have been one of the approach I tried during this implementation. Removing in  5f7df1c0643270af5d0cce6dbf7812ed8c3477be

controller/konnect/reconciler_kongplugin.go
czeslavo: Something like that should work:
```go
		serviceRef := pb.Spec.Targets.ServiceReference
		routeRef := pb.Spec.Targets.RouteReference
		
		s, serviceExists := getIfRefNotNil[*configurationv1alpha1.KongService](ctx, clientWithNamespace, serviceRef)
		r, routeExists := getIfRefNotNil[*configurationv1alpha1.KongRoute](ctx, clientWithNamespace, routeRef)
```

```go
func getIfRefNotNil[T client.Object](ctx context.Context, c client.Client, ref *configurationv1alpha1.TargetRefWithGroupKind) (T, bool) {
	var obj T
	if ref == nil {
		return obj, false
	}
	err := c.Get(ctx, client.ObjectKey{Name: ref.Name} , obj)
	return obj, err == nil
}
```

controller/konnect/reconciler_kongplugin.go
czeslavo: Note: To be clear, I'm fine with adding this in a separate PR.

controller/konnect/index_kongroute.go
pmalek: Good call. I went ahead and refactored plugin names extraction into a dedicated function in `pkg/annotation/` package: 7a0192bf03ec1d326b3c8edcf1eb22e95ec987b2

PTAL

controller/konnect/kongpluginbuilder.go
pmalek: I believe that for now we can keep this simple as any error stemming from an invalid `KongPluginBinding` will get returned from the reconcile loop anyway. WDYT?

Adding this validation would also require from us to update it with all the validation rules that we'd add on the CRD itself.

controller/konnect/reconciler_kongplugin.go
```diff
@@ -88,130 +103,124 @@ func (r *KongPluginReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+						pb.Spec.Targets.RouteReference.Name == rel.Route
+				})
+				builder.WithRouteTarget(rel.Route)
+			}
+
+			// TODO consumers and consumer groups
+
+			kongPluginBinding := builder.Build()
+			if err := controllerutil.SetOwnerReference(&kongPlugin, kongPluginBinding, clientWithNamespace.Scheme(), controllerutil.WithBlockOwnerDeletion(true)); err != nil {
+				return ctrl.Result{}, fmt.Errorf("failed to set owner reference: %w", err)
```
czeslavo: I think this, as a part of building the object, could be done by the builder with `WithOwnerReference(kongPluginBinding)`. It will require passing a client to `NewKongPluginBindingBuilder`, ~but that might be required for validating the references anyway~.

controller/konnect/kongpluginbuilder.go
pmalek: I've removed this and added a comment for `WithControlPlaneRef` in 4ab47d48b00483028cca58823ae69197c555e0f0.

With current implementation this cannot really happen as we're grouping the relations based in `ForeignRelations.GroupByControlPlane()`.

controller/konnect/reconciler_kongplugin.go
```diff
@@ -88,130 +103,124 @@ func (r *KongPluginReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+				WithPluginRef(kongPlugin.Name).
+				WithControlPlaneRef(&configurationv1alpha1.ControlPlaneRef{
+					Type: configurationv1alpha1.ControlPlaneRefKonnectNamespacedRef,
+					KonnectNamespacedRef: &configurationv1alpha1.KonnectNamespacedRef{
+						Namespace: cpNN.Namespace,
+						Name:      cpNN.Name,
+					},
+				})
+
+			l := kongPluginBindingList.Items
```
czeslavo: Taking into account the volume of this function, naming this variable more verbose might be helpful for readers (I caught myself not understanding what `l` is in the switch below).
```suggestion
			kpbList := kongPluginBindingList.Items
```

controller/konnect/ops/ops_kongpluginbinding.go
pmalek: 0d231de3880e7cd3ed4382df0513c8261ea5b26f

controller/konnect/index_kongroute.go
czeslavo: Very nice 👍

controller/konnect/reconciler_kongplugin.go
pmalek: fc141e0f75fb49b8623b1c04d37eaa81587ee419

controller/konnect/kongpluginbuilder.go
czeslavo: You're right, let's leave it.

controller/konnect/reconciler_kongplugin.go
pmalek: 57ece90da2fd8801225fdc9e678179f837666926

controller/konnect/reconciler_kongplugin.go
pmalek: Makes sense 👍 I was wondering if listing and then mapping vs just getting would be better (given that we use a cached client) but without proper performance tests that's just speculation I guess.

Implemented the above in 2321541450fea901a9d75f28596cfdcf99ade0d9 (Needed to adjust the type params a bit)

controller/konnect/reconciler_kongplugin.go
pmalek: 👍 Let's add that in a separate PR together with an envtest based test.

controller/konnect/reconciler_kongplugin.go
pmalek: c19d953d444b6b220d794613c0b7ba01be80dbc3

controller/konnect/reconciler_kongplugin.go
```diff
@@ -242,31 +251,187 @@ func (r *KongPluginReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+				break
+			}
+
+			if !combinationFound {
+				pluginBindingsToDelete[client.ObjectKeyFromObject(&pb)] = pb
+				continue
+			}
+
+			s, serviceExists := getIfRefNotNil[*configurationv1alpha1.KongService](ctx, clientWithNamespace, serviceRef)
+			r, routeExists := getIfRefNotNil[*configurationv1alpha1.KongRoute](ctx, clientWithNamespace, routeRef)
```
czeslavo, line 349: This could be moved up, out of the switch, same for other `getIfRefNotNil` calls.

pmalek: But then we could end up getting objects that we don't need, right? Do you think this (moving this up) is worth the potential unnecessary object lookup?

czeslavo: Yeah, makes sense. Let's leave it.

czeslavo (approved):

PR #656, opened 2024-09-26
chore(tests): disable KIC admission webhook for `TestKongPluginInstallationEssentials`

programmer04:

Flakiness of `TestKongPluginInstallationEssentials` has been observed, e.g. [here](https://github.com/Kong/gateway-operator/actions/runs/11049041892/job/30693756663#step:5:559) and [here](https://github.com/Kong/gateway-operator/actions/runs/11049419786/job/30694931382?pr=651#step:5:446). It's due to how our webhook in KIC tries to reject as soon as possible that leads to requirement of strict ordering.

**Which issue this PR fixes**

Followup for https://github.com/Kong/gateway-operator/issues/380

Remediation from the original PR with disabling a webhook was incorrect because it took into account only the webhook of KGO which in this case is irrelevant.

### DIFF
--- a/test/helpers/generators.go
+++ b/test/helpers/generators.go
@@ -65,9 +65,11 @@ func GenerateGateway(gatewayNSN types.NamespacedName, gatewayClass *gatewayv1.Ga
 	return gateway
 }
 
+type gatewayConfigurationOption func(*operatorv1beta1.GatewayConfiguration)
+
 // GenerateGatewayConfiguration generates a GatewayConfiguration to be used in tests
-func GenerateGatewayConfiguration(namespace string) *operatorv1beta1.GatewayConfiguration {
-	return &operatorv1beta1.GatewayConfiguration{
+func GenerateGatewayConfiguration(namespace string, opts ...gatewayConfigurationOption) *operatorv1beta1.GatewayConfiguration {
+	gwc := &operatorv1beta1.GatewayConfiguration{
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: namespace,
 			Name:      uuid.NewString(),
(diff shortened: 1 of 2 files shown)


test/helpers/generators.go
```diff
@@ -139,6 +141,20 @@ func GenerateGatewayConfiguration(namespace string) *operatorv1beta1.GatewayConf
...
 		},
 	}
+	for _, opt := range opts {
+		opt(gwc)
+	}
+	return gwc
+}
+
+// WithWebhookDisabled disables the admission webhook for the control plane
+func WithWebhookDisabled() func(*operatorv1beta1.GatewayConfiguration) {
```
pmalek: I'd consider calling this `WithControlPlaneWebhookDisabled` just to be a bit more explicit, WDYT?

test/integration/test_kongplugininstallation.go
```diff
@@ -212,7 +207,9 @@ func TestKongPluginInstallationEssentials(t *testing.T) {
 func deployGatewayWithKPI(
 	t *testing.T, cleaner *clusters.Cleaner, namespace string,
 ) (gatewayIPAddress string, gatewayConfigNN, httpRouteNN k8stypes.NamespacedName) {
-	gatewayConfig := helpers.GenerateGatewayConfiguration(namespace)
+	// NOTE: Disable webhook for KIC, because it checks for the plugin in Kong Gateway and rejects,
+	// thus it requires strict order of deployment.
```
pmalek:
```suggestion
	// NOTE: Disable webhook for KIC, because it checks for the plugin in Kong Gateway and rejects,
	// thus it requires strict order of deployment which is not guaranteed.
```

pmalek (approved):

pmalek (approved):

PR #662, opened 2024-09-27
chore: use IsZero() for ctr.Result to make check

programmer04:
**What this PR does / why we need it**

During the review of https://github.com/Kong/gateway-operator/pull/476 @pmalek mentioned that for `ctrl.Result` not only the field `Requeue` but also `RequeueAfter` indicates a need of requeue see https://github.com/Kong/gateway-operator/pull/476#discussion_r1771453386. This PR ensures that both fields are checked in all cases with a built-in method `IsZero()`. 

**Which issue this PR fixes**

Spotted during the review of PR for the issue https://github.com/Kong/gateway-operator/issues/380


czeslavo (commented):
Wouldn't `!res.IsZero()` be the equivalent?

czeslavo (approved):

pmalek (approved):

PR #663, opened 2024-09-27
feat(konnect): add KongKey - KongKeySet binding

### LINKED ISSUE #585: Konnect entities: Key Set
### Problem statement

Introduce `KongKeySet`. As introduced here https://docs.konghq.com/gateway/latest/reference/key-management/, a `Key` requires a `KeySet` which it belongs to.

### Proposed solution

- Introduce `KongKeySet` CRD in https://github.com/Kong/kubernetes-configuration
  - This CRD will contain a ControlPlane reference. Only references that refer to Konnect will result in reconciliation against Konnect.

### Acceptance criteria

- [ ] KGO can create, delete and update `KeySet`s in Konnect based on the CRD spec.
- [ ] Status fields of this CRD are filled with information such as
  - [ ] Indication whether the entity in Konnect has been created via a `Programmed` status condition

czeslavo:

Implements the KongKey - KongKeySet binding by handling the `KeySetRef` in the `KongKey`.

**Which issue this PR fixes**

Fixes https://github.com/Kong/gateway-operator/issues/585.

**PR Readiness Checklist**:

### DIFF
--- a/controller/konnect/ops/ops_kongkey_test.go
+++ b/controller/konnect/ops/ops_kongkey_test.go
@@ -1,9 +1,10 @@
 package ops
 
 import (
+	"sort"
 	"testing"
 
-	"github.com/google/uuid"
+	sdkkonnectcomp "github.com/Kong/sdk-konnect-go/models/components"
 	"github.com/samber/lo"
 	"github.com/stretchr/testify/require"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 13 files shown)


controller/konnect/ops/ops_kongkey_test.go
```diff
@@ -12,55 +13,138 @@ import (
...
-		"tag2",
-		"tag3",
-		"tag4",
-		"duplicate",
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			output := kongKeyToKeyInput(tc.key)
+
+			// Tags order is not guaranteed, so we need to sort them before comparing.
```
pmalek, line 145: Wouldn't it be better to sort in `kongKeyToKeyInput` instead and have deterministic output from that function?

I can see that we do not do this anywhere (or have I missed something?) and we mostly just use `Tags: lo.Uniq(slices.Concat(...))`. WDYT about sorting in functions that create the `*Input` objects to get deterministic ordering?

controller/konnect/reconciler_generic.go
```diff
@@ -1147,6 +1167,105 @@ func handleControlPlaneRef[T constraints.SupportedKonnectEntityType, TEnt constr
...
+	// Verify that the KongKeySet is programmed.
+	cond, ok := k8sutils.GetCondition(conditions.KonnectEntityProgrammedConditionType, &keySet)
+	if !ok || cond.Status != metav1.ConditionTrue {
+		if res, err := updateStatusWithCondition(
+			ctx, cl, ent,
+			conditions.KeySetRefValidConditionType,
+			metav1.ConditionFalse,
+			conditions.KeySetRefReasonInvalid,
+			fmt.Sprintf("Referenced KongKeySet %s is not programmed yet", nn),
+		); err != nil || res.Requeue {
```
pmalek:
```suggestion
		); err != nil || !res.IsZero() {
```

controller/konnect/reconciler_generic.go
```diff
@@ -1147,6 +1167,105 @@ func handleControlPlaneRef[T constraints.SupportedKonnectEntityType, TEnt constr
...
+			conditions.KeySetRefReasonInvalid,
+			err.Error(),
+		); errStatus != nil || res.Requeue {
+			return res, errStatus
+		}
+
+		// If it was not found, let's requeue.
+		if k8serrors.IsNotFound(err) {
+			return ctrl.Result{Requeue: true}, nil
+		}
```
pmalek: Should we really requeue here? I believe we should add a watch for related `KongKeySet` in `KongKeyReconciliationWatchOptions` https://github.com/Kong/gateway-operator/blob/a8465d59fb2798a95403a0e25f096b954906339c/controller/konnect/watch_kongkey.go#L23 so that events on referenced `KongKeySet` enqueue `KongKeys` that refer to them. WDYT?

controller/konnect/reconciler_generic.go
```diff
@@ -1147,6 +1167,105 @@ func handleControlPlaneRef[T constraints.SupportedKonnectEntityType, TEnt constr
...
+	}
+
+	if err := cl.Get(ctx, nn, &keySet); err != nil {
+		if res, errStatus := updateStatusWithCondition(
+			ctx, cl, ent,
+			conditions.KeySetRefValidConditionType,
+			metav1.ConditionFalse,
+			conditions.KeySetRefReasonInvalid,
+			err.Error(),
+		); errStatus != nil || res.Requeue {
```
pmalek:
```suggestion
		); errStatus != nil || !res.IsZero() {
```

controller/konnect/ops/ops_kongkey_test.go
czeslavo: Yeah, let's make sure the output is deterministic in all those functions: https://github.com/Kong/gateway-operator/pull/675

pmalek: 👍 Approved and merged #675. Thanks for working on that.

controller/konnect/reconciler_generic.go
czeslavo: Yes, you're right. Handled this and more edge cases (removing already resolved KeySetRef from the spec, ensuring an owner KeySet removal triggers removal of a Key). PTAL

test/envtest/konnect_entities_key_test.go
```diff
@@ -71,50 +73,145 @@ func TestKongKey(t *testing.T) {
...
+			if c.GetName() != createdKey.GetName() {
+				return false
+			}
+			return lo.ContainsBy(c.Status.Conditions, func(condition metav1.Condition) bool {
+				return condition.Type == conditions.KonnectEntityProgrammedConditionType &&
+					condition.Status == metav1.ConditionTrue
+			})
+		}, "KongKey's Programmed condition should be true eventually")
+
+		t.Log("Waiting for KongKey to be created in the SDK")
```
pmalek: Nit: the log here is not 100% accurate because when the object already got `Programmed` (checked above) then it means that the SDK operation has already taken place.

Perhaps something like this would be more accurate: 

```suggestion
		t.Log("Checking SDK KongKey operations")
```

controller/konnect/watch_kongkey.go
```diff
@@ -176,3 +184,43 @@ func enqueueKongKeyForKonnectControlPlane(
...
+			}
+			ret = append(ret, reconcile.Request{
+				NamespacedName: types.NamespacedName{
+					Namespace: key.Namespace,
+					Name:      key.Name,
+				},
+			})
+		}
+
+		return ret
```
pmalek: I believe all of this can be changed into a list using an index (which would have to be added) and a  [`objectListToReconcileRequests`](https://github.com/Kong/gateway-operator/blob/820f2c87a21189a8aa3157e573d8b0819de50a53/controller/konnect/watch.go#L112) call. WDYT?

controller/konnect/reconciler_keysetref.go
```diff
@@ -0,0 +1,186 @@
...
+			metav1.ConditionFalse,
+			conditions.KeySetRefReasonInvalid,
+			fmt.Sprintf("Referenced KongKeySet %s is not programmed yet", nn),
+		); err != nil || !res.IsZero() {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{Requeue: true}, nil
+	}
+
+	if res, err := passOwnershipExclusivelyTo(ctx, cl, ent, &keySet); err != nil || !res.IsZero() {
```
pmalek, line 123: Can you elaborate on why we need this 🤔 ?

controller/konnect/reconciler_keysetref.go
```diff
@@ -0,0 +1,186 @@
...
+			// If the entity has a resolved reference, but the spec has changed, we need to adjust the status
+			// and transfer the ownership back from the KeySet to the ControlPlane.
+			if key.Status.Konnect != nil && key.Status.Konnect.GetKeySetID() != "" {
+				// Reset the KeySetID in the status and set the condition to True.
+				key.Status.Konnect.KeySetID = ""
+				if res, err := updateStatusWithCondition(ctx, cl, key,
+					conditions.KeySetRefValidConditionType,
+					metav1.ConditionTrue,
+					conditions.KeySetRefReasonValid,
+					"KeySetRef is nil",
```
pmalek:
```suggestion
					"KeySetRef is unset",
```

`nil` might be considered that this is an error condition where in fact it's not.

controller/konnect/reconciler_keysetref.go
czeslavo: TIL that as long as an object has at least one owner reference pointing to an existing object, it [will not get garbage collected](https://github.com/kubernetes/kubernetes/blob/b42772c3b1b5464838761372042891b6888fb2af/pkg/controller/garbagecollector/garbagecollector.go#L567) (which makes sense if you think more about it). 

So, in our case that means that if we keep owner references both to the ControlPlane and KeySet and we remove KeySet object, the dependent Key won't get deleted because it will still have a solid reference to the ControlPlane. That's not the expected behavior, as removal of the associated KeySet deletes a Key on the Konnect side by cascading the delete. We need to reflect this in the Kubernetes world by explicitly passing the ownership to the KeySet *only*.

pmalek: Thanks for the explanation 🙇 

Do you mind adding at least some of those details in the code as comments?

controller/konnect/watch_kongkey.go
czeslavo: 11c59a4

pmalek: 👍

controller/konnect/reconciler_keysetref.go
czeslavo: [68f59b5](https://github.com/Kong/gateway-operator/pull/663/commits/68f59b596746a89b5f3c3e358144ab3b1797711e)

pmalek (approved):

PR #674, opened 2024-10-01
deps: gateway API bumped to v1.2.0

mlavacca:

**Which issue this PR fixes**

Fixes #

Conformance tests have been run locally and they are successful.

**PR Readiness Checklist**:


pmalek:
Is 1.2.0 aimed to be released before our release target date for 1.4.0? Otherwise we'd end up with using RC in 1.4.0 which I'd say is something we don't want to do.

Given that we might not be sure about the release date of 1.2.0 I suggest to keep this one as is until that happens so that we don't need to revert or rely on RC in a released version of KGO. WDYT?

mlavacca:
> Is 1.2.0 aimed to be released before our release target date for 1.4.0? Otherwise we'd end up with using RC in 1.4.0 which I'd say is something we don't want to do.
> 
> Given that we might not be sure about the release date of 1.2.0 I suggest to keep this one as is until that happens so that we don't need to revert or rely on RC in a released version of KGO. WDYT?

Yep, sounds good to me. I've asked the Gateway API folks about the candidate date for the actual 1.2 release, as I've been out for a while. Let's keep this PR as is and bump it once 1.2 is out 👍

pmalek:
With 1.2.0 out https://github.com/kubernetes-sigs/gateway-api/releases/tag/v1.2.0 I believe we can update this PR and merge it 👍

pmalek (approved):

PR #680, opened 2024-10-02
refactor(konnect): add generic `objectListToReconcileRequests`

pmalek:

This PR adds `objectListToReconcileRequests` which generically converts list of objects into `[]ctrl.Request`.


czeslavo (approved):
Very nice 👍

PR #681, opened 2024-10-02
refactor(konnect): add indices for objects that reference `KonnectGatewayControlPlane`

pmalek:

This PR adds index fields for some of the entities that refer to `KonnectGatewayControlPlane` and uses that for listing the objects.

### DIFF
--- a/controller/konnect/index.go
+++ b/controller/konnect/index.go
@@ -13,13 +13,13 @@ type ReconciliationIndexOption struct {
 	ExtractValue client.IndexerFunc
 }
 
-// controlPlaneKonnectNamespacedRefRefAsSlice returns a slice of strings representing
+// controlPlaneKonnectNamespacedRefAsSlice returns a slice of strings representing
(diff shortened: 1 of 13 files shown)


controller/konnect/index.go
```diff
@@ -10,3 +12,22 @@ type ReconciliationIndexOption struct {
 	IndexField   string
 	ExtractValue client.IndexerFunc
 }
+
+// controlPlaneKonnectNamespacedRefRefAsSlice returns a slice of strings representing
+// the KonnectNamespacedRef of the object.
+func controlPlaneKonnectNamespacedRefRefAsSlice[
```
czeslavo:
```suggestion
func controlPlaneKonnectNamespacedRefAsSlice[
```

czeslavo (approved):
🎖️

controller/konnect/index.go
pmalek: 👍 5d927d61a9b89885213f7887d32994dbe4574a41

PR #685, opened 2024-10-02
feat(konnect): add indices for `KongPluginBinding` -> `KonnectGatewayControlPlane` and `KongRoute` -> `KongService`

pmalek:

Add indices for `KongPluginBinding` -> `KonnectGatewayControlPlane` and `KongRoute` -> `KongService`

### DIFF
--- a/controller/konnect/index_kongroute.go
+++ b/controller/konnect/index_kongroute.go
@@ -51,10 +51,8 @@ func kongRouteRefersToKongService(object client.Object) []string {
 		return nil
 	}
 
-	namespace := route.Namespace
-	if svcRef.NamespacedRef.Namespace != "" {
-		namespace = svcRef.NamespacedRef.Namespace
-	}
+	// NOTE: We currently do not allow cross namespace references between KongRoute and KongService.
+	// https://github.com/Kong/kubernetes-configuration/issues/106 tracks the implementation.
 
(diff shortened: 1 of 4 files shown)


controller/konnect/index_kongroute.go
```diff
@@ -31,3 +38,23 @@ func kongRouteUsesPlugins(object client.Object) []string {
...
+	if svcRef == nil ||
+		svcRef.Type != configurationv1alpha1.ServiceRefNamespacedRef ||
+		svcRef.NamespacedRef == nil {
+		return nil
+	}
+
+	namespace := route.Namespace
+	if svcRef.NamespacedRef.Namespace != "" {
+		namespace = svcRef.NamespacedRef.Namespace
+	}
```
czeslavo: We might leave a similar note regarding the namespaced references support here.

pmalek: Good call. I was thinking about it but eventually I believe we should settle on removing the `Namespace` field from `ServiceRef` (which is what https://github.com/Kong/kubernetes-configuration/pull/107 does) and track the implementation of those in https://github.com/Kong/kubernetes-configuration/pull/106).

ac6dd7bc98403cfab39af774e1631b4ec9badd84 adds the comment about that.

WDYT?

czeslavo (approved):

PR #689, opened 2024-10-03
chore: adapt apidocs-gen script to use kconf definitions

czeslavo:

It makes the `scripts/apidocs-gen` script include `kubernetes-configuration` types that KGO support in the generated markdown reference.

From now on, all new types that shall be included in the documentation need to be annotated with the new `+apireference:kgo:include`.

It depends on https://github.com/Kong/kubernetes-configuration/pull/110.

**Which issue this PR fixes**

Part of https://github.com/Kong/kubernetes-configuration/issues/105.


README.md
```diff
@@ -85,11 +85,14 @@ machine against the cluster that you have configured via your `KUBECONFIG`.
...
 ### Adding new CRDs
 
-Whenever you add a new CRD ensure that it is included in project's [`PROJECT`](./PROJECT) file.
-
-This is necessary for creation of a bundle for external hubs like [Operator Hub's community operators][community-operators].
+Whenever you add a new CRD:
+- Ensure that it is included in project's [`PROJECT`](./PROJECT) file. This is necessary for creation of
+  a bundle for external hubs like [Operator Hub's community operators][community-operators].
+- Annotate the CRD and any new type it depends on with the right markers to make sure it will be included
+  in the generated documentation. See [available markers][available-markers].
```
pmalek, line 92: Is there an option `crd-ref-docs` that would allow us to just add this annotation on a package instead of adding it to every type (which can be error prone)?

czeslavo: I initially tried that and it didn't work in `crd-ref-docs` so I gave up this idea. I verified it's a feature it lacks and contributed it upstream: https://github.com/elastic/crd-ref-docs/pull/95. 

I've created an issue to track an improvement we should be able to make once it's merged: https://github.com/Kong/gateway-operator/issues/695

pmalek: Thanks for looking into this!

pmalek (approved):

PR #697, opened 2024-10-04
feat(konnect): allow setting ControlPlane group members

### LINKED ISSUE #449: Konnect entities: Gateway Control Plane Group - Control Plane Group Memberships
### Problem statement

Introduce ControlPlane group membership for `KonnectControlPlane`.

For details see https://docs.google.com/document/d/1TxW2ovqPRm-w1td1e_V0CyaET4lpiEBCJ7hiYMZJqng/edit#heading=h.i2q1r7wb62uu

### Proposed solution

- Introduce ControlPlane group membership for `KonnectControlPlane`.

### Prerequisites

- #432 
- PUT for CP group membership
  - https://kongstrong.slack.com/archives/C03LRB400TC/p1719493489848099, getting 404 even though implemented via https://github.com/Kong/koko-private/pull/3754 and https://github.com/Kong/koko-private/pull/3772

### Acceptance criteria

- [ ] KGO can set ControlPlane group membership on Konnect's ControlPlanes.

pmalek:

Allow setting konnect ControlPlane group members.

**Which issue this PR fixes**

~Requires https://github.com/Kong/kubernetes-configuration/pull/114~

Closes: #449 

**PR Readiness Checklist**:


controller/konnect/ops/ops_controlplane.go
```diff
@@ -113,7 +131,104 @@ func updateControlPlane(
...
+		cp.Spec.ClusterType == nil ||
+		*cp.Spec.ClusterType != sdkkonnectcomp.ClusterTypeClusterTypeControlPlaneGroup {
+		return nil
+	}
+
+	wg := sync.WaitGroup{}
+	wg.Add(len(cp.Spec.Members))
+	ch := make(chan string)
+	chErr := make(chan error)
+	errs := make([]error, 0)
```
czeslavo: Please consider moving this declaration closer to the code using it (the consuming goroutine).

controller/konnect/ops/ops_controlplane.go
```diff
@@ -113,7 +131,104 @@ func updateControlPlane(
...
+	for _, member := range cp.Spec.Members {
+		go func() {
+			var memberCP konnectv1alpha1.KonnectGatewayControlPlane
+			err := cl.Get(ctx, client.ObjectKey{Namespace: cp.Namespace, Name: member.Name}, &memberCP)
+			if err != nil {
+				if k8serrors.IsNotFound(err) {
+					chErr <- err
+					return
+				}
+				chErr <- fmt.Errorf("failed to get control plane group member %s: %w", member.Name, err)
```
czeslavo: I believe this could be simplified as we do not change behavior when it is 404. 
```suggestion
				chErr <- fmt.Errorf("failed to get control plane group member %s: %w", member.Name, err)
```

controller/konnect/ops/ops_controlplane.go
```diff
@@ -113,7 +131,104 @@ func updateControlPlane(
...
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case err, ok := <-chErr:
+				if !ok {
+					return
+				}
+				errs = append(errs, err)
+				wg.Done()
```
czeslavo: Wouldn't it make more sense to call `wg.Done()` as a deferred function in the producer?

controller/konnect/ops/ops_controlplane.go
```diff
@@ -113,7 +131,104 @@ func updateControlPlane(
...
+		SetKonnectEntityProgrammedConditionFalse(cp, conditions.KonnectGatewayControlPlaneProgrammedReasonFailedToSetControlPlaneGroupMembers, err.Error())
+		return err
+	}
+
 	SetKonnectEntityProgrammedCondition(cp)
 
 	return nil
 }
+
+func setGroupMembers(
```
czeslavo, line 145: Have you considered using a library to handle concurrency instead of implementing the producer/consumer here? I think that [iter.MapErr](https://pkg.go.dev/github.com/sourcegraph/conc@v0.3.0/iter#MapErr) from `sourcegraph/conc` that we used for quite a long time in [KIC](https://github.com/Kong/kubernetes-ingress-controller/blob/f8039f19b0f2e0bd3d35ba9a0a688fe208d8f6f8/internal/dataplane/kong_client.go#L698) might simplify this code significantly.

controller/konnect/reconciler_generic.go
```diff
@@ -526,13 +525,20 @@ func (r *KonnectEntityReconciler[T, TEnt]) Reconcile(
...
+			return ctrl.Result{}, fmt.Errorf("failed to update in cluster resource after Konnect update: %w %w", errUpd, err)
+		}
+
 		return ctrl.Result{}, fmt.Errorf("failed to update object: %w", err)
 	} else if !res.IsZero() {
 		return res, nil
 	}
 
-	ent.GetKonnectStatus().ServerURL = apiAuth.Spec.ServerURL
-	ent.GetKonnectStatus().OrgID = apiAuth.Status.OrganizationID
```
czeslavo: Shouldn't we leave them here as well so the `ServerURL` and `OrgID` are populated on the entity update success? 🤔

controller/konnect/ops/ops_controlplane.go
pmalek: 👍 12f64ac5b574d73f2302787f1263309b89a19721

controller/konnect/ops/ops_controlplane.go
pmalek: It seems that we do not import `sourcegraph.conc` just yet in KGO so I'd suggest to not pull in that dependency just for the use case above. WDYT?

controller/konnect/ops/ops_controlplane.go
pmalek: I thought about that but then I realized that there's a chance that we can call `.Done()` `n` times thus making the `.Wait()` unblock while the goroutine receiving on the channel could still write to the `members` slice while line 208 would read from the slice.

I'm pretty sure there is very little chance for this to happen but I believe it's still possible. WDYT?

controller/konnect/reconciler_generic.go
pmalek: Yes, I'm not sure why I removed this here 🤔 

5bc6b0dfc6843c29b58b43aacb058fd12369d7b0

controller/konnect/ops/ops_controlplane.go
czeslavo: I think it's a valid use case to at least consider pulling it in - IMO if we can use a building block for concurrency stuff, it's cheaper and safer to use it than worry about maintaining our scattered implementations (where the chances of making a mistake rise).

controller/konnect/ops/ops_controlplane.go
czeslavo: Maybe we could simplify this a little bit by not running the consumer in a goroutine, but in the main thread like so:
```go
	// Spawn producer goroutines that will fetch the Konnect ID of each member. They will push the IDs to the result
	// channel or any errors encountered to the error channel.
	var (
		ch    = make(chan string)
		chErr = make(chan error)
		wg    = sync.WaitGroup{}
	)
	wg.Add(len(cp.Spec.Members))
	for _, member := range cp.Spec.Members {
		go func() {
			defer wg.Done()

			var memberCP konnectv1alpha1.KonnectGatewayControlPlane
			err := cl.Get(ctx, client.ObjectKey{Namespace: cp.Namespace, Name: member.Name}, &memberCP)
			if err != nil {
				chErr <- fmt.Errorf("failed to get control plane group member %s: %w", member.Name, err)
				return
			}

			if memberCP.GetKonnectID() == "" {
				chErr <- fmt.Errorf("control plane group %s member %s has no Konnect ID", cp.Name, member.Name)
				return
			}
			ch <- memberCP.GetKonnectID()
		}()
	}
	// Wait for all producer goroutines to finish and close the channels.
	go func() {
		wg.Wait()
		close(ch)
		close(chErr)
	}()

	// Collect the results.
	var (
		members = make([]sdkkonnectcomp.Members, 0, len(cp.Spec.Members))
		errs    = make([]error, 0)
	)
	for {
		select {
		case <-ctx.Done():
			break
		case err, ok := <-chErr:
			if ok {
				errs = append(errs, err)
			} else {
				chErr = nil
			}

		case memberID, ok := <-ch:
			if ok {
				members = append(members, sdkkonnectcomp.Members{
					ID: lo.ToPtr(memberID),
				})
			} else {
				ch = nil
			}
		}

		// Break the loop if both channels are closed.
		if ch == nil && chErr == nil {
			break
		}
	}
```

This way we ensure that all values are received from both channels before proceeding.

controller/konnect/ops/ops_controlplane.go
pmalek: Makes sense 👍 801651d8a247901ab189628aab7fa8b6ef3e2cde should handle that.

czeslavo (approved):

controller/konnect/ops/ops_controlplane.go
pmalek: This could be an improvement over what was proposed in the PR but given that we went with using `conc` this is not going to be needed.

Thanks for taking time for looking into this though 🙇

PR #710, opened 2024-10-07
chore: make object counter conditional on CRD existence

### LINKED ISSUE #448: Make object counters conditional on the presence of CRDs in the cluster
### Problem statement

Objects counters should not report errors when the respective CRDs are missing from the cluster.

#360 addressed that for AIGateway but we should do the same for `ControlPlane`s.

### Acceptance criteria

- [ ] No errors should be reported from objects counters when the respective controllers are disabled and CRDs missing.

pmalek:

Fixes #448

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -111,11 +111,15 @@
   [#454](https://github.com/Kong/gateway-operator/pull/454)
 - Requeue instead of reporting an error when a GatewayClass status update yields a conflict.
   [#612](https://github.com/Kong/gateway-operator/pull/612)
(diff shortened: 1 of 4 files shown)


czeslavo (changes requested):
This might be changelog-worthy, no? We're adding new datapoints for the controllers and fixing a bug.

pmalek:
> This might be changelog-worthy, no? We're adding new datapoints for the controllers and fixing a bug.

I've mentioned the change in 633d5c9459db7e618b9aef812f2e2ce18823c67a. WDYT?

czeslavo (approved):

PR #714, opened 2024-10-08
feat: dataplane konnect extension reconciler

### LINKED ISSUE #203: Native support for Konnect integration in DataPlane

mlavacca:

**Which issue this PR fixes**

Fixes #203 

**PR Readiness Checklist**:


CHANGELOG.md
```diff
@@ -101,6 +101,9 @@
   [#669](https://github.com/Kong/gateway-operator/pull/669)
 - Allow setting `KonnectGatewayControlPlane`s group membership
   [#697](https://github.com/Kong/gateway-operator/pull/697)
+- Apply konnect-related customization to `DataPlane`s that properly reference `DataPlaneKonnectExtension`
+  resources.
+  [#714](https://github.com/Kong/gateway-operator/pull/714)
```
```suggestion
- Apply Konnect-related customizations to `DataPlane`s that properly reference `DataPlaneKonnectExtension`
  resources.
  [#714](https://github.com/Kong/gateway-operator/pull/714)
```

controller/dataplane/konnect_extension.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/kong/gateway-operator/api/v1alpha1"
+	"github.com/kong/gateway-operator/api/v1beta1"
+	dputils "github.com/kong/gateway-operator/internal/utils/dataplane"
+	"github.com/kong/gateway-operator/pkg/consts"
+	k8sutils "github.com/kong/gateway-operator/pkg/utils/kubernetes"
+	k8sresources "github.com/kong/gateway-operator/pkg/utils/kubernetes/resources"
+)
+
+// applyDataPlaneKonnectExtension gets the DataPlane as argument, and in case it references a KonnectExtension, it
+// fetch the referenced extension and applies the necessary changes to the DataPlane spec.
```
czeslavo:
```suggestion
// fetches the referenced extension and applies the necessary changes to the DataPlane spec.
```

controller/dataplane/konnect_extension.go
```diff
@@ -0,0 +1,123 @@
...
+func applyDataPlaneKonnectExtension(ctx context.Context, cl client.Client, dataplane *v1beta1.DataPlane) error {
+	if len(dataplane.Spec.Extensions) == 0 {
+		return nil
+	}
+
```
czeslavo: Could be dropped since iterating over an empty slice below will give the same result.
```suggestion
```

controller/dataplane/konnect_extension.go
```diff
@@ -0,0 +1,123 @@
...
+	for _, extensionRef := range dataplane.Spec.Extensions {
+		if extensionRef.Group != v1alpha1.SchemeGroupVersion.Group || extensionRef.Kind != "DataPlaneKonnectExtension" {
+			continue
+		}
+		namespace := dataplane.Namespace
+		if extensionRef.Namespace != nil {
+			namespace = *extensionRef.Namespace
```
czeslavo: What was the final conclusion about handling cross-namespace references here? Should we allow them without explicit grants?

internal/utils/dataplane/env.go
```diff
@@ -0,0 +1,52 @@
...
+	"KONG_CLUSTER_CONTROL_PLANE":         "<CP-ID>.<REGION>.cp0.<SERVER>:443",
+	"KONG_CLUSTER_SERVER_NAME":           "<CP-ID>.<REGION>.cp0.<SERVER>",
+	"KONG_CLUSTER_TELEMETRY_ENDPOINT":    "<CP-ID>.<REGION>.tp0.<SERVER>:443",
+	"KONG_CLUSTER_TELEMETRY_SERVER_NAME": "<CP-ID>.<REGION>.tp0.<SERVER>",
+	"KONG_CLUSTER_CERT":                  "/etc/secrets/kong-cluster-cert/tls.crt",
+	"KONG_CLUSTER_CERT_KEY":              "/etc/secrets/kong-cluster-cert/tls.key",
+	"KONG_LUA_SSL_TRUSTED_CERTIFICATE":   "system",
+	"KONG_KONNECT_MODE":                  "on",
+	"KONG_VITALS":                        "off",
+}
```
czeslavo: How about turning this into a function that will accept a struct, something like

```go
func KongInKonnectEnvVariables(konnectSrvCfg KonnectServerConfiguration) map[string]string
```

and do the replacements in here so we do not scatter this and `customizeKongInKonnectDefaults` in different places while they are both tightly coupled?

controller/dataplane/konnect_extension.go
mlavacca: good point. I blocked cross-namespace refs in b6c30196fc97337b6d32db88e838618576383e0c

internal/utils/index/index.go
pmalek, line 1: I've noticed that we have now this file, https://github.com/Kong/gateway-operator/blob/733f22103699e07ccfb18b76b6982b61854dd553/modules/manager/controller_setup.go#L162-L182, and https://github.com/Kong/gateway-operator/blob/733f22103699e07ccfb18b76b6982b61854dd553/modules/manager/controller_setup.go#L614-L716. 

I've created #721 to manage indices from 1 place so that the code responsible for this is not scattered in several places.

internal/utils/index/index.go
```diff
@@ -71,3 +76,39 @@ func KongPluginInstallationsOnDataPlane(ctx context.Context, c cache.Cache) erro
...
+					namespace := dp.Namespace
+					if ext.Group == operatorv1alpha1.SchemeGroupVersion.Group &&
+						ext.Kind == "DataPlaneKonnectExtension" {
+						if ext.Namespace != nil {
+							namespace = *ext.Namespace
+						}
+						result = append(result, namespace+"/"+ext.NamespacedRef.Name)
+					}
+				}
+			}
```
pmalek: Small suggestion to refactor this.

```suggestion
			for _, ext := range dp.Spec.Extensions {
				namespace := dp.Namespace
				if ext.Group != operatorv1alpha1.SchemeGroupVersion.Group ||
					ext.Kind != "DataPlaneKonnectExtension" {
					continue
				}
				if ext.Namespace != nil {
					namespace = *ext.Namespace
				}
				result = append(result, namespace+"/"+ext.NamespacedRef.Name)
			}
```

Additionally: as per https://github.com/Kong/gateway-operator/pull/714#discussion_r1793189672 I believe we don't want to allow cross namespace references, right? If that's the case we should only index based on name for now and drop extensions (that is call `continue` in the loop) for those that specify if (even though https://github.com/Kong/gateway-operator/commit/b6c30196fc97337b6d32db88e838618576383e0c should block them). WDYT?

controller/dataplane/dataplanekonnectextension_controller.go
```diff
@@ -0,0 +1,117 @@
...
+			if extension.Namespace != nil {
+				namespace = *extension.Namespace
+			}
+			recs = append(recs, reconcile.Request{
+				NamespacedName: client.ObjectKey{
+					Namespace: namespace,
+					Name:      extension.Name,
+				},
+			})
+		}
```
pmalek: WDYT about refactoring this so that happy path is the least indented one (similarly like suggested in another comment)?

pkg/consts/dataplane.go
```diff
@@ -178,3 +178,13 @@ const (
...
 )
+
+// -----------------------------------------------------------------------------
+// Consts - DataPlaneKonnectExtension Finalizers
+// -----------------------------------------------------------------------------
+
+const (
+	// DataPlaneExtensionFinalizer is the finalizer added to DataPlaneKonnectExtension
+	// to ensure that the resources are not deleted when in use by a dataplane.
+	DataPlaneExtensionFinalizer = "gateway-operator.konghq.com/extension-in-use"
```
pmalek, line 189: Is the use case for this finalizer simply to block deletion of the extension when referenced by a DP? I'm wondering what are we gaining with this vs allowing them to be deleted and reconciling the DP accordingly (with a proper status condition indicating such a situation). 🤔 

LMKWYT

mlavacca: This extension completely changes the behavior and the configuration of a DataPlane. From the user perspective, I'd be baffled to see my DataPlane entirely reconfigured with a completely different setup when deleting a resource. I strongly think that we should block the deletion of an extension in use, to avoid the above.

internal/utils/dataplane/env.go
mlavacca: I've gathered the env template and the customization helper. I haven't created the `konnectSrvCfg KonnectServerConfiguration` struct though, as I don't see a big deal having three string params.

internal/utils/index/index.go
mlavacca: yep, I agree. I changed this part in 1158067a53198f99abb1bcee4643600aee2e9756

controller/dataplane/dataplanekonnectextension_controller.go
mlavacca: 1843dc609f432eccc6d9d895ac0f7aef0a2b15b1

internal/utils/dataplane/env.go
czeslavo: 👍 I suggested using a struct instead of 3 strings as it's quite easy to make an order mistake if you need to pass 3 arguments of the same type beside each other.

mlavacca: sounds good, here it is c4adc17275da26aac1de8f1040b46581a5e1e32e

pkg/consts/dataplane.go
pmalek: Thanks for explanation. Now this makes more sense to me. 👍

internal/utils/dataplane/config.go
```diff
@@ -57,7 +27,7 @@ var KongDefaults = map[string]string{
 // PodTemplateSpec.
 // EnvVars are sorted lexographically as a side effect.
 // It also returns the updated EnvVar slice.
-func FillDataPlaneProxyContainerEnvs(existing []corev1.EnvVar, podTemplateSpec *corev1.PodTemplateSpec) {
+func FillDataPlaneProxyContainerEnvs(existing []corev1.EnvVar, podTemplateSpec *corev1.PodTemplateSpec, envSet map[string]string) {
```
pmalek, line 30: Now that we're passing the third argument I believe we should also add a mention of it to the godoc. WDYT?

controller/dataplane/konnect_extension.go
```diff
@@ -0,0 +1,104 @@
...
+		})
+		if container := k8sutils.GetPodContainerByName(&d.Spec.Template.Spec, consts.DataPlaneProxyContainerName); container == nil {
+			d.Spec.Template.Spec.Containers = append(d.Spec.Template.Spec.Containers, corev1.Container{
+				Name: consts.DataPlaneProxyContainerName,
+			})
+		}
+
+		d.WithVolume(kongInKonnectClusterCertificateVolume())
+		d.WithVolumeMount(kongInKonnectClusterCertificateVolumeMount(), consts.DataPlaneProxyContainerName)
+		d.WithVolume(kongInKonnectClusterCertVolume(konnectExt.Spec.AuthConfiguration.ClusterCertificateSecretRef.Name))
```
pmalek, line 84: Is the existence of the referred secret checked at any point and communicated to the user in e.g. `DataPlane`'s status conditions that it's e.g. missing?

controller/dataplane/konnect_extension_test.go
```diff
@@ -0,0 +1,272 @@
...
+					}
+				}
+
+				if tt.konnectExt != nil {
+					requiredEnv = append(requiredEnv, getKongInKonnectEnvVars(*tt.konnectExt)...)
+					sort.Sort(k8sutils.SortableEnvVars(requiredEnv))
+					assert.NotNil(t, dataplane.Spec.Deployment.PodTemplateSpec)
+					assert.Equal(t, requiredEnv, dataplane.Spec.Deployment.PodTemplateSpec.Spec.Containers[0].Env)
+				}
+			}
```
pmalek:
```suggestion
			if tt.expectedError {
				require.Error(t, err)
				return
			}
			require.NoError(t, err)
			requiredEnv := []corev1.EnvVar{}
			if tt.dataplane.Spec.Deployment.PodTemplateSpec != nil {
				if container := k8sutils.GetPodContainerByName(&tt.dataplane.Spec.Deployment.PodTemplateSpec.Spec, consts.DataPlaneProxyContainerName); container != nil {
					requiredEnv = container.Env
				}
			}
			if tt.konnectExt != nil {
				requiredEnv = append(requiredEnv, getKongInKonnectEnvVars(*tt.konnectExt)...)
				sort.Sort(k8sutils.SortableEnvVars(requiredEnv))
				assert.NotNil(t, dataplane.Spec.Deployment.PodTemplateSpec)
				assert.Equal(t, requiredEnv, dataplane.Spec.Deployment.PodTemplateSpec.Spec.Containers[0].Env)
			}
```

controller/dataplane/konnect_extension_test.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+}
+
+func getKongInKonnectEnvVars(konnectExt operatorv1alpha1.DataPlaneKonnectExtension) []corev1.EnvVar {
+	envSet := []corev1.EnvVar{}
+	for k, v := range dputils.KongInKonnectDefaults(dputils.KongInKonnectParams{
+		ControlPlane: *konnectExt.Spec.ControlPlaneRef.KonnectID,
+		Region:       konnectExt.Spec.ControlPlaneRegion,
+		Server:       konnectExt.Spec.ServerHostname,
+	}) {
```
pmalek, line 323: Nit: IMHO we could put the return value from `KongInKonnectDefaults` into a var.

controller/dataplane/watch.go
```diff
@@ -79,3 +87,28 @@ func listDataPlanesReferencingKongPluginInstallation(
...
+		ctx context.Context, ext *operatorv1alpha1.DataPlaneKonnectExtension,
+	) []reconcile.Request {
+		logger := ctrllog.FromContext(ctx)
+
+		// Find all DataPlane resources referencing DataPlaneKonnectExtension
+		// that maps to the DataPlaneKonnectExtension enqueued for reconciliation.
+		var dataPlaneList operatorv1beta1.DataPlaneList
+		if err := c.List(ctx, &dataPlaneList, client.MatchingFields{
+			index.DataPlaneKonnectExtensionIndex: ext.Namespace + "/" + ext.Name,
+		}); err != nil {
```
pmalek, line 104: Since we're disallowing cross namespace refs we can add the `client.InNamespace(ext.Namespace)` option here, right?

```suggestion
		if err := c.List(ctx, &dataPlaneList,
			client.InNamespace(ext.Namespace),
			client.MatchingFields{
				index.DataPlaneKonnectExtensionIndex: ext.Namespace + "/" + ext.Name,
			},
		); err != nil {
```

controller/dataplane/watch.go
```diff
@@ -79,3 +87,28 @@ func listDataPlanesReferencingKongPluginInstallation(
...
+			logger.Error(err, "Failed to list DataPlanes in watch", operatorv1alpha1.DataPlaneKonnectExtensionKind)
+			return nil
+		}
+		return lo.Map(dataPlaneList.Items, func(dp operatorv1beta1.DataPlane, _ int) reconcile.Request {
+			return reconcile.Request{
+				NamespacedName: client.ObjectKeyFromObject(&dp),
+			}
+		})
```
pmalek, line 112: We could use `objectListToReconcileRequests` but sadly that's in `konnect` package unexported now 😞 We should move that at some point to a watch dedicated package.

https://github.com/Kong/gateway-operator/blob/0d9d721bfca68961e6b7f6ab176f6feff8d5566e/controller/konnect/watch.go#L118-L140

pmalek (commented):
Looks good apart from some nits.

One thing that I believe we're missing here are some tests, specifically a `envtest` based test or an integration test would definitely make this complete. WDYT?

controller/dataplane/watch.go
mlavacca: It does not really matter, as the indexing is already done only for objects already living in the same namespace. Hence, all the objects returned by this list are already in the same ns.

controller/dataplane/watch.go
mlavacca: Honestly, I am not a big fan of using generics for everything. There are cases where it is definitely worth it but in cases such as this one, where 3 lines of easy code are replaced by 1, I don't know...

pmalek: Understood. I just want to point out that that these 3 lines (5 if we're precise) become:

```suggestion
		return objectListToReconcileRequests(dataPlaneList)
```

Which I believe is pretty elegant. Nonetheless we can leave it at what it is now 👍

czeslavo (approved):

controller/dataplane/konnect_extension.go
mlavacca: That part was not implemented, but it's actually a good point. I implemented the `ResolvedRefs` condition for the `DataPlane` in e6ba10117baf8f3c43f9b50a0db6f867f9bbda2c

mlavacca:
> Looks good apart from some nits.
> 
> One thing that I believe we're missing here are some tests, specifically a `envtest` based test or an integration test would definitely make this complete. WDYT?

Agreed, I'll add an integration test

mlavacca:
> > Looks good apart from some nits.
> > One thing that I believe we're missing here are some tests, specifically a `envtest` based test or an integration test would definitely make this complete. WDYT?
> 
> Agreed, I'll add an integration test

I've created https://github.com/Kong/gateway-operator/issues/726 to track the addition of such an integration test.

controller/dataplane/controller_utils.go
```diff
@@ -308,3 +309,47 @@ func isDeploymentReady(deploymentStatus appsv1.DeploymentStatus) (metav1.Conditi
...
+			condition.Reason = string(consts.RefNotPermittedReason)
+			condition.Message = consts.RefNotPermittedMessage
+		case errors.Is(err, ErrKonnectExtensionNotFound):
+			condition.Status = metav1.ConditionFalse
+			condition.Reason = string(consts.InvalidExtensionRefReason)
+			condition.Message = consts.InvalidExtensionRefMessage
+		case errors.Is(err, ErrClusterCertificateNotFound):
+			condition.Status = metav1.ConditionFalse
+			condition.Reason = string(consts.InvalidSecretRefReason)
+			condition.Message = consts.InvalidSecretRefMessage
```
pmalek: This is a nice starting point already and I'm OK with this being merged as is but I believe as a next step we could include the offending extension name in the condition's message. WDYT?

controller/dataplane/controller_utils.go
```diff
@@ -308,3 +309,47 @@ func isDeploymentReady(deploymentStatus appsv1.DeploymentStatus) (metav1.Conditi
...
+	if err != nil {
+		switch {
+		case errors.Is(err, ErrCrossNamespaceReference):
+			condition.Status = metav1.ConditionFalse
+			condition.Reason = string(consts.RefNotPermittedReason)
+			condition.Message = consts.RefNotPermittedMessage
+		case errors.Is(err, ErrKonnectExtensionNotFound):
+			condition.Status = metav1.ConditionFalse
+			condition.Reason = string(consts.InvalidExtensionRefReason)
+			condition.Message = consts.InvalidExtensionRefMessage
```
pmalek: WDYT about changing the error message ever so slightly to explicitly say that the extension was not found? Receiving an error message that it's invalid could indicate that it does exist but something's wrong with it.

controller/dataplane/dataplanekonnectextension_controller.go
```diff
@@ -0,0 +1,119 @@
...
+				Namespace: namespace,
+				Name:      ext.Name,
+			},
+		})
+	}
+	return recs
+}
+
+// Reconcile reconciles a DataPlaneKonnectExtension object.
+func (r *DataPlaneKonnectExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
pmalek, line 85: Can we also create an issue to add an `envtest` based test for this reconciler?

pmalek (commented):
Looks pretty much ready to me. I've left 1 comment with a proposal to change the condition message and 2 nits.

controller/dataplane/controller_utils.go
mlavacca: Yep, I agree. I've changed the message to include info about the offending object.

controller/dataplane/dataplanekonnectextension_controller.go
mlavacca: https://github.com/Kong/gateway-operator/issues/734

pmalek (commented):
1 thing I've just realized is that we're missing RBAC permissions for accessing `DataPlaneKonnectExtension`s (soon to be renamed to `KonnectExtension`s).

Do you mind adding `kubebuilder` RBAC markers to https://github.com/Kong/gateway-operator/blob/6b9a627106115acc27048f5fe57b0f063fbb9449/controller/dataplane/controller_rbac.go, adding said markers also for `DataPlaneKonnectExtensionReconciler` in `controller/dataplane/dataplanekonnectextension_controller_rbac.go` 
 and regenerating the manifests?

pmalek (approved):

PR #715, opened 2024-10-08
chore(refactor): do not hardcode certs for tests

programmer04:

There are dedicated helpers for dealing with certs for tests in the repo, thus there is no need to hardcode certificates anywhere (besides YAMLs).


pmalek (approved):
We could at some point refactor that so that each test creates its own cert and not share it.

PR #722, opened 2024-10-09
chore(konnect): rely on CEL rules to validate server url

czeslavo:

Adds less restrictive `NewServerURL` constructor that considers the value passed from the object always valid as the validation is done on the CRD CEL rules level. 

Depends on https://github.com/Kong/kubernetes-configuration/pull/121.

**Which issue this PR fixes**

Part of https://github.com/Kong/kubernetes-configuration/issues/63.

### DIFF
--- a/controller/konnect/reconciler_generic.go
+++ b/controller/konnect/reconciler_generic.go
@@ -47,8 +47,8 @@ type KonnectEntityReconciler[T constraints.SupportedKonnectEntityType, TEnt cons
 
 // KonnectEntityReconcilerOption is a functional option for the KonnectEntityReconciler.
 type KonnectEntityReconcilerOption[
-	T constraints.SupportedKonnectEntityType,
-	TEnt constraints.EntityType[T],
+T constraints.SupportedKonnectEntityType,
+TEnt constraints.EntityType[T],
 ] func(*KonnectEntityReconciler[T, TEnt])
 
 // WithKonnectEntitySyncPeriod sets the sync period for the reconciler.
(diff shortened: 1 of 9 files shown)


controller/konnect/reconciler_konnectapiauth.go
```diff
@@ -128,12 +126,12 @@ func (r *KonnectAPIAuthConfigurationReconciler) Reconcile(
 		return ctrl.Result{}, err
 	}
 
-	serverURL, err := getKonnectServerURL(apiAuth.Spec.ServerURL)
+	serverURL := ops.NewServerURL(apiAuth.Spec.ServerURL)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
```
pmalek:
```suggestion
```

czeslavo: Good catch 😅

randmonkey (approved):

controller/konnect/reconciler_generic.go
```diff
@@ -574,7 +575,7 @@ func setServerURLAndOrgIDFromAPIAuthConfiguration(
 	},
 	apiAuth konnectv1alpha1.KonnectAPIAuthConfiguration,
 ) {
-	ent.GetKonnectStatus().ServerURL = apiAuth.Spec.ServerURL
+	ent.GetKonnectStatus().ServerURL = ops.NewServerURL(apiAuth.Spec.ServerURL).String()
```
pmalek:
```suggestion
	ent.GetKonnectStatus().ServerURL = serverURL.String()
```

pmalek (approved):

controller/konnect/reconciler_generic.go
czeslavo: [fbab025](https://github.com/Kong/gateway-operator/pull/722/commits/fbab02505ddc2b52e8cdb9d61d7ffb4612423d5b)

controller/konnect/reconciler_generic.go
```diff
@@ -572,9 +573,10 @@ func setServerURLAndOrgIDFromAPIAuthConfiguration(
 	ent interface {
 		GetKonnectStatus() *konnectv1alpha1.KonnectEntityStatus
 	},
+	serverURL ops.ServerURL,
 	apiAuth konnectv1alpha1.KonnectAPIAuthConfiguration,
```
pmalek: Given that we're not extracting server URL from provided `apiAuth` anymore and we only need it to get the org ID I believe we can change the 3rd arg to be just that: orgID string. WDYT?

Issue #461, opened 2024-08-02
Add configuration options documentation


pmalek:
### Problem statement

KGO is missing docs on the configuration options.

This has already been asked by users in 

- https://github.com/Kong/gateway-operator/issues/425#issuecomment-2250124324

### Proposed solution

- Add a page under https://docs.konghq.com/gateway-operator/ similar to the one we have for KIC https://docs.konghq.com/kubernetes-ingress-controller/latest/reference/cli-arguments/

### Acceptance criteria

- [ ] As a User I have access to documentation page that lists configuration options for KGO

pmalek:
This is already available in newest (unpublished) version of the docs, e.g. in here: https://deploy-preview-8084--kongdocs.netlify.app/gateway-operator/unreleased/reference/cli-arguments/

Will be published when all 1.4.x related docs are merged (e.g. https://github.com/Kong/docs.konghq.com/pull/8084, https://github.com/Kong/docs.konghq.com/pull/8083)

PR #724, opened 2024-10-10
Chore(docs): add scripts to generate CLI argument docs

randmonkey:
Move scripts to generate CLI argument docs here from https://github.com/Kong/gateway-operator-enterprise.
**Which issue this PR fixes**

Part of #461.

We need to update the release template doc about generating CLI argument docs.

**PR Readiness Checklist**:


czeslavo:
Didn't we agree to have this only in the EE repo as OSS args are only a subset of all available ones? I think we should aim for a solution that will enable us to distinguish between OSS vs. EE but have a single page with those in docs, which could probably be generated on the EE side, WDYT?

randmonkey:
Close as not required.

PR #738, opened 2024-10-11
feat: make KonnectExtension conditionally enabled

mlavacca:

The KonnectExtension functionality is enabled only in case the --enable-controller-konnect flag is set.

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:


pmalek (commented):
I'm wondering if it would be reasonable to separate the KonnectExtension controller from Konnect entities controllers 🤔 

Those are 2 different functionalities (yet centered around 1 platform. WDYT?

mlavacca:
> I'm wondering if it would be reasonable to separate the KonnectExtension controller from Konnect entities controllers 🤔
> 
> Those are 2 different functionalities (yet centered around 1 platform. WDYT?

I've been uncertain about it myself, but eventually produced this reasoning:
- `konnectExtension` references a  Konnect ControlPlane. Today such a reference is KonnectID only, but we have the space to make it reference an actual `KonnectGatewayControlPlane`
- in the near future we'll likely implement the `KonnectExtension` for KGO controlPlanes as well. They will likely need to access a `KonnectGatewayControlPlane` object of type KIC that will have to be created in the cluster, and they will need even more likely to be bound to a `konnectapiauthconfigurations.konnect.konghq.com` to retrieve the access token from.
- this means that in the future the KonnectExtension will need to access Konnect-specific resources in the cluster

I think that if we add this constraint now we can always relax it, but in case the above happens, adding this constraint afterward can be problematic for users.

Does it make sense to you?

pmalek:
> > I'm wondering if it would be reasonable to separate the KonnectExtension controller from Konnect entities controllers 🤔
> > Those are 2 different functionalities (yet centered around 1 platform. WDYT?
> 
> I've been uncertain about it myself, but eventually produced this reasoning:
> 
> * `konnectExtension` references a  Konnect ControlPlane. Today such a reference is KonnectID only, but we have the space to make it reference an actual `KonnectGatewayControlPlane`
> * in the near future we'll likely implement the `KonnectExtension` for KGO controlPlanes as well. They will likely need to access a `KonnectGatewayControlPlane` object of type KIC that will have to be created in the cluster, and they will need even more likely to be bound to a `konnectapiauthconfigurations.konnect.konghq.com` to retrieve the access token from.
> * this means that in the future the KonnectExtension will need to access Konnect-specific resources in the cluster
> 
> I think that if we add this constraint now we can always relax it, but in case the above happens, adding this constraint afterward can be problematic for users.
> 
> Does it make sense to you?

That's reasonable 👍 Thanks for the explanation.

CHANGELOG.md
```diff
@@ -106,6 +106,9 @@
 - Apply Konnect-related customizations to `DataPlane`s that properly reference `KonnectExtension`
   resources.
   [#714](https://github.com/Kong/gateway-operator/pull/714)
+- The KonnectExtension functionality is enabled only when the `--enable-controller-konnect`
+  flag is set.
```
pmalek: Might also mention the env `GATEWAY_OPERATOR_ENABLE_CONTROLLER_KONNECT` which is equivalent.

pmalek (approved):

pmalek (approved):

pmalek (approved):

pmalek:
The e2e tests will fail (as they are already failing after merging #736) because helm chart (that's use in e2e tests) does not account yet for new RBAC rules (or CRDs):

```
manager W1011 10:42:01.674635       1 reflector.go:561] pkg/mod/k8s.io/client-go@v0.31.1/tools/cache/reflector.go:243: failed to list *v1alpha1.DataPlaneKon
nectExtension: dataplanekonnectextensions.gateway-operator.konghq.com is forbidden: User "system:serviceaccount:kong-kgo:controller-manager" cannot list res
ource "dataplanekonnectextensions" in API group "gateway-operator.konghq.com" at the cluster scope
manager E1011 10:42:01.674721       1 reflector.go:158] "Unhandled Error" err="pkg/mod/k8s.io/client-go@v0.31.1/tools/cache/reflector.go:243: Failed to watc
h *v1alpha1.DataPlaneKonnectExtension: failed to list *v1alpha1.DataPlaneKonnectExtension: dataplanekonnectextensions.gateway-operator.konghq.com is forbidd
en: User \"system:serviceaccount:kong-kgo:controller-manager\" cannot list resource \"dataplanekonnectextensions\" in API group \"gateway-operator.konghq.co
m\" at the cluster scope" logger="UnhandledError"
```

https://github.com/Kong/charts/pull/1138 adds new policy rules (but that will still fail until we release new chart version).

pmalek:
The fact that the https://github.com/Kong/gateway-operator/blob/4bffb353c8b8fab8fa3a3371e8587710a596596c/test/e2e/test_helm_install_upgrade.go test passed here means that the predicates for

- upgrade from 1.3.0 to current
- upgrade from nightly to current

are too loose 🤔 This might require some investigation

PR #740, opened 2024-10-11
chore: impersonate KGO RBACs with `make _run`

mlavacca:

When running `make run`, a new context that uses the KGO RBACs is set and used. This way, `make run` will allow us to properly test with the actual KGO RBACs, instead of using the admin user.

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -507,6 +507,7 @@ impersonate-kgo:
 	KUBECONFIG=$(TMP_KUBECONFIG) kubectl config set-credentials kgo --token=$(shell kubectl create token --namespace=kong-system controller-manager)
 	KUBECONFIG=$(TMP_KUBECONFIG) kubectl config set-context kgo --cluster=$(shell kubectl config get-contexts | grep '^\*' | tr -s ' ' | cut -d ' ' -f 3) --user=kgo --namespace=kong-system
 	KUBECONFIG=$(TMP_KUBECONFIG) kubectl config use-context kgo
+
 # Run a controller from your host.
 .PHONY: run
 run: webhook-certs-dir manifests generate install.all _ensure-kong-system-namespace install.rbacs
--- a/Makefile
+++ b/Makefile
@@ -510,7 +510,7 @@ impersonate-kgo:
 # Run a controller from your host.
 .PHONY: run
 run: webhook-certs-dir manifests generate install.all _ensure-kong-system-namespace install.rbacs
-	$(MAKE) _run
+	@$(MAKE) _run
 
 # Run the operator without checking any preconditions, installing CRDs etc.
 # This is mostly useful when 'run' was run at least once on a server and CRDs, RBACs
--- a/Makefile
+++ b/Makefile
@@ -497,20 +497,31 @@ webhook-certs-dir:
 _ensure-kong-system-namespace:
 	@kubectl create ns kong-system 2>/dev/null || true
 
+TMP_DIR := $(shell mktemp -d)
+KUBECONFIG ?= $(HOME)/.kube/config
+TMP_KUBECONFIG := $(TMP_DIR)/kubeconfig
+
+.PHONY: impersonate-kgo
+impersonate-kgo:
+	cp $(KUBECONFIG) $(TMP_KUBECONFIG)
+	KUBECONFIG=$(TMP_KUBECONFIG) kubectl config set-credentials kgo --token=$(shell kubectl create token --namespace=kong-system controller-manager)
+	KUBECONFIG=$(TMP_KUBECONFIG) kubectl config set-context kgo --cluster=$(shell kubectl config get-contexts | grep '^\*' | tr -s ' ' | cut -d ' ' -f 3) --user=kgo --namespace=kong-system
+	KUBECONFIG=$(TMP_KUBECONFIG) kubectl config use-context kgo
+
+
+
 # Run a controller from your host.
-# TODO: In order not to rely on 'main' version of Gateway API CRDs address but
-# on the tag that is used in code (defined in go.mod) address this by solving
-# https://github.com/Kong/gateway-operator/pull/480.
 .PHONY: run
-run: webhook-certs-dir manifests generate install.all _ensure-kong-system-namespace
-	@$(MAKE) _run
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -501,9 +501,19 @@ _ensure-kong-system-namespace:
 # TODO: In order not to rely on 'main' version of Gateway API CRDs address but
 # on the tag that is used in code (defined in go.mod) address this by solving
 # https://github.com/Kong/gateway-operator/pull/480.
```
pmalek: This comment is out of date already so we might as well remove it.

```suggestion
```

Makefile
```diff
@@ -501,9 +501,19 @@ _ensure-kong-system-namespace:
...
 .PHONY: run
-run: webhook-certs-dir manifests generate install.all _ensure-kong-system-namespace
-	@$(MAKE) _run
+run: webhook-certs-dir manifests generate install.all install.rbacs _ensure-kong-system-namespace
+	@(CONTEXT=$$(kubectl config current-context); \
+	trap 'kubectl config use-context $$CONTEXT; exit 1' INT ERR EXIT; \
+	kubectl config set-credentials kgo --token=$(shell kubectl create token --namespace=kong-system controller-manager); \
+	kubectl config set-context kgo --cluster=$(shell kubectl config get-contexts | grep '^\*' | tr -s ' ' | cut -d ' ' -f 3) --user=kgo --namespace=kong-system; \
+	kubectl config use-context kgo; \
+	$(MAKE) _run)
```
pmalek: Could we put this into a separate Makefile target so that we can also run something like `make _run` but with this added in? Otherwise we end up running all the generators etc which takes a while. WDYT? 

We could achieve this by either putting this code inside `_run` or into another target which wraps `_run`.

Makefile
```diff
@@ -501,9 +501,19 @@ _ensure-kong-system-namespace:
...
+# Once the operator is stopped, the context will be switched back to the original one.
 .PHONY: run
-run: webhook-certs-dir manifests generate install.all _ensure-kong-system-namespace
-	@$(MAKE) _run
+run: webhook-certs-dir manifests generate install.all install.rbacs _ensure-kong-system-namespace
+	@(CONTEXT=$$(kubectl config current-context); \
+	trap 'kubectl config use-context $$CONTEXT; exit 1' INT ERR EXIT; \
+	kubectl config set-credentials kgo --token=$(shell kubectl create token --namespace=kong-system controller-manager); \
+	kubectl config set-context kgo --cluster=$(shell kubectl config get-contexts | grep '^\*' | tr -s ' ' | cut -d ' ' -f 3) --user=kgo --namespace=kong-system; \
+	kubectl config use-context kgo; \
```
pmalek: Instead of manipulating current context how about we just run the operator using a curated context? We could specify e.g. the cli arg `kubeconfig` / env `GATEWAY_OPERATOR_KUBECONFIG` to a kubeconfig stored somewhere in `/tmp`. WDYT?

This will not require us to change the context if we'd like to apply objects against the cluster in a separate terminal.

mlavacca:
@pmalek I've addressed both of your comments. Now the target copies the KUBECONFIG into a tmp folder, and sets the new context. Such a TMP_KUBECONFIG is used in _run. Also, the impersonation is a prerequisite to _run.

Makefile
```diff
@@ -497,20 +497,30 @@ webhook-certs-dir:
 _ensure-kong-system-namespace:
 	@kubectl create ns kong-system 2>/dev/null || true
 
+TMP_DIR=$(shell pwd)/tmp
```
pmalek: Why not use `mktemp`? This will pollute the working dir and won't be cleaned after e.g. host restart.

Makefile
```diff
@@ -497,20 +497,31 @@ webhook-certs-dir:
...
+
+.PHONY: impersonate-kgo
+impersonate-kgo:
+	cp $(KUBECONFIG) $(TMP_KUBECONFIG)
+	KUBECONFIG=$(TMP_KUBECONFIG) kubectl config set-credentials kgo --token=$(shell kubectl create token --namespace=kong-system controller-manager)
+	KUBECONFIG=$(TMP_KUBECONFIG) kubectl config set-context kgo --cluster=$(shell kubectl config get-contexts | grep '^\*' | tr -s ' ' | cut -d ' ' -f 3) --user=kgo --namespace=kong-system
+	KUBECONFIG=$(TMP_KUBECONFIG) kubectl config use-context kgo
+
+
+
```
```suggestion

```

Makefile
```diff
@@ -569,6 +580,11 @@ KUBERNETES_CONFIGURATION_CRDS_CRDS_LOCAL_PATH = $(shell go env GOPATH)/pkg/mod/$
 install.kubernetes-configuration-crds: kustomize
 	$(KUSTOMIZE) build $(KUBERNETES_CONFIGURATION_CRDS_CRDS_LOCAL_PATH) | kubectl apply -f -
 
+# Install RBACs from config/rbac into the K8s cluster specified in ~/.kube/config.
+.PHONY: install.rbacs
+install.rbacs: kustomize
+	$(KUSTOMIZE) build config/rbac | kubectl apply -f -
```
pmalek, line 596: 👍

pmalek (commented):
Minor whitespace nit. Otherwise 👍

Makefile
```diff
@@ -497,20 +497,28 @@ webhook-certs-dir:
 _ensure-kong-system-namespace:
 	@kubectl create ns kong-system 2>/dev/null || true
 
+TMP_DIR := $(shell mktemp -d)
```
pmalek: I just realized this: this will cause a new temp dir to be created every time `make` is called right? Can we put this into `impersonate-kgo` to prevent this?

pmalek: To solve this we could use something like: https://github.com/Kong/gateway-operator/compare/mlavacca/make-run-impersonate...pmalek/make-run-impersonate?expand=1

mlavacca: For posterity, @pmalek implemented this in d246518872ffcb668b40440eee747f031fd24867

pmalek (approved):
👍

PR #745, opened 2024-10-14
feat(konnect): make KonnectGatewayControlPlane creation more robust

pmalek:

This PR improves the `KonnectGatewayControlPlane` creation (if accepted this should be implemented for all types) by reacting to `HTTP 409 Conflict` errors with listing in Konnect CPs with matching UID label.

When that's successful this Konnect ControlPlane is adopted. The label that's used, means that the originating Kubernetes resource is the one that is being reconciled. This prevents adoption of Konnect entities created by other Kubernetes resources.

**Which issue this PR fixes**

Fixes #

This can be tested using a simple manifest file for `KonnectGatewayControlPlane`:

```
kind: KonnectGatewayControlPlane
metadata:
  name: test1
  namespace: default
spec:
  name: test1
  konnect:
    authRef:
      name: konnect-api-auth-dev-1
```

and this little script:

```
kubectl delete konnectgatewaycontrolplanes.konnect.konghq.com test1
while true ; do
  kubectl wait --timeout=-1s --for=delete konnectgatewaycontrolplane test1 &&
  kubectl apply -f f.yaml && 
  kubectl wait --for=condition=Programmed=true --timeout=30s konnectgatewaycontrolplane test1 && 
  kubectl delete konnectgatewaycontrolplanes.konnect.konghq.com test1;
done
```

**PR Readiness Checklist**:

### DIFF
--- a/controller/konnect/ops/ops.go
+++ b/controller/konnect/ops/ops.go
@@ -108,6 +108,8 @@ func Create[
 	var errConflict *sdkkonnecterrs.ConflictError
 	switch {
 	case errors.As(err, &errConflict):
+                // If there was a conflict on the create request, we can assume the entity already exists.
+                // We'll get its Konnect ID by listing all entities of its type filtered by the Kubernetes object UID.
 		switch ent := any(e).(type) {
 		case *konnectv1alpha1.KonnectGatewayControlPlane:
 			id, err = getControlPlaneForUID(ctx, sdk.GetControlPlaneSDK(), ent)
--- a/controller/konnect/ops/ops.go
+++ b/controller/konnect/ops/ops.go
@@ -257,19 +257,16 @@ func Update[
 	cl client.Client,
 	e TEnt,
 ) (ctrl.Result, error) {
-	var (
-		ent = TEnt(e)
-		now = time.Now()
-	)
+	var now = time.Now()
 
-	if ok, res := shouldUpdate(ctx, ent, syncPeriod, now); !ok {
+	if ok, res := shouldUpdate(ctx, e, syncPeriod, now); !ok {
 		return res, nil
 	}
 
-	if ent.GetKonnectStatus().GetKonnectID() == "" {
+	if e.GetKonnectStatus().GetKonnectID() == "" {
 		return ctrl.Result{}, fmt.Errorf(
 			"can't update %T %s when it does not have the Konnect ID",
-			ent, client.ObjectKeyFromObject(ent),
+			e, client.ObjectKeyFromObject(e),
 		)
 	}
 
--- a/controller/konnect/ops/ops.go
+++ b/controller/konnect/ops/ops.go
@@ -62,13 +62,13 @@ func Create[
 		id, reason, err = createControlPlane(ctx, sdk.GetControlPlaneSDK(), sdk.GetControlPlaneGroupSDK(), cl, ent)
 		// TODO: modify the create* operation wrappers to return Konnect ID and error reason.
 	case *configurationv1alpha1.KongService:
-		id, reason, err = createService(ctx, sdk.GetServicesSDK(), ent)
+		err = createService(ctx, sdk.GetServicesSDK(), ent)
 	case *configurationv1alpha1.KongRoute:
-		id, reason, err = createRoute(ctx, sdk.GetRoutesSDK(), ent)
+		err = createRoute(ctx, sdk.GetRoutesSDK(), ent)
 	case *configurationv1.KongConsumer:
-		id, reason, err = createConsumer(ctx, sdk.GetConsumersSDK(), sdk.GetConsumerGroupsSDK(), cl, ent)
+		err = createConsumer(ctx, sdk.GetConsumersSDK(), sdk.GetConsumerGroupsSDK(), cl, ent)
 	case *configurationv1beta1.KongConsumerGroup:
-		id, reason, err = createConsumerGroup(ctx, sdk.GetConsumerGroupsSDK(), ent)
+		err = createConsumerGroup(ctx, sdk.GetConsumerGroupsSDK(), ent)
 	case *configurationv1alpha1.KongPluginBinding:
 		err = createPlugin(ctx, cl, sdk.GetPluginSDK(), ent)
 	case *configurationv1alpha1.KongUpstream:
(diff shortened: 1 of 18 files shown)


controller/konnect/ops/ops.go
```diff
@@ -99,6 +104,35 @@ func Create[
...
 	}
 
+	var errConflict *sdkkonnecterrs.ConflictError
+	switch {
+	case errors.As(err, &errConflict):
+		switch ent := any(e).(type) {
+		case *konnectv1alpha1.KonnectGatewayControlPlane:
+			id, err = getControlPlaneForUID(ctx, sdk.GetControlPlaneSDK(), ent)
+			// ---------------------------------------------------------------------
+			// TODO: add other Konnect types
```
randmonkey, line 121: Are we going to implement the same thing for other types in this PR ore in another PR?

controller/konnect/watch_konnectcontrolplane.go
```diff
@@ -26,7 +29,22 @@ func KonnectGatewayControlPlaneReconciliationWatchOptions(
 ) []func(*ctrl.Builder) *ctrl.Builder {
 	return []func(*ctrl.Builder) *ctrl.Builder{
 		func(b *ctrl.Builder) *ctrl.Builder {
-			return b.For(&konnectv1alpha1.KonnectGatewayControlPlane{})
+			return b.For(
+				&konnectv1alpha1.KonnectGatewayControlPlane{},
+				builder.WithPredicates(
+					predicate.Or(
+						predicate.GenerationChangedPredicate{},
```
randmonkey: So this means only enqueue `update`s of controlplanes when spec changed? I think comments should be added here.

controller/konnect/ops/ops.go
```diff
@@ -216,7 +250,13 @@ func shouldUpdate[
...
-](ctx context.Context, sdk SDKWrapper, syncPeriod time.Duration, cl client.Client, e *T) (ctrl.Result, error) {
+](
+	ctx context.Context,
+	sdk SDKWrapper,
+	syncPeriod time.Duration,
+	cl client.Client,
+	e TEnt,
+) (ctrl.Result, error) {
 	var (
 		ent = TEnt(e)
```
randmonkey:
```suggestion
```
Since we changed the function signature, the `ent` is not required and the appearance of `ent` should be changed to `e`.

controller/konnect/ops/ops.go
pmalek: I'd suggest we tackle controlplane first and review this approach and then tackle other types. WDYT?

controller/konnect/watch_konnectcontrolplane.go
pmalek: This shouldn't be part of the this pr. I'll remove this when I manage to get back to this pr.

controller/konnect/ops/ops_controlplane.go
```diff
@@ -8,45 +8,88 @@ import (
...
+	}
+
+	if respList == nil || respList.ListControlPlanesResponse == nil {
+		return "", fmt.Errorf("failed listing ControlPlanes: %w", ErrNilResponse)
+	}
+
+	var id string
+	for _, listCP := range respList.ListControlPlanesResponse.Data {
+		id = *listCP.ID
+		break
```
czeslavo:
```suggestion
		if listCP.ID != nil && *listCP.ID != "" {
			id = *listCP.ID
			break
		}
```

controller/konnect/watch_konnectcontrolplane.go
randmonkey: Removed these.

controller/konnect/ops/ops.go
```diff
@@ -99,6 +105,40 @@ func Create[
 		return nil, fmt.Errorf("unsupported entity type %T", ent)
 	}
 
+	var errConflict *sdkkonnecterrs.ConflictError
+	switch {
+	case errors.As(err, &errConflict):
+		switch ent := any(e).(type) {
```
czeslavo, line 117:
```suggestion
                // If there was a conflict on the create request, we can assume the entity already exists.
                // We'll get its Konnect ID by listing all entities of its type filtered by the Kubernetes object UID.
		switch ent := any(e).(type) {
```

controller/konnect/ops/ops.go
```diff
@@ -99,6 +105,40 @@ func Create[
...
+		if err == nil && id != "" {
+			e.SetKonnectID(id)
+			SetKonnectEntityProgrammedCondition(e)
+		} else {
+			SetKonnectEntityProgrammedConditionFalse(e, consts.KonnectEntitiesFailedToCreateReason, err.Error())
+		}
+	case err != nil:
+		if id != "" {
+			e.SetKonnectID(id)
+		}
```
czeslavo: Should we expect the `id` to be non-empty if the error is not `nil`? It looks like breaking the convention of not returning a valid value when an error is returned.

controller/konnect/ops/ops.go
```diff
@@ -282,6 +330,22 @@ func Update[
 		return ctrl.Result{}, fmt.Errorf("unsupported entity type %T", ent)
 	}
 
+	if err != nil {
+		if id != "" {
```
czeslavo: Ditto regarding the convention of not returning a valid object when an error is not nil.

czeslavo (commented):
A couple of nits, but in general I think the approach of looking up an entity filtering by its Kubernetes UID is correct.

controller/konnect/ops/ops.go
pmalek: So this is somewhat of a special case when we create a resource and then set associations like we do for CPs with group membership: https://github.com/Kong/gateway-operator/blob/9093830f84aedfe5e96aa3c812aef72d8f2c541c/controller/konnect/ops/ops_controlplane.go#L88-L90.

So the object at this point would exist but there would be an error that we'd surface in its conditions. Do you think we can improve this somehow?

czeslavo (approved):

controller/konnect/ops/ops.go
czeslavo: We discussed this on a call, and the conclusion was to change the `[create|update]ControlPlane` signatures to not return an ID and a condition reason, but instead return a specific error type when the relations' setting fails and handle it in the top-level `Create` or `Update` functions accordingly (set the ID on an object and propagate the `Programmed` condition with `False` and the returned reason).

PR #751, opened 2024-10-17
fix: handle KongSNI conflict on creation

czeslavo:

Handles conflicts on KongSNI creation.

**Which issue this PR fixes**

Follow up on https://github.com/Kong/gateway-operator/pull/745.


controller/konnect/ops/ops.go
```diff
@@ -148,7 +150,6 @@ func Create[
 		}
 	case errors.As(err, &errRelationsFailed):
 		SetKonnectEntityProgrammedConditionFalse(e, errRelationsFailed.Reason, err.Error())
-		e.SetKonnectID(errRelationsFailed.KonnectID)
```
czeslavo, line 155: This can be dropped as we set the ID in `create*` functions anyway.

pmalek (approved):

PR #752, opened 2024-10-17
fix(konnect): handle KongRoute conflict on creation

pmalek:

Follow up on #749 for `KongRoute`.

### DIFF
--- a/controller/konnect/ops/ops.go
+++ b/controller/konnect/ops/ops.go
@@ -413,20 +413,18 @@ type entityWithID interface {
 }
 
 // sliceToEntityWithIDSlice converts a slice of entities to a slice of entityWithID.
-// The conversion is done by asserting the type of the entity to entityWithID.
-// It panics if the conversion fails.
 func sliceToEntityWithIDSlice[
 	T any,
+	TPtr interface {
+		*T
+		GetID() *string
+	},
 ](
 	slice []T,
 ) []entityWithID {
 	result := make([]entityWithID, 0, len(slice))
 	for _, item := range slice {
-		converted, ok := any(&item).(entityWithID)
-		if !ok {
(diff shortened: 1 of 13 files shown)


controller/konnect/ops/ops.go
```diff
@@ -419,3 +421,47 @@ func logEntityNotFoundRecreating[
 			"id", id,
 		)
 }
+
+type entityWithID interface {
+	GetID() *string
+}
+
+func fromSliceToSlice[T any](
```
randmonkey: Could we some comments and give it a clearer name? `fromSliceToSlice` is really hard to understand.

controller/konnect/ops/ops.go
```diff
@@ -126,7 +126,7 @@ func Create[
 
 	switch {
 	case errors.As(err, &errConflict) ||
-		// Service API returns 400 for conflicts
+		// ControlPlane resource APIs return 400 for conflicts
 		errSdkBody.Code == 3 && errSdkBody.Message == "data constraint error":
```
randmonkey: Could we define some function or constant to eliminate the magic number and string here?

randmonkey (commented):
Generally LGTM, some minor comments remaining.

controller/konnect/ops/ops.go
pmalek: f71487ec2e4f0ce959131a05eb51b3486a77d822 PTAL

controller/konnect/ops/ops.go
pmalek: I've changed this to `sliceToEntityWithIDSlice` and added a comment.

f71487ec2e4f0ce959131a05eb51b3486a77d822

randmonkey (approved):

randmonkey (approved):

PR #756, opened 2024-10-18
fix(konnect): handle KongUpstream conflict on creation

pmalek:

Follow up on #752 for `KongUpstream`.

This PR also adds a separate type `CantPerformOperationWithoutControlPlaneIDError` that's reused by all entities having a CP ref, when the CP ID is not set.

### DIFF
--- a/controller/konnect/ops/ops_kongupstream.go
+++ b/controller/konnect/ops/ops_kongupstream.go
@@ -8,7 +8,7 @@ import (
 	sdkkonnectcomp "github.com/Kong/sdk-konnect-go/models/components"
 	sdkkonnectops "github.com/Kong/sdk-konnect-go/models/operations"
 	sdkkonnecterrs "github.com/Kong/sdk-konnect-go/models/sdkerrors"
-	"sigs.k8s.io/controller-runtime/pkg/client"
+	"github.com/samber/lo"
 	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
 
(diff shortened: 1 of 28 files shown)


controller/konnect/ops/ops_kongupstream.go
```diff
@@ -69,38 +63,10 @@ func updateUpstream(
...
-				// Create succeeded, createUpstream sets the status so no need to do this here.
-
-				return nil
-			default:
-				return FailedKonnectOpError[configurationv1alpha1.KongUpstream]{
-					Op:  UpdateOp,
-					Err: sdkError,
-				}
-			}
-		}
```
czeslavo: Why removing `NotFound` handling code?

pmalek: I've realized while working on those PRs that some entities have `Upsert*` functions (like upstream) while some have `Update*` (like controlplane). 

We can either do it here or create a separate PR to handle that ( the latter might actually be better). WDYT?

czeslavo: Ah, I see. Makes sense now. Let's keep it in this PR, but for others, it would be better to have separate one(s).

czeslavo (approved):

controller/konnect/ops/ops_kongupstream.go
pmalek: Needed to revert that as tests would need changing. Let's keep the original scope for this PR and handle those unnecessary error handling bits separately. They do not harm currently as they are pretty much noop.

PR #759, opened 2024-10-18
fix: handle KongVault creation conflicts

czeslavo:

Follow up on https://github.com/Kong/gateway-operator/pull/749 for `KongVault`.

### DIFF
--- a/test/envtest/konnect_entities_kongvault_test.go
+++ b/test/envtest/konnect_entities_kongvault_test.go
@@ -2,10 +2,12 @@ package envtest
 
 import (
 	"context"
+	"strings"
 	"testing"
 
 	sdkkonnectcomp "github.com/Kong/sdk-konnect-go/models/components"
 	sdkkonnectops "github.com/Kong/sdk-konnect-go/models/operations"
+	sdkkonnecterrs "github.com/Kong/sdk-konnect-go/models/sdkerrors"
 	"github.com/samber/lo"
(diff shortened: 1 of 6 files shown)


test/envtest/konnect_entities_kongvault_test.go
```diff
@@ -121,4 +123,57 @@ func TestKongVault(t *testing.T) {
...
+		)
+
+		t.Log("Setting up mock SDK for vault creation with conflict")
+		sdk.VaultSDK.EXPECT().CreateVault(mock.Anything, cp.GetKonnectStatus().GetKonnectID(), mock.MatchedBy(func(input sdkkonnectcomp.VaultInput) bool {
+			return input.Name == vaultBackend && input.Prefix == vaultPrefix
+		})).Return(&sdkkonnectops.CreateVaultResponse{
+			Vault: &sdkkonnectcomp.Vault{
+				ID: lo.ToPtr(vaultID),
+			},
+		}, &sdkkonnecterrs.ConflictError{})
```
pmalek: This should return just the error, no? Secondly, I believe this API won't return Conflict error but SDKError as for other entities living inside the CP.

```suggestion
		})).Return(nil, &sdkkonnecterrs.ConflictError{})
```

czeslavo: bd9a356

pmalek (approved):

PR #760, opened 2024-10-18
refactor(konnect): add handleDeleteError to uniformly handle delete errors

pmalek:

Introduces `handleDeleteError` which handles delete errors uniformly in every object type delete operation.

### DIFF
--- a/controller/konnect/ops/ops_errors.go
+++ b/controller/konnect/ops/ops_errors.go
@@ -152,7 +152,7 @@ func handleDeleteError[
 
 	var sdkError *sdkkonnecterrs.SDKError
 	if errors.As(err, &sdkError) {
-		if sdkError.StatusCode == 404 {
+		if sdkError.StatusCode == http.StatusNotFound {
 			logDeleteSkipped()
 			return nil
 		}
--- a/controller/konnect/ops/ops_errors.go
+++ b/controller/konnect/ops/ops_errors.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net/http"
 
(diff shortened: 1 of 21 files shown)


controller/konnect/ops/ops_errors.go
```diff
@@ -123,3 +127,42 @@ func SDKErrorIsConflict(sdkError *sdkkonnecterrs.SDKError) bool {
...
+
+	var sdkNotFoundError *sdkkonnecterrs.NotFoundError
+	if errors.As(err, &sdkNotFoundError) {
+		logDeleteSkipped()
+		return nil
+	}
+
+	var sdkError *sdkkonnecterrs.SDKError
+	if errors.As(err, &sdkError) {
+		if sdkError.StatusCode == 404 {
```
programmer04:
```suggestion
		if sdkError.StatusCode == http.StatusNotFound {
```
It is interesting that linters miss this

mlavacca (approved):

controller/konnect/ops/ops_errors.go
randmonkey: There is such a linter that checks for "magic numbers". But in its default settings, only `0` and `1` are not treated as a magic number: https://github.com/tommy-muehle/go-mnd

pmalek: It can be enabled with `golangci-lint`: https://golangci-lint.run/usage/linters/#mnd 👍

programmer04: Maybe it is worth enabling it and see how it behaves

programmer04 (approved):

czeslavo (approved):

PR #765, opened 2024-10-21
fix(konnect): make setting status conditions not update on every reconciliation

pmalek:

When testing Konnect entities with multiple entities applied e.g. 10s of consumers, using a script like this:

```
#!/bin/bash

function id() {
	cat /dev/urandom | LC_ALL=C tr -dc 'a-z0-9' | fold -w 10 | head -n 1
}

function deploy_consumer() {
	ID=$(id)

	kubectl apply -f - <<- EOF
	kind: KongConsumer
	metadata:
	  name: test-${ID}
	username: test-${ID}
	spec:
	  controlPlaneRef:
	    type: konnectNamespacedRef
	    konnectNamespacedRef:
	      name: test1
	EOF
}

function deploy_route() {
	ID=$(id)

	kubectl apply -f - <<- EOF
	kind: KongRoute
	metadata:
	  name: test-${ID}
	spec:
	  name: test-${ID}
	  protocols:
	  - http
	  hosts:
	  - example1.com
	  serviceRef:
	    type: namespacedRef
	    namespacedRef:
	      name: test1
	EOF
}


for i in $(seq 1 ${2});
do
	case ${1} in

	  consumer)
	    deploy_consumer &
	    ;;
	  route)
	    deploy_route &
	    ;;

	  *)
	    STATEMENTS
	    ;;
	esac
done
```

```
./main.sh consumer 50
```

I've noticed that status conditions responsible for indicating the referenced entity (e.g. consumer or CP) are repeatedly updated with `LastTransitionTime` (every 2s or so).

This PR ensures that this is not happening by checking if the condition has the status, reason, message and type as expected.

**Special notes**

Not sure (yet) what's the cause of requeues that can be observed. This can be easily triggered with multiple entities. Having just 1 or 2 doesn't trigger this behavior.

### DIFF
--- a/controller/controlplane/controller_reconciler_utils.go
+++ b/controller/controlplane/controller_reconciler_utils.go
@@ -230,7 +230,7 @@ func (r *Reconciler) ensureDeployment(
 			}
 		}
 
-		return patch.ApplyPatchIfNonEmpty(ctx, r.Client, logger, existingDeployment, oldExistingDeployment, params.ControlPlane, updated)
+		return patch.ApplyPatchIfNotEmpty(ctx, r.Client, logger, existingDeployment, oldExistingDeployment, updated)
 	}
 
 	if !dataplaneIsSet {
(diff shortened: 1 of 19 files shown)


czeslavo (approved):
🎖️

PR #772, opened 2024-10-22
chore: remove status update from plugin ops

mlavacca:

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:

### DIFF
--- a/controller/konnect/ops/ops_kongpluginbinding.go
+++ b/controller/konnect/ops/ops_kongpluginbinding.go
@@ -3,12 +3,10 @@ package ops
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 
 	sdkkonnectcomp "github.com/Kong/sdk-konnect-go/models/components"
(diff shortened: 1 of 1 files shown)


controller/konnect/ops/ops_kongpluginbinding.go
```diff
@@ -66,34 +67,51 @@ func updatePlugin(
...
+			default:
+				return FailedKonnectOpError[configurationv1alpha1.KongPluginBinding]{
+					Op:  UpdateOp,
+					Err: sdkError,
+				}
+			}
+		}
+
 		return errWrap
 	}
```
czeslavo: I think that `createPlugin` step was here before but was removed intentionally as it is not really needed due to `UpsertPlugin` method we call here which does the same already logically.

mlavacca: Yep, that's a good catch. I've reverted the change 👍

czeslavo (approved):

PR #776, opened 2024-10-23
chore: log when reducing KongPluginBindings

mlavacca:

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:

### DIFF
--- a/controller/konnect/reconciler_kongplugin.go
+++ b/controller/konnect/reconciler_kongplugin.go
@@ -219,6 +219,7 @@ func (r *KongPluginReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 				if err := k8sreduce.ReduceKongPluginBindings(ctx, clientWithNamespace, kpbList); err != nil {
 					return ctrl.Result{}, fmt.Errorf("failed to reduce KongPluginBindings: %w", err)
 				}
+				log.Info(logger, "deleted duplicated KongPluginBindings for kongPlugin", kongPlugin)
 			}
 
 		}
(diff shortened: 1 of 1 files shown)


controller/konnect/reconciler_kongplugin.go
```diff
@@ -217,6 +217,7 @@ func (r *KongPluginReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 
 			default:
 				if err := k8sreduce.ReduceKongPluginBindings(ctx, clientWithNamespace, kpbList); err != nil {
+					log.Info(logger, "deleted duplicated KongPluginBindings for kongPlugin", kongPlugin)
```
czeslavo: I think we should log inside of `ReduceKongPluginBindings` - here we are in the `err != nil` branch which is actually a reduce failure.

mlavacca: 🤦 I did it in a hurry and made this stupid mistake. I've put it out of the error branch, I think it's fine. Adding logs inside the reduce function would make it incoherent with the approach we already have with plenty of other reduce functions. I'm not saying it's not a valid point but it is probably something we may want to address separately by considering all the other functions as well. WDYT?

controller/konnect/reconciler_kongplugin.go
```diff
@@ -219,6 +219,7 @@ func (r *KongPluginReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 				if err := k8sreduce.ReduceKongPluginBindings(ctx, clientWithNamespace, kpbList); err != nil {
 					return ctrl.Result{}, fmt.Errorf("failed to reduce KongPluginBindings: %w", err)
 				}
+				log.Info(logger, "deleted duplicated KongPluginBindings for kongPlugin", kongPlugin)
```
programmer04:
```suggestion
				log.Info(logger, "deleted duplicated KongPluginBindings for KongPlugin", kongPlugin)
```

czeslavo (approved):

PR #784, opened 2024-10-23
chore(ci): collect diagnostics in release workflow

czeslavo:

Collect diagnostics in the reusable release workflow so we can investigate failing tests.


randmonkey (commented):
Why was `go.mod` updated with no update on go code? Is it because `apimachinery` get bumped in this PR?

czeslavo:
> Why was `go.mod` updated with no update on go code? Is it because `apimachinery` get bumped in this PR?

It's because in the meantime (between last PR merged and now) there was an update to k8s libraries. Our [updater](https://github.com/Kong/gateway-operator/blob/cdcfea4d31d772b9f16f1776dda89d748b6aefbc/hack/update-k8sio-gomod-replace.sh) script keeps them up-to-date and is also run in CI in the `verify` step. Because of that, I had to run it and commit changes to pass the `verify` step in CI.

programmer04 (approved):

PR #787, opened 2024-10-24
chore(konnect): do not handle upsert errors

pmalek:

`Upsert*` operations create entities if they do not exist so there's no need to handle that explicitly.

This is what this PR does.

The only entity thus far that requires this to be handled is ControlPlane which is using the newly added `handleUpdateError()`

Additionally, this PR also removes setting object status conditions in `ops_*.go` as that's done in `ops.go` for all types.

### DIFF
--- a/controller/konnect/ops/ops_errors.go
+++ b/controller/konnect/ops/ops_errors.go
@@ -129,6 +129,59 @@ func SDKErrorIsConflict(sdkError *sdkkonnecterrs.SDKError) bool {
 	return false
 }
 
+// handleUpdateError handles errors that occur during an update operation.
+// If the entity is not found, then it uses the provided create function to
+// recreate the it.
+func handleUpdateError[
+	T constraints.SupportedKonnectEntityType,
+	TEnt constraints.EntityType[T],
+](
+	ctx context.Context,
+	err error,
+	ent TEnt,
+	createFunc func(ctx context.Context) error,
+) error {
+	var (
+		sdkError *sdkkonnecterrs.SDKError
+		id       = ent.GetKonnectStatus().GetKonnectID()
+	)
+	if errors.As(err, &sdkError) {
+		switch sdkError.StatusCode {
+		case http.StatusNotFound:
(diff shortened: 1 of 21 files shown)


controller/konnect/ops/ops_errors.go
```diff
@@ -129,6 +129,59 @@ func SDKErrorIsConflict(sdkError *sdkkonnecterrs.SDKError) bool {
 	return false
 }
 
+// handleUpdateError handles errors that occur during an update operation.
+// If the entity is not found, then it uses the provided create function to
+// recreate the it.
```
czeslavo, line 134:
```suggestion
// recreate it.
```

controller/konnect/ops/ops_errors.go
```diff
@@ -129,6 +129,59 @@ func SDKErrorIsConflict(sdkError *sdkkonnecterrs.SDKError) bool {
...
+				Err: err,
+			}
+		}
+		return nil
+	}
+
+	return FailedKonnectOpError[T]{
+		Op:  UpdateOp,
+		Err: err,
+	}
```
czeslavo, line 182: Just to limit unnecessary branching:
```suggestion
	isNotFound := func() bool {
		var (
			notFoundError *sdkkonnecterrs.NotFoundError
			sdkError      *sdkkonnecterrs.SDKError
		)
		return errors.As(err, &notFoundError) ||
			errors.As(err, &sdkError) && sdkError.StatusCode == http.StatusNotFound
	}()

	if isNotFound {
		id := ent.GetKonnectStatus().GetKonnectID()
		logEntityNotFoundRecreating(ctx, ent, id)
		if createErr := createFunc(ctx); createErr != nil {
			return FailedKonnectOpError[T]{
				Op:  CreateOp,
				Err: fmt.Errorf("failed to create %s %s: %w", ent.GetTypeName(), id, createErr),
			}
		}
	}
	return FailedKonnectOpError[T]{
		Op:  UpdateOp,
		Err: err,
	}
```

czeslavo (approved):

PR #792, opened 2024-10-24
fix(konnect): don't block deletion when entity has no Konnect ID and is not Programmed

pmalek:

Deletion of entities should not be blocked when they do not have the Konnect ID assigned and are not Programmed.

### DIFF
--- a/controller/konnect/ops/ops.go
+++ b/controller/konnect/ops/ops.go
@@ -182,10 +182,14 @@ func Delete[
 ](ctx context.Context, sdk sdkops.SDKWrapper, cl client.Client, e *T) error {
 	ent := TEnt(e)
 	if ent.GetKonnectStatus().GetKonnectID() == "" {
-		return fmt.Errorf(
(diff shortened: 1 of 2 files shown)


controller/konnect/ops/ops.go
```diff
@@ -182,10 +182,14 @@ func Delete[
...
-		)
+		cond, ok := k8sutils.GetCondition(konnectv1alpha1.KonnectEntityProgrammedConditionType, ent)
+		if ok && cond.Status == metav1.ConditionTrue {
+			return fmt.Errorf(
+				"can't delete %T %s when it does not have the Konnect ID",
+				ent, client.ObjectKeyFromObject(ent),
+			)
+		}
+		return nil
 	}
```
czeslavo, line 192: Could we cover this logic in some unit test?

pmalek: 4642341dac30cb00d04eab84550d07629eb13216 adds tests.

czeslavo (approved):

PR #796, opened 2024-10-24
feat(konnect): handle BadRequest errors and other unrecoverable errors to prevent endless reconciliation

pmalek:

This PR add handling of BadRequest errors and other unrecoverable errors to prevent endless reconciliation.

When those types of errors are caught, operator will set object's Programmed status condition and not propagate the error to the caller, prevent it from bubbling up in the reconciliation loop.

### DIFF
--- a/controller/konnect/ops/ops.go
+++ b/controller/konnect/ops/ops.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"time"
 
+	sdkkonnecterrs "github.com/Kong/sdk-konnect-go/models/sdkerrors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -102,7 +103,10 @@ func Create[
 		return nil, fmt.Errorf("unsupported entity type %T", ent)
 	}
 
-	var errRelationsFailed KonnectEntityCreatedButRelationsFailedError
+	var (
+		errRelationsFailed KonnectEntityCreatedButRelationsFailedError
+		errSDK             *sdkkonnecterrs.SDKError
+	)
 	switch {
 	case ErrorIsCreateConflict(err):
 		// If there was a conflict on the create request, we can assume the entity already exists.
@@ -161,8 +165,17 @@ func Create[
 		} else {
 			SetKonnectEntityProgrammedConditionFalse(e, consts.KonnectEntitiesFailedToCreateReason, err.Error())
 		}
+
+	case errors.As(err, &errSDK):
+		switch {
+		case ErrorIsSDKErrorTypeField(err):
+			SetKonnectEntityProgrammedConditionFalse(e, consts.KonnectEntitiesFailedToCreateReason, errSDK.Error())
+		default:
+			SetKonnectEntityProgrammedConditionFalse(e, consts.KonnectEntitiesFailedToCreateReason, err.Error())
+		}
+
 	case errors.As(err, &errRelationsFailed):
-		SetKonnectEntityProgrammedConditionFalse(e, errRelationsFailed.Reason, err.Error())
+		SetKonnectEntityProgrammedConditionFalse(e, errRelationsFailed.Reason, errRelationsFailed.Err.Error())
 	case err != nil:
 		SetKonnectEntityProgrammedConditionFalse(e, consts.KonnectEntitiesFailedToCreateReason, err.Error())
 	default:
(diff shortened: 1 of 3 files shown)


controller/konnect/ops/ops.go
```diff
@@ -171,6 +184,14 @@ func Create[
...
 	logOpComplete(ctx, start, CreateOp, e, err)
 
+	// If the error is a type field error or bad request error, then don't propagate
+	// it to the caller.
+	// We cannot recover from this error as this requires user to change object's
+	// manifest. The entity's status is already updated with the error.
+	if ErrorIsSDKErrorTypeField(err) || ErrorIsSDKBadRequestError(err) {
+		err = nil
+	}
+
```
programmer04: This check is in a couple of places, maybe a dedicated helper like `IgnoreUnrecovableErr` and use it like that
```go
    return e, IgnoreUnrecovableErr(err)
```
wdyt?

controller/konnect/ops/ops_controlplane_test.go
```diff
@@ -408,92 +408,92 @@ func TestUpdateControlPlane(t *testing.T) {
...
+		// 				&sdkkonnecterrs.BadRequestError{
+		// 					Status: 400,
+		// 					Detail: "bad request",
+		// 				},
+		// 			)
+
+		// 		return sdk, sdkGroups, cp
+		// 	},
+		// 	expectedErr: true,
+		// },
```
programmer04: ?

pmalek: 🤦 

a1f22eb1cdd67e3a8f2a880b772bee92eb06cca6 fixes that.

pmalek: Thanks for staying vigilant 🏅

controller/konnect/ops/ops.go
pmalek: Good call. 62eb3f5e2d1e5a9b8603a03dd58fbd5fcb180eaf should handle that.

programmer04 (approved):

controller/konnect/ops/ops.go
```diff
@@ -161,8 +165,17 @@ func Create[
...
 			SetKonnectEntityProgrammedConditionFalse(e, consts.KonnectEntitiesFailedToCreateReason, err.Error())
 		}
+
+	case errors.As(err, &errSDK):
+		switch {
+		case ErrorIsSDKErrorTypeField(err):
+			SetKonnectEntityProgrammedConditionFalse(e, consts.KonnectEntitiesFailedToCreateReason, errSDK.Error())
+		default:
+			SetKonnectEntityProgrammedConditionFalse(e, consts.KonnectEntitiesFailedToCreateReason, err.Error())
+		}
```
czeslavo: Why do we need this as a separate case while it's doing the same as the `err != nil` one?

controller/konnect/ops/ops_errors.go
```diff
@@ -189,34 +229,32 @@ func handleDeleteError[
...
+
+// IgnoreUnrecoverableAPIErr ignores unrecoverable errors that would cause the
+// reconciler to endlessly requeue.
+func IgnoreUnrecoverableAPIErr(err error) error {
+	// If the error is a type field error or bad request error, then don't propagate
+	// it to the caller.
+	// We cannot recover from this error as this requires user to change object's
+	// manifest. The entity's status is already updated with the error.
+	if ErrorIsSDKErrorTypeField(err) || ErrorIsSDKBadRequestError(err) {
+		return nil
```
czeslavo, line 259: How about adding a debug log here?

controller/konnect/ops/ops_test.go
```diff
@@ -13,10 +18,239 @@ import (
...
+	TEnt constraints.EntityType[T],
+] struct {
+	name                  string
+	entity                TEnt
+	sdkFunc               func(t *testing.T, sdk *sdkmocks.MockSDKWrapper) *sdkmocks.MockSDKWrapper
+	expectedErrorContains string
+	assertions            func(t *testing.T, ent TEnt)
+}
+
+func TestCreate(t *testing.T) {
```
czeslavo, line 37: 👍

controller/konnect/ops/ops_errors.go
pmalek: 👍 a80100866c5e19ffe540136f34f7f4ee9077521a

controller/konnect/ops/ops.go
pmalek: That makes sense. I thought initially that we could use this branch here for handling different error types but let's do that when it comes to that. a80100866c5e19ffe540136f34f7f4ee9077521a removes the nested `switch`.

PR #801, opened 2024-10-25
chore: parallel k8sio go.mod replace update script

czeslavo:

It makes the k8sio modules replace directives script concurrent to save some time. 

With N=10 this gave me the following results (total time):

|                 | **Cold cache** | **Warm cache**  | 
|-----------------|------------|----------------|
| Serial          | 11.982s     | 3.596s         |
| Parallel        | 4.316s      | 1.365s         |
| _**Diff**_            | 7.66s (~63%)| 2.23s (~62%)  |

### DIFF
--- a/hack/update-k8sio-gomod-replace.sh
+++ b/hack/update-k8sio-gomod-replace.sh
@@ -11,26 +11,34 @@ MODS=($(
     sed -n 's|.*k8s.io/\(.*\) => ./staging/src/k8s.io/.*|k8s.io/\1|p'
 ))
 
-# Export variables for access in subshells run by xargs
+# Set concurrency level to the number of available CPU cores
+CONCURRENCY=$(nproc)
 export VERSION
 
-# Function to update go.mod replace directive for a module
-update_mod() {
+# Create an empty array to store replace directives
+declare -a REPLACE_COMMANDS
+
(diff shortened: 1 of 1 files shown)


hack/update-k8sio-gomod-replace.sh
```diff
@@ -10,12 +10,27 @@ MODS=($(
...
+
+# Function to update go.mod replace directive for a module
+update_mod() {
+    local MOD=$1
+    local V=$(
         go mod download -json "${MOD}@kubernetes-${VERSION}" |
         sed -n 's|.*"Version": "\(.*\)".*|\1|p'
     )
     echo "Updating go.mod replace directive for ${MOD}"
     go mod edit "-replace=${MOD}=${MOD}@${V}"
```
pmalek: Wouldn't calling `go mod edit` in parallel cause unwanted artifacts in `go.mod` and/or `go.sum` 🤔 ? Cursory google search doesn't yield any results on this.

czeslavo: Good point. The `go mod replace` commands are quick enough so we can safely parallelize the `go mod download`s only, still getting similar gains: 5d1c151

pmalek (approved):
🪄

PR #802, opened 2024-10-25
fix: Filter managed KongPluginBinding in finalizer reconciler

### LINKED ISSUE #786: When a target of an unmanaged `KongPluginBinding` is deleted, the `KongPluginBinding` is deleted without logs mentioning that
## Current Behavior

When I create a `KongPluginBinding` and delete one of the entities in its `targets`, for example:
```
kind: KongPluginBinding
metadata:
  name: plugin-binding-kongservice
spec:
  controlPlaneRef:
    type: konnectNamespacedRef
    konnectNamespacedRef:
      name: test1
  pluginRef:
    name: rate-limit-5-min
  targets:
    serviceRef:
      name: service-1
      kind: KongService
    consumerRef:
      name: consumer-1
```
When I delete the `KongConsumer`, the consumer gets deleted in Konnect as well as the plugin. After a `Reconciliation Error` telling that failed to update the plugin in Konnect because the plugin's foreign entity is not found, the `KongPluginBinding` is also deleted without a log to mention this.

## Expected Behavior

- Updating the status of `KongPluginBinding` to show that one of referenced targets is not found. Or
- Deleting the `KongPluginBinding` and add logs or k8s events to notice that it was deleted

randmonkey:
Only add finalizers to target objects and remove `KongPluginBinding` on removal of targets for managed `KongPluginBinding`s.

**Which issue this PR fixes**

Fixes #786 

**PR Readiness Checklist**:

### DIFF
--- a/controller/konnect/reconciler_generic_pluginbindingfinalizer.go
+++ b/controller/konnect/reconciler_generic_pluginbindingfinalizer.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/samber/lo"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
@@ -164,10 +166,14 @@ func (r *KonnectEntityPluginBindingFinalizerReconciler[T, TEnt]) Reconcile(
 		return ctrl.Result{}, err
 	}
 
+	manangedPluginBindings := lo.Filter(kongPluginBindingList.Items, func(kpb configurationv1alpha1.KongPluginBinding, _ int) bool {
+		return isManagedKongPluginBinding(&kpb)
+	})
+
 	var finalizersChangedAction string
 	// if the entity is marked for deletion, we need to delete all the PluginBindings that reference it.
 	if !ent.GetDeletionTimestamp().IsZero() {
-		for _, kpb := range kongPluginBindingList.Items {
+		for _, kpb := range manangedPluginBindings {
 			if err := cl.Delete(ctx, &kpb); err != nil {
 				if k8serrors.IsNotFound(err) {
 					continue
@@ -183,7 +189,7 @@ func (r *KonnectEntityPluginBindingFinalizerReconciler[T, TEnt]) Reconcile(
 		}
 	} else {
 		// if the entity is not marked for deletion, update the finalizers accordingly.
-		switch len(kongPluginBindingList.Items) {
+		switch len(manangedPluginBindings) {
 		case 0:
 			// in case no KongPluginBindings are referencing the entity, but it has the finalizer,
 			// we need to remove the finalizer.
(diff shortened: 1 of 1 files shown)


controller/konnect/reconciler_generic_pluginbindingfinalizer.go
```diff
@@ -237,6 +243,14 @@ func (r *KonnectEntityPluginBindingFinalizerReconciler[T, TEnt]) getKongPluginBi
 	}
 }
 
+// isManagedKongPluginBinding returns true if the KongPluginBinding is managed by the controller.
+// REVIEW: Here we judge whether KongPluginBinding is managed by checking its owner references and treat it as managed if it has a plugin as owner reference.
```
czeslavo, line 247: That makes sense to me.
```suggestion
// We judge whether KongPluginBinding is managed by checking its owner references and treat it as managed if it has a plugin as owner reference.
```

czeslavo (commented):
Is it possible to cover this in an envtest?

randmonkey:
> Is it possible to cover this in an envtest?

Current envtest do the reverse. They create unmanaged plugin bindings and check for finalizers.

pmalek (changes requested):
I believe we should filter out the unmanaged `KongPluginBinding`s earlier so that they are not even reconciled by this controller.

I'd suggest something like:

```diff
diff --git a/controller/konnect/reconciler_generic_pluginbindingfinalizer.go b/controller/konnect/reconciler_generic_pluginbindingfinalizer.go
index ac1d7526..8cbe4879 100644
--- a/controller/konnect/reconciler_generic_pluginbindingfinalizer.go
+++ b/controller/konnect/reconciler_generic_pluginbindingfinalizer.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/samber/lo"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
@@ -64,6 +66,14 @@ func (r *KonnectEntityPluginBindingFinalizerReconciler[T, TEnt]) SetupWithManage
 	return b.Complete(r)
 }
 
+func ownerRefIsAnyKongPlugin() func(ownerRef metav1.OwnerReference) bool {
+	return func(ownerRef metav1.OwnerReference) bool {
+		return ownerRef.Kind == "KongPlugin" ||
+			// NOTE: We currently do not support KongClusterPlugin, but we keep this here for future use.
+			ownerRef.Kind == "KongClusterPlugin"
+	}
+}
+
 // enqueueObjectReferencedByKongPluginBinding watches for KongPluginBinding objects
 // that reference the given Konnect entity.
 // It returns the reconcile.Request slice containing the entity that the KongPluginBinding references.
@@ -74,6 +84,11 @@ func (r *KonnectEntityPluginBindingFinalizerReconciler[T, TEnt]) enqueueObjectRe
 			return nil
 		}
 
+		// If the KongPluginBinding is unmanaged (created not using an annotation), skip it, do not delete it.
+		if !lo.ContainsBy(kpb.OwnerReferences, ownerRefIsAnyKongPlugin()) {
+			return nil
+		}
+
 		var (
 			name string
 			e    T
@@ -125,6 +140,9 @@ func (r *KonnectEntityPluginBindingFinalizerReconciler[T, TEnt]) enqueueObjectRe
 }
 
 // Reconcile reconciles the Konnect entity that can be set as KongPluginBinding target.
+// It reconciles only entities that are referenced by managed KongPluginBindings,
+// i.e. those that are created by the controller out of konghq.com/plugins annotations.
+//
 // Its purpose is to:
 //   - check if the entity is marked for deletion and mark KongPluginBindings
 //     that reference it.
```

WDYT?

pmalek:
> > Is it possible to cover this in an envtest?
> 
> Current envtest do the reverse. They create unmanaged plugin bindings and check for finalizers.

Tao's right. https://github.com/Kong/gateway-operator/blob/e3cd706111df252197188ae89157fd811bab1364/test/envtest/kongplugincleanupfinalizer_test.go should use the `konghq.com/plugins` annotations on these entities and check the finalizers. At the same time we should check that unmanaged KPB's do not cause those entities to have the finalizer: these entities are fully managed by the user and we should not interfere in the lifecycle of those.

WDYT?

pmalek:
Solved in #805

Issue #645, opened 2024-09-25
KongConsumerGroup status changes should trigger depending KongConsumer reconciliation


czeslavo:
## Problem Statement

When the status of `KongConsumerGroup` that a given `KongConsumer` depends on changes (e.g. its `Programmed` condition changes to `True`), it should trigger `KongConsumer`'s reconciliation despite the fact the `KongConsumer`'s `generation` and `Programmed` condition's `observedGeneration` are equal.

## Proposed Solution

Keep the last observed generations of the `KongConsumerGroup`s the `KongConsumer` references in the `KongConsumer`'s `status`. Compare those with the current generation of every `KongConsumerGroup` referenced by a `KongConsumer` in the [shouldUpdate](https://github.com/Kong/gateway-operator/blob/23b649b157b72a8e494ba378b52001396b811cfc/controller/konnect/ops/ops.go#L150) function and decide based on that (observed generation != current generation => should update).

## Additional Information

The issue was worked around in https://github.com/Kong/gateway-operator/pull/638 by returning an error from the `KongConsumer` reconciler when any of the referenced `KongConsumerGroup`s is not `Programmed` yet. It should be changed once we fix this issue.

## Acceptance Criteria
- [ ] `KongConsumer` reconciliation is triggered on every `KongConsumerGroup` change (update is not skipped)
- [ ] `KongConsumer` can become `Programmed` even if the `KongConsumerGroup` is not programmed yet

pmalek:
The same issue affects `KonnectGatewayControlPlane` and their groups: when e.g. one of the CP group members gets deleted or get its `Programmed` status condition changed from `False` to `True` we should reconcile the group as well and reflect its state in its status.

pmalek:
This also affects removing targets of `KongPluginBinding`, re #805

lahabana:
No user has complained about this yet so punting

sandromodarelli:
closing due to low priority

PR #805, opened 2024-10-25
fix(konnect): do not manage finalizers and lifecycle for unmanaged KongPluginBindings

### LINKED ISSUE #786: When a target of an unmanaged `KongPluginBinding` is deleted, the `KongPluginBinding` is deleted without logs mentioning that

pmalek:

Do not manage finalizers and lifecycle for unmanaged `KongPluginBinding`s.

**Which issue this PR fixes**

Fixes #

The approach proposed in this PR will suffer from #645: having KPB with `KongService` and `KongRoute` targets and deleting `KongRoute` we can observe:

- `KongRoute` being deleted on Konnect
- Plugin being deleted on Konnect
- `KongPluginBinding` doesn't get its status updated until next sync period tick (this is caused by #645)

**PR Readiness Checklist**:

### DIFF
--- a/controller/konnect/plugins.go
+++ b/controller/konnect/plugins.go
@@ -1,6 +1,8 @@
 package konnect
 
 import (
+	"github.com/samber/lo"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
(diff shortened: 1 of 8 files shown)


mlavacca (approved):
It makes sense to me 👍

PR #810, opened 2024-10-28
Fix(konnect): reject reconciliation for incompatible cluster type of control plane

### LINKED ISSUE #791: KIC type control plane can be created by `KonnectGatewayControlPlane` and entities cannot attached to the control plane
## Current Behavior
We can set `cluster_type` of `CLUSTER_TYPE_K8S_INGRESS_CONTROLLER`, a KIC controlled cluster is created in Konnect. Then if we attach entities like `KongService` to the control plane, Konnect will return 403.

## Expected Behavior

Attaching other entities to the KIC controlled cp should be forbidden on validation or rejected to send to Konnect.

randmonkey:

Pop a `ControlPlaneRefValid=false` condition to entities attached to control plane with incompatible type to prevent repeated failure of creation.  Now the entities will have the following condition if they are attached to an incompatible control plane:

```
  - lastTransitionTime: "2024-10-28T03:18:11Z"
    message: Attaching to ControlPlane default/group1 with cluster type CLUSTER_TYPE_CONTROL_PLANE_GROUP
      is not supported
    observedGeneration: 1
    reason: Invalid
    status: "False"
    type: ControlPlaneRefValid
```

Also move `handleControlPlaneRef` to a separate file because the `reconciler_generic` is too looong (>1000 lines).

**Which issue this PR fixes**

Fixes #791. Also fixes entities attaching to control plane group would cause repeated failure on creation.

**PR Readiness Checklist**:


controller/konnect/reconciler_controlplaneref_test.go
```diff
@@ -0,0 +1,289 @@
...
+](
+	t *testing.T, testCases []handleControlPlaneRefTestCase[T, TEnt],
+) {
+	t.Helper()
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			scheme := runtime.NewScheme()
+			require.NoError(t, configurationv1alpha1.AddToScheme(scheme))
+			require.NoError(t, konnectv1alpha1.AddToScheme(scheme))
```
pmalek: Might resort to using `scheme.Get()` https://github.com/Kong/gateway-operator/blob/d5f3d3672f8eba3b82185c39a5b2c993bb6ca191/modules/manager/scheme/scheme.go#L19-L37.

Doing it this way won't require updating this test when we move types between API versions.

controller/konnect/reconciler_controlplaneref.go
```diff
@@ -0,0 +1,239 @@
...
+		if err := cl.Patch(ctx, ent, client.MergeFrom(old)); err != nil {
+			if k8serrors.IsConflict(err) {
+				return ctrl.Result{Requeue: true}, nil
+			}
+			return ctrl.Result{}, fmt.Errorf("failed to update status: %w", err)
+		}
+
+		// TODO(pmalek): make this generic.
+		// CP ID is not stored in KonnectEntityStatus because not all entities
+		// have a ControlPlaneRef, hence the type constraints in the reconciler can't be used.
```
pmalek:
```suggestion
```

controller/konnect/reconciler_controlplaneref.go
```diff
@@ -0,0 +1,239 @@
...
+				ctx, cl, ent,
+				konnectv1alpha1.ControlPlaneRefValidConditionType,
+				metav1.ConditionFalse,
+				konnectv1alpha1.ControlPlaneRefReasonInvalid,
+				fmt.Sprintf("Attaching to ControlPlane %s with cluster type %s is not supported", nn, *cp.Spec.ClusterType),
+			); errStatus != nil || !res.IsZero() {
+				return res, errStatus
+			}
+			return ctrl.Result{}, nil
+		}
```
pmalek: ~I believe that instead of doing it here we should block the reconciliation at the watch functions level. So ideally we'd add these predicates in https://github.com/Kong/gateway-operator/blob/7d67f04feaafc31146b02b87a3fcd363014319cc/controller/konnect/watch.go#L104-L117. That would require the `client.Client` to be provided there.~

⬆️ That won't allow us to add status condition.

👍

controller/konnect/reconciler_controlplaneref_test.go
```diff
@@ -0,0 +1,289 @@
...
+			updatedEntAssertions: []func(svc *configurationv1alpha1.KongService) (ok bool, message string){
+				func(svc *configurationv1alpha1.KongService) (bool, string) {
+					return lo.ContainsBy(svc.Status.Conditions, func(c metav1.Condition) bool {
+						return c.Type == konnectv1alpha1.ControlPlaneRefValidConditionType && c.Status == metav1.ConditionFalse
+					}), "service should have ControlPlaneRefValid set to False"
+				},
+			},
+		},
+		{
+			name: "control plane with incompatible cluster type",
```
pmalek:
```suggestion
			name: "control plane with incompatible cluster type (ControlPlane Group)",
```

controller/konnect/reconciler_controlplaneref.go
```diff
@@ -0,0 +1,239 @@
...
+				return ctrl.Result{}, ReferencedControlPlaneDoesNotExistError{
+					Reference: nn,
+					Err:       err,
+				}
+			}
+			return ctrl.Result{}, err
+		}
+
+		// Do not continue reconciling of the control plane has incompatible cluster type to prevent repeated failure of creation.
+		// Only CLUSTER_TYPE_CONTROL_PLANE and CLUSTER_TYPE_HYBRID are supported.
```
pmalek, line 143: We might mention here that attaching entities to CP groups is unsupported because they configuration is read only and perhaps link https://docs.konghq.com/konnect/gateway-manager/control-plane-groups/#limitations in the comment.

WDYT?

controller/konnect/reconciler_controlplaneref_test.go
randmonkey: done.

controller/konnect/reconciler_controlplaneref_test.go
randmonkey: done.

controller/konnect/reconciler_controlplaneref.go
randmonkey: Added comments.

pmalek (approved):

PR #816, opened 2024-10-28
refactor(konnect): add enqueueObjectForKonnectGatewayControlPlane to generically list objects that refer to KonnectGatewayControlPlane

pmalek:

Generically list objects that refer to `KonnectGatewayControlPlane`.

### DIFF
--- a/controller/konnect/watch.go
+++ b/controller/konnect/watch.go
@@ -9,6 +9,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"github.com/kong/gateway-operator/controller/konnect/constraints"
 	operatorerrors "github.com/kong/gateway-operator/internal/errors"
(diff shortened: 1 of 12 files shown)


controller/konnect/constraints/constraints.go
```diff
@@ -75,3 +75,14 @@ type SupportedKonnectEntityPluginReferenceableType interface {
...
+
+// ClientObject represents an object that has a name and a namespace
+// (might be empty for cluster-scoped objects).
+type ClientObject[T any] interface {
+	*T
+
+	GetName() string
+	GetNamespace() string
+
+	client.Object
```
czeslavo: Doesn't `client.Object` already provide `GetName` and `GetNamespace`?

czeslavo: BTW Wouldn't it be possible to omit this interface and use the existing `EntityType`?

pmalek: Ah yes 🤦 27e38911dad8aee4c5b3fbab88d93c5900742655

pmalek: > BTW Wouldn't it be possible to omit this interface and use the existing EntityType?

This way it's less restrictive and can be used with other types, while `EntityType` already has some assumptions about konnect related methods that this type has to have.

czeslavo: I agree it's less restrictive, but we only use it now for Konnect entities (namely, in `enqueueObjectForKonnectGatewayControlPlane`, which implies a Konnect entity use case). We might as well keep it simple, not introduce more types, and add it when we find it required.

pmalek: 👍 f382e23f8fd1547593f6371f3b25777e01746034

czeslavo (approved):

PR #824, opened 2024-10-29
Fix: Add MembersReferenceResolved condition for cp groups

### LINKED ISSUE #797: KonnectGatewayControlPlane have no status for CP group members references
## Current Behavior

When creating a CP group like so:

```
echo '
kind: KonnectGatewayControlPlane
metadata:
  name: control-plane-group
  namespace: default
spec:
  name: control-plane-group # Name used to identify the Gateway Control Plane in Konnect
  cluster_type: CLUSTER_TYPE_CONTROL_PLANE_GROUP # Type of the Gateway Control Plane
  members:
    - name: gateway-control-plane # Reference to the KonnectGatewayControlPlane object
  konnect:
    authRef:
      name: konnect-api-auth # Reference to the KonnectAPIAuthConfiguration object
  ' | kubectl apply -f -
```

there is no status populated to signal whether the reference to the group members was resolved successfully or not.

## Expected Behavior

A conditions is populated for every group member (could be a new `map[name]condition` status field).

randmonkey:
Add a `MembersReferenceResolved` condition in status of control plane groups.
**Which issue this PR fixes**

Fixes #797 

**PR Readiness Checklist**:

### DIFF
--- a/controller/konnect/ops/ops_controlplane.go
+++ b/controller/konnect/ops/ops_controlplane.go
@@ -135,6 +135,16 @@ func setGroupMembers(
 		return nil
 	}
 
+	// Set MembersReferenceResolved condition to False if there are no members in the CP group.
+	if len(cp.Spec.Members) == 0 {
+		SetControlPlaneGroupMembersReferenceResolvedConditionFalse(
+			cp,
+			ControlPlaneGroupMembersReferenceResolvedReasonNoMembers,
+			"No members in the control plane group",
+		)
+		return nil
+	}
+
 	members, err := iter.MapErr(cp.Spec.Members,
 		func(member *corev1.LocalObjectReference) (sdkkonnectcomp.Members, error) {
 			var (
@@ -165,7 +175,7 @@ func setGroupMembers(
 	if err != nil {
 		SetControlPlaneGroupMembersReferenceResolvedConditionFalse(
 			cp,
-			"SomeMemberNotResolved",
+			ControlPlaneGroupMembersReferenceResolvedReasonPartialNotResolved,
 			err.Error(),
 		)
 		return fmt.Errorf("failed to set group members, some members couldn't be found: %w", err)
@@ -179,7 +189,7 @@ func setGroupMembers(
 	if err != nil {
 		SetControlPlaneGroupMembersReferenceResolvedConditionFalse(
 			cp,
-			"SetGroupMemberFailed",
+			ControlPlaneGroupMembersReferenceResolvedReasonFailedToSet,
 			err.Error(),
 		)
 		return fmt.Errorf("failed to set members on control plane group %s: %w",
@@ -187,7 +197,9 @@ func setGroupMembers(
 		)
 	}
 
-	SetControlPlaneGroupMembersReferenceResolvedCondition(cp)
+	SetControlPlaneGroupMembersReferenceResolvedCondition(
+		cp,
+	)
 	return nil
 }
 
--- a/controller/konnect/ops/ops_controlplane.go
+++ b/controller/konnect/ops/ops_controlplane.go
@@ -128,9 +128,7 @@ func setGroupMembers(
 	id string,
 	sdkGroups sdkops.ControlPlaneGroupSDK,
 ) error {
-	// REVIEW: should we add MembersReferenceResolved condition for CP groups with 0 members?
-	if len(cp.Spec.Members) == 0 ||
-		cp.Spec.ClusterType == nil ||
+	if cp.Spec.ClusterType == nil ||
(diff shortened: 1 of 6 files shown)


controller/konnect/ops/ops_controlplane.go
```diff
@@ -145,17 +146,28 @@ func setGroupMembers(
...
+					}
 			}
 			return sdkkonnectcomp.Members{
 				ID: lo.ToPtr(memberCP.GetKonnectID()),
 			}, nil
 		})
 	if err != nil {
+		SetControlPlaneGroupMembersReferenceResolvedConditionFalse(
+			cp,
+			"SomeMemberNotResolved",
```
pmalek: Can we put this into a named `const` with a comment?

controller/konnect/ops/ops_controlplane.go
```diff
@@ -128,6 +128,7 @@ func setGroupMembers(
 	id string,
 	sdkGroups sdkops.ControlPlaneGroupSDK,
 ) error {
+	// REVIEW: should we add MembersReferenceResolved condition for CP groups with 0 members?
```
pmalek: I believe we can set this one with a message stating that no members set. Just to be coherent on the API level. WDYT?

pmalek (commented):
Can we please cover this with a test? Perhaps we can extend the test cases in https://github.com/Kong/gateway-operator/blob/5d0b4b78c138f606b1eb5116e2649f35cf780bdb/test/envtest/konnect_entities_gatewaycontrolplane_test.go

controller/konnect/ops/ops_controlplane.go
randmonkey: Done.

controller/konnect/ops/ops_controlplane.go
randmonkey: Added a `NoMembers` reason for such case.

randmonkey:
> Can we please cover this with a test? Perhaps we can extend the test cases in https://github.com/Kong/gateway-operator/blob/5d0b4b78c138f606b1eb5116e2649f35cf780bdb/test/envtest/konnect_entities_gatewaycontrolplane_test.go

Added unit tests.

pmalek (approved):

pmalek (changes requested):
I'm doing some testing on this one and I do not see the condition set consistently. Let's put this on pause until that's resolved.

pmalek (changes requested):
This PR is missing setting the condition when we encounter a conflict against the API. Specifically when we hit https://github.com/Kong/gateway-operator/blob/170469ac707c458f88c97f4c271d71078fa3f353/controller/konnect/ops/ops.go#L118 we have to handle the members there as well (or requeue for update which will handle that, but that's only going to happen when the sync period passes).

randmonkey:
> This PR is missing setting the condition when we encounter a conflict against the API. Specifically when we hit
> 
> https://github.com/Kong/gateway-operator/blob/170469ac707c458f88c97f4c271d71078fa3f353/controller/konnect/ops/ops.go#L118
> 
> we have to handle the members there as well (or requeue for update which will handle that, but that's only going to happen when the sync period passes).

Added `setGroupMembers` in `getControlPlaneForUID`.

controller/konnect/ops/ops_controlplane.go
```diff
@@ -202,5 +232,20 @@ func getControlPlaneForUID(
...
+		// If we failed to set group membership, we should return a specific error with a reason
+		// so the downstream can handle it properly.
+		return id, KonnectEntityCreatedButRelationsFailedError{
+			KonnectID: id,
+			Err:       err,
+			Reason:    konnectv1alpha1.KonnectGatewayControlPlaneProgrammedReasonFailedToSetControlPlaneGroupMembers,
+		}
+	}
+
+	return id, nil
```
pmalek, line 250: Do you mind adding a test for this as well? Ideally extending the test suite in https://github.com/Kong/gateway-operator/blob/8c7fbd40e945bb95c5247288131eac3df9cc7113/test/envtest/konnect_entities_gatewaycontrolplane_test.go.

There's already 1 testcase for conflict https://github.com/Kong/gateway-operator/blob/8c7fbd40e945bb95c5247288131eac3df9cc7113/test/envtest/konnect_entities_gatewaycontrolplane_test.go#L375-L450. 1 more for CP group with members would be ideal.

pmalek (commented):
Please resolve conflicts that arose since last review.

Overall this looks good. Just 1 nit about adding a test case for this.

controller/konnect/ops/ops_controlplane.go
randmonkey: Done.

pmalek (approved):

PR #838, opened 2024-10-31
tests: bump version in e2e upgrade tests

czeslavo:

Bumps the KGO version in the upgrade tests to the latest v1.4.0.

### DIFF
--- a/test/e2e/test_helm_install_upgrade.go
+++ b/test/e2e/test_helm_install_upgrade.go
@@ -62,111 +62,42 @@ func TestHelmUpgrade(t *testing.T) {
 		assertionsAfterInstall []assertion
 		assertionsAfterUpgrade []assertion
 	}{
-		// NOTE: We do not support versions earlier than 1.2 with the helm chart.
-		// The initial version of the chart contained CRDs from KGO 1.2. which
-		// introduced a breaking change which makes it impossible to upgrade from
-		// automatically (without manually deleting the CRDs).
 		{
-			name:        "upgrade from 1.2.0 to 1.2.3",
-			fromVersion: "1.2.0",
-			toVersion:   "1.2.3",
-			objectsToDeploy: []client.Object{
-				&operatorv1beta1.GatewayConfiguration{
-					ObjectMeta: metav1.ObjectMeta{
-						Name: "gwconf-upgrade-120-123",
-					},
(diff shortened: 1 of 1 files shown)


pmalek (commented):
Can we please add a test case for 1.3.0 to 1.4.0 as well?

Additionally while at it I believe we can remove the case for 1.2.0 to 1.2.3 migration at this point (to cut down the CI run time).

czeslavo:
> Can we please add a test case for 1.3.0 to 1.4.0 as well?
> 
> Additionally while at it I believe we can remove the case for 1.2.0 to 1.2.3 migration at this point (to cut down the CI run time).

I adjusted the test cases' names so they are generic enough to be valid at any point in time. I also reopened https://github.com/Kong/gateway-operator/issues/121 to handle automatic version updates and will handle that in a next PR.

test/e2e/test_helm_install_upgrade.go
```diff
@@ -62,85 +62,16 @@ func TestHelmUpgrade(t *testing.T) {
 		assertionsAfterInstall []assertion
 		assertionsAfterUpgrade []assertion
 	}{
-		// NOTE: We do not support versions earlier than 1.2 with the helm chart.
-		// The initial version of the chart contained CRDs from KGO 1.2. which
-		// introduced a breaking change which makes it impossible to upgrade from
-		// automatically (without manually deleting the CRDs).
```
pmalek, line 68: I guess this could have been left as is but meh. We can also remove it.

pmalek (commented):
I think we're OK making the step towards making the testcases a bit more generic for the sake of small readability decrease.

BTW: tests seem to have failed 🤔

test/e2e/test_helm_install_upgrade.go
czeslavo: I think that we could document this in the chart's `UPGRADE.md` as that's where users would seek such information, WDYT?

pmalek: Added this in https://github.com/Kong/charts/pull/1158. PTAL

czeslavo:
@pmalek I've fixed the failures - it turned out to be a wrong assertion requiring the DP image to be patched.

pmalek (approved):

PR #847, opened 2024-11-05
fix: diff DataPlane log

pmalek:

3rd parameter of log functions (including `Trace`) is `rawObj`, not `string`. This has been reported by user in https://github.com/Kong/gateway-operator/issues/522#issuecomment-2309964408.

Current code yields:

```
unexpected type processed for trace logging: string, this is a bug!
```

This PR fixes it.

### DIFF
--- a/controller/dataplane/owned_deployment.go
+++ b/controller/dataplane/owned_deployment.go
@@ -342,7 +342,7 @@ func reconcileDataPlaneDeployment(
 		}
 		if updated {
 			diff := cmp.Diff(original.Spec.Template, desired.Spec.Template, opts...)
-			log.Trace(logger, "Deployment diff detected", diff)
+			log.Trace(logger, "DataPlane Deployment diff detected", dataplane, "diff", diff)
 		}
 
 		return patch.ApplyPatchIfNotEmpty(ctx, cl, logger, existing, original, updated)


randmonkey (approved):

czeslavo:
Can we make the log functions type-safe to ensure the `rawObj` is `client.Object` to avoid such situations? BTW looking into the [source code](https://github.com/Kong/gateway-operator/blob/62f05fc05116b602d8c36d69e9e720a326593e89/controller/pkg/log/log.go#L32), `rawObj` is unused. Are these logging helpers all right?

pmalek:
> Can we make the log functions type-safe to ensure the `rawObj` is `client.Object` to avoid such situations?

We can't do this today (without refactoring) as not all objects passed as `rawObj` are `client.Object`s (some of those only implement this interface for the pointer type and are passed by value).

> BTW looking into the [source code](https://github.com/Kong/gateway-operator/blob/62f05fc05116b602d8c36d69e9e720a326593e89/controller/pkg/log/log.go#L32), `rawObj` is unused. Are these logging helpers all right?

This has been changed in #727. Currently the `rawObj` is indeed unused as that information was being duplicated. Let me try to propose something that will fix this issue.

pmalek:
@czeslavo PTAL at #848

PR #848, opened 2024-11-05
refactor: drop rawObj parameter from logger's methods

pmalek:

This PR removes the `rawObj` parameter - which was unused anyway since #727 has been merged -  from logger's methods.

1 important change here is that now, `Info()`, `Debug()`, `Trace()` and `Error()` accept only the (optional error for `Error()`) logger instance, message (`string`) and key value pairs as variadic parameter. The last part makes this error prone and susceptible for providing an invalid argument lists (i.e. odd number of arguments).

We could potentially alleviate this with making these functions accept a strongly typed:

```
type KV struct {
	K string
	V any
}
```

but then we'd need to allocate on each log call like so:

```
func Info(logger logr.Logger, msg string, kvs ... KV) {
	keyValues := make([]interface{}, 0, len(kvs)*2)
	for _, kv := range kvs {
		keyValues = append(keyValues, kv.K, kv.V)
	}
	_log(logger, logging.InfoLevel, msg, keyValues...)
}
```

Feel free to drop your comments on this performance vs type safety trade off.

**Which issue this PR fixes**

Fixes #

[Code handling the development mode logger setting](https://github.com/Kong/gateway-operator/blob/37f6d5eef6d8c2d3cbd249f67d3cf4bc054ca99f/controller/pkg/log/log.go#L54-L61) makes the dev mode logger not have several fields which might be beneficial for debugging/observability.

This makes the following difference between development mode enabled and disabled on logs:

- development mode disabled

```json
{
	"level": "debug",
	"ts": "2024-11-05T12:24:42.434+0100",
	"logger": "dataplane",
	"msg": "reconciliation complete for DataPlane resource",
	"controller": "dataplane",
	"controllerGroup": "gateway-operator.konghq.com",
	"controllerKind": "DataPlane",
	"DataPlane": {
		"name": "dataplane-example",
		"namespace": "default"
	},
	"namespace": "default",
	"name": "dataplane-example",
	"reconcileID": "2641ef78-cebe-4f69-a907-5e2cd84eb846"
}
```

- development mode enabled

```json
{
	"level": "\u001b[35mDEBUG\u001b[0m",
	"ts": "2024-11-05T12:31:13.633+0100",
	"logger": "DATAPLANE",
	"msg": "reconciliation complete for DataPlane resource"
}
```

we might want to remove that code to make those uniform.

**PR Readiness Checklist**:

### DIFF
--- a/controller/controlplane/controller.go
+++ b/controller/controlplane/controller.go
@@ -115,7 +115,7 @@ func (r *Reconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager) err
 func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	logger := log.GetLogger(ctx, "controlplane", r.DevelopmentMode)
 
-	log.Trace(logger, "reconciling ControlPlane resource", req)
+	log.Trace(logger, "reconciling ControlPlane resource")
 	cp := new(operatorv1beta1.ControlPlane)
 	if err := r.Client.Get(ctx, req.NamespacedName, cp); err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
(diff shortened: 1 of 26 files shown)


czeslavo (approved):
Yeah, variadic key value pairs are hard to tackle without performance penalty... :/ We've lived with them up until now, so I guess we can leave it as is. I was looking for a linter that could potentially detect this, but I haven't found anything suitable.

PR #872, opened 2024-11-20
fix: set DataPlane's extensions through GatewayConfiguration

pmalek:

#453 missed configuring extensions for `Gateway`'s `DataPlane` so this is adding the missing bit.

**Which issue this PR fixes**

Fixes #

This requires tests to be added.

**PR Readiness Checklist**:

### DIFF
--- a/controller/gateway/controller_reconciler_utils.go
+++ b/controller/gateway/controller_reconciler_utils.go
@@ -137,6 +137,7 @@ func gatewayConfigDataPlaneOptionsToDataPlaneOptions(
 
 	dataPlaneOptions := &operatorv1beta1.DataPlaneOptions{
 		Deployment:       opts.Deployment,
+		Extensions:       opts.Extensions,
 		PluginsToInstall: pluginsToInstall,
 	}
 


mlavacca:
I have been exploring this issue when implementing #453, but it's not as straightforward as it might look. If we provision the DataPlane with Konnect configuration and a regular ControlPlane, we end up with KIC not being able to push configuration to the DataPlane (as it is not configured to accept the config from KIC), and it never gets ready. We likely need to improve the Controlplane as well to be able to receive KonnectExtension, to configure it to work in Konnect.

I've just created https://github.com/Kong/gateway-operator/issues/874 to track this effort

pmalek:
> I have been exploring this issue when implementing #453, but it's not as straightforward as it might look. If we provision the DataPlane with Konnect configuration and a regular ControlPlane, we end up with KIC not being able to push configuration to the DataPlane (as it is not configured to accept the config from KIC), and it never gets ready. We likely need to improve the Controlplane as well to be able to receive KonnectExtension, to configure it to work in Konnect.
> 
> I've just created #874 to track this effort

Ah indeed. I missed this because I only glanced at the fact that this field is not set for `DataPlane`s when deploying with `Gateway`.

I was also thinking about making `ControlPlane` able to get the Konnect extension. Thanks for creating the issue for this one 🙇

PR #881, opened 2024-11-22
chore: enhance KongPluginInstallation example and document tests

### LINKED ISSUE #465: Streamline and automatize/document testing strategy for `KongPluginInstallation` CRDs

programmer04:

Tests for `KongPluginInstallation` rely on container registries configured in k8s-team's GCP project and images built according to specifications. This PR leaves some description for future us to learn how those were built and what they contain. Moreover, it simplifies and enhances examplar to make it more informative. 

**Which issue this PR fixes**

Closes https://github.com/Kong/gateway-operator/issues/465

Building some kind of machinery/scripts for auto bootstrap, etc. seems like an overkill because they won't change often (also, we've already used remote registers in our tests, e.g. KIC image, http-echo, etc.). Furthermore, building and uploading a container image according to the docs is not the hardest / most time-consuming job in the world - [see the cost/saving ratio](https://imgs.xkcd.com/comics/is_it_worth_the_time_2x.png)

> Reviewer [please check](https://console.cloud.google.com/artifacts?pli=1&invt=AbiK-A&inv=1&project=k8s-team-playground) if you have access. As a member of k8s-team you should.

### DIFF
--- a/config/samples/gateway-httproute.yaml
+++ b/config/samples/gateway-httproute.yaml
@@ -115,8 +115,9 @@ spec:
                 fieldRef:
                   fieldPath: status.podIP
           resources:
-            requests:
-              cpu: 10m
+            limits:
+              cpu: 100m
+              memory: 100Mi
 ---
 kind: GatewayConfiguration
 apiVersion: gateway-operator.konghq.com/v1beta1
--- a/config/samples/gateway-kongplugininstallation-httproute.yaml
+++ b/config/samples/gateway-kongplugininstallation-httproute.yaml
@@ -19,8 +19,6 @@ spec:
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  labels:
-    app: echo
   name: echo
 spec:
   replicas: 1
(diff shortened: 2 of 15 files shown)


czeslavo:
> Building some kind of machinery/scripts for auto bootstrap, etc. seems like an overkill because they won't change often

I partially agree, but IMO it wouldn't be a huge cost to have source Dockerfiles for every image used in these tests stored in this repo just in case we delete those images by mistake or something. I don't expect an end-to-end automation that syncs the images with the remote registry but to have a living reference to what these images are underneath.

programmer04:
I thought about it @czeslavo. There is a manual in the official docs that are [mentioned in the comment](https://docs.konghq.com/gateway-operator/latest/guides/plugin-distribution/#create-a-custom-plugin) as baseline + each test case name and assertion shows what should be inside (but yep it may change). For each image we have to have a `Dockerfile` + plugin files (with dir structure) to copy inside this. So, for instance, to test exceeding size putting a 1MiB garbage file in the repository similar with two files of 512KiB, etc. when the whole repo currently is 8.6M. This is why I didn't put them. But ofc I may put some placeholder files for them. WDYT?

czeslavo:
Yeah, I imagine there could be a directory dedicated to these images, e.g., under `hack/plugin-images` and a separate dir for every image. As for the huge files, we can generate these in the Dockerfile itself randomly and not store them in the repo. I believe `RUN dd if=/dev/urandom of=random_file bs=1M count=2` should add a random 2MB file to the image.

czeslavo (approved):
Thank you for adding Dockerfiles 🎖️

PR #966, opened 2025-01-07
Feat: Add Name in ServiceOptions and allow specifying ingress service name of dataplane

### LINKED ISSUE #233: Allow user to define proxy service name
## Problem Statement

KGO generates proxy service name as `dataplane-ingress-<Gateway_Name>-s2vb6-lhz28` which is difficult to predict the real service name to use.

## Proposed Solution

Add support for `nameOverride` or something similar to allow user define their proxy service name so they have a predictable name to send requests to. It can be as simple as  `dataplane-ingress-<Gateway_Name>_<user_defined_name>`

## Additional Information

This feature is also valuable for Kong mesh users to select their gateway dataplanes.

randmonkey:
Add `Name` field in `ServiceOptions` to allow specifying ingress service name of dataplane
**Which issue this PR fixes**
Fixes #233 

**PR Readiness Checklist**:


api/v1beta1/dataplane_types.go
```diff
@@ -260,6 +260,10 @@ type ServiceOptions struct {
 	// +kubebuilder:validation:Enum=LoadBalancer;ClusterIP
 	Type corev1.ServiceType `json:"type,omitempty" protobuf:"bytes,4,opt,name=type,casttype=ServiceType"`
 
+	// Name defines the name of the ingress service.
```
pmalek: `ServiceOptions` should be agnostic as to which service they describe. I can now see that line 239 should also be fixed to not mention `ingress`.

```suggestion
	// Name defines the name of the service.
```

api/v1beta1/dataplane_types.go
```diff
@@ -260,6 +260,10 @@ type ServiceOptions struct {
...
+	// If Name is empty, the controller will generate a service name from the owning object.
+	Name *string `json:"name,omitempty"` // REVIEW: should we constrain the name to be immutable?
```
pmalek: I believe it should be doable to implement this so that the name is mutable.
In order to achieve this we would use the label selector to select the service and then enforce the name as per this field value. WDYT?

pmalek (commented):
Since that should be relatively easy to implement I'd suggest to include the implementation in the PR and not split this.

WDYT?

api/v1beta1/dataplane_types.go
randmonkey: Done.

api/v1beta1/dataplane_types.go
randmonkey: done.

test/integration/test_dataplane.go
```diff
@@ -1033,3 +1033,111 @@ func TestDataPlaneServiceExternalTrafficPolicy(t *testing.T) {
...
+		}),
+		time.Minute, time.Second,
+	)
+
+	t.Log("verifying that ingress service with the new name is created")
+	require.Eventually(t, testutils.DataPlaneHasActiveServiceWithName(t, GetCtx(), dataplaneName, nil, clients, client.MatchingLabels{
+		consts.GatewayOperatorManagedByLabel: consts.DataPlaneManagedLabelValue,
+		consts.DataPlaneServiceTypeLabel:     string(consts.DataPlaneIngressServiceLabelValue),
+	}, serviceName,
+	), waitTime, tickTime)
```
pmalek: Could we also check that the old service has been deleted?

test/integration/test_dataplane.go
```diff
@@ -1033,3 +1033,111 @@ func TestDataPlaneServiceExternalTrafficPolicy(t *testing.T) {
...
+		if len(dataplaneService.Status.LoadBalancer.Ingress) > 0 {
+			dataplaneIP = dataplaneService.Status.LoadBalancer.Ingress[0].IP
+			return true
+		}
+		return false
+	}, waitTime, tickTime)
+
+	require.Eventually(t, Expect404WithNoRouteFunc(t, GetCtx(), "http://"+dataplaneIP), waitTime, tickTime)
+
+	t.Logf("updating ingress service name in dataplane")
```
pmalek: Can we log the old and new service names for easier tests debugging when tests fail?

pkg/utils/kubernetes/reduce/reduce.go
```diff
@@ -132,6 +133,28 @@ func ReduceServices(ctx context.Context, k8sClient client.Client, services []cor
 	return nil
 }
 
+// +kubebuilder:rbac:groups="discovery.k8s.io",resources=endpointslices,verbs=list;watch
```
pmalek: Why do we need the permissions for `EndpointSlices` here? I can't see them being used

test/integration/test_dataplane.go
randmonkey: Added the checks.

test/integration/test_dataplane.go
randmonkey: Added service names in the logs.

pkg/utils/kubernetes/reduce/reduce.go
randmonkey: Removed them.

pmalek (approved):

PR #981, opened 2025-01-09
chore: fix upgrade test to work with EE

czeslavo:

We should use `helpers.GetDefaultDataPlaneImageBase()` to make this work properly in KGO EE tests (it overrides the default image on its own: https://github.com/Kong/gateway-operator-enterprise/blob/27c99e34d6e6fc6bd3ee99be885be5e8a71ad0c4/test/e2e/run_e2e_test.go#L15).

Should fix https://github.com/Kong/gateway-operator-enterprise/actions/runs/12686258105/job/35358212920?pr=383.

### DIFF
--- a/test/e2e/test_helm_install_upgrade.go
+++ b/test/e2e/test_helm_install_upgrade.go
@@ -120,10 +120,10 @@ func TestHelmUpgrade(t *testing.T) {
 					},
 				},
 				{
-					Name: fmt.Sprintf("DataPlane deployment is patched after operator upgrade (due to change in default Kong image version to %q)", helpers.GetDefaultDataPlaneImagePreviousMinor()),
+					Name: "DataPlane deployment is patched after operator upgrade (due to change in default Kong image version to 3.8)",
 					Func: func(c *assert.CollectT, cl *testutils.K8sClients) {
 						gatewayDataPlaneDeploymentIsPatched("gw-upgrade-onebeforelatestminor-latestminor=true")(ctx, c, cl.MgrClient)
-						gatewayDataPlaneDeploymentHasImageSetTo("gw-upgrade-onebeforelatestminor-latestminor=true", helpers.GetDefaultDataPlaneImagePreviousMinor())(ctx, c, cl.MgrClient)
+						gatewayDataPlaneDeploymentHasImageSetTo("gw-upgrade-onebeforelatestminor-latestminor=true", helpers.GetDefaultDataPlaneBaseImage()+":3.8")(ctx, c, cl.MgrClient)
 					},
 				},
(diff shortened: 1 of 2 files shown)


programmer04 (dismissed):

test/e2e/test_helm_install_upgrade.go
```diff
@@ -120,10 +120,10 @@ func TestHelmUpgrade(t *testing.T) {
 					},
 				},
 				{
-					Name: "DataPlane deployment is patched after operator upgrade (due to change in default Kong image version to kong:3.8)",
+					Name: fmt.Sprintf("DataPlane deployment is patched after operator upgrade (due to change in default Kong image version to %q)", helpers.GetDefaultDataPlaneImagePreviousMinor()),
 					Func: func(c *assert.CollectT, cl *testutils.K8sClients) {
 						gatewayDataPlaneDeploymentIsPatched("gw-upgrade-onebeforelatestminor-latestminor=true")(ctx, c, cl.MgrClient)
-						gatewayDataPlaneDeploymentHasImageSetTo("gw-upgrade-onebeforelatestminor-latestminor=true", "kong:3.8")(ctx, c, cl.MgrClient)
+						gatewayDataPlaneDeploymentHasImageSetTo("gw-upgrade-onebeforelatestminor-latestminor=true", helpers.GetDefaultDataPlaneImagePreviousMinor())(ctx, c, cl.MgrClient)
```
pmalek: I do see the need to use `GetDefaultDataPlaneImage()` to make the test work in KGO EE test harness 👍 but I'm not completely on board with using `GetDefaultDataPlaneImagePreviousMinor()` and similar here.

The reason I'm raising this is that for instance, imaging next Kong version is released and [`DefaultDataPlaneTag`](https://github.com/Kong/gateway-operator/blob/dd397cf447c2d06b3ec44dbc5de116d5192ea543/pkg/consts/dataplane.go#L87-L94) is bumped.

At that point this test starts failing because 1.4.0's dp version didn't change but the test's expectation did.

My proposal here is to use `GetDefaultDataPlaneImage()` ( to make EE tests pass) but with a hardcoded tag in the expectation. 

WDYT?

czeslavo: I see your point. It makes sense to me to hardcode the GW tag for a fixed KGO version which is the case in these tests. How about this: https://github.com/Kong/gateway-operator/pull/981/commits/ed2637d31b9b407a2074b210bf012d4d38178011?

pmalek: Looks good to me 👍

pmalek (approved):

PR #983, opened 2025-01-09
fix: odd number of arguments passed to logger in Konnect ops

czeslavo:

Fixes an odd number of arguments passed to logger:

```text
2025-01-09T11:37:37.477+0100 - ERROR - KONGCONSUMER - reconciling ConsumerGroups for KongConsumer - {"konnect_id": "1b073b9e-ec16-420e-be45-6cccbd72ec13", "error": "log message has odd number of arguments"}
```

Also, enables `loggercheck` linter that should detect similar issues if we use a non-decorated `logr.Logger` (unfortunately, in most cases we use our `log.Debug/Info` wrapper function).

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -33,6 +33,7 @@ linters:
   - unparam
   - unused
   - wastedassign
+  - loggercheck
 linters-settings:
   gci:
     sections:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,11 +25,6 @@
 
 ## Unreleased
 
-### Fixes
-
-- Fix watch predicates for types shared between KGO and KIC.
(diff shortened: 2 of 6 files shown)


.golangci.yaml
```diff
@@ -33,6 +33,7 @@ linters:
   - unparam
   - unused
   - wastedassign
+  - loggercheck
```
pmalek, line 36: 👍

pmalek (approved):

.golangci.yaml
czeslavo: Unfortunately, it will only detect issues where `logr.Logger` is directly used - won't help where our wrapper functions are used (`log.Debug`, `log.Info`, etc.)

Issue #949, opened 2024-12-20
Migrate from webhook to CRD CEL expressions, `ValidatingWebhookPolicy` and `ValidatingWebhookPolicyBinding`


pmalek:
### Problem statement

Creating and managing webhook resources like certificates and coordinating the `kube-webhook-certgen` `Job` to properly update the webhook's `caBundle` has proven to be difficult and unreliable.

This setup has already caused issues for a number of users (internal and external):

- #428 
- #122 
- #921 

### Proposed solution

Migrate away from `ValidatingWebhookConfiguration` and managing certificates and `kube-webhook-certgen` `Job`, towards using `ValidatingWebhookPolicy` and `ValidatingWebhookPolicyBinding`s which will be deployed as part of [KGO's helm chart](https://github.com/Kong/charts/tree/main/charts/gateway-operator).

All the rules that are currently defined in 

- https://github.com/Kong/gateway-operator/blob/d6b69e3366089fda45386e3289138b5832b97c96/internal/validation/dataplane/validation.go
- and https://github.com/Kong/gateway-operator/blob/3876430e09e61edce58bd8464989e33236bd1872/internal/validation/controlplane/validation.go

are either trivial to implement at CEL expressions on CRDs or through CEL expressions in `ValidatingAdmissionPolicy`.

Some rules would be more complex than the others and would require to be added as the latter, e.g. [validation of `KONG_PORT_MAPS` and `KONG_PROXY_LISTEN` against `DataPlane`'s `spec.network.services.ingress.ports](https://github.com/Kong/gateway-operator/blob/d6b69e3366089fda45386e3289138b5832b97c96/internal/validation/dataplane/validation.go#L133-L248), when added as CRD CEL expression:

```
// +kubebuilder:validation:XValidation:message="Each port from spec.network.services.ingress.ports has to have an accompanying port in KONG_PORT_MAPS env",rule="self.spec.network.services.ingress.ports.all(p, self.spec.deployment.podTemplateSpec.spec.containers.filter(c, c.name == 'proxy')[0].env.filter(e, e.name == 'KONG_PORT_MAPS')[0].value.split(',').exists(pm, pm.split(':')[1] == string(p.targetPort)))"
```

yields the following

```
The CustomResourceDefinition "dataplanes.gateway-operator.konghq.com" is invalid:
[...]

PR #984, opened 2025-01-09
feat(controlplane): move ControlPlane image validation to CRD CEL rule

pmalek:

Move ControlPlane image validation to CRD CEL rule

**Which issue this PR fixes**

Part of #949 

**PR Readiness Checklist**:

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -174,6 +174,22 @@ jobs:
       with:
         name: tests-report-unit-tests
         path: unit-tests.xml
+
+  CRDs-validation:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+
+    - uses: actions/setup-go@v5
+      with:
+        go-version-file: go.mod
+
+    - uses: jdx/mise-action@5083fe46898c414b2475087cc79da59e7da859e8 # v2.1.11
+      with:
+        install: false
+
+    - name: Run the crds validation tests
+      run: make test.crds-validation
   
   envtest-tests:
     runs-on: ubuntu-latest
@@ -505,6 +521,7 @@ jobs:
       - build
       - unit-tests
       - envtest-tests
+      - CRDs-validation
       - samples
       - conformance-tests
       - integration-tests
--- a/api/v1beta1/controlplane_types.go
+++ b/api/v1beta1/controlplane_types.go
@@ -56,6 +56,8 @@ type ControlPlaneList struct {
 	Items           []ControlPlane `json:"items"`
 }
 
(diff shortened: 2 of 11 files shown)


api/v1beta1/controlplane_types.go
```diff
@@ -56,6 +56,8 @@ type ControlPlaneList struct {
 	Items           []ControlPlane `json:"items"`
 }
 
+// +kubebuilder:validation:XValidation:message="ControlPlane requires an image to be set on controller container",rule="has(self.deployment.podTemplateSpec) && has(self.deployment.podTemplateSpec.spec.containers) && self.deployment.podTemplateSpec.spec.containers.exists(c, c.name == 'controller' && has(c.image))"
```
programmer04, line 59: I think we should have some testing strategy for those CEL rules. It's easy to make even simple mistakes in a plain text DSL included in a comment. Is it possible to cover them with K8s envtest? It should be quite lightweight in terms of resources and execution times, but robust

pmalek: That's a good idea. I'll try to prepare something.

Given that we do plan to move the CRDs to kcfg this would ultimately be covered by https://github.com/Kong/kubernetes-configuration/tree/main/test/crdsvalidation but I believe it's good to have something now to prevent errors from creeping in.

programmer04 (dismissed):

api/v1beta1/controlplane_types.go
pmalek: I propose to use kcfg `crdsvalidation` package for this purpose. https://github.com/Kong/kubernetes-configuration/pull/220 exports this (up until now the types there have been located under `crdsvalidation_test` package). When that gets merged I'll work on using it for changes here.

pmalek: Added https://github.com/Kong/gateway-operator/pull/984/commits/20d8936afbc595b0c7c10f4a49008aae6247d151. 

Uses commit from https://github.com/Kong/kubernetes-configuration/pull/221.

.github/workflows/tests.yaml
```diff
@@ -174,6 +174,22 @@ jobs:
       with:
         name: tests-report-unit-tests
         path: unit-tests.xml
+
+  CRDs-validation:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
```
programmer04, line 181: Those officials Actions from GH we consider as trusted and do not pin by hash?

pmalek: I believe this is something worth discussing. Do we want to make an exception for `actions/*` actions? 

On one hand better to be consistent and have the same rules for all actions. 

OTOH this could create a lot of dependabot spam bumping each time a patch version of e.g. `actions/checkout` is released.

programmer04: You have my thoughts. Let's bring it up on today's team-k8s sync

programmer04 (dismissed):

programmer04 (approved):

PR #997, opened 2025-01-13
Feat(konnect): adopt existing konnect entities

### LINKED ISSUE #460: ☔ Konnect entities: existing entity adoption
### Aha! Task

[KUB-65](https://konghq.aha.io/features/KUB-65)

### Problem statement

Users might want to adopt already existing Konnect entities into their Kubernetes resources.

Currently (as implemented in Kong/kubernetes-configuration#446), `status.konnect.id` drives the operator decision whether to create or update the Konnect entity. If that field is empty, the entity is created, otherwise its updated.

This task tracks the effort of introducing a mechanism which will allow users to adopt already existing entities.

### Proposed solution

Implement a mechanism which will react to an annotation on Kubernetes resources mapping Konnect entities which will express user's intent to adopt the already existing entity.

This annotation's value would indicate which Konnect ID this resource is supposed to adopt.

```
konnect.konghq.com/adopt: c66a3c89-9bd4-4e2e-a6fa-1111111111111
```

### Acceptance

- [ ] As a User I have means to create a Kubernetes resource and adopt already existing Konnect entity.

### Sub-tasks
- [ ] https://github.com/Kong/kong-operator/issues/2277

randmonkey:

Support adopting existing konnect entity by annotation `konnect.konghq.com/adopt`.
**Which issue this PR fixes**

Fixes #460

This PR starts from `KonnectGatewayControlPlane`. The following work in mainly on adding functions to implement `GetByID` for other types of entities.
**PR Readiness Checklist**:

### DIFF
--- a/controller/konnect/annotations.go
+++ b/controller/konnect/annotations.go
@@ -1,0 +1,14 @@
+package konnect
+
+const (
+	// AnnotationKeyAdoptEntity is the annotation key to adopt an existing Konnect entity.
+	AnnotationKeyAdoptEntity = "konnect.konghq.com/adopt"
+)
+
+// getAdoptEntityID gets the Konnect ID of the adopted entity in the value of the annotation.
+func getAdoptEntityID(annotations map[string]string) string {
+	if annotations == nil {
+		return ""
+	}
+	return annotations[AnnotationKeyAdoptEntity]
+}
--- a/controller/konnect/ops/ops.go
+++ b/controller/konnect/ops/ops.go
@@ -36,6 +36,8 @@ const (
 	UpdateOp Op = "update"
 	// DeleteOp is the operation type for deleting a Konnect entity.
 	DeleteOp Op = "delete"
+	// GetOp is the operation type for getting a Konnect entity.
+	GetOp Op = "get"
 )
 
 // Create creates a Konnect entity.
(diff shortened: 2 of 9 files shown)


pmalek:
Can we create a doc / more elaborate PR description for this feature before we move forward with it?

I'd like to make sure that we understand the corner cases that can potentially arise with using this feature (e.g. what happens if 2 unrelated entities - of the same type - in 2 namespaces of clusters use the same `konnect.konghq.com/adopt` label value). That would be a conflict that would have to be resolved somehow. The second entity to use the label would potentially (I'm not 100% sure) get a conflict condition of sorts to indicate a failure.

There can be more corner cases, this is just one from the top of my head.

Let me know if it makes sense.

randmonkey:
@pmalek The issue is much more complicated than it looks like after some consideration. I can come up with the following problem:
- Can user change the adopted entity ID? I think we should make it immutable at least after it is `Programmed` since it is problematic to point a resource to another Konnect entity after it is already pointing to one.
- Which one should be the source of truth, the spec of the k8s resource or the existing Konnect entity? I think we should let the k8s resource to be source of the truth.
- My opinion of conflicting resources: We can put all resources with the same adopted entity ID into a "priority queue" ordered by creation time and only reconcile the first one. The others should have `Programmed=false` condition in the status.
- How about deleting a resource adopting an existing entity? 
  - What if we delete a resource adopting an existing entity, but there are other resources adopting the same entity ID not deleted?


Do you think these opinions make sense, or do we need some other one to look at these problems?

pmalek:
> @pmalek The issue is much more complicated than it looks like after some consideration.

For this reason I'd suggest us to take a step back and consider drafting a design for this so that we can review it before implementation.

> * Can user change the adopted entity ID? I think we should make it immutable at least after it is `Programmed` since it is problematic to point a resource to another Konnect entity after it is already pointing to one.

This would be a problem of potentially migrating between CPs which is not implemented today. This would require deleting the entity with the old ID and creating a new one, potentially with a different CP. I'd be inclined to say this is out of scope for this feature.

> * Which one should be the source of truth, the spec of the k8s resource or the existing Konnect entity? I think we should let the k8s resource to be source of the truth.

K8s resource IMO.

> * My opinion of conflicting resources: We can put all resources with the same adopted entity ID into a "priority queue" ordered by creation time and only reconcile the first one. The others should have `Programmed=false` condition in the status.

Agreed on the condition. The simplest mechanism I'd have for this is to decide this based on the tag/label set on the Konnect entity indicating this is managed by a k8s resource. This is already done with [these labels](https://github.com/Kong/gateway-operator/blob/fad649a79a91de3033ebf79124619c41c2cb3623/controller/konnect/ops/objects_metadata.go#L16-L37), so simply a presence of `k8s-uid` would already indicate that this resource is already managed from a k8s cluster.

> * How about deleting a resource adopting an existing entity?
>   
>   * What if we delete a resource adopting an existing entity, but there are other resources adopting the same entity ID not deleted?

I believe this comes down to 3 distinct stages:

- first, when there's more than 1 k8s resource attempting to adopt, 1 will eventually win and set the tags. Others should fail (with appropriate status condition)
- secondly, when k8s resource which successfully adopted gets deleted, Konnect entity also gets deleted. This is already the case for all the Konnect entities that are managed by k8s resources.
- last but not least, when Konnect entity gets deleted all the other k8s resources attempting to adopt it will receive an appropriate status condition indicating they can't adopt because the remote resource doesn't exist.

---

These are excellent questions and I believe there would be more. We just need to write them down and get them peer reviewed.

pmalek:
Moved this to 1.6 milestone as [we've decided](https://github.com/Kong/gateway-operator/issues/460#issuecomment-2633312915) to punt it.

pmalek:
Since #460 has been moved to KO 2.0 punting this as well.

randmonkey:
Close this as the code is stale and may not satisfy the current requirement and design of the issue.

PR #999, opened 2025-01-13
feat(dataplane): move DataPlane image validation to CRD CEL rule

pmalek:

**Which issue this PR fixes**

Part of #949

**PR Readiness Checklist**:

### DIFF
--- a/api/v1beta1/controlplane_types.go
+++ b/api/v1beta1/controlplane_types.go
@@ -35,6 +35,7 @@ func init() {
 // +kubebuilder:resource:shortName=kocp,categories=kong;all
 // +kubebuilder:printcolumn:name="Ready",description="The Resource is ready",type=string,JSONPath=`.status.conditions[?(@.type=='Ready')].status`
 // +kubebuilder:printcolumn:name="Provisioned",description="The Resource is provisioned",type=string,JSONPath=`.status.conditions[?(@.type=='Provisioned')].status`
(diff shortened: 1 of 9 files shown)


randmonkey (approved):
LGTM generally.
Should we define the `spec.deployent.podTemplateSpec` in `ControlPlane` and `DataPlane` to `required` insstead of `optional`? The pod templates looks mandatory.

pmalek:
> Should we define the `spec.deployent.podTemplateSpec` in `ControlPlane` and `DataPlane` to `required` instead of `optional`?

That would be a breaking change technically so we shouldn't do that in a minor release.

We didn't mark it as `required` because the type that encloses it is reused in `GatewayConfiguration` (specifically in https://github.com/Kong/gateway-operator/blob/f2ac0c8ab6e8d049da4e9fcd88dda94e74d7a626/api/v1beta1/gatewayconfiguration_types.go#L66) which doesn't require the podTemplateSpec (or image) to be specified. That allows simple manifests  using `Gateway` like:

https://github.com/Kong/gateway-operator/blob/dd74abd7598f4e30f4a8c1f77d2288b90137530b/config/samples/gateway.yaml

```
kind: GatewayClass
metadata:
  name: kong
spec:
  controllerName: konghq.com/gateway-operator
---
kind: Gateway
metadata:
  name: kong
  namespace: default
spec:
  gatewayClassName: kong
  listeners:
  - name: http
    protocol: HTTP
    port: 80
```

to be used.

PR #1007, opened 2025-01-13
feat(dataplane): move DataPlane ports validation to ValidationAdmissionPolicy and ValidationAdmissionPolicyBinding

pmalek:

**Which issue this PR fixes**

Part of #949.

This will get an accompanying charts PR where `ValidationAdmissionPolicy and `ValidationAdmissionPolicyBinding` are added there.

This all sets the max items on the `DataPlaneServicePort` array to 4 to potentially cut the added validations cost down.

An initial attempt on this was to add CEL rules on the CRD but API server was complaining about the cost of these. With `ValidationAdmissionPolicy` rules it was possible.

> NOTE: This will require k8s 1.30 as [`ValidatingAdmissionPolicy`](https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/) graduated to stable in 1.30  but users will only get this deployed in the helm chart when `.Capabilities.APIVersions.Has` check will pass, ref https://github.com/Kong/charts/pull/1215.

**PR Readiness Checklist**:


czeslavo (dismissed):
Given that we plan on making this repo a source of truth for ValidationAdmissionPolicies, should we add tests verifying them in here?

pmalek:
> Given that we plan on making this repo a source of truth for ValidationAdmissionPolicies, should we add tests verifying them in here?

I agree, as agreed on slack I added tests by extending the crd validation tests in b03db6a04767fbace0c9613da17a4c0d32d12ec5

PTAL.

pmalek:
@czeslavo This is ready for review now. Feel free to take a look and comment.

config/default/validation_policies/dataplane_validation_policies.yaml
czeslavo, line 1: Nit: to be consistent with the Kubernetes nomenclature we might want to name the directory `validating_policies`. Same applies to the file name.

test/helpers/kustomize/apply.go
```diff
@@ -0,0 +1,153 @@
...
+	"sigs.k8s.io/kustomize/api/krusty"
+	"sigs.k8s.io/kustomize/kyaml/filesys"
+
+	"github.com/kong/gateway-operator/pkg/utils/test"
+)
+
+var decUnstructured = yaml.NewDecodingSerializer(unstructured.UnstructuredJSONScheme)
+
+// Apply applies a kustomization to the cluster using the given rest.Config.
+func Apply(ctx context.Context, t *testing.T, cfg *rest.Config, dir string) {
```
czeslavo, line 34: That's quite a lot of code to write to support `kustomize apply`. 😅

test/crdsvalidation/dataplane_validatingadmissionpolicy_test.go
```diff
@@ -0,0 +1,261 @@
...
+
+	kcfgcrdsvalidation "github.com/kong/kubernetes-configuration/test/crdsvalidation"
+)
+
+const (
+	// KustomizePathValidationPolicies is the path to the Kustomize directory containing the validation policies.
+	KustomizePathValidationPolicies = "config/default/validation_policies/"
+)
+
+func TestDataPlaneValidatingAdmissionPolicy(t *testing.T) {
```
czeslavo, line 26: 🎖️

czeslavo (dismissed):

config/default/validation_policies/dataplane_validation_policies.yaml
pmalek: 4a4729f85503e2f975d6e02ee6cb7c126730f989

test/helpers/kustomize/apply.go
programmer04: We have something like this

https://github.com/Kong/gateway-operator/blob/d8b42dd60620e9969db375361015775794cd3256/pkg/utils/test/setup_helpers.go#L214-L244

and it uses `clusters.KustomizeDeployForCluster` from KTF, maybe there is an opprotunity to reuse some bits  to reduce the core implementation to a common denominator. WDYT? Ofc it can be handled as a separte issue

pmalek: We could do 2 things:

- provide an implementation for the envtest cluster by implementing the `Cluster` interface and thus making it possible to create a cluster with `envtest.Setup()` and then pass it to `clusters.KustomizeDeployForCluster()` and the like
- or decrease the interface requirements on https://github.com/Kong/kubernetes-testing-framework/blob/efb08ea99cc422e457a0cc3cca93431fd388dbfc/pkg/clusters/utils.go#L211-L239 by requiring only the `Name() string` and `Config() *rest.Config` methods thus making it very easy to provide a wrapper type for envtest cluster

I'm not sure I'd like to commit to either of these ways for now and that's why I implemented it using the kustomize API. I'm open to comments on how you'd like to see this. Perhaps the second option wouldn't be that bad. LMK.

programmer04: 👍 for the second option, making interfaces as narrow as possible is the soul of the good Go code, IMO

pmalek: I've created https://github.com/Kong/kubernetes-testing-framework/issues/1262 to track this. In the meantime this PR will require eventual CRD assertions which are provided by https://github.com/Kong/kubernetes-configuration/pull/235. PTAL

czeslavo: Hm, it would be nice if `ValidatingAdmissionPolicyBinding` supported reporting in its status whether it's successfully activated. CRDs implement status, and that allowed us to not implement the assertion with eventual consistency. Looking at the [API](https://github.com/kubernetes/kubernetes/blob/f64b651ebae643d422f4625161dc415970e2c166/pkg/apis/admissionregistration/types.go#L426) though, there's no such capability as of now :(

czeslavo (approved):

juangon:
Hi @pmalek . I am trying to upgrade from KGO 1.4 to KGO 1.5 or 1.6, but it fails on this PR's validation hook, because KGO already set KONG_PORT_MAPS and KONG_PROXY_LISTEN. As these ports arent' the same as service`s, it fails. 
I've tried to change these env variables values manually so they're the same, but it makes my current environment unusable before the upgrade. Any idea? Thanks very much!!

pmalek:
> Hi @pmalek . I am trying to upgrade from KGO 1.4 to KGO 1.5 or 1.6, but it fails on this PR's validation hook, because KGO already set KONG_PORT_MAPS and KONG_PROXY_LISTEN. As these ports arent' the same as service`s, it fails. I've tried to change these env variables values manually so they're the same, but it makes my current environment unusable before the upgrade. Any idea? Thanks very much!!

Hi @juangon. Can you create an issue for this? It's not clear to me how this would fail. I'd assume you're using the [`gateway-operator` helm chart](https://github.com/Kong/charts/tree/main/charts/gateway-operator) for installing and upgrade, correct?

juangon:
> > Hi @pmalek . I am trying to upgrade from KGO 1.4 to KGO 1.5 or 1.6, but it fails on this PR's validation hook, because KGO already set KONG_PORT_MAPS and KONG_PROXY_LISTEN. As these ports arent' the same as service`s, it fails. I've tried to change these env variables values manually so they're the same, but it makes my current environment unusable before the upgrade. Any idea? Thanks very much!!
> 
> Hi @juangon. Can you create an issue for this? It's not clear to me how this would fail. I'd assume you're using the [`gateway-operator` helm chart](https://github.com/Kong/charts/tree/main/charts/gateway-operator) for installing and upgrade, correct?

Yes, using helm charts.

pmalek:
> > > Hi @pmalek . I am trying to upgrade from KGO 1.4 to KGO 1.5 or 1.6, but it fails on this PR's validation hook, because KGO already set KONG_PORT_MAPS and KONG_PROXY_LISTEN. As these ports arent' the same as service`s, it fails. I've tried to change these env variables values manually so they're the same, but it makes my current environment unusable before the upgrade. Any idea? Thanks very much!!
> > 
> > 
> > Hi @juangon. Can you create an issue for this? It's not clear to me how this would fail. I'd assume you're using the [`gateway-operator` helm chart](https://github.com/Kong/charts/tree/main/charts/gateway-operator) for installing and upgrade, correct?
> 
> Yes, using helm charts.

Please, create an issue for this so that we can understand more about this.

juangon:
Thanks @pmalek ! Here it is: https://github.com/Kong/kong-operator/issues/2482

EDIT: it seems now I was able to upgrade to 1.5.x Thanks for your help!

PR #1013, opened 2025-01-14
chore: remove redundant dependency `golang.org/x/net`

programmer04:

Redo 

- https://github.com/Kong/gateway-operator/pull/1002

because

- https://github.com/Kong/gateway-operator/pull/999

introduced this again by mistake, [see](https://github.com/Kong/gateway-operator/pull/999/files#diff-0106af3724df5586d4928666ade8185dea56e53cbb0480c2c326552ed1a321faR7)

### DIFF
--- a/test/crdsvalidation/dataplane_test.go
+++ b/test/crdsvalidation/dataplane_test.go
@@ -1,10 +1,10 @@
 package crdsvalidation
 
 import (
+	"context"
 	"testing"
 
 	"github.com/samber/lo"
-	"golang.org/x/net/context"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 


pmalek:
Maybe we introduce a simple script that will check for uses of `"golang.org/x/net/context"`?
Could be a simple bash script or something that parses the AST tree and checks the imports (if such a tool is not yet available)

programmer04:
Hmmm, maybe it was bad luck because this is the first time we have had such a problem. I don't recall fighting with it in the past, just bad timing and rolling a tool may be [not worth it](https://imgs.xkcd.com/comics/is_it_worth_the_time_2x.png).  


But contributing to `gomodguard` to overcome the limitation described in https://github.com/Kong/gateway-operator/pull/1002 may be a nice little side quest that I may investigate in a spare time 😄

pmalek:
> Hmmm, maybe it was bad luck because this is the first time we have had such a problem. I don't recall fighting with it in the past, just bad timing and rolling a tool may be [not worth it](https://imgs.xkcd.com/comics/is_it_worth_the_time_2x.png).
> 
> But contributing to `gomodguard` to overcome the limitation described in #1002 may be a nice little side quest that I may investigate in a spare time 😄

I'll leave the decision to you but I believe a simple `grep "golang.org/x/net/context"` would suffice (at least for now).

pmalek (approved):

PR #1021, opened 2025-01-14
feat: set GWC accepted condition properly

### LINKED ISSUE #982: Validate the `parametersRef` referenced by `GatewayClass`
## Problem Statement

The `.spec.parametersRef` field of the `GatewayClass` objects can be used to customize the `Gateway`s behaviour. KGO accepts only `GatewayConfiguration` objects as `parametersRef`. We need to validate that in case a `GatewayClass` is managed by Kong, such a field must be unset or properly set.

## Proposed Solution

Use the `GatewayClassReasonInvalidParameters` `Accepted` reason to reflect the status of the `ParametersRef` field.

## Acceptance Criteria
- [ ] In case the `GatewayClass` is managed by Kong and references an unsupported `parametersRef`, the `GatewayClass` is marked as not accepted with reason `GatewayClassReasonInvalidParameters`.

mlavacca:

The GWC accepted condition is now properly set to False with reason InvalidParameters in case the ParametersRef field of the GatewayClass spec is not a valid GatewayConfiguration. 

**Which issue this PR fixes**

Fixes #982 

**PR Readiness Checklist**:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,13 +67,12 @@
 - Move `DataPlane` promotion in progress validation to CRD CEL validation expressions.
   This is relevant for `DataPlane`s with BlueGreen rollouts enabled only.
   [#1054](https://github.com/Kong/gateway-operator/pull/1054)
-- The `GatewayClass` Accepted Condition is set to False with reason `InvalidParameters`
+- The `GatewayClass` Accepted Condition is set to `False` with reason `InvalidParameters`
   in case the `.spec.parametersRef` field is not a valid reference to an existing
   `GatewayConfiguration` object.
   [#1021](https://github.com/Kong/gateway-operator/pull/1021)
 
 [kubebuilder_3907]: https://github.com/kubernetes-sigs/kubebuilder/discussions/3907
-
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,8 +67,13 @@
 - Move `DataPlane` promotion in progress validation to CRD CEL validation expressions.
   This is relevant for `DataPlane`s with BlueGreen rollouts enabled only.
   [#1054](https://github.com/Kong/gateway-operator/pull/1054)
+- The `GatewayClass` Accepted Condition is set to False with reason `InvalidParameters`
+  in case the `.spec.parametersRef` field is not a valid reference to an existing
+  `GatewayConfiguration` object.
+  [#1021](https://github.com/Kong/gateway-operator/pull/1021)
 
 [kubebuilder_3907]: https://github.com/kubernetes-sigs/kubebuilder/discussions/3907
+
 
 ### Fixes
 
--- a/controller/gatewayclass/controller.go
+++ b/controller/gatewayclass/controller.go
@@ -59,10 +59,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	if err != nil {
 		return ctrl.Result{}, fmt.Errorf("failed to get accepted condition: %w", err)
 	}
-	k8sutils.SetCondition(
-		*condition,
-		gwc,
-	)
+	k8sutils.SetCondition(*condition, gwc)
 
 	if err := r.Status().Patch(ctx, gwc.GatewayClass, client.MergeFrom(oldGwc)); err != nil {
 		if k8serrors.IsConflict(err) {
--- a/controller/gatewayclass/controller.go
+++ b/controller/gatewayclass/controller.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
@@ -16,7 +15,6 @@ import (
 	"github.com/kong/gateway-operator/controller"
 	"github.com/kong/gateway-operator/controller/pkg/log"
 	"github.com/kong/gateway-operator/internal/utils/gatewayclass"
-	"github.com/kong/gateway-operator/pkg/consts"
 	k8sutils "github.com/kong/gateway-operator/pkg/utils/kubernetes"
 )
 
@@ -55,29 +53,26 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{}, nil
 	}
 
-	if !gwc.IsAccepted() {
-		oldGwc := gwc.DeepCopy()
+	oldGwc := gwc.DeepCopy()
 
-		k8sutils.SetCondition(
-			k8sutils.NewConditionWithGeneration(
-				consts.ConditionType(gatewayv1.GatewayClassConditionStatusAccepted),
-				metav1.ConditionTrue,
-				consts.ConditionReason(gatewayv1.GatewayClassReasonAccepted),
-				"the gatewayclass has been accepted by the operator",
-				gwc.GetGeneration(),
-			),
-			gwc,
-		)
-		if err := r.Status().Patch(ctx, gwc.GatewayClass, client.MergeFrom(oldGwc)); err != nil {
-			if k8serrors.IsConflict(err) {
-				log.Debug(logger, "conflict found when updating GatewayClass, retrying")
-				return ctrl.Result{
-					Requeue:      true,
-					RequeueAfter: controller.RequeueWithoutBackoff,
-				}, nil
-			}
-			return ctrl.Result{}, fmt.Errorf("failed patching GatewayClass: %w", err)
+	condition, err := getAcceptedCondition(ctx, r.Client, gwc.GatewayClass)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to get accepted condition: %w", err)
+	}
+	k8sutils.SetCondition(
+		*condition,
+		gwc,
+	)
+
+	if err := r.Status().Patch(ctx, gwc.GatewayClass, client.MergeFrom(oldGwc)); err != nil {
+		if k8serrors.IsConflict(err) {
+			log.Debug(logger, "conflict found when updating GatewayClass, retrying")
+			return ctrl.Result{
+				Requeue:      true,
+				RequeueAfter: controller.RequeueWithoutBackoff,
+			}, nil
 		}
+		return ctrl.Result{}, fmt.Errorf("failed patching GatewayClass: %w", err)
 	}
 
 	return ctrl.Result{}, nil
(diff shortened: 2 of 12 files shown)


controller/gatewayclass/controller.go
```diff
@@ -55,29 +53,26 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
-			}
-			return ctrl.Result{}, fmt.Errorf("failed patching GatewayClass: %w", err)
+	condition, err := getAcceptedCondition(ctx, r.Client, gwc.GatewayClass)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to get accepted condition: %w", err)
+	}
+	k8sutils.SetCondition(
+		*condition,
+		gwc,
+	)
```
programmer04:
```suggestion
	k8sutils.SetCondition(*condition, gwc)
```

CHANGELOG.md
```diff
@@ -55,9 +55,14 @@
...
   [kubernetes-sigs/kubebuilder#3907][kubebuilder_3907].
   [#956](https://github.com/Kong/gateway-operator/pull/956)
+- The `GatewayClass` Accepted Condition is set to False with reason `InvalidParameters`
+  in case the `.spec.parametersRef` field is not a valid reference to an existing
+  `GatewayConfiguration` object.
+  [#1021](https://github.com/Kong/gateway-operator/pull/1021)
 
 [kubebuilder_3907]: https://github.com/kubernetes-sigs/kubebuilder/discussions/3907
 
+
```
```suggestion
```

CHANGELOG.md
```diff
@@ -55,9 +55,14 @@
   For more information on this migration please consult
   [kubernetes-sigs/kubebuilder#3907][kubebuilder_3907].
   [#956](https://github.com/Kong/gateway-operator/pull/956)
+- The `GatewayClass` Accepted Condition is set to False with reason `InvalidParameters`
```
```suggestion
- The `GatewayClass` Accepted Condition is set to `False` with reason `InvalidParameters`
```

controller/gatewayclass/controller.go
```diff
@@ -55,29 +53,26 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
-				log.Debug(logger, "conflict found when updating GatewayClass, retrying")
-				return ctrl.Result{
-					Requeue:      true,
-					RequeueAfter: controller.RequeueWithoutBackoff,
-				}, nil
-			}
-			return ctrl.Result{}, fmt.Errorf("failed patching GatewayClass: %w", err)
+	condition, err := getAcceptedCondition(ctx, r.Client, gwc.GatewayClass)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to get accepted condition: %w", err)
```
pmalek, line 60: Shouldn't we set the condition here with some error message so that users do not have to look into operators logs? Ideally I'd consider adding the error's message into the condition message so that users know what happened. WDYT?

mlavacca: The `AcceptedCondition` message is properly set in the `getAcceptedCondition` function. Isn't this what you are referring to?

pmalek: I was specifically alluding here to lines 42-44 https://github.com/Kong/gateway-operator/pull/1021/files/31ff623213f0fed16947060e8a0eccec5e02f4ff#diff-f4e83572e27854d40d56ef54e8f982cdca3ccaaa3f17900c3ab6e3f4fb04e971R42-R44 where the `err` is simply returned. Then we seem to return it here as well. Not setting anything on the `GatewayClass`. Or am I missing something here?

mlavacca: That can be only a transient error caused by the get function. I don't think we should change the status of the resource when such an error is raised. Do you?

controller/gatewayclass/controller_reconciler_utils.go
```diff
@@ -0,0 +1,65 @@
...
+		if validRef {
+			gatewayConfig := operatorv1beta1.GatewayConfiguration{}
+			err := cl.Get(ctx, client.ObjectKey{Name: gwc.Spec.ParametersRef.Name, Namespace: string(*gwc.Spec.ParametersRef.Namespace)}, &gatewayConfig)
+			if client.IgnoreNotFound(err) != nil {
+				return nil, err
+			}
+			if k8serrors.IsNotFound(err) {
+				reason = string(gatewayv1.GatewayClassReasonInvalidParameters)
+				message = append(message, "The referenced GatewayConfiguration does not exist")
+			}
```
programmer04:
```suggestion
			if k8serrors.IsNotFound(err) {
				reason = string(gatewayv1.GatewayClassReasonInvalidParameters)
				message = append(message, "The referenced GatewayConfiguration does not exist")
			} else {
			        return nil, err
			}
```
Wouldn't it be simpler? WDYT?

controller/gatewayclass/controller_reconciler_utils.go
```diff
@@ -0,0 +1,65 @@
...
+	operatorv1beta1 "github.com/kong/gateway-operator/api/v1beta1"
+	"github.com/kong/gateway-operator/pkg/consts"
+	k8sutils "github.com/kong/gateway-operator/pkg/utils/kubernetes"
+)
+
+// getAcceptedCondition returns the accepted condition for the GatewayClass, with
+// the proper status, reason and message.
+func getAcceptedCondition(ctx context.Context, cl client.Client, gwc *gatewayv1.GatewayClass) (*metav1.Condition, error) {
+	reason := string(gatewayv1.GatewayClassReasonAccepted)
+	message := []string{}
```
programmer04:
```suggestion
	messages := []string{}
```
Since it's a list, I suggest usage of plural

controller/gatewayclass/controller.go
pmalek: It doesn't have to be transient. It can also be a problem with e.g. permissions. I believe it would be better UX to make the error visible in the status rather than forcing users to go through operator's logs. WDYT?

controller/gatewayclass/controller_reconciler_utils.go
```diff
@@ -0,0 +1,65 @@
...
+
+	operatorv1beta1 "github.com/kong/gateway-operator/api/v1beta1"
+	"github.com/kong/gateway-operator/pkg/consts"
+	k8sutils "github.com/kong/gateway-operator/pkg/utils/kubernetes"
+)
+
+// getAcceptedCondition returns the accepted condition for the GatewayClass, with
+// the proper status, reason and message.
+func getAcceptedCondition(ctx context.Context, cl client.Client, gwc *gatewayv1.GatewayClass) (*metav1.Condition, error) {
+	reason := string(gatewayv1.GatewayClassReasonAccepted)
```
programmer04: Why do you want this `reason` var to be a `string` type? Later you always convert from `GatewayClassConditionType` to `string` and at the end in the line `59` you make a conversion again

controller/gatewayclass/controller.go
mlavacca: I don't think so honestly. If that's an error related to permissions, for example, that isn't related to the resource, but to the operator itself. An object status should include information related to the resource(s) themselves, not to the operators managing it IMO.

controller/gatewayclass/controller_reconciler_utils_test.go
```diff
@@ -0,0 +1,131 @@
...
+			},
+			expectedStatus: metav1.ConditionTrue,
+			expectedReason: string(gatewayv1.GatewayClassReasonAccepted),
+			expectedMsg:    "GatewayClass is accepted",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ctx := context.TODO()
```
programmer04:
```suggestion
			ctx := context.Background()
```
Are you going to do something with a TODO context in the future? If not, I'm suggesting the above. You can even pull it above for loop

internal/utils/gatewayclass/get.go
```diff
@@ -30,6 +35,10 @@ func Get(ctx context.Context, cl client.Client, gatewayClassName string) (*Decor
 			vars.ControllerName(),
 		))
 	}
+	acceptedCondition, found := k8sutils.GetCondition(consts.ConditionType(gatewayv1.GatewayClassConditionStatusAccepted), gwc)
+	if acceptedCondition.Status != metav1.ConditionTrue || !found {
```
programmer04:
```suggestion
	if !found || acceptedCondition.Status != metav1.ConditionTrue {
```
I suggest this order, because as I understand it, when `!found` is `true` value of `acceptedCondition` should be discarded

controller/gatewayclass/controller.go
pmalek: > If that's an error related to permissions, for example, that isn't related to the resource, but to the operator itself. 

I agree but the error in object's status is a breadcrumb which can help you as a user debug this problem faster.

The permission might not be directly related to the operator. There might exist policies in the cluster which would prevent the `Get` from succeeding.

I have already seen reports of several issues where users were confused because an object didn't get any status condition set (for different reasons but still). For this reason I do believe that putting information that can help debug a problem in status condition is going to allow people to quickly resolve an issue that caused it.

controller/gatewayclass/controller_reconciler_utils.go
mlavacca: It's a different behavior. With your proposed solution we return even if the error is nil

controller/gatewayclass/controller.go
mlavacca: I hear you and tend to agree with this general statement of helping users to debug problems. But here we are discussing about putting information that has nothing to do with the resource itself. I think we should instead try to educate users about what kind of information can be found where. I am very opinionated on this matter, and for the reasons above I think this is not the way to go honestly.

controller/gatewayclass/controller_reconciler_utils.go
programmer04: Yep, my mistake

controller/gatewayclass/controller_reconciler_utils.go
mlavacca: good point, I've removed the cast to a string

controller/gatewayclass/controller.go
pmalek: > But here we are discussing about putting information that has nothing to do with the resource itself.

Agreed. Nevertheless this information is beneficial to the user to quickly resolve an issue with this object not being accepted.

> I think we should instead try to educate users about what kind of information can be found where.

Agreed on this as well. Do you have an idea how we could educate users about this one so that they can better debug problems at we discuss here?

---

I see your view and I understand it yet I do not share it. Feel free to continue with the way this is proposed as of now 👍

internal/errors/errors.go
```diff
@@ -31,6 +33,21 @@ func (e ErrUnsupportedGatewayClass) Error() string {
...
 }
 
+// ErrNotAcceptedGatewayClass is an error which indicates that a provided GatewayClass
+// is not accepted.
+type ErrNotAcceptedGatewayClass struct {
+	gatewayClass string
+	condition    metav1.Condition
+}
+
+func NewErrNotAcceptedGatewayClass(gatewayClass string, condition metav1.Condition) ErrNotAcceptedGatewayClass {
```
pmalek, line 46: This func should have a godoc

internal/errors/errors.go
```diff
@@ -31,6 +33,21 @@ func (e ErrUnsupportedGatewayClass) Error() string {
...
+	return ErrNotAcceptedGatewayClass{gatewayClass: gatewayClass, condition: condition}
+}
+
+func (e ErrNotAcceptedGatewayClass) Error() string {
```
pmalek, line 51: Ditto

pmalek (commented):
Minor nits but overall 👍 .

I'm OK with this PR moving forward even without writing the error from `Get()` in object status.

pmalek (dismissed):

pmalek (approved):

PR #1049, opened 2025-01-17
feat(dataplane): move DataPlane dbmode validation to CRD CEL validation rules

pmalek:

Move `DataPlane` dbmode validation to CRD CEL rules.

Caveat:

CEL expressions do not allow cross object validation which has been done in webhook. This validation was evaluating `EnvFrom` from Secrets and/or ConfigMaps. This PR removes these check so only the environment variables set on the object itself like so:

```
spec:
  deployment:
    podTemplateSpec:
      spec:
        containers:
        - name: proxy
          image: kong/kong-gateway:3.9
          env:
          - name: KONG_DATABASE
            value: postgres
```

are validated.

**Which issue this PR fixes**

Part of #949 

**PR Readiness Checklist**:


test/crdsvalidation/dataplane_test.go
```diff
@@ -60,6 +60,73 @@ func TestDataPlane(t *testing.T) {
 					},
 				},
 			},
+			{
+				Name: "dbmode off is supported",
```
czeslavo, line 97: How about verifying that an empty string is allowed as well?

test/crdsvalidation/dataplane_test.go
```diff
@@ -60,6 +60,106 @@ func TestDataPlane(t *testing.T) {
...
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			{
+				Name: "dbmode '' is supported",
```
czeslavo: Nit: `dbmode '' is supported` and `dbmode 'off' is supported` test cases' names should be the other way around 🙂

test/crdsvalidation/dataplane_test.go
pmalek: 👍 7c6ccc6db6670630650db94451113656d8658c2e

test/crdsvalidation/dataplane_test.go
pmalek: 🤦 6b95094b29ab74d03519c932cad00729b90ce05e

czeslavo (approved):
I think that removing the cross-object validations is a good step forward.

PR #1082, opened 2025-01-23
chore(ci): avoid infinite release-bot loop by amending commit msg

czeslavo:

As we want to cherry-pick a commit that triggers the workflow onto main, we need to amend its commit message to avoid it being triggered again.

`carloscastrojumo/github-cherry-pick-action` doesn't allow amending commit messages so this PR changes the job to make it manually using git commands.


pmalek (approved):
Looks sensible 👍

PR #1094, opened 2025-01-24
fix: make controller manager ready regardless of leader election or webhook server being started

pmalek:

When webhook is enabled, only the elected leader will get ready, causing the other operator Pods to be marked as non ready.
This can cause alerts to be triggered in some systems (e.g in ArgoCD).

To mitigate this we start the webhook server not looking at the leader election results.

To do this in our current framework (creating webhook resources within the operator code due to redhat bundle legacy reasons) we add starting of the webhook server to readiness probe (via `sync.Once`) - since we need to access the `Secret` with webhooks TLS certs.

Currently this code should be removed and replaced with webhook configuration being deployed by the helm chart.

Since at the same time webhook is being disabled by default in #1066 (and it doesn't perform any validations as of today) this should be treated as a stopgap solution before we rip this code out of operator code and add (when necessary at any point in the future) deploying `ValidatingWebhookConfiguration` from the chart.

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:


pmalek:
As discussed on today's sync meeting: we decided to drop the webhook altogether given that there are no more validations in it.

A follow up PR will handle removing the webhook code.

PR #1099, opened 2025-01-27
fix(konnect): do not set owner relationship between ControlPlane and its config entities

### LINKED ISSUE #788: Konnect entities: All entities get deleted when referenced ControlPlane gets deleted
## Current Behavior

When a `KonnectGatewayControlPlane` is deleted, the control plane is deleted on the Konnect side and all attached entities (services, routes, consumers, ...) gets deleted as well. Then, all corresponding k8s resources (`KongService`,`KongRoute`,`KongConsumer`,...) gets deleted silently.

## Expected Behavior
Similar to #786,  If cascade deletion is expected, we should add logs to tell that the resource are deleted.

pmalek:

**Which issue this PR fixes**

Uses condition reason from : https://github.com/Kong/kubernetes-configuration/pull/261

**PR Readiness Checklist**:


controller/konnect/reconciler_generic.go
```diff
@@ -174,6 +173,13 @@ func (r *KonnectEntityReconciler[T, TEnt]) Reconcile(
 				}
 			}
 		}
+
+		if res, err := setProgrammedStatusConditionBasedOnOtherConditions(ctx, r.Client, ent); err != nil {
+			return res, err
+		} else if !res.IsZero() {
```
randmonkey: Here the `Reconcile` returns `ctrl.Result{}, nil` when the returned `res`. Do you think it is simpler to directly return the return values from calling `setProgrammedStatusConditionBasedOnOtherConditions` here, like 
```
return setProgrammedStatusConditionBasedOnOtherConditions(ctx, r.Client, ent)
```
?

pmalek: 👍 0fe2fbf64511d97dd9ecc96e81da7ca0d0b9784a

pkg/utils/kubernetes/status.go
```diff
@@ -195,7 +195,10 @@ func SetAcceptedConditionOnGateway(resource ConditionsAndListenerConditionsAndGe
 	}
 }
 
-func areAllConditionsHaveTrueStatus(resource ConditionsAware) bool {
+// AreAllConditionsHaveTrueStatus checks if all the conditions on the resource are in the True state.
+// It skips the Programmed condition as that particular condition will be set based on
+// the return value of this function.
+func AreAllConditionsHaveTrueStatus(resource ConditionsAware) bool {
 	for _, condition := range resource.GetConditions() {
 		if condition.Type == string(gatewayv1.GatewayConditionProgrammed) {
```
randmonkey: The logic is not so intuitive here. I think the criteria to return false should be "Type is not `Programmed` or `Ready` and Status is `False`". So it looks clearer to move the `condition.Type != string(consts.ReadyType)` to
 
```
// Skip check if the condition type is "Programmed" or "Ready"
if condition.Type == string(gatewayv1.GatewayConditionProgrammed) || condition.Type == string(consts.ReadyType) {
    continue
} 
```

randmonkey (dismissed):
Genrally LGTM, but small places can be improved (not related to this PR though).

pkg/utils/kubernetes/status.go
pmalek: Addressed in 00f6973af3a0f6b8f8f6f443609b79d34159432b

randmonkey (dismissed):

tao12345666333 (dismissed):

tao12345666333 (dismissed):

tao12345666333 (approved):

PR #1128, opened 2025-02-05
chore: fix e2e upgrade test - add missing effective semver

czeslavo:

https://github.com/Kong/charts/pull/1243 made `image.effectiveSemver` required to be set if the `image.tag` is not a valid semver. This PR addresses this in the upgrade tests that use a non-semver tag for an image built locally.

### DIFF
--- a/test/e2e/test_helm_install_upgrade.go
+++ b/test/e2e/test_helm_install_upgrade.go
@@ -54,13 +54,16 @@ func TestHelmUpgrade(t *testing.T) {
 	}
 
 	testCases := []struct {
-		name                   string
-		fromVersion            string
-		toVersion              string
-		objectsToDeploy        []client.Object
-		upgradeToCurrent       bool
-		assertionsAfterInstall []assertion
-		assertionsAfterUpgrade []assertion
+		name             string
+		fromVersion      string
+		toVersion        string
+		objectsToDeploy  []client.Object
+		upgradeToCurrent bool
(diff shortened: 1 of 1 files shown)


test/e2e/test_helm_install_upgrade.go
```diff
@@ -135,9 +136,10 @@ func TestHelmUpgrade(t *testing.T) {
 			},
 		},
 		{
-			name:             "upgrade from latest minor to current",
-			fromVersion:      "1.4.0", // renovate: datasource=docker packageName=kong/gateway-operator-oss
-			upgradeToCurrent: true,
+			name:                     "upgrade from latest minor to current",
+			fromVersion:              "1.4.0", // renovate: datasource=docker packageName=kong/gateway-operator-oss
+			upgradeToCurrent:         true,
+			upgradeToEffectiveSemver: "1.5.0",
```
pmalek, line 146: Could we add a comment here that this should be the next version to be released to get the charts templates rendered?

czeslavo: Added comments. 👍

pmalek (approved):

pmalek:
Just a note: this should be fixed via https://github.com/Kong/charts/pull/1246 (chart version 0.4.9) as well (defaulting the `effectiveSemver` to `.Chart.AppVersion`) but it's good to merge this PR as well as we want to test against effective semver that ahead of the current version to test chart rendering 👍

PR #1161, opened 2025-02-17
fix(aigateway): check reference grant aigateway credential

### LINKED ISSUE #1150: `cloudProviderCredentials` in `AIGateway` can reference AI cloud credentials in other namespace without permission
## Current Behavior

The `spec.cloudProviderCredentials` can reference `Secret`s in another namespace as AI cloud credential. Currently the `AIGateway` controller do not check the permission of reference, so users can use AI cloud credentials in other namesapces without permission.

## Expected Behavior

User cannot reference credentials in other namespaces unless granted by `ReferenceGrant`.

## Steps To Reproduce

1. Deploy KGO with AI gateway controller enabled
2. Create a `Secret` with an `openai` key in namespace `ns1` 
3. Create an `AIGateway` in namespace `ns2` referencing the `Secret` in namespace `ns1`
4. a `KongPlugin` that is using the AI cloud credential from secret in `ns1` is created in `ns2`.

## Operator Version
1.4.2

randmonkey:
Checks if `AIGateway` using a `Secret` as AI cloud provider credential is allowed by `ReferenceGrant`.
**Which issue this PR fixes**

Fixes #1150

This may break the current usage of using `Secret` in the different namespace with `AIGateway`. So I listed it in `Breaking Changes` in the CHANGELOG.
**PR Readiness Checklist**:

### DIFF
--- a/controller/pkg/secrets/ref/ref.go
+++ b/controller/pkg/secrets/ref/ref.go
@@ -12,6 +12,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
 	gatewayv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+
+	"github.com/kong/gateway-operator/pkg/utils/kubernetes"
 )
 
 // Predicates to filter only the ReferenceGrants that allow a Gateway cross-namespace reference.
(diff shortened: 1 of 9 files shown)


controller/specialized/aigateway_controller_reconciler_owned_resources.go
```diff
@@ -194,6 +197,15 @@ func (r *AIGatewayReconciler) configurePlugins(
 	if aiGateway.Spec.CloudProviderCredentials.Namespace != nil {
 		credentialSecretNamespace = *aiGateway.Spec.CloudProviderCredentials.Namespace
 	}
+
+	// check if referencing the credential secret is allowed by referencegrants.
+	if !r.secretReferenceAllowedByReferenceGrants(ctx, logger, aiGateway, credentialSecretNamespace, credentialSecretName) {
+		log.Info(logger, "Referencing Secret is not allowed by ReferenceGrants",
+			"secret_namespace", credentialSecretNamespace, "secret_name", credentialSecretName)
+		return false, fmt.Errorf("Referencing Secret %s/%s is not allowed by ReferenceGrants",
```
czeslavo:
```suggestion
		return false, fmt.Errorf("referencing Secret %s/%s is not allowed by ReferenceGrants",
```

controller/specialized/aigateway_controller_reconciler_owned_resources.go
```diff
@@ -194,6 +197,15 @@ func (r *AIGatewayReconciler) configurePlugins(
 	if aiGateway.Spec.CloudProviderCredentials.Namespace != nil {
 		credentialSecretNamespace = *aiGateway.Spec.CloudProviderCredentials.Namespace
 	}
+
+	// check if referencing the credential secret is allowed by referencegrants.
+	if !r.secretReferenceAllowedByReferenceGrants(ctx, logger, aiGateway, credentialSecretNamespace, credentialSecretName) {
+		log.Info(logger, "Referencing Secret is not allowed by ReferenceGrants",
+			"secret_namespace", credentialSecretNamespace, "secret_name", credentialSecretName)
```
czeslavo: Given we're returning the error from here, I don't think a separate `Info` log is needed. The controller runtime will log the error, right?
```suggestion
```

pkg/utils/kubernetes/referencegrant.go
```diff
@@ -0,0 +1,67 @@
...
+	// Same namespace is always allowed.
+	if from.Namespace == gatewayv1beta1.Namespace(targetNamespace) {
+		return true, nil
+	}
+	referenceGrantList := gatewayv1beta1.ReferenceGrantList{}
+	namespacedClient := client.NewNamespacedClient(cl, targetNamespace)
+	err := namespacedClient.List(
+		ctx,
+		&referenceGrantList,
+	)
```
czeslavo: I believe we could limit the scope of listing by indexing the reference grants by `from` entries. How about leaving a note here that we could improve it?

pkg/utils/kubernetes/referencegrant.go
```diff
@@ -0,0 +1,67 @@
...
+	"context"
+
+	"github.com/samber/lo"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	gatewayv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+)
+
+// AllowedByReferenceGrants checks if the reference from the input `from` to the object(s)
+// in namespace `targetNamespace` with group, kind, name given in the input `to` is allowed by any ReferenceGrant.
+func AllowedByReferenceGrants(
```
czeslavo, line 13: Please consider reusing this new generic function in https://github.com/Kong/gateway-operator/blob/3d5b5061ad25cba0d261e75bc88a6d75857d0a9d/controller/pkg/secrets/ref/ref.go#L75 to limit logic duplication.

czeslavo (commented):
LGTM in general, just some nits 👍

pkg/utils/kubernetes/referencegrant.go
randmonkey: :+1:  I think it would make code clearer and more extensible if we refactor `CheckReferenceGrantForSecret` to call `AllowedByReferenceGrants` here. There will be more usage of checking reference permissions by `ReferenceGrant` between other kinds of resources.

controller/specialized/aigateway_controller_reconciler_owned_resources.go
randmonkey: Done.

pkg/utils/kubernetes/referencegrant.go
randmonkey: Added a `// TODO` comment.

controller/pkg/secrets/ref/ref.go
```diff
@@ -82,44 +84,25 @@ func CheckReferenceGrantForSecret(
...
+			Namespace: gatewayv1.Namespace(fromObj.GetNamespace()),
+		},
+		string(*secretRef.Namespace),
+		gatewayv1beta1.ReferenceGrantTo{
+			Kind: "Secret",
+			Name: lo.ToPtr(secretRef.Name),
+		},
+	)
+	if err != nil {
+		return "", false, fmt.Errorf("Failed to check if Secret  %s/%s is allowed by ReferenceGrants: %w",
```
czeslavo:
```suggestion
		return "", false, fmt.Errorf("failed to check if Secret  %s/%s is allowed by ReferenceGrants: %w",
```

czeslavo (approved):

PR #1176, opened 2025-02-18
fix(konnect): fix handling conflicts

pmalek:

- Some conflict errors coming from Konnect contain yet unhandled error message: `(type: unique) constraint failed`. This PR fixes that.

  This error happens when e.g. a route tries to get created with another route already existing in Konnect with the same name.

- This also handles "conflicts" when creating certificates: something changed (not sure what at this stage) and we've started to notice errors in konnect integration tests (documented in #1180). This PR adds a workaround for this by first listing the matching certificates using the UID matching tag before it attempts to create the certificate.

  A long term solution can be implemented as part of #1180.

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:

### DIFF
--- a/controller/konnect/ops/ops_errors.go
+++ b/controller/konnect/ops/ops_errors.go
@@ -196,24 +196,48 @@ func ErrorIsCreateConflict(err error) bool {
 }
 
 // SDKErrorIsConflict returns true if the provided SDKError indicates a conflict.
+// We currently handle two codes (as mapped in
+// https://grpc.io/docs/guides/status-codes/#the-full-list-of-status-codes)
+// from SDK error:
+// - 3: INVALID_ARGUMENT
+// - 6: ALREADY_EXISTS
+//
+// Exemplary body:
+//
+//	{
+//	"code": 3,
+//	"message": "name (type: unique) constraint failed",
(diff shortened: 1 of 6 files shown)


controller/konnect/ops/ops_errors.go
```diff
@@ -202,14 +202,20 @@ func SDKErrorIsConflict(sdkError *sdkkonnecterrs.SDKError) bool {
...
+	)
 
-	switch sdkErrorBody.Code {
-	case 3, 6:
-		return true
+	if sdkErrorBody.Message == dataConstraintMesasge ||
+		strings.Contains(sdkErrorBody.Message, typeUniqueConstraintFailed) {
+
+		switch sdkErrorBody.Code {
+		case 3, 6:
```
randmonkey: Could we add some comments (better if we can provide some references) about the error codes? `case 3, 6` looks too magic.

pmalek: Good point, added more docs and exemplary body in 28eaa105fd7150a446e8c8d4d35f11c092df1bae

PTAL

mlavacca (approved):

PR #1232, opened 2025-02-25
ci: update workflow to use ubuntu-24.04-arm for tests

fahedouch:

GHA now provides ARM runners for free

https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

**Which issue this PR fixes**
- set the runner version to avoid unexpected breaking changes
- trying to accelerate GHA CI

**PR Readiness Checklist**:

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -28,13 +28,13 @@ permissions:
 
 jobs:
   ensure-actions-sha-pin:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
     steps:
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
     - uses: zgosalvez/github-actions-ensure-sha-pinned-actions@25ed13d0628a1601b4b44048e63cc4328ed03633 # v3.0.22
 
   ossf-scorecard:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


fahedouch:
Hi Folks, need CI workflows approval to validate my changes please :)

pmalek:
Hi @fahedouch, 

Thanks for the PR. This is quite nice and we'd definitely want to leverage that (at some point). I'm worried though that this will require a bit more work as:

- We don't just want to change to ARM environment in each workflow. Most likely we'd like to leverage both arm64 and x86_64 in build and/or test workflows.
- The docker build workflows might need some adjustments as I believe they have been relying on the fact that they run on x86_64 and emulate arm64 in build

There might be more that I'm not aware of as of now but these should be enough to pause and consider how we'd like to leverage that capability.

fahedouch:
Hi @pmalek,

thank you for your feedback.

> Most likely we'd like to leverage both arm64 and x86_64 in build and/or test workflows.

> The docker build workflows might need some adjustments as I believe they have been relying on the fact that they run on x86_64 and emulate arm64 in build

LGTM for both architectures. I suggest enabling double architecture just for test workflows for the moment, just to see the behavior, and then expanding the scope for the build.

programmer04:
Stale, not for now

PR #1241, opened 2025-02-26
feat(konnect): set KongConsumer conditions when specified credentials are missing

### LINKED ISSUE #1127: Konnect entities: Update `KongConsumer` conditions when secrets specified as `credentials` are missing
## Problem Statement

..

## Proposed Solution

...

## Additional Information

...

## Acceptance Criteria
- [ ] When `Secret` specified as `KongConsumer` `credential` is missing, as a user I can see that reflected in consumer's status conditions

pmalek:

Apart from setting the status conditions in `KongConsumer` this PR also moves the ref handling code for 
- `KonnectAPIAuthConfiguration` to `controller/konnect/reconciler_apiauthref.go`
- `KongConsumer` to `controller/konnect/reconciler_consumerref.go`

**Which issue this PR fixes**

Fixes #1127 

Uses conditions and reasons from https://github.com/Kong/kubernetes-configuration/pull/318

**PR Readiness Checklist**:


controller/konnect/reconciler_generic_type_specific.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+func handleTypeSpecific[
+	T constraints.SupportedKonnectEntityType,
+	TEnt constraints.EntityType[T],
+](
+	ctx context.Context,
+	cl client.Client,
+	ent TEnt,
+) (bool, ctrl.Result, error) {
```
czeslavo, line 30: As this is function is the entry point that is called from the generic reconciler, it would be nice to have named/documented return values here so it's clear what possible combinations mean.

controller/konnect/reconciler_generic_type_specific.go
```diff
@@ -0,0 +1,90 @@
...
+	case *configurationv1.KongConsumer:
+		updated, isProblem = handleKongConsumerSpecific(ctx, cl, e)
+		var (
+			res ctrl.Result
+			err error
+		)
+		if updated {
+			res, err = setProgrammedStatusConditionBasedOnOtherConditions(ctx, cl, e)
+		}
+		return isProblem, res, err
```
czeslavo: If everything is handled in the same case, why not put it all directly to `handleKongConsumerSpecific` and have just `return handleKongConsumerSpecific(ctx, cl, e)`?

controller/konnect/reconciler_generic.go
```diff
@@ -412,6 +409,13 @@ func (r *KonnectEntityReconciler[T, TEnt]) Reconcile(
 		return ctrl.Result{}, nil
 	}
 
+	// Handle type specific operations and stop reconciliation if needed.
+	// This can happen for instance when KongConsumer references credentials Secrets
```
czeslavo, line 413: Does it mean that we will halt the reconciliation at this stage if a `KongConsumer` has multiple credential references and one is invalid (e.g., it doesn't exist)? Couldn't we populate the condition but keep on rolling to create the entity in Konnect and not behave using the "all-or-nothing" strategy? WDYT?

controller/konnect/reconciler_generic_type_specific.go
pmalek: Makes sense, 👍 07eee1af0f3b10d4a9b8c758d0a380d7a3d03bfe

controller/konnect/reconciler_generic_type_specific.go
pmalek: 👍 6b26d786521825ca6c96ff5cac9f0fe6b5eceec4

controller/konnect/reconciler_generic.go
pmalek: > Does it mean that we will halt the reconciliation at this stage if a KongConsumer has multiple credential references and one is invalid (e.g., it doesn't exist)?

Correct. I can be persuaded on this but my thinking was that it'd be better to stop the reconciliation rather than create a consumer without the requested credentials.

czeslavo: From a security standpoint, it shouldn't be a threat to create a consumer without credentials, thus my question. I don't have a strong opinion on that, to be fair. One potential problem we could hit is that if there was an already existing consumer, adding an invalid credential ref would block its updates until the credential references get fixed.

czeslavo: I'm approving the PR and leaving it up to you to decide what is the right behavior here.

czeslavo (approved):

programmer04 (approved):

controller/konnect/reconciler_generic.go
pmalek: > One potential problem we could hit is that if there was an already existing consumer, adding an invalid credential ref would block its updates until the credential references get fixed.

That's a valid concern. I'm actually thinking what should be the result of this:

- reconciliation: whether it should stop or not?
  - if it should stop (current status) I believe it should set both `CredentialSecretRefsValid` and `Programmed` to false. (current status of this PR)
  - if it should continue: I _think_ it should set `Programmed` to `True` and only `CredentialSecretRefsValid` to false. That can make this error hard to spot

I think I'm leaning towards leaving this as is but I'm open to have a discussion on this. My reasoning is that the resource has a bad state and should be fixed ( even though it might be considered usable in this half broken state).

pmalek: I'm going to merge this PR to unblock #1259 but I'm happy to change this if we get to a consensus that this shouldn't stop the reconciliation.

czeslavo: > I think I'm leaning towards leaving this as is but I'm open to have a discussion on this. My reasoning is that the resource has a bad state and should be fixed ( even though it might be considered usable in this half broken state).

I agree with this reasoning. 👍

PR #1242, opened 2025-02-26
feat: Manual `KonnectExtension` certificate provisioning

### LINKED ISSUE #1167: Manual provisioning of DP certificates
## Problem Statement

The `KonnectExtension` API allows the `DataPlane` certificate to be either manually provided by the user or automatically created by KGO. We need to implement a logic to push a manually provided certificate to Konnect via the `KongDataPlaneClientCertificate` resource.

## Proposed Solution

- A new controller watches all the secrets labeled `foo`. For each of such secrets, it creates a new `KongDataPlaneClientCertificate` resource In the cluster, setting the certificate as the `cert` field and the original secret as the owner.
- The `KongDataPlaneClientCertificate` controller pushes the provisioned certificates to Konnect (already implemented)

## Acceptance Criteria
- [ ] As a user, I can have my `DataPlane` certificates automatically provisioned and pushed to Konnect.

mlavacca:

**Which issue this PR fixes**

Fixes #1167 

**PR Readiness Checklist**:

### DIFF
--- a/controller/konnect/handle_status_conditions.go
+++ b/controller/konnect/handle_status_conditions.go
@@ -16,10 +16,10 @@ import (
 )
 
 // handleAPIAuthStatusCondition handles the status conditions for the APIAuthConfiguration reference.
-// The last variadic parameter is related to the optional depending conditions, which is set only in case
-// some check fails and the related condition gets marked as False.
+// The last variadic parameter is related to the optional depending conditions. This parameter is set only when
+// a check fails and the related condition gets marked as False.
 // Example: The KonnectExtension resource has a ready condition that must be set to False if any of the
-// conditions set here are false. Passing it as a depending condition will ensure that the ready condition
+// conditions set here are false. Passing it as a depending condition will ensure that the it
 // is set to False if any of the conditions set here are false.
 func handleAPIAuthStatusCondition[T interface {
 	client.Object
--- a/controller/konnect/handle_status_conditions.go
+++ b/controller/konnect/handle_status_conditions.go
@@ -16,6 +16,11 @@ import (
 )
 
 // handleAPIAuthStatusCondition handles the status conditions for the APIAuthConfiguration reference.
+// The last variadic parameter is related to the optional depending conditions, which is set only in case
+// some check fails and the related condition gets marked as False.
+// Example: The KonnectExtension resource has a ready condition that must be set to False if any of the
+// conditions set here are false. Passing it as a depending condition will ensure that the ready condition
+// is set to False if any of the conditions set here are false.
 func handleAPIAuthStatusCondition[T interface {
 	client.Object
 	k8sutils.ConditionsAware
@@ -25,32 +30,40 @@ func handleAPIAuthStatusCondition[T interface {
 	ent T,
 	apiAuth konnectv1alpha1.KonnectAPIAuthConfiguration,
 	err error,
+	dependingConditions ...metav1.Condition,
 ) (requeue bool, res ctrl.Result, retErr error) {
+	resolvedRefCondition := metav1.Condition{
+		Type:    konnectv1alpha1.KonnectEntityAPIAuthConfigurationResolvedRefConditionType,
+		Status:  metav1.ConditionTrue,
+		Reason:  konnectv1alpha1.KonnectEntityAPIAuthConfigurationResolvedRefReasonResolvedRef,
+		Message: fmt.Sprintf("KonnectAPIAuthConfiguration reference %s is resolved", client.ObjectKeyFromObject(&apiAuth)),
+	}
 	if err != nil {
+		resolvedRefCondition.Status = metav1.ConditionFalse
 		if k8serrors.IsNotFound(err) {
-			if res, err := patch.StatusWithCondition(
-				ctx, cl, ent,
-				konnectv1alpha1.KonnectEntityAPIAuthConfigurationResolvedRefConditionType,
-				metav1.ConditionFalse,
-				konnectv1alpha1.KonnectEntityAPIAuthConfigurationResolvedRefReasonRefNotFound,
-				fmt.Sprintf("Referenced KonnectAPIAuthConfiguration %s not found", client.ObjectKeyFromObject(&apiAuth)),
+			resolvedRefCondition.Reason = konnectv1alpha1.KonnectEntityAPIAuthConfigurationResolvedRefReasonRefNotFound
+			resolvedRefCondition.Message = fmt.Sprintf("Referenced KonnectAPIAuthConfiguration %s not found", client.ObjectKeyFromObject(&apiAuth))
+			if res, _, err := patch.StatusWithConditions(
+				ctx,
+				cl,
+				ent,
+				append(dependingConditions, resolvedRefCondition)...,
 			); err != nil || !res.IsZero() {
 				return true, ctrl.Result{}, err
 			}
-
 			return true, ctrl.Result{}, nil
 		}
 
-		if res, err := patch.StatusWithCondition(
-			ctx, cl, ent,
-			konnectv1alpha1.KonnectEntityAPIAuthConfigurationResolvedRefConditionType,
-			metav1.ConditionFalse,
-			konnectv1alpha1.KonnectEntityAPIAuthConfigurationResolvedRefReasonRefInvalid,
-			fmt.Sprintf("KonnectAPIAuthConfiguration reference %s is invalid: %v", client.ObjectKeyFromObject(&apiAuth), err),
+		resolvedRefCondition.Reason = konnectv1alpha1.KonnectEntityAPIAuthConfigurationResolvedRefReasonRefInvalid
+		resolvedRefCondition.Message = fmt.Sprintf("KonnectAPIAuthConfiguration reference %s is invalid: %v", client.ObjectKeyFromObject(&apiAuth), err)
+		if res, _, err := patch.StatusWithConditions(
+			ctx,
+			cl,
+			ent,
+			append(dependingConditions, resolvedRefCondition)...,
 		); err != nil || !res.IsZero() {
 			return true, ctrl.Result{}, err
 		}
-
 		return true, ctrl.Result{}, fmt.Errorf("failed to get KonnectAPIAuthConfiguration: %w", err)
 	}
 
@@ -59,16 +72,22 @@ func handleAPIAuthStatusCondition[T interface {
 		cond.Status != metav1.ConditionTrue ||
 		cond.ObservedGeneration != ent.GetGeneration() ||
 		cond.Reason != konnectv1alpha1.KonnectEntityAPIAuthConfigurationResolvedRefReasonResolvedRef {
-		if res, err := patch.StatusWithCondition(
-			ctx, cl, ent,
-			konnectv1alpha1.KonnectEntityAPIAuthConfigurationResolvedRefConditionType,
-			metav1.ConditionTrue,
-			konnectv1alpha1.KonnectEntityAPIAuthConfigurationResolvedRefReasonResolvedRef,
-			fmt.Sprintf("KonnectAPIAuthConfiguration reference %s is resolved", client.ObjectKeyFromObject(&apiAuth)),
+		if res, _, err := patch.StatusWithConditions(
+			ctx,
+			cl,
+			ent,
+			resolvedRefCondition,
 		); err != nil || !res.IsZero() {
 			return true, res, err
 		}
 		return true, ctrl.Result{}, nil
+	}
+
+	apiAuthValidCondition := metav1.Condition{
+		Type:    konnectv1alpha1.KonnectEntityAPIAuthConfigurationValidConditionType,
+		Status:  metav1.ConditionTrue,
+		Reason:  konnectv1alpha1.KonnectEntityAPIAuthConfigurationReasonValid,
+		Message: conditionMessageReferenceKonnectAPIAuthConfigurationValid(client.ObjectKeyFromObject(&apiAuth)),
 	}
 
 	// Check if the referenced APIAuthConfiguration is valid.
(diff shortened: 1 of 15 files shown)


mlavacca:
This PR requires https://github.com/Kong/platform-api/pull/1053 to be merged, as the SDK's `DataplaneClientCertificate` schema differs from the actual one in Konnect.

pkg/utils/konnect/resources/kongdataplaneclientcertificate.go
```diff
@@ -0,0 +1,30 @@
+package resources
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	commonv1alpha1 "github.com/kong/kubernetes-configuration/api/common/v1alpha1"
+	configurationv1alpha1 "github.com/kong/kubernetes-configuration/api/configuration/v1alpha1"
+)
+
+// GenerateKongDataPlaneClientCertificate generates a KongDataPlaneClientCertificate object .
```
pmalek:
```suggestion
// GenerateKongDataPlaneClientCertificate generates a KongDataPlaneClientCertificate object
// using the provided controlPlaneRef.
```

- can we also comment on the cert param here, mentioning that this is the actual cert (its contents)?

pkg/consts/konnect.go
```diff
@@ -11,4 +15,17 @@ const (
...
+	// referenced by KonnectExtension to ensure that the secret is not deleted
+	// when in use by an active KonnectExtension.
+	SecretKonnectExtensionFinalizer = "gateway.konghq.com/secret-in-use"
+)
+
+// -----------------------------------------------------------------------------
+// Consts - Labels
+// -----------------------------------------------------------------------------
+const (
+	// DataPlaneCertificateIDLabelKey is the label key used to store the certificate ID
```
pmalek: Can we elaborate a bit more on the `certificate ID` here? This is the Konnect ID right? If so, I'd suggest to mention it in the comment explicitly.

controller/konnect/konnectextension_controller_utils.go
```diff
@@ -0,0 +1,81 @@
...
+func getCertificateSecret(ctx context.Context, cl client.Client, ext konnectv1alpha1.KonnectExtension) (*corev1.Secret, error) {
+	var certificateSecret corev1.Secret
+	switch *ext.Spec.DataPlaneClientAuth.CertificateSecret.Provisioning {
+	case konnectv1alpha1.ManualSecretProvisioning:
+		// No need to check CertificateSecretRef is nil, as it is enforced at the CRD level.
+		if err := cl.Get(ctx, types.NamespacedName{
+			Namespace: ext.Namespace,
+			Name:      ext.Spec.DataPlaneClientAuth.CertificateSecret.CertificateSecretRef.Name,
+		}, &certificateSecret); err != nil {
+			return &certificateSecret, err
```
pmalek:
```suggestion
			return nil, err
```

controller/konnect/konnectextension_controller_utils.go
```diff
@@ -0,0 +1,81 @@
...
+	}, kgcp)
+	if err != nil {
+		return types.NamespacedName{}, err
+	}
+	return types.NamespacedName{
+		Namespace: kgcp.Namespace,
+		Name:      kgcp.Spec.KonnectConfiguration.APIAuthConfigurationRef.Name,
+		// NOTE: We requeue here to keep enforcing the state of the resource in Konnect.
+		// Konnect does not allow subscribing to changes so we need to keep pushing the
+		// desired state periodically.
```
pmalek: ?

controller/konnect/konnectextension_controller.go
```diff
@@ -201,24 +271,44 @@ func (r *KonnectExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+			readyCondition,
+			apiAuthConfigValidCond,
+		); errStatus != nil || updated || !res.IsZero() {
 			return res, errStatus
 		}
+		log.Debug(logger, "token retrieval failed")
 		return ctrl.Result{}, err
 	}
 
+	log.Debug(logger, "token retrieved")
```
pmalek: Suggestion might not compile but I'd suggest adding this information to the log entry.

```suggestion
	log.Debug(logger, "API token retrieved from KonnectAPIAuthConfiguration", "apiAuthRef", apiAuth.Name)
```

controller/konnect/konnectextension_controller.go
pmalek, line 1: The Reconcile() func is getting rather long and hard to follow, any chance that we could extract some of its bits to functions?

controller/konnect/handle_status_conditions.go
```diff
@@ -25,32 +25,40 @@ func handleAPIAuthStatusCondition[T interface {
 	ent T,
 	apiAuth konnectv1alpha1.KonnectAPIAuthConfiguration,
 	err error,
+	dependingConditions ...metav1.Condition,
```
pmalek, line 33: Can we add a godoc about this parameter so that it's clear what it's used for?

controller/konnect/konnectextension_controller_utils.go
mlavacca: Copy-pasta from the generic reconciler. Deleting it.

controller/konnect/handle_status_conditions.go
mlavacca: Left a comment about it

controller/konnect/konnectextension_controller.go
mlavacca: Let me see what I can do to improve readability 👍🏻

pkg/consts/konnect.go
```diff
@@ -11,4 +15,20 @@ const (
...
+	SecretKonnectExtensionFinalizer = "gateway.konghq.com/secret-in-use"
+)
+
+// -----------------------------------------------------------------------------
+// Consts - Labels
+// -----------------------------------------------------------------------------
+const (
+	// DataPlaneCertificateIDLabelKey is the label key used to store the certificate IDs
+	// associated with the secret resource. Since multiple Konnect Certificates can be
+	// created out a single secret, this label is used to store the certificate ID
```
pmalek:
```suggestion
	// created out of a single secret, this label is used to store the certificate ID
```

controller/konnect/handle_status_conditions.go
```diff
@@ -16,6 +16,11 @@ import (
 )
 
 // handleAPIAuthStatusCondition handles the status conditions for the APIAuthConfiguration reference.
+// The last variadic parameter is related to the optional depending conditions, which is set only in case
+// some check fails and the related condition gets marked as False.
```
pmalek:
```suggestion
// The last variadic parameter is related to the optional depending conditions. This parameter is set only when
// a check fails and the related condition gets marked as False.
```

controller/konnect/handle_status_conditions.go
```diff
@@ -16,6 +16,11 @@ import (
...
+// Example: The KonnectExtension resource has a ready condition that must be set to False if any of the
+// conditions set here are false. Passing it as a depending condition will ensure that the ready condition
+// is set to False if any of the conditions set here are false.
```
pmalek:
```suggestion
// conditions set here are false. Passing it as a depending condition will ensure that the it
// is set to False if any of the conditions set here are false.
```

controller/konnect/konnectextension_controller.go
```diff
@@ -26,31 +27,37 @@ import (
...
 	k8sutils "github.com/kong/gateway-operator/pkg/utils/kubernetes"
 
-	commonv1alpha1 "github.com/kong/kubernetes-configuration/api/common/v1alpha1"
+	configurationv1alpha1 "github.com/kong/kubernetes-configuration/api/configuration/v1alpha1"
 	operatorv1alpha1 "github.com/kong/kubernetes-configuration/api/gateway-operator/v1alpha1"
 	operatorv1beta1 "github.com/kong/kubernetes-configuration/api/gateway-operator/v1beta1"
 	konnectv1alpha1 "github.com/kong/kubernetes-configuration/api/konnect/v1alpha1"
 )
 
+const certificateIDsAnnotationKey = "konnect.konghq.com/certificate-ids"
```
pmalek: - Can we add a godoc for this to describe where/when this annotation is used?
- BTW: this has the same value as `DataPlaneCertificateIDLabelKey`, can we reuse the const (add a separate one but use `=` to assign the value)?

controller/konnect/konnectextension_controller_utils.go
```diff
@@ -0,0 +1,198 @@
...
+) (cp *sdkkonnectcomp.ControlPlane, res ctrl.Result, err error) {
+	var konnectCPID string
+
+	// init the controlPlaneRefValidCond with the assumption that the ControlPlaneRef is valid
+	controlPlaneRefValidCond := metav1.Condition{
+		Type:    konnectv1alpha1.ControlPlaneRefValidConditionType,
+		Status:  metav1.ConditionTrue,
+		Reason:  konnectv1alpha1.ControlPlaneRefReasonValid,
+		Message: "ControlPlaneRef is valid",
+	}
```
pmalek: Nit: this could use the same `var` block

pmalek (commented):
Generally looks good. Left some minor comments mostly around adding godocs.

I'm still a little worried about the size of the `Reconcile()` but I can see that we're pressed for time and I'm OK for merging this in this shape. Perhaps we could set some time aside to consider refactoring this bit of code after the release?

controller/konnect/konnectextension_controller.go
mlavacca: Good catch, these two consts were actually duplicates. I got rid of this one in favor of the already documented in the consts package.

mlavacca:
> Generally looks good. Left some minor comments mostly around adding godocs.
> 
> I'm still a little worried about the size of the `Reconcile()` but I can see that we're pressed for time and I'm OK for merging this in this shape. Perhaps we could set some time aside to consider refactoring this bit of code after the release?

I agree with you. This reconcile function is definitely too long and not in the best possible shape. This is because of the impossibility of using the `KongDataPlaneClientCertificate` entity directly (because of the problem with KonnectID references). To overcome that limitation, this controller takes care of two workflows (which is not exactly a best practice): `KonnectExtension` and `Secret`s for certificates. Once we implement `KonnectIDs` (planned for next cycle), we'll ba able to massively simplify this reconcile and make the controller single purpose (for `KonnectExtension`s only)

pmalek:
> > Generally looks good. Left some minor comments mostly around adding godocs.
> > I'm still a little worried about the size of the `Reconcile()` but I can see that we're pressed for time and I'm OK for merging this in this shape. Perhaps we could set some time aside to consider refactoring this bit of code after the release?
> 
> I agree with you. This reconcile function is definitely too long and not in the best possible shape. This is because of the impossibility of using the `KongDataPlaneClientCertificate` entity directly (because of the problem with KonnectID references). To overcome that limitation, this controller takes care of two workflows (which is not exactly a best practice): `KonnectExtension` and `Secret`s for certificates. Once we implement `KonnectIDs` (planned for next cycle), we'll ba able to massively simplify this reconcile and make the controller single purpose (for `KonnectExtension`s only)

Thanks for the explanation 🙇. Do we have an issue tracking this (that would be blocked on KonnectID support implementation)?

controller/konnect/konnectextension_controller.go
```diff
@@ -326,82 +562,45 @@ func (r *KonnectExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
-	case sdkkonnectcomp.ControlPlaneClusterTypeClusterTypeK8SIngressController:
-		return konnectv1alpha1.ClusterTypeK8sIngressController
-	default:
-		// default never happens as the validation is at the CRD level
-		return ""
-	}
+	log.Debug(logger, "reconciled")
+	return ctrl.Result{
+		RequeueAfter: r.syncPeriod,
+	}, nil
```
pmalek, line 615: Nit: I think we could add a comment here (similar to the one in generic reconciler) stating this is is to enforce the config in Konnect. WDYT?

controller/konnect/konnectextension_controller.go
```diff
@@ -326,82 +562,45 @@ func (r *KonnectExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
-
-	if ext.Spec.KonnectConfiguration != nil {
-		// TODO: handle cross namespace refs when supported.
-		return types.NamespacedName{
-			Namespace: ext.Namespace,
-			Name:      ext.Spec.KonnectConfiguration.APIAuthConfigurationRef.Name,
-		}, nil
+	if updateExtensionStatus {
+		log.Debug(logger, "updating KonnectExtension status")
+		return ctrl.Result{}, r.Client.Status().Update(ctx, &ext)
```
pmalek: Could status update also return a conflict 🤔 ? If so I think we can use `k8serrors.IsConflict(err)` and requeue to prevent returning the error and to just requeue.

mlavacca:
> Thanks for the explanation 🙇. Do we have an issue tracking this (that would be blocked on KonnectID support implementation)?

Just linked the issue, you can find the reference above this comment ⬆️

controller/konnect/konnectextension_controller.go
```diff
@@ -231,90 +317,240 @@ func (r *KonnectExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
-		_, err := patch.StatusWithCondition(
-			ctx, r.Client, &ext,
-			kcfgconsts.ConditionType(konnectv1alpha1.ControlPlaneRefValidConditionType),
-			metav1.ConditionFalse,
-			kcfgconsts.ConditionReason(konnectv1alpha1.ControlPlaneRefReasonInvalid),
-			err.Error(),
-		)
+		certProvisionedCond.Status = metav1.ConditionFalse
+		certProvisionedCond.Reason = konnectv1alpha1.DataPlaneCertificateProvisionedReasonRefNotFound
+		certProvisionedCond.Message = err.Error()
```
pmalek, line 340: I can see this approach is used in several places in this reconciler and I have to say that this is slightly hard to follow: instead of creating the condition whenever it's needed, we create 1 object, and then set some fields on it when it's being used. This is hard to reason about what field that object already has set and requires from the reader to look for the definition of used object somewhere else than the immediate surrounding scope.

I'd suggest to consider using a builder pattern like in [KIC](https://github.com/Kong/kubernetes-ingress-controller/tree/513db87cbf94ce66207f74365b502e8cde841357/internal/util/builder) or in [KGO](https://github.com/Kong/gateway-operator/blob/082464d44557aebe01d56db1801382f421c0f335/controller/pkg/builder/builder.go).

This also has a potential of making this a bit less verbose.

I'm OK not doing this in this PR but I think that would greatly improving readability here.

mlavacca: Yeah, I can see the value in using the builder pattern for conditions. However, this is something completely new, as we don't use said pattern to set conditions in any place, neither in KGO nor KIC. I'd say not to do it right now and maybe evaluate it afterward: https://github.com/Kong/gateway-operator/issues/1298

controller/konnect/konnectextension_controller.go
mlavacca: yep, added an addional check

pmalek (approved):

PR #1250, opened 2025-02-28
feat(konnect): Add watch for secret -> KonnectExtension

### LINKED ISSUE #1210: Enqueue `KonnectExtension` upon events on the refererenced secret
## Problem Statement

`KonnectExtension` references a Kubernetes secret that contains the DP certificate. Events on this kind of secret must trigger reconciliation loops of the `KonnectExtension`.

## Acceptance Criteria
- [ ] When a referenced secret changes, the referencing `KonnectExtension` gets enqueued.

randmonkey:
Add watch for `Secret` in `KonnectExtension` controller. A `Secret` must have the label `konghq.com/konnect-dp-cert` to be watched.
**Which issue this PR fixes**

Fixes #1210 

**PR Readiness Checklist**:

### DIFF
--- a/controller/konnect/index_konnectextension.go
+++ b/controller/konnect/index_konnectextension.go
@@ -9,6 +9,8 @@ import (
 const (
 	// IndexFieldKonnectExtensionOnAPIAuthConfiguration is the index field for KonnectExtension -> APIAuthConfiguration.
 	IndexFieldKonnectExtensionOnAPIAuthConfiguration = "konnectExtensionAPIAuthConfigurationRef"
+	// IndexFieldKonnectExtensionOnSecrets is the index field for KonnectExtension -> Secret.
+	IndexFieldKonnectExtensionOnSecrets = "konnectExtensionSecretRef"
 )
 
 // IndexOptionsForKonnectExtension returns required Index options for KonnectExtension reconciler.
@@ -18,6 +20,11 @@ func IndexOptionsForKonnectExtension() []ReconciliationIndexOption {
 			IndexObject:  &konnectv1alpha1.KonnectExtension{},
 			IndexField:   IndexFieldKonnectExtensionOnAPIAuthConfiguration,
 			ExtractValue: konnectExtensionAPIAuthConfigurationRef,
+		},
+		{
+			IndexObject:  &konnectv1alpha1.KonnectExtension{},
+			IndexField:   IndexFieldKonnectExtensionOnSecrets,
+			ExtractValue: konnectExensionSecertRef,
 		},
 	}
 }
@@ -29,4 +36,18 @@ func konnectExtensionAPIAuthConfigurationRef(object client.Object) []string {
 	}
 
 	return []string{ext.Spec.KonnectConfiguration.APIAuthConfigurationRef.Name}
+}
+
+func konnectExensionSecertRef(obj client.Object) []string {
+	ext, ok := obj.(*konnectv1alpha1.KonnectExtension)
+	if !ok {
+		return nil
+	}
+
+	if ext.Spec.DataPlaneClientAuth == nil ||
+		ext.Spec.DataPlaneClientAuth.CertificateSecret.CertificateSecretRef == nil {
+		return nil
+	}
+
+	return []string{ext.Spec.DataPlaneClientAuth.CertificateSecret.CertificateSecretRef.Name}
 }
--- a/controller/konnect/index_konnectextension.go
+++ b/controller/konnect/index_konnectextension.go
@@ -24,7 +24,7 @@ func IndexOptionsForKonnectExtension() []ReconciliationIndexOption {
 		{
 			IndexObject:  &konnectv1alpha1.KonnectExtension{},
 			IndexField:   IndexFieldKonnectExtensionOnSecrets,
-			ExtractValue: konnectExensionSecertRef,
+			ExtractValue: konnectExtensionSecretRef,
 		},
 	}
 }
@@ -38,7 +38,7 @@ func konnectExtensionAPIAuthConfigurationRef(object client.Object) []string {
 	return []string{ext.Spec.KonnectConfiguration.APIAuthConfigurationRef.Name}
 }
 
-func konnectExensionSecertRef(obj client.Object) []string {
+func konnectExtensionSecretRef(obj client.Object) []string {
 	ext, ok := obj.(*konnectv1alpha1.KonnectExtension)
 	if !ok {
 		return nil
--- a/controller/konnect/konnectextension_controller.go
+++ b/controller/konnect/konnectextension_controller.go
@@ -58,6 +58,7 @@ func (r *KonnectExtensionReconciler) SetupWithManager(ctx context.Context, mgr c
 	ls := metav1.LabelSelector{
 		// A secret ust have `konghq.com/konnect-dp-cert` label to be watched by the controller.
 		// This constraint is added to prevent from watching all secrets which may cause high resource consumption.
+		// TODO: https://github.com/Kong/gateway-operator/issues/1255 set label constraints of `Secret`s on manager level if possible.
 		MatchExpressions: []metav1.LabelSelectorRequirement{
 			{
 				Key:      SecretKonnectDataPlaneCertificateLabel,
--- a/controller/konnect/konnectextension_controller.go
+++ b/controller/konnect/konnectextension_controller.go
@@ -11,10 +11,12 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"github.com/kong/gateway-operator/controller/konnect/ops"
@@ -53,6 +55,21 @@ func NewKonnectExtensionReconciler(
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *KonnectExtensionReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager) error {
+	ls := metav1.LabelSelector{
+		// A secret must has `konghq.com/konnect-dp-cert` label to be watched by the controller.
+		// This constraint is added to prevent from watching all secrets which may cause high resource consumption.
+		MatchExpressions: []metav1.LabelSelectorRequirement{
+			{
+				Key:      SecretKonnectDataPlaneCertificateLabel,
+				Operator: metav1.LabelSelectorOpExists,
+			},
+		},
+	}
+	labelSelectorPredicate, err := predicate.LabelSelectorPredicate(ls)
+	if err != nil {
+		return err
+	}
+
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&konnectv1alpha1.KonnectExtension{}).
 		Watches(
@@ -68,7 +85,16 @@ func (r *KonnectExtensionReconciler) SetupWithManager(ctx context.Context, mgr c
 				),
 			),
 		).
-		// TODO: watch secrets https://github.com/Kong/gateway-operator/issues/1210
+		Watches(
+			&corev1.Secret{},
+			handler.EnqueueRequestsFromMapFunc(
+				konnectExtensionReconcileRequestsForSecret(mgr.GetClient()),
+			),
+			builder.WithPredicates(
+				labelSelectorPredicate,
+				predicate.NewPredicateFuncs(secretUsedByKonnectExtension(mgr.GetClient())),
+			),
+		).
 		Complete(r)
 }
 
--- a/controller/konnect/konnectextension_controller.go
+++ b/controller/konnect/konnectextension_controller.go
@@ -56,7 +56,7 @@ func NewKonnectExtensionReconciler(
 // SetupWithManager sets up the controller with the Manager.
 func (r *KonnectExtensionReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager) error {
 	ls := metav1.LabelSelector{
-		// A secret must has `konghq.com/konnect-dp-cert` label to be watched by the controller.
+		// A secret ust have `konghq.com/konnect-dp-cert` label to be watched by the controller.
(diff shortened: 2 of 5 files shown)


randmonkey:
Unit tests fails because the order of DP labels is changed in the `TestClusterDataPlaneLabelStringFromLabels/multiple_labels` test when code changed. Requires #1252.

controller/konnect/index_konnectextension.go
```diff
@@ -30,3 +37,17 @@ func konnectExtensionAPIAuthConfigurationRef(object client.Object) []string {
 
 	return []string{ext.Spec.KonnectConfiguration.APIAuthConfigurationRef.Name}
 }
+
+func konnectExensionSecertRef(obj client.Object) []string {
```
czeslavo:
```suggestion
func konnectExtensionSecretRef(obj client.Object) []string {
```

controller/konnect/konnectextension_controller.go
```diff
@@ -53,6 +55,21 @@ func NewKonnectExtensionReconciler(
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *KonnectExtensionReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager) error {
+	ls := metav1.LabelSelector{
+		// A secret must has `konghq.com/konnect-dp-cert` label to be watched by the controller.
```
czeslavo:
```suggestion
		// A secret must have `konghq.com/konnect-dp-cert` label to be watched by the controller.
```

controller/konnect/konnectextension_controller.go
```diff
@@ -53,6 +55,21 @@ func NewKonnectExtensionReconciler(
...
+		// This constraint is added to prevent from watching all secrets which may cause high resource consumption.
```
czeslavo, line 60: Please note that filtering on this level (predicates filter events client-side) won't help with reducing resource consumption (we will still cache all Secrets). If we wanted to only watch Secrets with labels, we'd have to modify the manager's `cache.Options` by defining `ByObject` explicitly to do so.

https://github.com/kubernetes-sigs/controller-runtime/issues/244 for reference.

Given this, I think we should either:
1. Consider modifying the `cache.Options` as suggested to not filter by a label (this has a wider impact - we'd have to ensure all Secrets we handle in the operator are labeled as this is a manager global setting - I'm not sure if this could be done without a breaking change).
2. Rely only on indexing KonnectExtensions by a Secret and a predicate that will ensure the reconciled Secret is referred by any KonnectExtension. This way we do not require the label on the Secret and make this more user-friendly, while still preserving the same performance that we'd have with the client-side label selector predicate.

randmonkey: (1) is not feasible given that the `Manager` is shared by all the controllers. But how could (2) be done? Is it supported to watch only a finite list of `Secret`s?

czeslavo: In 2) I meant that given that we already cache all of the Secrets from the cluster (that happens if we do not modify the manager's `cache.Options`), we may as well not require the label in the controller's predicates and rely only on the references from KonnectExtensions only.

czeslavo: > (1) is not feasible given that the Manager is shared by all the controllers

If all of the Secrets we want to reconcile would have a `konghq.com/credential` label, we could make this on the manager level and not cache other Secrets.

randmonkey: So Let's create an issue and discuss whether we should implement (1) to restrain all watched `Secret`s in KGO to have a certain label. I'll do (2) to remove the constraint in this PR first.

controller/konnect/konnectextension_controller.go
```diff
@@ -68,7 +85,16 @@ func (r *KonnectExtensionReconciler) SetupWithManager(ctx context.Context, mgr c
...
-		// TODO: watch secrets https://github.com/Kong/gateway-operator/issues/1210
+		Watches(
+			&corev1.Secret{},
+			handler.EnqueueRequestsFromMapFunc(
+				konnectExtensionReconcileRequestsForSecret(mgr.GetClient()),
+			),
+			builder.WithPredicates(
+				labelSelectorPredicate,
+				predicate.NewPredicateFuncs(secretUsedByKonnectExtension(mgr.GetClient())),
+			),
```
czeslavo, line 96: That's essentially what I proposed in 1) above.

```suggestion
			&corev1.Secret{},
			handler.EnqueueRequestsFromMapFunc(
				konnectExtensionReconcileRequestsForSecret(mgr.GetClient()),
			),
			builder.WithPredicates(				predicate.NewPredicateFuncs(secretUsedByKonnectExtension(mgr.GetClient())),
			),
```

controller/konnect/konnectextension_controller.go
```diff
@@ -68,7 +85,16 @@ func (r *KonnectExtensionReconciler) SetupWithManager(ctx context.Context, mgr c
...
 			),
 		).
-		// TODO: watch secrets https://github.com/Kong/gateway-operator/issues/1210
+		Watches(
+			&corev1.Secret{},
+			handler.EnqueueRequestsFromMapFunc(
+				konnectExtensionReconcileRequestsForSecret(mgr.GetClient()),
+			),
+			builder.WithPredicates(
+				labelSelectorPredicate,
```
czeslavo, line 95: That's essentially what I proposed in 2) above.


```suggestion
```

controller/konnect/konnectextension_controller.go
randmonkey: For the topic of adding label selector of watched secrets, please move to https://github.com/Kong/gateway-operator/issues/1255. Removed the label selector predicates here.

pmalek: Sorry to be late to comment here but I'd argue that we keep the label requirement here.

The reason for this is that we can always relax the requirements later - which would not be a breaking change - but we cannot make the more strict - which would be considered a breaking change.

This label can be then used when we want to address #1255.

pmalek (changes requested):

controller/konnect/konnectextension_controller.go
czeslavo: > The reason for this is that we can always relax the requirements later - which would not be a breaking change - but we cannot make the more strict - which would be considered a breaking change.

That's a good point. If we implement https://github.com/Kong/gateway-operator/issues/1255, we'll need a label anyway, so it's better to require it from the user now. Let's just use the same label key as we use for different Secrets (which is `konghq.com/credential`) so that we can create a label selector that verifies its existence (AFAIK, you cannot have ORed requirements for labels).

pmalek: > AFAIK, you cannot have ORed requirements for labels

I was pretty sure that this was possible but apparently it's not: https://github.com/kubernetes/kubernetes/issues/90549. Good call out. That will make the #1255 need to rely on a user provided label (that they want to use for all resources, per type or not) as we already use several different labels for secrets.

czeslavo: So wouldn't the conclusion be we don't need to add this label now, as it won't be necessary for #1255? 🤔

controller/konnect/konnectextension_controller.go
```diff
@@ -53,6 +55,22 @@ func NewKonnectExtensionReconciler(
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *KonnectExtensionReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager) error {
+	ls := metav1.LabelSelector{
+		// A secret ust have `konghq.com/konnect-dp-cert` label to be watched by the controller.
```
pmalek:
```suggestion
		// A secret must have `konghq.com/konnect-dp-cert` label to be watched by the controller.
```

controller/konnect/konnectextension_controller_secrets.go
```diff
@@ -0,0 +1,79 @@
...
+			return ke.Spec.DataPlaneClientAuth != nil &&
+				ke.Spec.DataPlaneClientAuth.CertificateSecret.CertificateSecretRef != nil &&
+				ke.Spec.DataPlaneClientAuth.CertificateSecret.CertificateSecretRef.Name == obj.GetName()
+		})
+	}
+}
+
+func konnectExtensionReconcileRequestsForSecret(cl client.Client) func(context.Context, client.Object) []reconcile.Request {
+	return func(ctx context.Context, obj client.Object) []reconcile.Request {
+		konnectExtensions, err := listKonnectExtensionsBySecret(ctx, cl, obj)
```
pmalek: Can we type assert at the top of this enqueue func and use the typed `Secret` throughout this func?

```suggestion
	return func(ctx context.Context, obj client.Object) []reconcile.Request {
		s, ok := obj.(*corev1.Secret)
		if !ok {
			return nil, fmt.Errorf("expect object to be secret, actually %T", obj)
		}
		konnectExtensions, err := listKonnectExtensionsBySecret(ctx, cl, s)
```

controller/konnect/konnectextension_controller_secrets.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+		return lo.ContainsBy(konnectExtensions, func(ke konnectv1alpha1.KonnectExtension) bool {
+			return ke.Spec.DataPlaneClientAuth != nil &&
+				ke.Spec.DataPlaneClientAuth.CertificateSecret.CertificateSecretRef != nil &&
+				ke.Spec.DataPlaneClientAuth.CertificateSecret.CertificateSecretRef.Name == obj.GetName()
+		})
+	}
+}
+
+func konnectExtensionReconcileRequestsForSecret(cl client.Client) func(context.Context, client.Object) []reconcile.Request {
```
pmalek: Perhaps this would be a slightly clearer name?

```suggestion
func enqueueKonnectExtensionsForSecret(cl client.Client) func(context.Context, client.Object) []reconcile.Request {
```

controller/konnect/konnectextension_controller.go
```diff
@@ -68,7 +86,16 @@ func (r *KonnectExtensionReconciler) SetupWithManager(ctx context.Context, mgr c
...
 		).
-		// TODO: watch secrets https://github.com/Kong/gateway-operator/issues/1210
+		Watches(
+			&corev1.Secret{},
+			handler.EnqueueRequestsFromMapFunc(
+				konnectExtensionReconcileRequestsForSecret(mgr.GetClient()),
+			),
+			builder.WithPredicates(
+				labelSelectorPredicate,
+				predicate.NewPredicateFuncs(secretUsedByKonnectExtension(mgr.GetClient())),
```
pmalek: I believe that this is unnecessary given that we use the same information in the enqueue func.

We're basically checking (and listing) the same thing twice:

- first on the `KonnectExtension` listing using the index that is based on the `.Spec.DataPlaneClientAuth.CertificateSecret.CertificateSecretRef.Name`
- secondly in the `Secret` predicate listing the `KonnectExtension`s using the very same index and checking the `.Spec.DataPlaneClientAuth.CertificateSecret.CertificateSecretRef.Name` field.

I'd suggest to drop this to prevent the unnecessary work. We can add a comment if we'd like to be verbose about this here. (That's where the secret label comes in handy to limit the amount of secrets we'll perform the listing for).

```suggestion
```

controller/konnect/konnectextension_controller.go
pmalek: IMHO it's better to require this now and then possibly relax this in the future than the other way around and perform a breaking change.

Additionally, this is useful because of https://github.com/Kong/gateway-operator/pull/1250#discussion_r1975208365

czeslavo: My main point is that if a user refers to a Secret from KonnectExtension, that should be enough for us to decide whether we should reconcile it. I think that https://github.com/Kong/gateway-operator/pull/1250/files/d1cfd1daed797851f03b51962a114fd948830b1b#r1975208365 boils down to deciding whether we want to reconcile a secret based on its label or it being referenced by KonnectExtension (it doesn't make much sense to do both at once, I agree). 

I'm ok with either. If you think any potential use cases will require this label in the future, let's add it now.

pmalek: > If you think any potential use cases will require this label in the future, let's add it now.

I think that might be the case and I'd prefer to be on the safe side and add it.

> I think that https://github.com/Kong/gateway-operator/pull/1250/files/d1cfd1daed797851f03b51962a114fd948830b1b#r1975208365 boils down to deciding whether we want to reconcile a secret based on its label or it being referenced by KonnectExtension

The builder watch criteria work so that a provided object is first matched (using predicates) and then it's handled in the provided enqueue func. If we use the label we can just use that as the predicate, and then enqueue func will further filter down the `Secret`s by listing the `KonnectExtensions` and comparing its spec with the `Secret` name.

This is where the label comes in useful. Without it we'd be listing more `KonnectExtension`s than necessary (albeit using an index but still).

controller/konnect/konnectextension_controller.go
randmonkey: Done.

controller/konnect/konnectextension_controller_secrets.go
```diff
@@ -0,0 +1,68 @@
...
+func enqueueKonnectExtensionsForSecret(cl client.Client) func(context.Context, client.Object) []reconcile.Request {
+	return func(ctx context.Context, obj client.Object) []reconcile.Request {
+		_, ok := obj.(*corev1.Secret)
+		if !ok {
+			return nil
+		}
+		konnectExtensions, err := listKonnectExtensionsBySecret(ctx, cl, obj)
+		if err != nil {
+			return nil
+		}
```
pmalek: Given that we already perform the type assert we could use the secret and get rid of the type assert inside the `listKonnectExtensionsBySecret`, right?

```suggestion
		s, ok := obj.(*corev1.Secret)
		if !ok {
			return nil
		}
		konnectExtensions, err := listKonnectExtensionsBySecret(ctx, cl, s)
		if err != nil {
			return nil
		}
```

controller/konnect/konnectextension_controller_secrets.go
```diff
@@ -0,0 +1,68 @@
...
+	// SecretKonnectDataPlaneCertificateLabel is the label to mark that the secret is used as a Konnect DP certificate.
+	// A secret must have the label to be watched by the KonnectExtension reconciler.
+	SecretKonnectDataPlaneCertificateLabel = "konghq.com/konnect-dp-cert" //nolint:gosec
+)
+
+func listKonnectExtensionsBySecret(ctx context.Context, cl client.Client, obj client.Object) ([]konnectv1alpha1.KonnectExtension, error) {
+	s, ok := obj.(*corev1.Secret)
+	if !ok {
+		return nil, fmt.Errorf("expect object to be secret, actually %T", obj)
+	}
```
pmalek:
```suggestion
func listKonnectExtensionsBySecret(ctx context.Context, cl client.Client, s *corev1.Secret) ([]konnectv1alpha1.KonnectExtension, error) {
```

pmalek (dismissed):

mlavacca (dismissed):

czeslavo (dismissed):

czeslavo (approved):

PR #1253, opened 2025-02-28
feat: enable status/ready endpoint in Konnect

### LINKED ISSUE #1235: Kong in Konnect readiness probe
## Problem Statement

When Kong is configured in Konnect, we need a readiness endpoint to check the configuration with the Konnect ControlPlane and ensure it has been properly set up. In case some problem is raised (e.g. an invalid certificate), we need to mark the pod as not ready.

## additional information

In case no endpoint is available at the gateway level, check whether we can implement a temporary workaround.

## Acceptance Criteria

- [ ] In the context of Kong in Konnect, the DataPlane pod gets ready only if the communication with the Konnect ControlPlane is successful.

mlavacca:

**Which issue this PR fixes**

Fixes #1235 

**PR Readiness Checklist**:

### DIFF
--- a/controller/dataplane/controller.go
+++ b/controller/dataplane/controller.go
@@ -181,6 +181,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		labelSelectorFromDataPlaneStatusSelectorDeploymentOpt(dataplane),
 	}
 
+	// if the dataplane is configured with Konnect, the status/ready endpoint should be set as the readiness probe.
+	if _, konnectApplied := k8sutils.GetCondition(consts.KonnectExtensionAppliedType, dataplane); konnectApplied {
+		deploymentOpts = append(deploymentOpts, statusReadyEndpointDeploymentOpt(dataplane))
+	}
+
 	log.Trace(logger, "ensuring generation of deployment configuration for KongPluginInstallations configured for DataPlane")
 	kpisForDeployment, requeue, err := ensureMappedConfigMapToKongPluginInstallationForDataPlane(ctx, logger, r.Client, dataplane)
 	if err != nil {
@@ -253,6 +258,14 @@ func labelSelectorFromDataPlaneStatusSelectorDeploymentOpt(dataplane *operatorv1
 			d.Labels[consts.OperatorLabelSelector] = dataplane.Status.Selector
 			d.Spec.Selector.MatchLabels[consts.OperatorLabelSelector] = dataplane.Status.Selector
 			d.Spec.Template.Labels[consts.OperatorLabelSelector] = dataplane.Status.Selector
+		}
+	}
+}
+
+func statusReadyEndpointDeploymentOpt(_ *operatorv1beta1.DataPlane) func(s *appsv1.Deployment) {
+	return func(d *appsv1.Deployment) {
+		if container := k8sutils.GetPodContainerByName(&d.Spec.Template.Spec, consts.DataPlaneProxyContainerName); container != nil {
+			container.ReadinessProbe = k8sresources.GenerateDataPlaneReadinessProbe(consts.DataPlaneStatusReadyEndpoint)
 		}
 	}
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,9 @@
 - Added support for `KonnectCloudGatewayNetwork` CRD which can manage Konnect
   Cloud Gateway Network entities.
   [#1136](https://github.com/Kong/gateway-operator/pull/1136)
+- When the `DataPlane` is configured in Konnect, the `/status/ready` endpoint
+  is set as the readiness probe.
+  [#1235](https://github.com/Kong/gateway-operator/pull/1253)
 
 ### Changed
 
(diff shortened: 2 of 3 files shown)


controller/dataplane/controller.go
```diff
@@ -181,6 +181,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		labelSelectorFromDataPlaneStatusSelectorDeploymentOpt(dataplane),
 	}
 
+	// if the dataplane is configured with Konnect, the status/ready endpoint should be set as the readiness probe.
+	if _, konnectApplied := k8sutils.GetCondition(consts.KonnectExtensionAppliedType, dataplane); konnectApplied {
+		deploymentOpts = append(deploymentOpts, statusReadyEndpointDeploymentOpt(dataplane))
+	}
```
pmalek, line 187: I believe this will also have to be added to the BG controller deployment options at https://github.com/Kong/gateway-operator/blob/292b536f7ff2f43eac7ce99b8668fee6180f4efa/controller/dataplane/bluegreen_controller.go#L515-L517

mlavacca: Good catch. See [1f235cb](https://github.com/Kong/gateway-operator/pull/1253/commits/1f235cb82313414ecfe08fd79a5ebeb65f351f44)

sandromodarelli:
since the readiness endpoint is always the same despite the kong mode, we should fix the generic behavior instead of acting with conditions like `if konnect`. IMO it worth to fix the default readiness probe here https://github.com/Kong/gateway-operator/blob/main/pkg/utils/kubernetes/resources/deployments.go#L340 using `consts.DataPlaneStatusReadyEndpoint` instead of `consts.DataPlaneStatusEndpoint`

mlavacca:
> since the readiness endpoint is always the same despite the kong mode, we should fix the generic behavior instead of acting with conditions like `if konnect`. IMO it worth to fix the default readiness probe here https://github.com/Kong/gateway-operator/blob/main/pkg/utils/kubernetes/resources/deployments.go#L340 using `consts.DataPlaneStatusReadyEndpoint` instead of `consts.DataPlaneStatusEndpoint`

I pretty much agree with you, but I think that we kept the  `/status` endpoint as the default readiness probe to make the DataPlane ready in case it's deployed in isolation and not configured by any `ControlPlane`. If you create just a `DataPlane` without any `ControlPlane` and set the `/status/ready` endpoint, the `DataPlane` pod will never get ready, as that endpoint expects an actual configuration to be loaded. For example, when the `DataPlane` is created out of a Gateway, we set the correct `/status/ready` endpoint as well: https://github.com/Kong/gateway-operator/blob/441b0e769461861be803dafb693ab016cf0d3acd/controller/gateway/controller.go#L661-L663

czeslavo:
Maybe we could switch to `/status/ready` as default, as it's now used in most use cases (when Konnect CP is configured and in Gateway-managed DPs), and have a special case only for standalone DPs?

mlavacca:
> Maybe we could switch to `/status/ready` as default, as it's now used in most use cases (when Konnect CP is configured and in Gateway-managed DPs), and have a special case only for standalone DPs?

It would make sense from a logical point of view. However, the complexity lies in the fact that the dataplane controller does not have the whole context: the Gateway controller has the whole context and knows there is a cp/dp pair. The DP controller, instead, is not able to determine if the dataplane is configured by a controlplane or running in isolation.

This is something that could be changed and properly designed, but I'm afraid it's far from being a small patch.

czeslavo:
> It would make sense from a logical point of view. However, the complexity lies in the fact that the dataplane controller does not have the whole context: the Gateway controller has the whole context and knows there is a cp/dp pair. The DP controller, instead, is not able to determine if the dataplane is configured by a controlplane or running in isolation.
> 
> This is something that could be changed and properly designed, but I'm afraid it's far from being a small patch.

Wouldn't checking if a DP has owner references be enough (no owners - a standalone DP)?

mlavacca:
> > It would make sense from a logical point of view. However, the complexity lies in the fact that the dataplane controller does not have the whole context: the Gateway controller has the whole context and knows there is a cp/dp pair. The DP controller, instead, is not able to determine if the dataplane is configured by a controlplane or running in isolation.
> > This is something that could be changed and properly designed, but I'm afraid it's far from being a small patch.
> 
> Wouldn't checking if a DP has owner references be enough (no owners - a standalone DP)?

It would probably work, sure, or we could check the labels - but both approaches sound a bit hacky tbh. And since this wouldn't change anything from the behavioral point of view, we would introduce hackiness just for the sake of code "structure". I am just not sure it is worth it.

mlavacca:
I think that it could make sense, if we want to have a proper definition of the environment, to do some effort in the API itself, and work on some proper design rather than using ownerRefs or labels to check the parents

czeslavo:
> since this wouldn't change anything from the behavioral point of view, we would introduce hackiness just for the sake of code "structure". I am just not sure it is worth it.

Yup, 💯  agree. It's not the time :)

pmalek (approved):

sandromodarelli:
> I think that it could make sense, if we want to have a proper definition of the environment, to do some effort in the API itself, and work on some proper design rather than using ownerRefs or labels to check the parents

i agree with @czeslavo that now is not the right time, but let's plan a discussion after next release to review ControlPlane and DataPlane APIs struct. ship it 🚀 LGTM

PR #1259, opened 2025-02-28
fix: delete owner relationship between KongConsumers and credentials

pmalek:

Requires: #1241

**Which issue this PR fixes**

Part of #788

**PR Readiness Checklist**:

### DIFF
--- a/controller/konnect/reconciler_consumerref.go
+++ b/controller/konnect/reconciler_consumerref.go
@@ -107,8 +107,8 @@ func handleKongConsumerRef[T constraints.SupportedKonnectEntityType, TEnt constr
 		ctrllog.FromContext(ctx).Info("failed to check if KongConsumer has owner reference", "error", err)
 	} else if ok {
 		// Make sure there is no ownership between the consumer and the entity by removing
-		// a reference if found. Such references may have been added by older versions 
-		// of KGO (< 1.5). 
+		// a reference if found. Such references may have been added by older versions
+		// of KGO (< 1.5).
 		// TODO: remove this after a couple of minor versions.
 		old := ent.DeepCopyObject().(TEnt)
 		if err := controllerutil.RemoveOwnerReference(&consumer, ent, cl.Scheme()); err != nil {
--- a/controller/konnect/reconciler_consumerref.go
+++ b/controller/konnect/reconciler_consumerref.go
@@ -103,6 +103,21 @@ func handleKongConsumerRef[T constraints.SupportedKonnectEntityType, TEnt constr
 		return ctrl.Result{Requeue: true}, nil
 	}
 
+	if ok, err := controllerutil.HasOwnerReference(consumer.OwnerReferences, ent, cl.Scheme()); err != nil {
+		ctrllog.FromContext(ctx).Info("failed to check if KongConsumer has owner reference", "error", err)
+	} else if ok {
+		old := ent.DeepCopyObject().(TEnt)
+		if err := controllerutil.RemoveOwnerReference(&consumer, ent, cl.Scheme()); err != nil {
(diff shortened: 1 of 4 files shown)


randmonkey (dismissed):

controller/konnect/reconciler_consumerref.go
```diff
@@ -103,17 +103,6 @@ func handleKongConsumerRef[T constraints.SupportedKonnectEntityType, TEnt constr
 		return ctrl.Result{Requeue: true}, nil
 	}
 
-	old := ent.DeepCopyObject().(TEnt)
-	if err := controllerutil.SetOwnerReference(&consumer, ent, cl.Scheme(), controllerutil.WithBlockOwnerDeletion(true)); err != nil {
```
czeslavo, line 107: Have we considered what our behavior should be if there are already existing credentials under consumer ownership (assigned by KGO 1.4)? I understand that in the current approach, we accept that objects reconciled by KGO 1.4 will end with cascade delete if a consumer gets deleted, right?

WDYT about an alternative where we migrate the already reconciled objects by removing their owner references?

pmalek: That's a good point. I'll address that. I think that simply calling `controllerutil.RemoveOwnerReference()` would do. WDYT?

controller/konnect/reconciler_consumerref.go
```diff
@@ -103,15 +103,19 @@ func handleKongConsumerRef[T constraints.SupportedKonnectEntityType, TEnt constr
...
+	} else if ok {
+		old := ent.DeepCopyObject().(TEnt)
+		if err := controllerutil.RemoveOwnerReference(&consumer, ent, cl.Scheme()); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to delete owner reference: %w", err)
+		}
+		if err := cl.Patch(ctx, ent, client.MergeFrom(old)); err != nil {
+			if k8serrors.IsConflict(err) {
+				return ctrl.Result{Requeue: true}, nil
+			}
+			return ctrl.Result{}, fmt.Errorf("failed to update status: %w", err)
```
czeslavo:
```suggestion
			return ctrl.Result{}, fmt.Errorf("failed to remove owner reference: %w", err)
```

controller/konnect/reconciler_consumerref.go
```diff
@@ -103,15 +103,19 @@ func handleKongConsumerRef[T constraints.SupportedKonnectEntityType, TEnt constr
...
-	if err := controllerutil.SetOwnerReference(&consumer, ent, cl.Scheme(), controllerutil.WithBlockOwnerDeletion(true)); err != nil {
-		return ctrl.Result{}, fmt.Errorf("failed to set owner reference: %w", err)
-	}
-	if err := cl.Patch(ctx, ent, client.MergeFrom(old)); err != nil {
-		if k8serrors.IsConflict(err) {
-			return ctrl.Result{Requeue: true}, nil
+	if ok, err := controllerutil.HasOwnerReference(consumer.OwnerReferences, ent, cl.Scheme()); err != nil {
+		ctrllog.FromContext(ctx).Info("failed to check if KongConsumer has owner reference", "error", err)
+	} else if ok {
+		old := ent.DeepCopyObject().(TEnt)
```
czeslavo, line 113: How about leaving some trace of why we want to do it?
```suggestion
		// Make sure there is no ownership between the consumer and the entity by removing
		// a reference if found. Such references may have been added by older versions 
		// of KGO (< 1.5). 
		// TODO: remove this after a couple of minor versions.
		old := ent.DeepCopyObject().(TEnt)
```

czeslavo (approved):

Issue #874, opened 2024-11-21
Add `KonnectExtension` capabilities to the `ControlPlane`


mlavacca:
## Problem Statement

So far, the `KonnectExtension` has been exclusively used with `DataPlane`. We need to provide the same functionality for `ControlPlane`s.

## Proposed Solution

- In the `ControlPlane` controller, add a logic very similar to [what we already have](https://github.com/Kong/gateway-operator/blob/292b536f7ff2f43eac7ce99b8668fee6180f4efa/controller/dataplane/controller_utils.go#L328-L329) In the `DataPlane` one to customize the underlying deployment with the Konnect-related configuration bits.
- A check must be performed: in case the `KonnectExtension` is not of type `K8SIngressController` the attachment must fail and the `KonnectExtensionApplied` condition must be set accordingly. We may need to introduce a new reason to highlight this kind of problem.

## Acceptance Criteria
- [ ] As a user, I can attach a `KonnectExtension` to a `ControlPlane` resource.

randmonkey:
Blocked by #1145

mlavacca:
Unblocked

Issue #726, opened 2024-10-10
Integration test for DataPlane KonnectExtension


mlavacca:
## Problem Statement

https://github.com/Kong/gateway-operator/pull/714 introduced the logic to patch dataplanes based on the referenced `KonnectExtension`. We do need to add an integration test for such a new feature.

## Acceptance Criteria
- [ ] An integration test for the `DataPlane` `KonnectExtension` feature exists.

mlavacca:
Blocked by #1166, as having automatic secret provisioning will massively help to implement this test.

PR #1272, opened 2025-03-04
test(integration): add integration tests and helper functions for KonnectExtension

randmonkey:

Add helper functions to deploy `KonnectExtension`s and integration test case (not complete yet) for testing `KonnectExtension`.

**Which issue this PR fixes**

Part of #726 

Waiting for #874 to test full function of KonnectExtension.
**PR Readiness Checklist**:

### DIFF
--- a/test/helpers/conditions.go
+++ b/test/helpers/conditions.go
@@ -1,0 +1,26 @@
+package helpers
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/samber/lo"
+
+	"github.com/kong/gateway-operator/pkg/consts"
+	k8sutils "github.com/kong/gateway-operator/pkg/utils/kubernetes"
+)
+
+// CheckAllConditionsTrue returns true if all the conditions with given type in `conditionTypes` are set to `True` in the given resource.
+// If it returns `false`, the second return value contains a message to tell what conditions are not `True`.
+func CheckAllConditionsTrue(resource k8sutils.ConditionsAware, conditionTypes []consts.ConditionType) (bool, string) {
+	failedConditions := make([]string, 0, len(conditionTypes))
+	lo.ForEach(conditionTypes, func(conditionType consts.ConditionType, _ int) {
+		if !k8sutils.HasConditionTrue(conditionType, resource) {
+			failedConditions = append(failedConditions, string(conditionType))
+		}
+	})
+	if len(failedConditions) > 0 {
+		return false, fmt.Sprintf("condition(s) %s not set to True", strings.Join(failedConditions, ", "))
+	}
+	return true, ""
+}
--- a/test/helpers/conditions.go
+++ b/test/helpers/conditions.go
@@ -4,21 +4,21 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/samber/lo"
-
-	"github.com/kong/gateway-operator/pkg/consts"
 	k8sutils "github.com/kong/gateway-operator/pkg/utils/kubernetes"
+
+	kcfgconsts "github.com/kong/kubernetes-configuration/api/common/consts"
 )
 
 // CheckAllConditionsTrue returns true if all the conditions with given type in `conditionTypes` are set to `True` in the given resource.
 // If it returns `false`, the second return value contains a message to tell what conditions are not `True`.
-func CheckAllConditionsTrue(resource k8sutils.ConditionsAware, conditionTypes []consts.ConditionType) (bool, string) {
+func CheckAllConditionsTrue(resource k8sutils.ConditionsAware, conditionTypes []kcfgconsts.ConditionType) (bool, string) {
 	var failedConditions []string
-	lo.ForEach(conditionTypes, func(conditionType consts.ConditionType, _ int) {
+	for _, conditionType := range conditionTypes {
 		if !k8sutils.HasConditionTrue(conditionType, resource) {
 			failedConditions = append(failedConditions, string(conditionType))
 		}
-	})
+	}
+
 	if len(failedConditions) > 0 {
 		return false, fmt.Sprintf("condition(s) %s not set to True", strings.Join(failedConditions, ", "))
 	}
--- a/test/helpers/conditions.go
+++ b/test/helpers/conditions.go
@@ -13,7 +13,7 @@ import (
 // CheckAllConditionsTrue returns true if all the conditions with given type in `conditionTypes` are set to `True` in the given resource.
 // If it returns `false`, the second return value contains a message to tell what conditions are not `True`.
 func CheckAllConditionsTrue(resource k8sutils.ConditionsAware, conditionTypes []consts.ConditionType) (bool, string) {
-	failedConditions := make([]string, 0, len(conditionTypes))
+	var failedConditions []string
 	lo.ForEach(conditionTypes, func(conditionType consts.ConditionType, _ int) {
 		if !k8sutils.HasConditionTrue(conditionType, resource) {
 			failedConditions = append(failedConditions, string(conditionType))
--- a/test/helpers/deploy/deploy_resources.go
+++ b/test/helpers/deploy/deploy_resources.go
@@ -1083,34 +1083,6 @@ func WithKonnectConfiguration[T ObjectSupportingKonnectConfiguration](
 	}
 }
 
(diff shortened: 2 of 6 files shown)


test/helpers/deploy/deploy_resources.go
```diff
@@ -1010,6 +1010,57 @@ func KongDataPlaneClientCertificateAttachedToCP(
...
+		ke.Spec.KonnectControlPlane.ControlPlaneRef = commonv1alpha1.ControlPlaneRef{
+			Type:      commonv1alpha1.ControlPlaneRefKonnectID,
+			KonnectID: lo.ToPtr(cpID),
+		}
+	})
+	return KonnectExtension(
+		t, ctx, cl,
+		opts...,
+	)
+}
```
pmalek: Wouldn't it be more composable (to be used by other types that support `KonnectConfiguration` field) to define an object option like `WithKonnectConfiguration` and use it similarly to 

```
		created := deploy.KonnectExtension(t, ctx, clientNamespaced,
			deploy.WithKonnectIDControlPlaneRef(cp),
			deploy.WithKonnectConfiguration(konnectConfiguration),
		)
```

WDYT?

test/integration/test_konnect_extension.go
```diff
@@ -0,0 +1,147 @@
...
+		if !ok || status != metav1.ConditionTrue {
+			failedConditionTypes = append(failedConditionTypes, conditionType)
+		}
+	}
+
+	if len(failedConditionTypes) > 0 {
+		return false, fmt.Sprintf("condition(s) %s not set to True", strings.Join(failedConditionTypes, ", "))
+	}
+	return true, ""
+}
```
pmalek: Can we make this a generic (not necessarily using Go generics) function which checks a provided slice of conditions types for being set to true? Something that could be placed in https://github.com/Kong/gateway-operator/blob/27e877a3fa9f023b2bed8ad0c6a1fb289bbb86bb/pkg/utils/kubernetes/status.go or similar to be shared by other places that check conditions.

WDYT?

test/helpers/deploy/deploy_resources.go
randmonkey: Extracted `WithKonnectConfiguration`.

test/integration/test_konnect_extension.go
randmonkey: Extracted a `CheckAllConditionsTrue`.

test/helpers/deploy/deploy_resources.go
randmonkey: @pmalek `KonnectExtension` does not support `GetControlPlaneRef`/`SetControlPlaneRef`. It requires updating of `kubernetes-configuartion`. I think we should do it later after https://github.com/Kong/kubernetes-configuration/issues/328.

test/helpers/conditions.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/samber/lo"
+
+	"github.com/kong/gateway-operator/pkg/consts"
+	k8sutils "github.com/kong/gateway-operator/pkg/utils/kubernetes"
+)
+
+// CheckAllConditionsTrue returns true if all the conditions with given type in `conditionTypes` are set to `True` in the given resource.
+// If it returns `false`, the second return value contains a message to tell what conditions are not `True`.
+func CheckAllConditionsTrue(resource k8sutils.ConditionsAware, conditionTypes []consts.ConditionType) (bool, string) {
+	failedConditions := make([]string, 0, len(conditionTypes))
```
pmalek: I _think_ that trying to optimize here might not be suitable: typically conditions are expected to have their status set to true most often than not and we should not penalize  situations (users) that do have have these circumstances met.

```suggestion
	var failedConditions []string
```

test/helpers/conditions.go
```diff
@@ -0,0 +1,26 @@
...
+	lo.ForEach(conditionTypes, func(conditionType consts.ConditionType, _ int) {
+		if !k8sutils.HasConditionTrue(conditionType, resource) {
+			failedConditions = append(failedConditions, string(conditionType))
+		}
+	})
```
pmalek: Nit: I believe that `ForEach` here doesn't provide us any benefits (shorter syntax etc) so I'd be inclined to suggest to use a plain `for` loop. WDYT?

test/helpers/deploy/deploy_resources.go
```diff
@@ -1010,6 +1010,80 @@ func KongDataPlaneClientCertificateAttachedToCP(
...
+// KonnectExtensionWithAPIAuthRefAndCPID deploys a KonnectExtension attaching to control plane
+// by the CP's ID and the API auth configuration to the Konnect server where the CP is in.
+func KonnectExtensionWithAPIAuthRefAndCPID(
+	t *testing.T,
+	ctx context.Context,
+	cl client.Client,
+	apiAuth konnectv1alpha1.KonnectAPIAuthConfigurationRef,
+	cpID string,
+	opts ...ObjOption,
+) *konnectv1alpha1.KonnectExtension {
```
pmalek: With these other functions and options in place I believe we want to remove this one, no? (especially since it appears unused now)

randmonkey: Done.

test/integration/test_konnect_extension.go
```diff
@@ -53,26 +57,140 @@ func TestKonnectExtensionKonnectGatewayControlPlaneNamespacedRef(t *testing.T) {
...
+func checkKonnectExtensionStatus(
+	ke *konnectv1alpha1.KonnectExtension,
+	expectedKonnectCPID string,
+	expectedDPCertificateSecretName string,
+) func(t *assert.CollectT) {
+	return func(t *assert.CollectT) {
+		err := GetClients().MgrClient.Get(GetCtx(), types.NamespacedName{Name: ke.Name, Namespace: ke.Namespace}, ke)
+		require.NoError(t, err)
+		// Check Konnect control plane ID
+		assert.NotNil(t, ke.Status.Konnect, "status.konnect should be present")
```
pmalek:
```suggestion
		require.NotNil(t, ke.Status.Konnect, "status.konnect should be present")
```

test/integration/test_konnect_extension.go
```diff
@@ -53,26 +57,140 @@ func TestKonnectExtensionKonnectGatewayControlPlaneNamespacedRef(t *testing.T) {
...
+		assert.Equal(t, expectedKonnectCPID, ke.Status.Konnect.ControlPlaneID, "Konnect control plane ID should be set in status")
+		// Check dataplane client auth
+		assert.NotNil(t, ke.Status.DataPlaneClientAuth, "status.dataPlaneClientAuth should be present")
+		assert.NotNil(t, ke.Status.DataPlaneClientAuth.CertificateSecretRef, "status.dataPlaneClientAuth.certiifcateSecretRef should be present")
```
pmalek:
```suggestion
		require.NotNil(t, ke.Status.DataPlaneClientAuth, "status.dataPlaneClientAuth should be present")
		require.NotNil(t, ke.Status.DataPlaneClientAuth.CertificateSecretRef, "status.dataPlaneClientAuth.certiifcateSecretRef should be present")
```

test/integration/test_konnect_extension.go
randmonkey: I don't think we should use `require` instead of `assert` here. `require` fails the test case immediately and do not give change to wait for the conditions to be met. The `status.Konnect` is not set initially so we should not use `require` here.

test/integration/test_konnect_extension.go
randmonkey: Ditto here for `status.dataplaneClientAuth`.

test/integration/test_konnect_extension.go
pmalek: If this fails, `ke.Status.Konnect.ControlPlaneID` below will panic anyway (since `ke.Status.Konnect` is `nil` so nothing is ever going to be checked besides this.

czeslavo: > require fails the test case immediately and do not give change to wait for the conditions to be met

Thanks to https://github.com/stretchr/testify/pull/1481 it's no longer true.

test/helpers/conditions.go
pmalek: Ping

PR #1279, opened 2025-03-04
fix: delete owner relationship between KongUpstreams and KongTargets

pmalek:

Follow up on #1259 

**Which issue this PR fixes**

Part of #788 

**PR Readiness Checklist**:

### DIFF
--- a/controller/konnect/reconciler_consumerref.go
+++ b/controller/konnect/reconciler_consumerref.go
@@ -105,7 +105,7 @@ func handleKongConsumerRef[T constraints.SupportedKonnectEntityType, TEnt constr
 
 	res, err := RemoveOwnerRefIfSet(ctx, cl, ent, &consumer)
 	if err != nil || !res.IsZero() {
-		return ctrl.Result{}, err
+		return res, err
 	}
 
 	type EntityWithConsumerRef interface {
--- a/controller/konnect/reconciler_consumerref.go
+++ b/controller/konnect/reconciler_consumerref.go
@@ -103,23 +103,9 @@ func handleKongConsumerRef[T constraints.SupportedKonnectEntityType, TEnt constr
 		return ctrl.Result{Requeue: true}, nil
 	}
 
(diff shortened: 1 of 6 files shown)


controller/konnect/reconciler_consumerref.go
```diff
@@ -103,23 +103,9 @@ func handleKongConsumerRef[T constraints.SupportedKonnectEntityType, TEnt constr
...
-		}
-		if err := cl.Patch(ctx, ent, client.MergeFrom(old)); err != nil {
-			if k8serrors.IsConflict(err) {
-				return ctrl.Result{Requeue: true}, nil
-			}
-			return ctrl.Result{}, fmt.Errorf("failed to remove owner reference: %w", err)
-		}
+	res, err := RemoveOwnerRefIfSet(ctx, cl, ent, &consumer)
+	if err != nil || !res.IsZero() {
+		return ctrl.Result{}, err
```
programmer04:
```suggestion
		return res, err
```
since `res` is not zero, it means it's not `ctrl.Result{}`, thus it contains e.g. requeue after. Or am I missing something?

controller/konnect/reconciler_upstreamref.go
```diff
@@ -106,6 +106,11 @@ func handleKongUpstreamRef[T constraints.SupportedKonnectEntityType, TEnt constr
 		return ctrl.Result{}, fmt.Errorf("failed to update status: %w", err)
 	}
 
+	res, err := RemoveOwnerRefIfSet(ctx, cl, ent, kongUpstream)
+	if err != nil || !res.IsZero() {
+		return ctrl.Result{}, err
+	}
```
programmer04: The same as mentioned in the previous comment

pmalek:
```suggestion
	if err != nil || !res.IsZero() {
		return res, err
	}
```

controller/konnect/reconciler_consumerref.go
pmalek: Thanks for catching this. Indeed this was supposed to return `res`. PTAL now

programmer04 (approved):

PR #1289, opened 2025-03-05
chore(tests): enable e2e `TestOperatorLogs`, case `upgrade from nightly to current` and adjustments

programmer04:

- **for case `upgrade from nightly to current` from `TestHelmUpgrade`**

   Nightly images are available in 

     - [kong/nightly-gateway-operator-oss](https://hub.docker.com/r/kong/nightly-gateway-operator-oss/tags)
     - [kong/nightly-gateway-operator](https://hub.docker.com/r/kong/nightly-gateway-operator/tags)

    so ensure that those registries are used in the enabled test.
    
    Furthermore spotted that `"DataPlane deployment is not patched after operator upgrade"` tried to assert `gatewayDataPlaneDeploymentIsPatched` instead of `gatewayDataPlaneDeploymentIsNotPatched` in two places. 
    Make cleanup logic more robust to ensure that every GWAPI object provisioned before is deleted properly (finalizers need some more time) before uninstalling the Helm chart (otherwise orphaned `ControlPlane` was left and namespace couldn't be deleted)
    
- **`TestOperatorLogs`**
   `TestOperatorLogs` used to be disabled quite long time ago, see
    - https://github.com/kong/gateway-operator-archive/issues/908
    - https://github.com/Kong/gateway-operator-archive/issues/1283#issuecomment-2018318653

   But it is no longer flaky, underlying issues have been addressed, so reenable it. 
   
Replaces https://github.com/Kong/gateway-operator/pull/990

### DIFF
--- a/test/e2e/test_helm_install_upgrade.go
+++ b/test/e2e/test_helm_install_upgrade.go
@@ -14,6 +14,7 @@ import (
 	"github.com/stretchr/testify/require"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -40,9 +41,7 @@ func TestHelmUpgrade(t *testing.T) {
 
 		waitTime = 3 * time.Minute
 	)
-
-	ctx, cancel := context.WithCancel(t.Context())
-	defer cancel()
+	ctx := t.Context()
 
 	// createEnvironment will queue up environment cleanup if necessary
 	// and dumping diagnostics if the test fails.
@@ -201,7 +200,7 @@ func TestHelmUpgrade(t *testing.T) {
 				{
 					Name: "DataPlane deployment is not patched after operator upgrade",
 					Func: func(c *assert.CollectT, cl *testutils.K8sClients) {
-						gatewayDataPlaneDeploymentIsPatched("gw-upgrade-latestminor-current=true")(ctx, c, cl.MgrClient)
+						gatewayDataPlaneDeploymentIsNotPatched("gw-upgrade-latestminor-current=true")(ctx, c, cl.MgrClient)
 					},
 				},
 				{
@@ -212,8 +211,6 @@ func TestHelmUpgrade(t *testing.T) {
 				},
 			},
 		},
-		/**
-		// TODO(Jintao): This test is disabled. After a new nightly image is available which uses KIC 3.4.1, we can enable it.
 		{
 			name:             "upgrade from nightly to current",
 			fromVersion:      "nightly",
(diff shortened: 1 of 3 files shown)


pmalek (commented):
This test checks the upgrade path between operator versions. Since we moved the nightly builds to https://hub.docker.com/r/kong/nightly-gateway-operator-oss/ this changeset will need to accomodate for that as well.

test/e2e/test_operator_logs.go
```diff
@@ -208,11 +216,6 @@ func TestOperatorLogs(t *testing.T) {
 
 func isReconcilerErrorAllowedByRegexMatch(errorMsg string) bool {
 	allowedReconcilerErrorRegexes := []string{
-		// For some reason this sometimes happen on CI. While this might be an actual
-		// issue, this should not fail the test on its own.
-		//
-		// Possibly related upstream issue:
-		// - https://github.com/kubernetes-sigs/controller-runtime/issues/1881
```
pmalek, line 215: Why remove this comment? I believe leaving this in makes sense to keep readers aware of the issue.

test/e2e/test_helm_install_upgrade.go
```diff
@@ -304,8 +300,10 @@ func TestHelmUpgrade(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			// Repository is different for OSS and Enterprise images and it should be set accordingly.
 			kgoImageRepository := "docker.io/kong/gateway-operator-oss"
+			kgoImageRepositoryNightly := "docker.io/kong/nightly-gateway-operator-oss"
 			if helpers.GetDefaultDataPlaneBaseImage() == consts.DefaultDataPlaneBaseEnterpriseImage {
 				kgoImageRepository = "docker.io/kong/gateway-operator"
+				kgoImageRepositoryNightly = "docker.io/kong/nightly-gateway-operator"
```
pmalek, line 306: Nit: could we extract these to `const`s?

pmalek (commented):
2 minor nits. Otherwise this is 🚀 . Thanks for handling this 🙇

test/e2e/test_helm_install_upgrade.go
programmer04: Here IMHO `const` adds only indirection that increases mental overhead, because I need to be able to overwrite it the same as it's for var `kgoImageRepository`

test/e2e/test_operator_logs.go
programmer04: Because it seems to be resolved, in the original issue I found https://github.com/kubernetes-sigs/controller-runtime/issues/1881#issuecomment-1492082960 and we have eventually, and I couldn't trigger that failure in the subsequent re-runs. But I can add it, your call
?

pmalek: Seems there's a k/k issue linked in that controller-runtime issue: https://github.com/kubernetes/kubernetes/issues/124347. Can we link that instead then?

test/e2e/test_helm_install_upgrade.go
pmalek: Fair enough. Given that we have the prod registries here as well let's leave the nightly ones as well here.

test/e2e/test_operator_logs.go
programmer04: sure, linked

pmalek (approved):

PR #1295, opened 2025-03-06
chore: pass `--devel` to helm e2e tests

programmer04:

To use `rc` version in helm, the flag `--devel` needs to be passed

**Which issue this PR fixes**

Follow-up for https://github.com/Kong/gateway-operator/issues/949

### DIFF
--- a/test/e2e/test_helm_install_upgrade.go
+++ b/test/e2e/test_helm_install_upgrade.go
@@ -349,10 +349,21 @@ func TestHelmUpgrade(t *testing.T) {
 					RestConfig: e.Environment.Cluster().Config(),
 				},
 				SetValues: values,
+				ExtraArgs: map[string][]string{
+					"install": {
+						"--devel",
+					},
+					"upgrade": {
+						"--devel",
+					},
+				},
 			}
 
 			require.NoError(t, helm.AddRepoE(t, opts, "kong", "https://charts.konghq.com"))
 			require.NoError(t, helm.InstallE(t, opts, chart, releaseName))
+			out, err := helm.RunHelmCommandAndGetOutputE(t, opts, "list")
+			require.NoError(t, err)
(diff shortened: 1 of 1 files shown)


test/e2e/test_helm_install_upgrade.go
```diff
@@ -349,6 +349,14 @@ func TestHelmUpgrade(t *testing.T) {
...
+					"install": {
+						"--devel",
+					},
+					"upgrade": {
+						"--devel",
+					},
+				},
 			}
 
 			require.NoError(t, helm.AddRepoE(t, opts, "kong", "https://charts.konghq.com"))
```
pmalek, line 362: Just a proposal: could we add a call to `helm ls` to show which version we're actually using in tests?

I think this could be done via 

```
			optsCommon := &helm.Options{
				KubectlOptions: &k8s.KubectlOptions{
					Namespace:  e.Namespace.Name,
					RestConfig: e.Environment.Cluster().Config(),
				},
			}

			helm.RunHelmCommandAndGetOutputE(t, optsCommon, "ls")
```

just below `helm.InstallE()` call. WDYT?

pmalek: Thanks! It works 🥳 

https://github.com/Kong/gateway-operator/actions/runs/13695706021/job/38297534599?pr=1295#step:6:191

```
    test_helm_install_upgrade.go:366: Helm list output after install:
          NAME              	NAMESPACE                           	REVISION	UPDATED                                	STATUS  	CHART                       	APP VERSION
        kgo-1-3-0-to-1-4-0	5e91aeec-57e6-4031-a5a5-b362c2924e43	1       	2025-03-06 09:50:36.234259209 +0000 UTC	deployed	gateway-operator-0.4.10-rc.1	1.4     
```

pmalek (approved):

PR #1303, opened 2025-03-06
chore: delete old owned CP resources, re-enable upgrade e2e tests

czeslavo:

Adds CP reconciler code responsible for deleting the old owned resources created by the previous version of KGO and re-enables upgrade E2E tests verifying the upgrade path.

**Which issue this PR fixes**

Fixes https://github.com/Kong/gateway-operator/issues/1188.


controller/controlplane/controller_watch.go
```diff
@@ -0,0 +1,237 @@
...
+		ctrllog.FromContext(ctx).Error(
+			operatorerrors.ErrUnexpectedObject,
+			"failed to run predicate function",
+			"expected", "ClusterRoleBinding", "found", reflect.TypeOf(obj),
+		)
+		return false
+	}
+
+	return r.clusterScopedObjHasControlPlaneOwner(ctx, clusterRoleBinding)
+}
```
pmalek, line 53: Nit: these 2 + `validatingWebhookConfigurationHasControlPlaneOwner` could be made to use generics to prevent code duplication.

controller/controlplane/controller_watch.go
```diff
@@ -0,0 +1,237 @@
...
+		ctrllog.FromContext(ctx).Error(
+			operatorerrors.ErrUnexpectedObject,
+			"failed to run map funcs",
+			"expected", "ValidatingWebhookConfiguration", "found", reflect.TypeOf(obj),
+		)
+		return nil
+	}
+
+	return r.getControlPlaneRequestFromManagedByNameLabel(ctx, validatingWebhookConfig)
+}
```
pmalek, line 132: Nit: ditto about using generics for these 3.

pmalek (approved):
Nice change. Just 2 nits but the general idea is 👍 .

I think we want to add it to `main` (and also backport it to 1.5) right?

czeslavo:
> I think we want to add it to `main` (and also backport it to 1.5) right?

I don't think so; it's only for the feature branch for now. This is exclusively KO behavior, we don't want to delete the resources in KGO as they're still needed.

pmalek:
> > I think we want to add it to `main` (and also backport it to 1.5) right?
> 
> I don't think so; it's only for the feature branch for now. This is exclusively KO behavior, we don't want to delete the resources in KGO as they're still needed.

You're right. I was thinking about the webhook resources which were removed in https://github.com/Kong/gateway-operator/pull/1100/files#diff-1b75b2a136efd06a1175041f5d638ecaf07a05badaefe9d36e0127fbc004ccb2 but that's separate from what's handled in here.

1.4.x and older had the [cleanup routine](https://github.com/Kong/gateway-operator/pull/1100/files#diff-1b75b2a136efd06a1175041f5d638ecaf07a05badaefe9d36e0127fbc004ccb2L292-L324) which should handle this.

PR #1304, opened 2025-03-06
feat: automatic secret provisioning

### LINKED ISSUE #1166: Automatic provisioning of DP Certificates
## Problem Statement

The `KonnectExtension` API allows the `DataPlane` certificate to be either manually provided by the user or automatically created by KGO. We need to implement a logic to provision a certificate and push it to Konnect via the `KongDataPlaneClientCertificate` resource.

## Proposed Solution

- Whenever a `KonnectExtension` with automatic certificate provision Is detected, a new certificate labeled `foo` (label to be defined) is created and stored in a Kubernetes secret.
- A new controller watches all the secrets labeled `foo`. For each of such secrets, it creates a new `KongDataPlaneClientCertificate` resource In the cluster, setting the certificate as the `cert` field and the original secret as the owner.
- The `KongDataPlaneClientCertificate` controller pushes the provisioned certificates to Konnect (already implemented)

## Acceptance Criteria
- [ ] As a user, I can have my `DataPlane` certificates automatically provisioned and pushed to Konnect.

mlavacca:

**Which issue this PR fixes**

Fixes #1166 

**PR Readiness Checklist**:


config/samples/dataplane-konnect-extension-automatic-provisioning.yaml
pmalek, line 1: oh, the automatic provisioning is the default: https://github.com/Kong/kubernetes-configuration/blob/da21a286cd79d5c7decf1e64c8f801567fb57d0b/api/konnect/v1alpha1/konnect_extension_types.go#L81-L82.

I missed that fact somehow. 👍

controller/konnect/konnectextension_controller.go
```diff
@@ -480,7 +478,20 @@ func (r *KonnectExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+					if k8serrors.IsConflict(err) {
+						return ctrl.Result{Requeue: true}, nil
+					}
+					return ctrl.Result{}, err
+				}
+
+				log.Info(logger, "konnect-cleanup finalizer on the referenced secret updated")
+				// Setting "Requeue: true" along with RequeueAfter makes the controller bulletproof, as
+				// if the syncPeriod is set to zero, the controller won't requeue.
+				return ctrl.Result{Requeue: true, RequeueAfter: r.SyncPeriod}, err
```
pmalek: Consider using https://github.com/Kong/gateway-operator/blob/43c548dbfeadcf56453b9b9cc2cdaf0c8dc3df53/controller/pkg/patch/finalizer.go#L15-L45

controller/konnect/konnectextension_controller_secrets.go
```diff
@@ -30,26 +32,43 @@ func listKonnectExtensionsBySecret(ctx context.Context, cl client.Client, s *cor
...
+			err := cl.Get(ctx, k8stypes.NamespacedName{
+				Namespace: s.Namespace,
+				Name:      ownerRef.Name,
+			}, owner)
+			if err != nil {
+				return nil, err
+			}
+			if owner.UID == ownerRef.UID {
+				l.Items = append(l.Items, *owner)
+			}
```
pmalek: 🤔 Do we actually need to compare the UIDs here? Wouldn't it be possible to just use: `controllerutil.HasOwnerReference` (possibly with `controllerutil.HasControllerReference`? Although I see that the latter just returns a bool, taking in no arguments 🤔 )?

`HasOwnerReference` verifies owners' group, kind and name and I think that should be enough for this case, no?

pkg/utils/kubernetes/resources/labels.go
```diff
@@ -33,6 +34,18 @@ func LabelObjectAsKongPluginInstallationManaged(obj metav1.Object) {
 	obj.SetLabels(labels)
 }
 
+// LabelObjectAsKonnectExtensionManaged ensures that labels are set on the
+// provided object to signal that it's owned by a KonnectExtension resource
+// and that its lifecycle is managed by this operator.
+func LabelObjectAsKonnectExtensionManaged(obj metav1.Object) {
```
pmalek, line 40: Could we get a UT for this func?

pkg/utils/kubernetes/resources/secrets.go
```diff
@@ -42,28 +43,32 @@ func WithAnnotation[T client.Object](k, v string) func(d T) {
 	}
 }
 
-type controlPlaneOrDataPlane interface {
-	*operatorv1beta1.ControlPlane | *operatorv1beta1.DataPlane
+type controlPlaneDataPlaneOrKonnectExtension interface {
```
pmalek: Can we export this type and move it somewhere where it can be shared and used in `controller/pkg/secrets/cert.go` where we redefine it?

config/samples/dataplane-konnect-extension-automatic-provisioning.yaml
mlavacca: Yep, it's set as default. In case users don't want to bother themselves with certificates at all, they can just leave that field empty

pkg/utils/kubernetes/resources/secrets.go
mlavacca: Why should we? That interface is used only here. I'd lean toward leaving it here and maybe moving it somewhere else when we'll need it somewhere else.

controller/konnect/konnectextension_controller_secrets.go
mlavacca: I have added the check to `Kind` and `APIVersion` and removed the UID check.

pkg/utils/kubernetes/resources/secrets.go
pmalek: All I'm saying is that https://github.com/Kong/gateway-operator/pull/1304/files#diff-465e5135e663acc85bb2b22849039cb833ae715282354383ac12ee3d638b67feR180 is also defining a type constraint based on these 3 types. I'm OK if we leave it as is but I think this could be simplified by reusing the type constraint.

controller/konnect/konnectextension_controller_secrets.go
```diff
@@ -30,26 +34,44 @@ func listKonnectExtensionsBySecret(ctx context.Context, cl client.Client, s *cor
 		return nil, err
 	}
 
+	// Add all the secrets automatically for the KonnectExtensions.
```
pmalek: This comment seems a little inaccurate: we're not really adding anything here. We're just looping over the owner references, checking if they exist right?

Can we fix that?

pmalek (dismissed):
Minor nit, otherwise LGTM.

pkg/utils/kubernetes/resources/secrets.go
mlavacca: I apologize, I didn't catch the point. It makes sense, here it is: 6ac83e72a66bc513afa5e4287a3d3ec0d4f9f5d9

controller/konnect/konnectextension_controller_secrets.go
mlavacca: Actually, in 3fd157adf1fe4f87b3012ed59d228c9801a53410 when addressing your comment I accidentally removed the append. aa77534d0ed8c2d3965b3373e97ac59b3f136a73 re-adds that.

pkg/utils/kubernetes/resources/labels_test.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/kong/gateway-operator/pkg/consts"
+
+	konnectv1alpha1 "github.com/kong/kubernetes-configuration/api/konnect/v1alpha1"
+)
+
+func TestLabelObjectAsKonnectExtensionManaged(t *testing.T) {
+	for _, tt := range []struct {
+		name   string
+		input  map[string]string
+		output map[string]string
```
pmalek, line 18: Nit: this could be called `expected` to make this a bit more clear to the reader.

pmalek (dismissed):

pmalek (approved):

PR #1312, opened 2025-03-07
fix: prevent panic in KE controller when dereferencing not existing secret

mlavacca:

In the KE controller, when dereferencing a secret to remove its finalizer, ensure the secret actually exists.

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:

### DIFF
--- a/controller/konnect/konnectextension_controller.go
+++ b/controller/konnect/konnectextension_controller.go
@@ -205,14 +205,13 @@ func (r *KonnectExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
 			}, nil
 		}
 
-		certificateSecret, err := getCertificateSecret(ctx, r.Client, ext)
+		certificateSecret, certExists, err := getCertificateSecret(ctx, r.Client, ext)
 		if client.IgnoreNotFound(err) != nil {
 			return ctrl.Result{}, err
 		}
 
-		// if the secret does not have the konnect-cleanup finalizer, the certificate has been properly cleaned up
-		// in Konnect.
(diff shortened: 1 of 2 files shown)


controller/konnect/konnectextension_controller.go
```diff
@@ -205,14 +205,13 @@ func (r *KonnectExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
 			}, nil
 		}
 
-		certificateSecret, err := getCertificateSecret(ctx, r.Client, ext)
+		certificateSecret, certExists, err := getCertificateSecret(ctx, r.Client, ext)
```
pmalek: Instead of `certExists`, couldn't we rely on `client.IgnoreNotFound(err)`? This is essentially doubling the information IMHO. This would require changing how the `if` statement are organized below but I think this makes sense.

WDYT?

mlavacca: Makes sense. Changed it to check the returned error type. PTAL

controller/konnect/konnectextension_controller_utils.go
```diff
@@ -238,7 +238,7 @@ func getKonnectAPIAuthRefNN(ctx context.Context, cl client.Client, ext *konnectv
 	}, nil
 }
 
-func getCertificateSecret(ctx context.Context, cl client.Client, ext konnectv1alpha1.KonnectExtension) (*corev1.Secret, error) {
+func getCertificateSecret(ctx context.Context, cl client.Client, ext konnectv1alpha1.KonnectExtension) (secret *corev1.Secret, err error) {
```
pmalek: These return values don't look like they are used

pmalek (approved):

PR #1326, opened 2025-03-11
Fix conflict for "deps: bump kubernetes-configuration to include last KonnectExtension schema changes"

randmonkey:
Follow up of #1319. Fixed the conflict in that PR.
**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:


controller/pkg/extensions/konnect/konnectextension_test.go
```diff
@@ -407,11 +411,16 @@ func getKongInKonnectEnvVars(
 	konnectExt konnectv1alpha1.KonnectExtension,
 ) []corev1.EnvVar {
 	envSet := []corev1.EnvVar{}
-	for k, v := range config.KongInKonnectDefaults(konnectExt.Spec.DataPlaneLabels, konnectExt.Status) {
+	var dataplaneLabels map[string]konnectv1alpha1.DataPlaneLabelValue
+	if konnectExt.Spec.Konnect.DataPlane != nil {
+		dataplaneLabels = konnectExt.Spec.Konnect.DataPlane.Labels
+	}
+	for k, v := range config.KongInKonnectDefaults(dataplaneLabels, konnectExt.Status) {
```
pmalek, line 418: Wouldn't this

```suggestion
	for k, v := range config.KongInKonnectDefaults(konnectExt.Spec.Konnect.DataPlane.Labels, konnectExt.Status) {
```

be equivalent? (same in the implementation in `controller/pkg/extensions/konnect/dataplane.go`)

I'm not sure I see the reason for using the additional variable here 🤔

pmalek:
@randmonkey Should we close #1319 in favor of this one?

controller/pkg/extensions/konnect/konnectextension_test.go
randmonkey: Looks like it is used for preventing `panic`s when `spec.konnect.dataPlane` is not specified since it is not required.

randmonkey:
> @randmonkey Should we close #1319 in favor of this one?

If @mlavacca thinks it is OK to continue on this one, we can close #1319.

controller/pkg/extensions/konnect/konnectextension_test.go
pmalek: Oh, yes. I missed that.

I think that would could change the `KongInKonnectDefaults` to accept `konnectExtension.Spec.Konnect.DataPlane` ( move this `nil` check to the function itself). This would make the call site a bit clearer. I'm fine if we don't do it in this PR though.

pmalek (approved):

controller/pkg/extensions/konnect/konnectextension_test.go
mlavacca: Yes, that's needed to prevent panic in case the dataplane is not set

mlavacca (approved):

PR #1332, opened 2025-03-11
chore: give write permissions to release workflows

mlavacca:

The release workflow https://github.com/Kong/gateway-operator/actions/runs/13796774023 failed because of the tight permissions given to it. This PR set the top level permissions to write

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:

### DIFF
--- a/.github/workflows/__release-workflow.yaml
+++ b/.github/workflows/__release-workflow.yaml
@@ -71,7 +71,7 @@ on:
         default: false
 
 permissions:
-  contents: read
+  contents: write
 
 jobs:
   semver:
--- a/.github/workflows/__release-workflow.yaml
+++ b/.github/workflows/__release-workflow.yaml
@@ -71,7 +71,7 @@ on:
         default: false
 
 permissions:
-  contents: write
+  contents: read
 
 jobs:
   semver:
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yaml
```diff
@@ -21,7 +21,7 @@ on:
         - release
 
 permissions:
-  contents: read
```
pmalek: To be on the safe side, I'd only add that in the job below. WDYT?

.github/workflows/release-bot.yaml
```diff
@@ -10,7 +10,7 @@ on:
       - 'release/*'
 
 permissions:
-  contents: read
```
pmalek: I don't think this should be required given that we do not call any workflows that would need to elevate permissions

.github/workflows/__release-workflow.yaml
```diff
@@ -71,7 +71,7 @@ on:
         default: false
 
 permissions:
-  contents: read
+  contents: write
```
pmalek: This also shouldn't be required. The only workflow that we call here is `.github/workflows/__build-workflow.yaml` which doesn't require any additional permissions.


```suggestion
  contents: read
```

pmalek (approved):

PR #1388, opened 2025-03-21
feat(controlplane): add watchNamespaces handling

### LINKED ISSUE #1347: Allow a ControlPlane to watch all namespaces or only specific ones
## Problem Statement

#SEC-1325
Currently a ControlPlane can watch all namespaces in the cluster, creating a ClusterRole and CRBinding that allows the ControlPlane's service account to read configmaps and secrets cluster wide. 

## Proposed Solution

As KIC supports an option to watch only specific namespaces, the ControlPlane should have in the spec an optional list field (empty by default - works clusterwide) where user can specify which namespaces the KIC deployment is going to watch.
Accordingly, if a namespaces list has been provided, the ControlPlane controller must create a Role and RoleBinding pointing to KIC service account in the specified namespaces and go with ClusterRole and CRBinding only if the namespaces list is empty

## Acceptance Criteria
- [x] ControlPlane controller reconciles KIC deployment against an optional watched namespaces list received in the specs
- [x] ControlPlane controller reconciles, if a watched namespaces list has been provided, Roles and RoleBindings in each namespace; with an empty list of watched namespaces it will reconciler ClusterRole and ClusterRoleBinding (as today)

pmalek:

Add handling of `watchNamespaces` field, added in https://github.com/Kong/kubernetes-configuration/pull/358.

This PR:

- adds setting of `CONTROLLER_WATCH_NAMESPACE` environment variable in `ControlPlane`s `Deployment` template based on `ControlPlane`'s `spec.watchNamespaces`
- adds a requirement to add `ReferenceGrant`s for each namespace in the list (when `spec.watchNamespaces.type` is `list`) which grant `ControlPlane` access to a namespace like so:

    ```
    apiVersion: gateway.networking.k8s.io/v1beta1
    kind: ReferenceGrant
    metadata:
      name: ref
      namespace: kong-system
    spec:
      from:
      - group: gateway-operator.konghq.com
        kind: ControlPlane
        namespace: default
      to:
      - group: ""
        kind: Namespace
        name: kong-system
    ```

    This manifest grants `kong-system` namespace access to `ControlPlane`s in `default` namespace.

**Which issue this PR fixes**

Part of #1347

Open questions:

- When a `ControlPlane` is deployed and then updated to have a list of watch namespaces, we'll set a `Ready` condition to `False` with a `Reason` `MissingReferenceGrants` but the `Deployment` (and `Pod`s) will stay in place allowing the reconciliation to continue.
  - Do we want to handle that differently? Removal of `Deployment` seems a bit harsh for a missing `ReferenceGrant`.
  - I **think** we want to remove the namespaces that have missing `ReferenceGrant`s from the `CONTROLLER_WATCH_NAMESPACE` env + set the status condition to `False`.

**PR Readiness Checklist**:


czeslavo:
> I think we want to remove the namespaces that have missing ReferenceGrants from the CONTROLLER_WATCH_NAMESPACE env + set the status condition to False.

+1. I'd update the deployment so it stops watching a non-granted namespace.

controller/controlplane/controller.go
```diff
@@ -340,6 +348,28 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+		)
+		res, err := r.patchStatus(ctx, logger, cp)
+		if err != nil {
+			log.Debug(logger, "unable to patch ControlPlane status", "error", err)
+			return ctrl.Result{}, err
+		}
+		if !res.IsZero() {
+			return res, nil
+		}
+		return ctrl.Result{}, nil
```
czeslavo: I think this could be simplified to the below snippet - the error will be logged anyway if we return an error from the `Reconcile`.
```suggestion
		if err != nil || !res.IsZero() {
			return res, err
		}
		return ctrl.Result{}, nil
```

controller/controlplane/controller_reconciler_utils.go
```diff
@@ -264,6 +266,8 @@ func (r *Reconciler) ensureDeployment(
 		generatedDeployment.Spec.Replicas = lo.ToPtr(int32(numReplicasWhenNoDataPlane))
 	}
 	if err := r.Client.Create(ctx, generatedDeployment); err != nil {
+		pretty.Print(generatedDeployment)
+		// fmt.Printf("failed creating ControlPlane Deployment %s: %w", generatedDeployment)
```
czeslavo:
```suggestion
```

controller/controlplane/controller_reconciler_utils.go
```diff
@@ -764,3 +797,70 @@ func (r *Reconciler) ensureValidatingWebhookConfiguration(
...
 	return op.Created, r.Client.Create(ctx, generatedWebhookConfiguration)
 }
+
+func (r *Reconciler) validateReferenceGrants(
+	ctx context.Context,
+	cp *operatorv1beta1.ControlPlane,
+) error {
+	if cp.Spec.WatchNamespaces == nil {
+		return nil
+	}
```
czeslavo: This technically shouldn't happen due to the default on the CRD, right? I think this will be more future-proof if we return an error here, as this is something we do not expect.
```suggestion
	if cp.Spec.WatchNamespaces == nil {
		return errors.New("spec.watchNamespaces cannot be empty")
	}
```

controller/controlplane/controller_reconciler_utils.go
```diff
@@ -764,3 +797,70 @@ func (r *Reconciler) ensureValidatingWebhookConfiguration(
...
+	case operatorv1beta1.WatchNamespacesTypeList:
+		for _, ns := range cp.Spec.WatchNamespaces.List {
+			if err := ensureReferenceGrantsForNamespace(ctx, r.Client, cp, ns); err != nil {
+				return err
+			}
+		}
+
+		return nil
+	default:
+		return nil
```
czeslavo:
```suggestion
	default:
		return fmt.Errorf("unexpected watchNamespaces.type: %q", cp.Spec.WatchNamespaces.Type)
```

controller/controlplane/controller_reconciler_utils.go
```diff
@@ -764,3 +797,70 @@ func (r *Reconciler) ensureValidatingWebhookConfiguration(
...
+	cp *operatorv1beta1.ControlPlane,
+	ns string,
+) error {
+	var refGrants gatewayv1beta1.ReferenceGrantList
+	if err := cl.List(ctx, &refGrants, client.InNamespace(ns)); err != nil {
+		return fmt.Errorf("failed listing ReferenceGrants in namespace %s: %w", ns, err)
+	}
+	for _, refGrant := range refGrants.Items {
+		if !lo.ContainsBy(refGrant.Spec.From, func(from gatewayv1beta1.ReferenceGrantFrom) bool {
+			return from.Group == "gateway-operator.konghq.com" &&
```
czeslavo: Could we use a group extracted from the package variable instead of the hardcoded constant?

czeslavo (dismissed):
Do I get this correctly that it will be extended according to the design doc so that we create separate Role/RoleBinding pairs per namespace + a single ClusterRole/RoleBinding for each CP?

pmalek:
> Do I get this correctly that it will be extended according to the design doc so that we create separate Role/RoleBinding pairs per namespace + a single ClusterRole/RoleBinding for each CP?

Yes, I intend to extend this with that + removing the namespaces from `CONTROLLER_WATCH_NAMESPACE` env where a `ReferenceGrant` is missing.

controller/controlplane/controller_reconciler_utils.go
```diff
@@ -7,6 +7,7 @@ import (
 
 	"github.com/go-logr/logr"
 	"github.com/google/go-cmp/cmp"
+	"github.com/kr/pretty"
```
pmalek:
```suggestion
```

controller/controlplane/controller_reconciler_utils.go
pmalek: Updated, PTAL.

pmalek:
> Do I get this correctly that it will be extended according to the design doc so that we create separate Role/RoleBinding pairs per namespace + a single ClusterRole/RoleBinding for each CP?

cab744887770040b159bbe6c6081467d7e06c910 contains the changes which filter down the watch namespaces when `ReferenceGrant`(s) are missing.

I suggest to review this as is now and then solve the `ClusterRole`/`Role` split in a separate PR as this will make the review easier to read and to reason about. WDYT?

czeslavo (approved):

PR #1396, opened 2025-03-25
chore: refactor cache indexes setup

### LINKED ISSUE #721: Manager's cache index related code is scattered in several places
### Problem statement

Currently manager's index related code is scattered across several places:

- https://github.com/Kong/gateway-operator/blob/fb23412717ccc423ce63d384f6a520d365d8faa1/internal/utils/index/index.go
- https://github.com/Kong/gateway-operator/blob/733f22103699e07ccfb18b76b6982b61854dd553/modules/manager/controller_setup.go#L614-L716
- https://github.com/Kong/gateway-operator/blob/733f22103699e07ccfb18b76b6982b61854dd553/modules/manager/controller_setup.go#L162-L182

This issue tracks the effort of putting all that code in 1 place and using single boilerplate for scaffolding.

### Acceptance criteria

- [ ] Manager's cache index related code is in 1 place.

czeslavo:

Refactors cache indexes setup code so that:

- There's a single function responsible for setting up the indexes: `manager.SetupCacheIndexes`
- All indexes are configured with a single `index.Option` struct that's used in the `manager.SetupCacheIndexes`
- There's a separate file for each indexed object type in the `internal//utils/index` package and each of them exposes an `index.OptionsFor*` function that's plugged into `manager.SetupCacheIndexes`

Other notes:

- `AdmissionRequestHandlerFunc` parameter was removed from `manager.Run` function as it's not used anymore
- `ReferencedControlPlaneDoesNotExistError` and `ReferencedKongGatewayControlPlaneIsUnsupported` error types had to be moved to `pkg/controlplane` package to avoid circular dependencies
- `GetCPForRef` and `GetControlPlaneRef` helper functions were moved from to `pkg/controlplane` as well because of the same reason
- All functions responsible for mapping an object to `[]string` to index on were made private and are only exposed indirectly by `index.OptionsFor*` functions

**Which issue this PR fixes**

Fixes https://github.com/Kong/gateway-operator/issues/721.


modules/manager/run.go
```diff
@@ -135,7 +135,6 @@ func Run(
 	cfg Config,
 	scheme *runtime.Scheme,
 	setupControllers SetupControllersFunc,
-	admissionRequestHandler AdmissionRequestHandlerFunc,
 	startedChan chan<- struct{},
```
randmonkey, line 138: Could we mention that we  removed the param `admissionRequestHandler` since we have removed admission webhooks in the PR comments?

modules/manager/controller_setup.go
```diff
@@ -165,42 +166,57 @@ func (c *ControllerDef) MaybeSetupWithManager(ctx context.Context, mgr ctrl.Mana
...
-				return fmt.Errorf("failed to setup index for ControlPlanes on KonnectExtensions: %w", err)
-			}
-			if err := index.ExtendableOnKonnectExtension(ctx, mgr.GetCache(), &operatorv1beta1.GatewayConfiguration{}); err != nil {
-				return fmt.Errorf("failed to setup index for ControlPlanes on KonnectExtensions: %w", err)
-			}
+	}
+	if cfg.KonnectControllersEnabled {
+		cl := mgr.GetClient()
+		indexOptions = slices.Concat(indexOptions,
+			index.OptionsForExtendables(),
```
randmonkey: I think the PR should include some comments about what is the `OptionsForExtendables` used for.

internal/utils/index/extendable.go
```diff
@@ -0,0 +1,123 @@
+package index
```
randmonkey, line 1: Could we add more comments to tell what `OptionsForExtendables` is used for?

randmonkey (commented):
Generally LGTM, but It would be clearer if more comments are added o tell what the "new" files (not directly migrated from the original `Index*`) used for.

internal/utils/index/extendable.go
czeslavo: To make the pattern consistent (each object type has its own `OptionsFor*` function), I split `OptionsForExtendables` into `OptionsForDataPlane`, `OptionsForGatewayConfiguration`, and `OptionsForControlPlane`. PTAL https://github.com/Kong/gateway-operator/compare/a6d4244a85e207c6fb5cd491a0cc424be98db329..459d80229b863ab93be00260dac19d7185f66b15

modules/manager/controller_setup.go
czeslavo: PTAL at https://github.com/Kong/gateway-operator/compare/a6d4244a85e207c6fb5cd491a0cc424be98db329..459d80229b863ab93be00260dac19d7185f66b15

modules/manager/run.go
czeslavo: Mentioned in the PR description 👍

randmonkey (approved):

PR #1403, opened 2025-03-26
chore: add govulncheck wrapper that allows excluding specific CVEs

czeslavo:

Adds a govulncheck wrapper script that allows excluding particular CVEs by their IDs. Already adds two CVEs to the exclude list as they do not affect us and block CI:

- GO-2025-3547 (https://github.com/kubernetes/kubernetes/issues/126587)
- GO-2025-3521 (https://github.com/kubernetes/kubernetes/issues/130786)

The script is an adapted version of https://raw.githubusercontent.com/tianon/gosu/refs/heads/master/govulncheck-with-excludes.sh.

If govulncheck fails, it is rerun with the `-json` option to capture the output in JSON format that is used for filtering.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -206,7 +206,7 @@ build: generate
 
 .PHONY: _govulncheck
 _govulncheck:
-	$(GOVULNCHECK) -scan $(SCAN) -show color,verbose ./...
+	GOVULNCHECK=$(GOVULNCHECK) SCAN=$(SCAN) ./hack/govulncheck-with-excludes.sh ./...
 
 .PHONY: govulncheck
 govulncheck: download.govulncheck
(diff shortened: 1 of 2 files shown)


pmalek (approved):
Tested locally by removing the first entry in `excludeVulns` and I have received a failure (as expected): 

```
make govulncheck
make _govulncheck SCAN=symbol
GOVULNCHECK=/Users/X/code_/gateway-operator/bin/installs/govulncheck/1.1.4/bin/govulncheck SCAN=symbol ./hack/govulncheck-with-excludes.sh ./...
- GO-2025-3547 (aka CVE-2024-7598, GHSA-r56h-j38w-hrqq)

        Kubernetes kube-apiserver Vulnerable to Race Condition in k8s.io/kubernetes
make[1]: *** [_govulncheck] Error 1
make: *** [Makefile:213: govulncheck] Error 2
```

PR #1404, opened 2025-03-26
tests: add envtest covering DP + KonnectExtension

### LINKED ISSUE #734: Env test for DataPlane KonnectExtension
Problem Statement
https://github.com/Kong/gateway-operator/pull/714 introduced the logic to patch dataplanes based on the referenced KonnectExtension. We do need to add an envtest for such a new feature.

Acceptance Criteria
- [x] An envtest for the DataPlane KonnectExtension feature exists.

czeslavo:

Adds an envtest covering `DataPlane` and `KonnectExtension` reconcilers' integration.

**Which issue this PR fixes**

Fixes https://github.com/Kong/gateway-operator/issues/734.

### DIFF
--- a/test/envtest/dataplane_konnectextensions_test.go
+++ b/test/envtest/dataplane_konnectextensions_test.go
@@ -112,7 +112,7 @@ func TestDataPlaneKonnectExtension(t *testing.T) {
 			StatusCode: http.StatusOK,
 		}, nil)
 
-	t.Logf("Setting pu expected CreateDataplaneCertificate SDK call")
+	t.Logf("Setting up expected CreateDataplaneCertificate SDK call")
 	factory.SDK.DataPlaneCertificatesSDK.EXPECT().CreateDataplaneCertificate(mock.Anything, konnectControlPlaneID, mock.Anything).
 		Return(&sdkkonnectops.CreateDataplaneCertificateResponse{
 			StatusCode: http.StatusCreated,
--- a/test/envtest/dataplane_konnectextensions_test.go
+++ b/test/envtest/dataplane_konnectextensions_test.go
@@ -8,7 +8,6 @@ import (
 
 	sdkkonnectcomp "github.com/Kong/sdk-konnect-go/models/components"
 	sdkkonnectops "github.com/Kong/sdk-konnect-go/models/operations"
-	"github.com/go-logr/logr"
 	"github.com/samber/lo"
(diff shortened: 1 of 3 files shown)


test/envtest/dataplane_konnectextensions_test.go
```diff
@@ -0,0 +1,315 @@
...
+				},
+			}, nil)
+
+	t.Logf("Setting up expected ListDpClientCertificates SDK call returning no certificates")
+	factory.SDK.DataPlaneCertificatesSDK.EXPECT().ListDpClientCertificates(mock.Anything, konnectControlPlaneID).
+		Return(&sdkkonnectops.ListDpClientCertificatesResponse{
+			StatusCode: http.StatusOK,
+		}, nil)
+
+	t.Logf("Setting pu expected CreateDataplaneCertificate SDK call")
```
pmalek:
```suggestion
	t.Logf("Setting up expected CreateDataplaneCertificate SDK call")
```

test/envtest/dataplane_konnectextensions_test.go
```diff
@@ -0,0 +1,315 @@
...
+
+	t.Logf("Creating cluster CA secret")
+	caManager := manager.CAManager{
+		Logger:          logr.Discard(),
+		Client:          cl,
+		SecretName:      clusterCASecretName,
+		SecretNamespace: ns.Name,
+		KeyConfig:       clusterCAKeyConfig,
+	}
+	require.NoError(t, caManager.Start(ctx))
```
pmalek: Since this is only used for creating the secret in https://github.com/Kong/gateway-operator/blob/5fe805419eba5479f5257a60df00a6e35c4bc6e2/controller/konnect/konnectextension_controller_utils.go#L257-L268 WDYT about creating said secret in the test and not starting the CAManager?

This way we get less moving parts in the test and more isolation.

czeslavo: I made a shortcut and used the manager here as it only creates the said secret. But I agree it's better to isolate this and do only what's necessary if the manager gets extended. 👍 I extracted a `CreateClusterCACertificate` function to `secrets` package so it can be used both in the `CAManager` and here. PTAL: https://github.com/Kong/gateway-operator/pull/1404/commits/a4b2f2168d39cfb8b935c3abae49f270477f8f16

pmalek (approved):

PR #1409, opened 2025-03-27
feat: deduce cloud gateway DP group geo based on API auth URL

czeslavo:

Replaces a hardcoded ControlPlane region in Cloud Gateway requests with one deduced from the `KonnectAPIAuthConfiguration`.

**PR Readiness Checklist**:


CHANGELOG.md
```diff
@@ -35,6 +35,9 @@
   a `ReferenceGrant` in the namespace of the `ControlPlane` that allows the `ControlPlane`
   to watch resources in the specified namespace.
   [#1388](https://github.com/Kong/gateway-operator/pull/1388)
+- Deduce `KonnectCloudGatewayDataPlaneGroupConfiguration` region based on the attached
+  `KonnectAPIAuthConfiguration` instead of using a hardcoded `ue` value.
```
```suggestion
  `KonnectAPIAuthConfiguration` instead of using a hardcoded `eu` value.
```

controller/konnect/server/region.go
pmalek, line 1: ~I believe that this slightly duplicates what https://github.com/Kong/sdk-konnect-go/blob/66224491dbdc903b74fa7ac647e72604b6a65635/models/components/entityregion.go does.~

~Do you think we could reuse these consts and that type here?~

Actually, it seems that there's no validation code other than what's in the `UnmarshalJSON()` func, which makes adding validation using that code cumbersome at best (we'd have to construct a JSON with that value and try to validate that). So I guess we'd want to leave the validation code as proposed here.

The only problem though is that we'll have to change this code when new region is supported in CG, which makes me want to consider something else 🤔

pmalek (dismissed):
Just minor nits, overall this looks good 👍

controller/konnect/server/region.go
czeslavo: We have to decide what we prefer:
1. Having little (e.g., verify a region satisfies a two letters regex) or no validation and thanks to that forward compatibility (when new regions are added, old KGO versions can handle them)
2. having strict validation and being forced to update code when new regions are added

The 3rd, the best option is to have good sides of the two worlds simultaneously - have strict validation, and be forward-compatible. But for that, we'd have to have a way to query what are the supported regions from Konnect API and validate against it. As I looked up, there's such an endpoint available (https://global.api.konghq.tech/v2/available-regions or https://global.api.konghq.com/v2/available-regions) but we'd have to take into account this depends on an environment we're using (`.tech` regions differ from `.com`).

I'd suggest going with the first option (loosening the validation to accept regions `global` or satisfying `^[a-z]{2}$` regex) and creating an issue to track using the `/available-regions` endpoint in the future if needed. WDYT @pmalek?

pmalek: That's a good idea 👍 

We'd have to take care when deciding when to check this against the API as we probably don't want to do that on every reconciliation but overall the idea seems good to me.

czeslavo: > We'd have to take care when deciding when to check this against the API as we probably don't want to do that on every reconciliation but overall the idea seems good to me.

Yeah, definitely. I'd go with caching the response in memory for the whole process lifetime. I don't think it's worth refreshing it, as available regions should change very infrequently.

czeslavo: PTAL at [e163767](https://github.com/Kong/gateway-operator/pull/1409/commits/e163767b94352d8aadcee95b89909266396ac081). I also created an issue to track the potential improvement https://github.com/Kong/gateway-operator/issues/1412.

controller/konnect/server/region.go
```diff
@@ -0,0 +1,50 @@
...
+	// RegionEU represents the European region.
+	RegionEU Region = "eu"
+	// RegionUS represents the United States region.
+	RegionUS Region = "us"
+	// RegionAU represents the Australian region.
+	RegionAU Region = "au"
+	// RegionME represents the Middle East region.
+	RegionME Region = "me"
+	// RegionIN represents the Indian region.
+	RegionIN Region = "in"
```
pmalek: These are now unused (apart from some usage in tests) so I guess we can get rid of them. WDYT?

controller/konnect/server/server.go
```diff
@@ -0,0 +1,70 @@
...
+
+	// NOTE: Code below tries to determine if we need a global API endpoint for the given type.
+	// Thus far there's no way to determine that programmatically (easily).
+	var t T
+	// TODO: add more types here that target the global API endpoint.
+	switch any(t).(type) {
+	case konnectv1alpha1.KonnectCloudGatewayNetwork:
+		region = RegionGlobal
+	default:
+	}
```
pmalek: We could define the `region` variable here and move this call to the `default` case below as this result is not needed for global types. WDYT?

```suggestion
	var region Region

	// NOTE: Code below tries to determine if we need a global API endpoint for the given type.
	// Thus far there's no way to determine that programmatically (easily).
	var t T
	// TODO: add more types here that target the global API endpoint.
	switch any(t).(type) {
	case konnectv1alpha1.KonnectCloudGatewayNetwork:
		region = RegionGlobal
	default:
		var err error
		region, err = NewRegion(regionPart)
		if err != nil {
			return Server{}, fmt.Errorf("failed to parse region from hostname: %w", err)
		}
	}
```

controller/konnect/server/region.go
pmalek: 👍

pmalek (approved):

PR #1410, opened 2025-03-27
feat(controlplane): split ClusterRole into ClusterRole and Role to separate permissions when watching a list of namespaces

### LINKED ISSUE #1347: Allow a ControlPlane to watch all namespaces or only specific ones

pmalek:

This PRs adds splitting of `ControlPlane`'s `ClusterRole` into `Role`s and a `ClusterRole` when `spec.watchNamespaces` is specified and `type` is `list` or `own`.

Along with the `Role`s, the operator will also create appropriate `RoleBinding`s.

**Which issue this PR fixes**

Fixes #1347

Adding integration tests is tracked in #1482 to prevent this PR getting out of hand and becoming too big to review.

Chart changes (granting `Role` and `RoleBinding` policy rules) are done in https://github.com/Kong/charts/pull/1295.

Updated API (`WatchNamespaceGrant`): https://github.com/Kong/kubernetes-configuration/pull/403

**PR Readiness Checklist**:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,9 +57,31 @@
 - Added `spec.watchNamespace` field to `ControlPlane` CRD to allow watching resources
   only in the specified namespace.
   When `spec.watchNamespace.type=list` is used, each specified namespace requires
-  a `ReferenceGrant` in the namespace of the `ControlPlane` that allows the `ControlPlane`
-  to watch resources in the specified namespace.
+  a `ReferenceGrant` that allows the `ControlPlane` to watch resources in the specified namespace.
+  For example:
+
+  ```yaml
+  apiVersion: gateway.networking.k8s.io/v1beta1
+  kind: ReferenceGrant
+  metadata:
+    name: ref
+    namespace: namespace-a
+  spec:
+    from:
+    - group: gateway-operator.konghq.com
+      kind: ControlPlane
+      namespace: controlplane-namespace
+    to:
+    - group: ""
+      kind: Namespace
+      name: namespace-a
+  ```
+
+  Aforementioned list is extended with `ControlPlane`'s own namespace which doesn't
+  require said `ReferenceGrant`.
   [#1388](https://github.com/Kong/gateway-operator/pull/1388)
+  [#1410](https://github.com/Kong/gateway-operator/pull/1410)
+
 - Deduce `KonnectCloudGatewayDataPlaneGroupConfiguration` region based on the attached
   `KonnectAPIAuthConfiguration` instead of using a hardcoded `eu` value.
   [#1409](https://github.com/Kong/gateway-operator/pull/1409)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,31 +57,13 @@
 - Added `spec.watchNamespace` field to `ControlPlane` CRD to allow watching resources
   only in the specified namespace.
   When `spec.watchNamespace.type=list` is used, each specified namespace requires
-  a `ReferenceGrant` that allows the `ControlPlane` to watch resources in the specified namespace.
-  For example:
-
-  ```yaml
-  apiVersion: gateway.networking.k8s.io/v1beta1
-  kind: ReferenceGrant
-  metadata:
-    name: ref
-    namespace: namespace-a
-  spec:
-    from:
-    - group: gateway-operator.konghq.com
-      kind: ControlPlane
-      namespace: controlplane-namespace
-    to:
-    - group: ""
-      kind: Namespace
-      name: namespace-a
-  ```
-
+  a `WatchNamespaceGrant` that allows the `ControlPlane` to watch resources in the specified namespace.
   Aforementioned list is extended with `ControlPlane`'s own namespace which doesn't
-  require said `ReferenceGrant`.
+  require said `WatchNamespaceGrant`.
   [#1388](https://github.com/Kong/gateway-operator/pull/1388)
   [#1410](https://github.com/Kong/gateway-operator/pull/1410)
-
+  <!-- TODO: https://github.com/Kong/gateway-operator/issues/1501 add link to guide from documentation. -->
+  For more information on this please see: https://docs.konghq.com/gateway-operator/latest/
 - Deduce `KonnectCloudGatewayDataPlaneGroupConfiguration` region based on the attached
   `KonnectAPIAuthConfiguration` instead of using a hardcoded `eu` value.
   [#1409](https://github.com/Kong/gateway-operator/pull/1409)
--- a/controller/controlplane/controller.go
+++ b/controller/controlplane/controller.go
@@ -135,12 +135,12 @@ func (r *Reconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager) err
 		// reconciliation for all supported ControlPlane objects which use this Role.
 		Watches(
 			&rbacv1.Role{},
-			handler.EnqueueRequestsFromMapFunc(r.listControlPlanesForRoles)).
+			handler.EnqueueRequestsFromMapFunc(listControlPlanesFor[*rbacv1.Role])).
 		// Watch for events on RoleBindings, if any RoleBinding event happens, enqueue
 		// reconciliation for all supported ControlPlane objects which use this RoleBinding.
 		Watches(
 			&rbacv1.RoleBinding{},
-			handler.EnqueueRequestsFromMapFunc(r.listControlPlanesForRoleBindings))
+			handler.EnqueueRequestsFromMapFunc(listControlPlanesFor[*rbacv1.RoleBinding]))
 
 	if r.KonnectEnabled {
 		// Watch for changes in KonnectExtension objects that are referenced by ControlPlane objects.
--- a/controller/controlplane/controller.go
+++ b/controller/controlplane/controller.go
@@ -131,12 +131,12 @@ func (r *Reconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager) err
 		Watches(
 			&gatewayv1beta1.ReferenceGrant{},
 			handler.EnqueueRequestsFromMapFunc(r.listControlPlanesForReferenceGrants)).
-		// watch for events on Roles, if any Role event happen, enqueue
+		// Watch for events on Roles, if any Role event happens, enqueue
 		// reconciliation for all supported ControlPlane objects which use this Role.
 		Watches(
 			&rbacv1.Role{},
 			handler.EnqueueRequestsFromMapFunc(r.listControlPlanesForRoles)).
-		// watch for events on RoleBindings, if any RoleBinding event happen, enqueue
+		// Watch for events on RoleBindings, if any RoleBinding event happens, enqueue
 		// reconciliation for all supported ControlPlane objects which use this RoleBinding.
 		Watches(
 			&rbacv1.RoleBinding{},
--- a/controller/controlplane/controller.go
+++ b/controller/controlplane/controller.go
@@ -363,18 +363,18 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	}
 
 	const (
-		ConditionTypeReferenceGrantsValid   = "ReferenceGrantsValid"
-		ConditionReasonReferenceGrantsValid = "ReferenceGrantsValid"
+		ConditionTypeWatchNamespaceGrantsValid   = "WatchNamespaceGrantsValid"
+		ConditionReasonWatchNamespaceGrantsValid = "WatchNamespaceGrantsValid"
 	)
-	log.Trace(logger, "validating ReferenceGrants exist for the ControlPlane")
-	validatedWatchNamespaces, err := r.validateReferenceGrants(ctx, cp)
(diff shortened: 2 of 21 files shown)


CHANGELOG.md
```diff
@@ -47,9 +47,31 @@
...
+  spec:
+    from:
+    - group: gateway-operator.konghq.com
+      kind: ControlPlane
+      namespace: controlplane-namespace
+    to:
+    - group: ""
+      kind: Namespace
+      name: namespace-a
+  ```
```
mlavacca: I am not sure about having this example directly in the CHANGELOG, as it can become more challenging to navigate. We could add it to the documentation and link the docs page here instead. WDYT?

pkg/utils/kubernetes/reduce/filters.go
```diff
@@ -94,24 +94,56 @@ func filterClusterRoles(clusterRoles []rbacv1.ClusterRole) []rbacv1.ClusterRole
...
 // -----------------------------------------------------------------------------
 
-// filterClusterRoleBindings filters out the ClusterRoleBinding to be kept and returns
-// all the ClusterRoleBindings to be deleted.
-// The filtered-out ClusterRoleBinding is decided as follows:
-// 1. creationTimestamp (older is better)
-func filterClusterRoleBindings(clusterRoleBindings []rbacv1.ClusterRoleBinding) []rbacv1.ClusterRoleBinding {
+// filterBindings filters out the ClusterRoleBinding or RoleBindings to be kept and returns
+// all the objects to be deleted.
+func filterBindings[
```
mlavacca:
```suggestion
func filterRoleBindings[
```

I think that explicitly stating we are talking about roles can help to quickly understand what this function is about.

mlavacca (commented):
What I am getting from this implementation is the following use of the `ReferenceGrant` resource: we are "abstracting" the whole set of resources (services, routes, ingress, etc.) that have to be reconciled by KGO with the Namespace kind. This means that if I want to cross the namespace boundaries with the CP, I need to create a RG very similar to the one you posted in the CHANGELOG, and that will act as a "allow-all-kinds" for the namespace where the RG lives.

Is that correct?

pmalek:
> What I am getting from this implementation is the following use of the `ReferenceGrant` resource: we are "abstracting" the whole set of resources (services, routes, ingress, etc.) that have to be reconciled by KGO with the Namespace kind. This means that if I want to cross the namespace boundaries with the CP, I need to create a RG very similar to the one you posted in the CHANGELOG, and that will act as a "allow-all-kinds" for the namespace where the RG lives.
> 
> Is that correct?

That's correct yes. Exactly as proposed in https://docs.google.com/document/d/1Xr0U5yxRWt0aJf310vmxOsrUx2uLfg3SLMkEASgZIXg/edit?pli=1&tab=t.0#heading=h.1pagv99uxvhp

CHANGELOG.md
pmalek: We'll certainly add a docs section (as a guide for this feature) but I thought that adding it here would help readers understand what sort of `ReferenceGrant` would be needed.

If you feel that's too much I can remove it and add a placeholder for the future docs section link. WDYT?

pkg/utils/kubernetes/reduce/filters.go
pmalek: 95c8d45a6d6fecad44a59ffea5990f475b4c38f6

mlavacca:
> > What I am getting from this implementation is the following use of the `ReferenceGrant` resource: we are "abstracting" the whole set of resources (services, routes, ingress, etc.) that have to be reconciled by KGO with the Namespace kind. This means that if I want to cross the namespace boundaries with the CP, I need to create a RG very similar to the one you posted in the CHANGELOG, and that will act as a "allow-all-kinds" for the namespace where the RG lives.
> > Is that correct?
> 
> That's correct yes. Exactly as proposed in https://docs.google.com/document/d/1Xr0U5yxRWt0aJf310vmxOsrUx2uLfg3SLMkEASgZIXg/edit?pli=1&tab=t.0#heading=h.1pagv99uxvhp

I am sorry I did not catch that earlier, but I think we are misusing the `ReferenceGrant` API here. Having the `ReferenceGrantTo` set to the `Namespace` kind means that the implementation has a grant on the `Namespace` object (which is problematic again, as the namespace is cluster-wide), not on all the objects needed in that namespace. I might be ok to leave it as it is now, but I strongly think we should reconsider and change our approach in the future to have a proper usage of the API.

CHANGELOG.md
mlavacca: yep, if we can add a placeholder and set the docs link here afterward, it would be great 👍🏻

pmalek:
> I might be ok to leave it as it is now, but I strongly think we should reconsider and change our approach in the future to have a proper usage of the API.

Do you have a proposal on how you'd like to this API to look like?

Requiring `ReferenceGrant`s for each type of resource that `ControlPlane` needs to access seems  clunky and would be a bad UX IMHO.

mlavacca:
> Requiring `ReferenceGrant`s for each type of resource that `ControlPlane` needs to access seems clunky and would be a bad UX IMHO.

100% agree. Even considering the target cap to 16, this is definitely a non-viable solution.

> Do you have a proposal on how you'd like to this API to look like?

At the moment, I don't. I think it would require a design of some sort.

Generally speaking, do you agree with this point I'm raising?

pmalek:
We've discussed this with @mlavacca and decided to take a step back and consider changing the API.

We'll bring this up on next team sync.

pmalek:
Updated API (`WatchNamespaceGrant`): https://github.com/Kong/kubernetes-configuration/pull/403

cc: @mlavacca

controller/controlplane/controller_watch.go
```diff
@@ -288,3 +288,81 @@ func (r *Reconciler) listControlPlanesForReferenceGrants(
...
+	if !ok {
+		return nil
+	}
+
+	return []reconcile.Request{
+		{
+			NamespacedName: nn,
+		},
+	}
+}
```
programmer04:
```suggestion
func listControlPlanesFor[T *rbacv1.Role | *rbacv1.RoleBinding](
	ctx context.Context,
	obj client.Object,
) []reconcile.Request {
	if _, ok := obj.(T); !ok {
		ctrllog.FromContext(ctx).Error(
			operatorerrors.ErrUnexpectedObject,
			fmt.Sprintf("failed to map %T on ControlPlane", T(nil)),
			"expected", "RoleBinding", "found", reflect.TypeOf(obj),
		)
		return nil
	}

	nn, ok := isManagedByControlPlane(obj)
	if !ok {
		return nil
	}

	return []reconcile.Request{
		{
			NamespacedName: nn,
		},
	}
}
```

Instead of having two specific method helpers, have one generic function helper. WDYT? 
On the caller side, it'll look like that 

```go
Watches(
			&rbacv1.RoleBinding{},
			handler.EnqueueRequestsFromMapFunc(listControlPlanesFor[*rbacv1.RoleBinding]))
```

pkg/utils/kubernetes/reduce/filters.go
```diff
@@ -94,24 +94,56 @@ func filterClusterRoles(clusterRoles []rbacv1.ClusterRole) []rbacv1.ClusterRole
...
+	}
+
+	best := 0
+	for i, cr := range roles {
+		if cr.CreationTimestamp.After(roles[best].CreationTimestamp.Time) {
+			best = i
+		}
+	}
+
+	return append(roles[:best], roles[best+1:]...)
```
programmer04: There is a stdlib helper for this
```suggestion
	return slices.Delete(roles, best, best+1)
```
But, still modifying the slice passed as an argument may lead to side effects that manifest in the code when the argument and/or returned slice is reused later and modified. 

So the bullet-proof solution is to use 

```go 
return slices.Concat(roles[:best], roles[best+1:])
```

which will allocate a new slice, but we have a dedicated issue to maybe enhance everywhere
- https://github.com/Kong/gateway-operator/issues/146 

> This comment is rather to acknowledge a fact than a call to action

controller/controlplane/controller.go
```diff
@@ -129,7 +130,17 @@ func (r *Reconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager) err
 		// are referenced in a "from" instance.
 		Watches(
 			&gatewayv1beta1.ReferenceGrant{},
-			handler.EnqueueRequestsFromMapFunc(r.listControlPlanesForReferenceGrants))
+			handler.EnqueueRequestsFromMapFunc(r.listControlPlanesForReferenceGrants)).
+		// watch for events on Roles, if any Role event happen, enqueue
```
programmer04:
```suggestion
		// Watch for events on Roles, if any Role event happens, enqueue
```

controller/controlplane/controller_reconciler_utils.go
```diff
@@ -348,12 +356,306 @@ func (r *Reconciler) ensureRoles(
...
+				roles = append(roles, existing)
+				createdOrUpdated = true
+				continue rolesLoop
+			}
+			roles = append(roles, existing)
+		default:
+			if err := k8sreduce.ReduceRoles(ctx, r.Client, existingRoles); err != nil {
+				return false, nil, err
+			}
+			return false, nil, errors.New("number of Roles reduced")
```
programmer04:
```suggestion
			return false, nil, fmt.Errorf("number of Roles reduced from: %d to 1", count)
```

a little bit more of extra context may be beneficial I think

controller/controlplane/controller_reconciler_utils.go
```diff
@@ -348,12 +356,306 @@ func (r *Reconciler) ensureRoles(
...
+			r.Client,
+			client.MatchingLabels(k8sutils.GetManagedByLabelSet(cp)),
+			client.InNamespace(generatedRole.Namespace),
+		)
+		if err != nil {
+			return false, nil, err
+		}
+
+		count := len(existingRoles)
+		switch count {
```
programmer04:
```suggestion
		switch count := len(existingRoles); count {
```

controller/controlplane/controller.go
```diff
@@ -129,7 +130,17 @@ func (r *Reconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager) err
...
+		// reconciliation for all supported ControlPlane objects which use this Role.
+		Watches(
+			&rbacv1.Role{},
+			handler.EnqueueRequestsFromMapFunc(r.listControlPlanesForRoles)).
+		// watch for events on RoleBindings, if any RoleBinding event happen, enqueue
```
programmer04:
```suggestion
		// Watch for events on RoleBindings, if any RoleBinding event happens, enqueue
```

controller/controlplane/controller_reconciler_utils.go
```diff
@@ -348,12 +356,306 @@ func (r *Reconciler) ensureRoles(
...
+		}
+		if createdOrUpdated {
+			log.Debug(logger, "Roles created/updated")
+			return op.Updated, nil // requeue will be triggered by the creation or update of Roles.
+		}
+
+		res, err := r.ensureRoleBindings(ctx, cp, controlplaneServiceAccount, controlplaneRoles)
+		if err != nil {
+			return op.Noop, err
+		} else if res != op.Noop {
```
programmer04:
```suggestion
		}
		if res != op.Noop {
```
to be consitent with what is above and below

programmer04 (commented):
LGTM some minor suggestions

programmer04 (dismissed):

controller/controlplane/controller_watch.go
pmalek: 👍 

Added that in 3f20b733182c2fca371d9c7beffa59d182d75704 (needed to change your snippet ever so slightly to make it compile).

programmer04 (dismissed):

CHANGELOG.md
pmalek: Added in 43874f5a642270e8202a983b2cfbe079cd727113

pmalek:
@programmer04 PTAL once more. Needed to adjust the changelog and sample manifest in https://github.com/Kong/gateway-operator/pull/1410/commits/ef4b9f1621aa5398465614b7d59c53141f8a37f1

programmer04 (approved):

PR #1414, opened 2025-03-28
feat(ke): set `ControlPlaneRefValid` condition to false

mlavacca:

If the `KonnectExtension` references the KE by `NamespacedRef`, but the CP does not exist, the `KonnectExtension` `ControlPlaneRefValid` condition is not properly set. This PR fixes that.

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:

### DIFF
--- a/controller/konnect/konnectextension_controller.go
+++ b/controller/konnect/konnectextension_controller.go
@@ -260,6 +260,17 @@ func (r *KonnectExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
 			return res, err
 		}
 	}
+
+	// get the Konnect Control Plane ID. Set the ControlPlaneRefValid condition accordingly.
+	_, res, err := r.getKonnectControlPlaneID(ctx, ext, readyCondition)
+	if err != nil || !res.IsZero() {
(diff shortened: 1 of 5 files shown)


pmalek (commented):
This looks good. One thing I think is worth pointing out is that we do not have an `envtest` based test to check this functionality which I think we should add (similarly as it's being done in #1404 for DPs). I think it could be done in a separate PR.

What I think should be done in this PR is to add code for the `TestKonnectExtension` integration test to test the changes here. WDYT?

mlavacca:
> This looks good. One thing I think is worth pointing out is that we do not have an `envtest` based test to check this functionality which I think we should add (similarly as it's being done in #1404 for DPs). I think it could be done in a separate PR.
> 
> What I think should be done in this PR is to add code for the `TestKonnectExtension` integration test to test the changes here. WDYT?

I've added an integration test. Let me know if that works for you

pmalek (approved):
👍

pmalek:
ℹ️ I've changed the milestone for this to 1.6 as this is where this will land when merged. We'll want to backport it to 1.5 to get it there.

PR #1435, opened 2025-04-03
chore: split development mode flag into granular ones

### LINKED ISSUE #118: Remove `CONTROLLER_DEVELOPMENT_MODE` in favor of specific settings
## Problem Statement

Using a one-size-fits-all dev mode approach violates one of the 12-factor principles https://12factor.net/dev-prod-parity and we're seeing it cause real-world problems.

## Proposed Solution

1. Get rid of the `CONTROLLER_DEVELOPMENT_MODE` setting in favor of specific settings for specific behaviors.
2. Determine the minimal set of those "specific behaviors" that must be different on CI than in production and only enable those.

## Additional Information

I speculate that allowing non-allowlisted image tags could move from a runtime knob to e.g. an annotation (or some other kind of config as code parameter).

## Acceptance Criteria
- [ ] `CONTROLLER_DEVELOPMENT_MODE` does not exist anymore
- [ ] CI has anonymous reports disabled and everything else in dev mode enabled unless there's a well understood reason for that particular bit to be non-default

czeslavo:

Instead of having a single rule-them-all kind of toggle - `GATEWAY_OPERATOR_DEVELOPMENT_MODE` - introduce granular ones that will allow changing the behavior more precisely depending on a use case needs (e.g., disable image validations in production without changing the logging mode and anonymous reports settings).

Newly introduced flags:

- `--logging-mode` (`GATEWAY_OPERATOR_LOGGING_MODE`) with default `production`
- `--validate-images` (`GATEWAY_OPERATOR_VALIDATE_IMAGES`) with default `true`

**Which issue this PR fixes**

Fixes https://github.com/Kong/gateway-operator/issues/118.

**PR Readiness Checklist**:

### DIFF
--- a/modules/cli/cli.go
+++ b/modules/cli/cli.go
@@ -30,7 +30,7 @@ func New(m metadata.Info) *CLI {
 	var deferCfg flagsForFurtherEvaluation
 
 	flagSet.BoolVar(&cfg.ValidateImages, "validate-images", true, "Validate the images set in ControlPlane and DataPlane specifications.")
-	flagSet.Var(NewValidatedValue[logging.LoggingMode](&cfg.LoggingMode, logging.NewLoggingMode, WithDefault(logging.ProductionMode)), "logging-mode", "Logging mode to use. Possible values: production, development.")
+	flagSet.Var(NewValidatedValue[logging.Mode](&cfg.LoggingMode, logging.NewMode, WithDefault(logging.ProductionMode)), "logging-mode", "Logging mode to use. Possible values: production, development.")
 
 	flagSet.BoolVar(&cfg.AnonymousReports, "anonymous-reports", true, "Send anonymized usage data to help improve Kong.")
 	flagSet.StringVar(&cfg.APIServerPath, "apiserver-host", "", "The Kubernetes API server URL. If not set, the operator will use cluster config discovery.")
--- a/modules/cli/cli.go
+++ b/modules/cli/cli.go
@@ -29,6 +29,9 @@ func New(m metadata.Info) *CLI {
 	}
 	var deferCfg flagsForFurtherEvaluation
 
+	flagSet.BoolVar(&cfg.ValidateImages, "validate-images", true, "Validate the images set in ControlPlane and DataPlane specifications.")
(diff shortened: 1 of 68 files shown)


modules/manager/logging/mode.go
```diff
@@ -0,0 +1,29 @@
+package logging
+
+import "fmt"
+
+// LoggingMode is the type for the logging mode.
+type LoggingMode string
```
pmalek: Nitpick: perhaps we could call this `Mode` given that it's located in the `logging` package. This will prevent the stuttering use sites like: `logging.LoggingMode`

WDYT?

modules/cli/cli.go
```diff
@@ -148,15 +144,6 @@ func (c *CLI) Metadata() metadata.Info {
 // Must be called after all additional flags in the FlagSet() are defined and before flags are accessed
 // by the program. It returns config for controller manager.
 func (c *CLI) Parse(arguments []string) manager.Config {
-	developmentModeEnabled := manager.DefaultConfig().DevelopmentMode
-	// TODO: clean env handling https://github.com/Kong/gateway-operator-archive/issues/19
-	if os.Getenv(envVarFlagPrefix+"DEVELOPMENT_MODE") == "true" ||
-		os.Getenv("CONTROLLER_DEVELOPMENT_MODE") == "true" {
-		developmentModeEnabled = true
-	}
```
pmalek, line 156: I think we should warn our users who still use either of these envs so that they migrate to using the new (introduced in this PR).

How about printing a log message stating that this env is now ineffective and users should use the new envs?

czeslavo: PTAL: https://github.com/Kong/gateway-operator/pull/1435/commits/70046304700aae6bb81b669a2d87c147e0faa080

modules/manager/logging/mode.go
czeslavo: PTAL: https://github.com/Kong/gateway-operator/pull/1435/commits/70046304700aae6bb81b669a2d87c147e0faa080

pmalek (approved):
👍

PR #1436, opened 2025-04-03
chore(refactor): in `KonnectExtension` controller do not fetch `KonnectGatewayControlPlane` from Konnect

### LINKED ISSUE #1275: Fill in the `KonnectGatewayControlPlane` endpoints and use them in the `KonnectExtension` controller
## Problem Statement

We should update the `KonnectGatewayControlPlane` endpoints with the Konnect values and use them in the `KonnectExtension` controller.

## Acceptance Criteria
- [ ] As per description, use the `KonnectGatewayControlPlane` status endpoint instead of fetching the `KonnectGatewayControlPlane` from the SDK.

programmer04:

Do not perform redundant fetches from Konnect API of `KonnectGatewayControlPlane` in `KonnectExtension`. Instead rely on a `KonnectGatewayControlPlane` resource reconciled in K8s, because all needed information is already here. 

**Which issue this PR fixes**

Fixes #1275 

`controller/konnect/konnectextension_controller.go:578` is done this way, because for creation of `KonnectGatewayControlPlane` in K8s [it's not required to fill `spec.cluster_type` field](https://github.com/Kong/kubernetes-configuration/blob/68327eadc8d2d88c9924831f4902144a05234a7d/api/konnect/v1alpha1/konnect_gateway_controlplane_types.go#L46). Hence this field may be empty, but such entity when requested from Konnect has it set to `CLUSTER_TYPE_CONTROL_PLANE` (at least it seems to be like that).

~It can be left as it is. Or a new field `clusterType` in `status` can be introduced for `KonnectGatewayControlPlane` and filled in the helper [handleKonnectGatewayControlPlaneSpecific](https://github.com/Kong/gateway-operator/blob/f984148cb41e14492f1b8cced7a1080e8106f39a/controller/konnect/reconciler_generic_type_specific.go#L60-L81) during reconciliation of `KonnectGatewayControlPlane`.  This solution seems to be better, because it does not need any assumption to consider~
> [see the comment](https://github.com/Kong/gateway-operator/pull/1436/files/f984148cb41e14492f1b8cced7a1080e8106f39a#r2026664014)

### DIFF
--- a/controller/konnect/konnectextension_controller.go
+++ b/controller/konnect/konnectextension_controller.go
@@ -261,11 +261,13 @@ func (r *KonnectExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		}
 	}
 
-	// get the Konnect Control Plane ID. Set the ControlPlaneRefValid condition accordingly.
-	_, res, err := r.getKonnectControlPlaneID(ctx, ext, readyCondition)
+	// Get the GatewayKonnectControlPlane and set conditions accordingly.
+	cp, res, err := r.getGatewayKonnectControlPlane(ctx, ext)
 	if err != nil || !res.IsZero() {
-		// don't return the error here to avoid noise. Status condition is properly set.
-		log.Debug(logger, "controlPlane reference has not properly resolved", "error", err)
+		if !k8serrors.IsNotFound(err) && !errors.Is(err, extensionserrors.ErrKonnectGatewayControlPlaneNotProgrammed) {
+			return res, err
+		}
+		log.Debug(logger, "controlPlane not ready yet")
 		return res, nil
 	}
 
@@ -326,16 +328,6 @@ func (r *KonnectExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		return ctrl.Result{}, fmt.Errorf("failed to parse server URL: %w", err)
 	}
 	sdk := r.SdkFactory.NewKonnectSDK(server, sdkops.SDKToken(token))
-
-	// get the Konnect Control Plane
-	cp, res, err := r.getKonnectControlPlane(ctx, logger, sdk.GetControlPlaneSDK(), ext)
-	if err != nil || !res.IsZero() {
-		if !k8serrors.IsNotFound(err) && !errors.Is(err, extensionserrors.ErrKonnectGatewayControlPlaneNotProgrammed) {
-			return res, err
-		}
-		log.Debug(logger, "controlPlane not ready yet")
-		return res, nil
-	}
 
 	certProvisionedCond := metav1.Condition{
 		Type:    konnectv1alpha1.DataPlaneCertificateProvisionedConditionType,
@@ -404,7 +396,7 @@ func (r *KonnectExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	log.Debug(logger, "DataPlane certificate validity checked")
 
 	// get the list of DataPlane client certificates in Konnect
-	dpCertificates, err := ops.ListKongDataPlaneClientCertificates(ctx, sdk.GetDataPlaneCertificatesSDK(), cp.ID)
+	dpCertificates, err := ops.ListKongDataPlaneClientCertificates(ctx, sdk.GetDataPlaneCertificatesSDK(), cp.Status.ID)
 	if err != nil {
 		certProvisionedCond.Status = metav1.ConditionFalse
 		certProvisionedCond.Reason = konnectv1alpha1.DataPlaneCertificateProvisionedReasonKonnectAPIOpFailed
@@ -470,7 +462,7 @@ func (r *KonnectExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
 					dpCert.Status.Konnect = &konnectv1alpha1.KonnectEntityStatusWithControlPlaneRef{
 						// setting the controlPlane ID in the status as a workaround for the GetControlPlaneID method,
 						// that expects the ControlPlaneID to be set in the status.
-						ControlPlaneID: cp.ID,
+						ControlPlaneID: cp.Status.ID,
 					}
 				},
 			)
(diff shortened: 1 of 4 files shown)


test/helpers/deploy/deploy_resources.go
```diff
@@ -269,6 +269,10 @@ func KonnectGatewayControlPlaneWithID(
 		},
 	}
 	cp.Status.ID = uuid.NewString()[:8]
+	cp.Status.Endpoints = &konnectv1alpha1.KonnectEndpoints{
+		ControlPlaneEndpoint: "cp.endpoint",
+		TelemetryEndpoint:    "tp.endpoint",
+	}
```
pmalek, line 275: I believe we can add using the provided options here (before `Update()`) like so:

```suggestion
	cp.Status.Endpoints = &konnectv1alpha1.KonnectEndpoints{
		ControlPlaneEndpoint: "cp.endpoint",
		TelemetryEndpoint:    "tp.endpoint",
	}
	for _, opt := range opts {
		opt(cp)
	}
```

So that the call site at `test/envtest/dataplane_konnectextensions_test.go` becomes:

```
	cp := deploy.KonnectGatewayControlPlaneWithID(t, ctx, cl, konnectAPIAuthConfiguration, WithKonnectID(konnectControlPlaneID))
	require.NoError(t, cl.Status().Update(ctx, cp))
```

controller/konnect/konnectextension_controller_utils.go
```diff
@@ -33,39 +29,95 @@ import (
...
+		}
+		cp = kgcp
+	case commonv1alpha1.ControlPlaneRefKonnectID:
+		kgcpList := &konnectv1alpha1.KonnectGatewayControlPlaneList{}
+		if err := r.List(ctx, kgcpList, client.InNamespace(ext.Namespace)); err != nil {
+			return nil, ctrl.Result{}, err
+		}
+		kgcps := lo.Filter(kgcpList.Items, func(kgcp konnectv1alpha1.KonnectGatewayControlPlane, _ int) bool {
+			return kgcp.Status.ID == *ext.Spec.Konnect.ControlPlane.Ref.KonnectID
+		})
```
pmalek: I believe we can use the `IndexFieldKonnectGatewayControlPlaneOnKonnectID` index here to make it a little more efficient, right?

controller/konnect/konnectextension_controller_utils.go
```diff
@@ -33,39 +29,95 @@ import (
...
+func (r *KonnectExtensionReconciler) getGatewayKonnectControlPlane(
 	ctx context.Context,
-	logger logr.Logger,
-	sdk sdkops.ControlPlaneSDK,
 	ext konnectv1alpha1.KonnectExtension,
 	dependingConditions ...metav1.Condition,
-) (cp *sdkkonnectcomp.ControlPlane, res ctrl.Result, err error) {
+) (cp *konnectv1alpha1.KonnectGatewayControlPlane, res ctrl.Result, err error) {
+	// Get respective KonnectGatewayControlPlane from K8s cluster.
+	var getFromK8sErr error
```
pmalek: According to https://go.dev/wiki/Errors#naming error variables should start with `err` or `Err`.

controller/konnect/konnectextension_controller.go
```diff
@@ -583,12 +575,22 @@ func (r *KonnectExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
 	if ext.Status.Konnect == nil {
+		// Since KonnectControlPlane can be configured in K8s without specifying the cluster type,
+		// this field may be empty and in such a case in Konnect it's set to CLUSTER_TYPE_CONTROL_PLANE
+		// hence assume this as the default value.
+		clusterType := konnectv1alpha1.ClusterTypeControlPlane
+		if cp.Spec.ClusterType != nil {
+			clusterType = konnectClusterTypeToCRDClusterType(
+				sdkkonnectcomp.ControlPlaneClusterType(*cp.Spec.ClusterType),
+			)
+		}
```
pmalek: Perhaps we could refactor that so that all this is handled inside `konnectClusterTypeToCRDClusterType`, WDYT?

controller/konnect/konnectextension_controller_utils.go
```diff
@@ -33,39 +29,95 @@ import (
...
 
-	// get the Konnect Control Plane from Konnect
-	konnectCP, err := ops.GetControlPlaneByID(ctx, sdk, konnectCPID)
-	// set the controlPlaneRefValidCond to false in case the Control Plane is not found in Konnect
-	if err != nil {
+	// Set the controlPlaneRefValidCond to false in case the KonnectGatewayControlPlane is not programmed yet.
+	if !lo.ContainsBy(cp.Status.Conditions, func(cond metav1.Condition) bool {
+		return cond.Type == konnectv1alpha1.KonnectEntityProgrammedConditionType &&
+			cond.Status == metav1.ConditionTrue
+	}) {
```
pmalek: For this we have `HasConditionTrue` and `HasConditionFalse` https://github.com/Kong/gateway-operator/blob/f769bfc62d6945f526b87abe668a989428842e14/pkg/utils/kubernetes/status.go#L85-L93

controller/konnect/konnectextension_controller.go
programmer04: Ok, so you're leaning towards doing this way, instead of the proposition articulated in _Special notes for your reviewer_? If yes, I agree it makes sense

pmalek: Yup, it does make sense given that an alternative would be to add a net new field to the status which can be extrapolated from the spec field (as it's done via `konnectClusterTypeToCRDClusterType`)

programmer04: OK, so since we're sure that this approach always works. let's do it this way

controller/konnect/konnectextension_controller_utils.go
```diff
@@ -18,54 +17,106 @@ import (
...
+	k8sutils "github.com/kong/gateway-operator/pkg/utils/kubernetes"
 
 	commonv1alpha1 "github.com/kong/kubernetes-configuration/api/common/v1alpha1"
 	operatorv1beta1 "github.com/kong/kubernetes-configuration/api/gateway-operator/v1beta1"
 	"github.com/kong/kubernetes-configuration/api/konnect"
 	konnectv1alpha1 "github.com/kong/kubernetes-configuration/api/konnect/v1alpha1"
 )
 
-// getKonnectControlPlane retrieves the Konnect Control Plane based on the provided KonnectExtension specification.
+// getGatewayKonnectControlPlane retrieves the Konnect Control Plane form K8s cluster
```
pmalek:
```suggestion
// getGatewayKonnectControlPlane retrieves the Konnect Control Plane from K8s cluster
```

pmalek (dismissed):
1 minor nit. Otherwise LGTM

pmalek (approved):

PR #1440, opened 2025-04-07
chore: bump `oras-go`, fix typos, use `slices` from stdlib

programmer04:

This PR improves a bunch of nits
- Still, there is no stable release of `oras-go` with required changes, but it shouldn't stop us from keeping it up to date
- Some typos in the source code have been spotted and fixed
- Use a standard library `slice` instead of `k8s.io/utils/strings/slices`


pmalek (approved):
🎖️ 


Do you use a specific linter / config to catch these typos?

programmer04:
https://cspell.org with a VS code extenison

PR #1489, opened 2025-04-14
feat(konnect): transit gateway reconciler

### LINKED ISSUE #1220: Add `KonnectCloudGatewayTransitGateway` reconciler
## Problem Statement

Add `KonnectCloudGatewayTransitGateway` reconciler to support creating/deleting Konnect transit gateways.

## Proposed Solution

Add a reconciles in the Konnect entity reconcilers to create/get/delete Konnect transit gateways when reconcile requests for `KonnectCloudGatewayTransitGateway` comes. This includes:
- Resolve network reference for `KonnectCloudGatewayTransitGateway` CR
- Implement ops for create/get/delete transit gateways in Konnect
- Update status and finalizers based on the state of the transit gateway in Konnect

## Additional Information

...

## Acceptance Criteria
- [ ] `KonnectCloudGatewayTransitGateway` CRs are reconciled by added reconciler

randmonkey:

Add the controller to reconcile `KonnectCloudGatewayTransitGateway`s.

**Which issue this PR fixes**

Fixes #1220 

**PR Readiness Checklist**:


controller/konnect/reconciler_generic.go
pmalek, line 1: 👍

controller/konnect/reconciler_apiauthref.go
```diff
@@ -152,6 +153,39 @@ func getAPIAuthRefNN[T constraints.SupportedKonnectEntityType, TEnt constraints.
 		return getCPAuthRefForRef(ctx, cl, cpRef, ent.GetNamespace())
 	}
 
+	// If the entity has a NetworkRef, get the KonnectAPIAuthConfiguration
+	// ref from the referenced KonnectCloudGatewayNetwork.
+	// TODO: This is used for reconciling transit gateways.
+	// Should we define interfaces to contrain the method only applies to KonnectCloudGatewayTransitGateway?
```
pmalek: Can you elaborate? I think this is OK for now.

controller/konnect/ops/sdk/konnecttransitgateway.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"context"
+
+	sdkkonnectcomp "github.com/Kong/sdk-konnect-go/models/components"
+	sdkkonnectops "github.com/Kong/sdk-konnect-go/models/operations"
+)
+
+// TransitGatewaysSDK is the interface for operating Konnect transit gateways.
+// REVIEW: define the transit gateway methods as a dedicated interface or add the methods to the `CloudGatewaysSDK` interface?
```
pmalek: These are all defined on `CloudGateways` type in the SDK https://github.com/Kong/sdk-konnect-go/blob/7b5878290ac230c0f1668f734af51fa6c073d3f7/cloudgateways.go#L5031 so I believe we should stick to that (given that networks and configurations already use this approach: https://github.com/Kong/gateway-operator/blob/ea346f51bd54c3c8ec3fe0dc17dd75c13c1679ea/controller/konnect/ops/sdk/dedicatedcloudgateways.go#L11-L21)

controller/konnect/ops/ops_konnecttransitgateway.go
```diff
@@ -0,0 +1,246 @@
...
+}
+
+func extractNameFromTransitGatewayResponse(resp *sdkkonnectcomp.TransitGatewayResponse) string {
+	switch resp.Type {
+	case sdkkonnectcomp.TransitGatewayResponseTypeAwsTransitGatewayResponse:
+		return resp.AwsTransitGatewayResponse.Name
+	case sdkkonnectcomp.TransitGatewayResponseTypeAzureTransitGatewayResponse:
+		return resp.AzureTransitGatewayResponse.Name
+	case sdkkonnectcomp.TransitGatewayResponseTypeAwsVpcPeeringGatewayResponse:
+		// AWS VPC peering gateway is not supported yet.
```
pmalek, line 235: AFAIR this will be caught on the CRD level so I think this is worth mentioning here so that future readers to not need to think: "Is this validated somewhere?"

controller/konnect/ops/ops_konnecttransitgateway.go
```diff
@@ -0,0 +1,246 @@
...
+	})
+
+	if err != nil {
+		return "", fmt.Errorf("failed listing %s: %w", tg.GetTypeName(), err)
+	}
+
+	if resp == nil || resp.ListTransitGatewaysResponse == nil {
+		return "", fmt.Errorf("failed listing %s: %w", tg.GetTypeName(), ErrNilResponse)
+	}
+	// TODO: update KonnectCloudGatewayTransitGateway to satisfy the GetID() interface to reuse getMatchingEntryFromListResponseData?
```
pmalek: 👍 on defining `GetID()` to reuse `getMatchingEntryFromListResponseData` here

controller/konnect/ops/ops_konnecttransitgateway.go
```diff
@@ -0,0 +1,246 @@
...
+	networkID := tg.GetNetworkID()
+	if networkID == "" {
+		return CantPerformOperationWithoutNetworkIDError{
+			Entity: tg,
+			Op:     CreateOp,
+		}
+	}
+	// Failures on creating transit gateway causes `sdk.CreateTransitGateway` stuck because the SDK performs a retry that can take up to 1 hour:
+	// https://github.com/Kong/gateway-operator/issues/1521
+	// So we disable the retry here. The reconciler retries the creation by the requeue mechanism of the controller runtime.
```
pmalek: I think we can make these comments shorter (leaving the link to the issue). Simply saying: "We need to set the empty retry config to prevent using the default retry which causes failures and blocks the reconciliation". 

WDYT?

controller/konnect/ops/ops_konnecttransitgateway.go
randmonkey: Sounds good. Since it requires updates on the `KonnectCloudGatewayTransitGateway` CRD, I would like to make the change in another PR. Created https://github.com/Kong/gateway-operator/issues/1527 to track.
WDYT if we reuse the `GetKonnectID()` in the interface used in `getMatchingEntryFromListResponseData` so we do not need to change the CRD?

controller/konnect/reconciler_apiauthref.go
randmonkey: There may be other CRs having `networkRef`s but does not fit the handling process (like `KonnectCloudGatewayDataPlaneConfiguration`). I am not sure if we want to skip it for these types.

controller/konnect/ops/sdk/konnecttransitgateway.go
randmonkey: Done. Moved the methods to `CloudGatewaysSDK` interface.

controller/konnect/ops/ops_konnecttransitgateway.go
randmonkey: Done.

controller/konnect/ops/ops_konnecttransitgateway.go
randmonkey: done.

controller/konnect/ops/ops_konnecttransitgateway.go
pmalek: I see that this is a bit of a mess now that we have both `GetID() string` and `GetKonnectID() string` defined on types (for some both, for some 1 of these).

I'd suggest that we try to go with `GetID()` for now (by adding it to `KonnectCloudGatewayTransitGateway`) and perhaps trying to simplify this (by removing `GetKonnectID()`? I'm not sure at this point).

Trying to use the latter now in `getMatchingEntryFromListResponseData()` will yield errors because not all types that are currently used with this `func` have this method defined, hence the proposal to use `GetID()` for now.

pmalek (approved):
LGTM 👍 .

Given that we'll address #1527 in 1.6 I'm ok merging this as is.

PR #1494, opened 2025-04-15
tests: add integration tests for watchNamespaces in ControlPlanes

pmalek:
**What this PR does / why we need it**:

**Which issue this PR fixes**

Fixes #1482

**Special notes for your reviewer**:

- [x] Requires changes from https://github.com/Kong/charts/pull/1299 to pass the e2e helm upgrade test.

- [x] ~Blocked by #1410~

- [x] Uses API changes from https://github.com/Kong/kubernetes-configuration/pull/412. Please review that one first 

**PR Readiness Checklist**:

Complete these before marking the PR as `ready to review`:

- [ ] the `CHANGELOG.md` release notes have been updated to reflect significant changes

PR #1496, opened 2025-04-16
feat: implement mirror KonnectGatewayControlPlanes

### LINKED ISSUE #1470: Support `mirror` `KonnectGatewayControlplane`s
## Problem Statement

When the new type field in `KonnectGatewayControlplane`s is set to `mirror`, the reconciliation logic must be implemented in a way that Konnect is the source of truth and the Kubernetes resource is treated as its mirror.

## Proposed Solution

In case the type field is `mirror`, the `KonnectGatewayControlPlane` controller:
- gets the `KonnectID` field from the `KonnectGatewayControlPlane` `MirrorSpec` field
- fetches the `ControlPlane` from Konnect by using the ID specified above and uses that resource as the source of truth
- sets the mirror-related condition (https://github.com/Kong/kubernetes-configuration/issues/384) according to the outcome of the fetch operation from Konnect

## Acceptance Criteria
- [ ] Setting the `Mirror` field corresponds to a mirroring behavior when operating with `KonnectGatewayControlPlane`s

mlavacca:

**Which issue this PR fixes**

Fixes #1470 

**PR Readiness Checklist**:


controller/konnect/ops/conditions.go
```diff
@@ -40,23 +41,58 @@ func SetKonnectEntityProgrammedConditionFalse(
...
 		obj,
+		konnectv1alpha1.KonnectEntityProgrammedConditionType,
 		metav1.ConditionFalse,
 		reason,
 		err.Error(),
 	)
 }
 
-func _setKonnectEntityProgrammedConditon(
+// SetKonnectEntityMirroredConditionTrue sets the KonnectEntityProgrammed condition to true
```
randmonkey: Seems some copy-pasta error on the comments here.

controller/konnect/ops/conditions.go
```diff
@@ -40,23 +41,58 @@ func SetKonnectEntityProgrammedConditionFalse(
...
+	_setKonnectEntityConditon(
+		obj,
+		konnectv1alpha1.ControlPlaneMirroredConditionType,
+		metav1.ConditionTrue,
+		konnectv1alpha1.ControlPlaneMirroredReasonMirrored,
+		"",
+	)
+}
+
+// SetKonnectEntityMirroredConditionFalse sets the KonnectEntityProgrammed condition
```
randmonkey: Ditto here.

controller/konnect/ops/ops.go
```diff
@@ -215,6 +217,19 @@ func Create[
 			time.Since(start),
 		)
 	}
+
+	// For ControlPlane entities, we need to check if the source type is mirror
+	// and set the mirrored condition accordingly.
+	if cp, ok := any(e).(*konnectv1alpha1.KonnectGatewayControlPlane); ok {
+		if *cp.Spec.Source == commonv1alpha1.EntitySourceMirror {
```
randmonkey: The `spec.Source` is optional (and legacy cps does not have such a field), I think it is clearer and less error prone to define some utility functions to get or judge the entity source.

controller/konnect/ops/ops.go
```diff
@@ -394,7 +409,10 @@ func Update[
 	)
 	switch ent := any(e).(type) {
 	case *konnectv1alpha1.KonnectGatewayControlPlane:
-		err = updateControlPlane(ctx, sdk.GetControlPlaneSDK(), sdk.GetControlPlaneGroupSDK(), cl, ent)
+		// if the ControlPlane is of type origin, enforce the spec on Konnect.
+		if *ent.Spec.Source == commonv1alpha1.EntitySourceOrigin {
```
randmonkey, line 419: Ditto here.

controller/konnect/ops/ops_controlplane.go
```diff
@@ -34,6 +35,36 @@ func convertCreateControlPlaneRequestToSDK(
...
+			sdk,
+			// not nilness is ensured by CEL rules
+			string(cp.Spec.Mirror.Konnect.ID),
+		); err != nil {
+			return err
+		}
+		cp.SetKonnectID(string(cp.Spec.Mirror.Konnect.ID))
+		return nil
+	default:
+		// This should never happen, as the source type is validated by CEL rules.
```
randmonkey, line 63: There would be legacy cps that does not have this field so this may happen. Could you please test running it with cps created based on the older version of CRD without the `spec.source` field? This may bring breaking changes,

randmonkey (commented):
Please note that `spec.source` field does not exist in the older version of `KonnectGatewayControlPlane`s. So please make sure what is the behavior of the new code running with existing `KonnectGatewayControlPlane`s and confirm if there are any breaking changes.

controller/konnect/ops/ops_controlplane.go
mlavacca: It cannot happen, as the source field is defaulted to `Source` if it does not exist. When upgrading the CRDs, the old CPs get automatically updated to be configured as source. For this reason, this is not a breaking change.

controller/konnect/ops/ops.go
mlavacca: The `spec.source` is optional but defaulted. All the CPs will have this field set after CRDs upgrade. I've added a helper nonetheless

mlavacca:
> Please note that `spec.source` field does not exist in the older version of `KonnectGatewayControlPlane`s. So please make sure what is the behavior of the new code running with existing `KonnectGatewayControlPlane`s and confirm if there are any breaking changes.

As replied inline, this is not a breaking change. The new field is optional and defaulted to origin. This means that all the old CPs will have the source field set to `origin` after the CRD upgrade

controller/controlplane/controller.go
```diff
@@ -373,7 +373,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 			k8sutils.NewConditionWithGeneration(
 				ConditionTypeWatchNamespaceGrantsValid,
 				metav1.ConditionFalse,
-				kcfgcontrolplane.ConditionReasonMissingReferenceGrant,
+				kcfgcontrolplane.ConditionReasonWatchNamespaceGrantInvalid,
```
pmalek: This (as well as condition definition above and used conditions below) is handled in https://github.com/Kong/gateway-operator/pull/1494/files#diff-64ee605326f88832bc39c8e4902efd8f4fe1c5072d0d6fa5476cf4d1f113f102.

I'm fine leaving this as is but we can also leave #1494 to handle it.

controller/konnect/ops/ops.go
```diff
@@ -219,6 +221,19 @@ func Create[
...
+	// and set the mirrored condition accordingly.
+	if cp, ok := any(e).(*konnectv1alpha1.KonnectGatewayControlPlane); ok {
+		if isMirrorControlPlane(cp) {
+			if err == nil {
+				SetKonnectEntityMirroredConditionTrue(e)
+			} else {
+				SetKonnectEntityMirroredConditionFalse(e, err)
+			}
+		}
+	}
```
pmalek: Could we put this into an extracted function (like `handleMirrorEntities`)? It will get easier in the future to add handling of new types that support mirror type.

controller/konnect/ops/ops.go
```diff
@@ -486,6 +504,19 @@ func Update[
...
```
pmalek: Ditto as above. Extracting will allow to reduce the code added.

controller/konnect/ops/ops.go
```diff
@@ -511,6 +542,15 @@ func logOpComplete[
...
 	TEnt constraints.EntityType[T],
 ](ctx context.Context, start time.Time, op Op, e TEnt, err error) {
+
+	// if the entity is a Mirror KonnectGatewayControlPlane, don't log the konnect operation,
+	// as no operation occurred.
+	if cp, ok := any(e).(*konnectv1alpha1.KonnectGatewayControlPlane); ok {
+		if isMirrorControlPlane(cp) {
+			return
+		}
+	}
```
pmalek: It would be nice to add a function which checks `isMirror(e)` and returns `bool`.

controller/controlplane/controller.go
mlavacca: Good to know, I needed to change it as I had to bump kconf. Let's wait for #1494 to be merged first 👍🏻

controller/konnect/ops/ops.go
mlavacca: I've improved the helpers according to your suggestions. Let me know what you think :)

pmalek (dismissed):

randmonkey (dismissed):

mlavacca:
Waiting for #1494 to be merged first

pmalek:
@mlavacca This is now unblocked

controller/konnect/ops/ops.go
```diff
@@ -669,3 +709,31 @@ func ClearInstanceFromError(err error) error {
...
+	T constraints.SupportedKonnectEntityType,
+	TEnt constraints.EntityType[T],
+](ent TEnt) bool {
+	switch cp := any(ent).(type) {
+	case *konnectv1alpha1.KonnectGatewayControlPlane:
+		return cp.Spec.Source != nil && *cp.Spec.Source == commonv1alpha1.EntitySourceMirror
+	default:
+		return false
+	}
+}
```
pmalek, line 739: Let's continue with this approach now but I believe this should be placed in kubernetes-configuration, ideally generating a method on each type that supports being mirrored.

This will become more important when we add other types that will fall into this category.

pmalek (approved):

controller/konnect/ops/ops.go
mlavacca: I agree, let's keep that in mind when we provide mirroring for additional entities 👍🏻

PR #1502, opened 2025-04-16
tests: add support for KONG_TEST_DISABLE_CRD_INSTALL which disables installing CRDs in integration and conformance tests

pmalek:

Useful in local development when we don't need to keep re-installing the CRDs over and over again.

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:

### DIFF
--- a/pkg/utils/test/setup_helpers.go
+++ b/pkg/utils/test/setup_helpers.go
@@ -26,6 +26,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
 	"github.com/kong/gateway-operator/modules/manager"
+	"github.com/kong/gateway-operator/test"
 
 	operatorclient "github.com/kong/kubernetes-configuration/pkg/clientset"
 )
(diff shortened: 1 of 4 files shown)


pkg/utils/test/setup_helpers.go
```diff
@@ -213,6 +214,10 @@ func ExtractModuleVersion(moduleName string) (string, error) {
 
 // DeployCRDs deploys the CRDs commonly used in tests.
 func DeployCRDs(ctx context.Context, crdPath string, operatorClient *operatorclient.Clientset, cluster clusters.Cluster) error {
+	if test.IsInstallingCRDsDisabled() {
```
mlavacca: Can we perform this check in the `DeployCRDs` caller rather than inside this function? I think it'd be easier to navigate.

pmalek: 👍 cf14ea85eda6ae5d26895218b2eb413284fa59ba. PTAL

mlavacca (approved):

PR #1535, opened 2025-04-24
feat(konnect): filter name in listing transit gateways and reuse getMatchingEntryFromListResponseData to extract Konnect ID

### LINKED ISSUE #1527: Reuse getMatchingEntryFromListResponseData for transit gateways to match names
## Problem Statement

For matching existing transit gateways in Konnect with `KonnectCloudGatewayTransitGateway` CRs, we need to match the names in Konnect with `spec.name` fields of CRs. The getMatchingEntryFromListResponseData can be reused but `KonnectCloudGatewayTransitGateway` does not have the `GetID` method so we cannot directly reuse it.

## Proposed Solution

Either one of the following:

- Update `KonnectCloudGatewayTransitGateway` CRD to implement the `GetID` interface
- Resue `GetKonnectID` method in the `getMatchingEntryFromListResponseData`

## Additional Information

...

## Acceptance Criteria
- [ ] Matching names for transit gateways reuses the `getMatchingEntryFromListResponseData` method

randmonkey:
filter name in listing transit gateways and extract ID of listing responses to reuse `getMatchingEntryFromListResponseData` to extract Konnect ID of transit gateways.
**Which issue this PR fixes**

Fixes #1527

**PR Readiness Checklist**:

~- [x] the `CHANGELOG.md` release notes have been updated to reflect significant changes~
No need to update CHANGELOG?

### DIFF
--- a/controller/konnect/ops/ops.go
+++ b/controller/konnect/ops/ops.go
@@ -594,6 +594,17 @@ func sliceToEntityWithIDSlice[
 	return result
 }
 
+// extractedEntityID represents the ID of an entity.
+// The type is defined to satisfy the interface in the parameter of getMatchingEntryFromListResponseData.
+// Some types of responses does not provide a GetID in sdk-konnect-go
+// so we need to extract the ID and return this type for getMatchingEntryFromListResponseData.
+type extractedEntityID string
+
+// GetID returns the extracted ID. Implements the interface in the parameter of getMatchingEntryFromListResponseData.
+func (e extractedEntityID) GetID() string {
+	return string(e)
+}
+
 // getMatchingEntryFromListResponseData returns the ID of the first entry in the list response data.
 // It returns an error if no entry with a non-empty ID was found.
 // It is used in conjunction with the list operation to get the ID of the entity that matches the UID
--- a/controller/konnect/ops/ops.go
+++ b/controller/konnect/ops/ops.go
@@ -596,7 +596,7 @@ func sliceToEntityWithIDSlice[
 
 // extractedEntityID represents the ID of an entity.
 // The type is defined to satisfy the interface in the parameter of getMatchingEntryFromListResponseData.
-// Some types of responses does not provide a GetID in sdk-konnect-go
+// Some types of responses do not provide a GetID in sdk-konnect-go
 // so we need to extract the ID and return this type for getMatchingEntryFromListResponseData.
 type extractedEntityID string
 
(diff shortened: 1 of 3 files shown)


pmalek (commented):
Didn't we want to add the `GetID() string` to `KonnectCloudGatewayTransitGateway` in kcfg? That would prevent the need for new type and conversion between `KonnectCloudGatewayTransitGateway` and `extractedEntityID`.

WDYT @randmonkey ?

> No need to update CHANGELOG?

I don't think we need to update the changelog as that's not a user facing change.

randmonkey:
> Didn't we want to add the `GetID() string` to `KonnectCloudGatewayTransitGateway` in kcfg? That would prevent the need for new type and conversion between `KonnectCloudGatewayTransitGateway` and `extractedEntityID`.
> 
> WDYT @randmonkey ?
> 
> > No need to update CHANGELOG?
> 
> I don't think we need to update the changelog as that's not a user facing change.

I checked with the `getMatchingEntryFromListResponseData` function, this does not apply on the types in kcfg but on the responses of Konnect entities in `sdk-konnect-go`. So do you think we should update `sdk-konnect-go` to add `GetID` method in `TransitGatewayResponse`?

Looks llike the `GetID` is generated somewhere and the content is `return resp.ID`. The generating method is not available for `TransitGatewayResponse` so the `GetID` cannot be generated. Should we implement that manually? @pmalek

controller/konnect/ops/ops.go
```diff
@@ -594,6 +594,17 @@ func sliceToEntityWithIDSlice[
 	return result
 }
 
+// extractedEntityID represents the ID of an entity.
+// The type is defined to satisfy the interface in the parameter of getMatchingEntryFromListResponseData.
+// Some types of responses does not provide a GetID in sdk-konnect-go
```
pmalek:
```suggestion
// Some types of responses do not provide a GetID in sdk-konnect-go
```

pmalek:
> getMatchingEntryFromListResponseData

I see. The problem here is that the ID is not stored directly in the [response type](https://github.com/Kong/sdk-konnect-go/blob/08223d25420edd301cacdad9a94630a4114cd829/models/operations/listtransitgateways.go#L53-L62) but in either of the [union fields](https://github.com/Kong/sdk-konnect-go/blob/03910600178e6e6b3c5f97ae8ff18dc5a41edf73/models/components/transitgatewayresponse.go#L20-L22) depending on the `.type`.

In that case I believe it makes sense to continue with the proposed approach.

Thanks for looking into this 🙇

pmalek (approved):

PR #1540, opened 2025-04-24
fix: dp deploy can be restart

### LINKED ISSUE #1390: Restarting Dataplane deployment should result in new dataplane pods to be created
## Problem Statement
When we restart Deployment resource created by operator for a DataPlane, restarting it results in new ReplicaSet be created but controller set's it replica count to 0,  thus not let the user restart DataPlane pods in a graceful manner.

## Setup Information
kgo version is 1.4.2

## Replication
- Install kgo 
- create a dataplane
- use  kubectl rollout restart deployment to restart deployment

tao12345666333:

**Which issue this PR fixes**

Fixes https://github.com/Kong/gateway-operator/issues/1390

My test cases cover the following scenarios:

- Test scaling replicas from 2 to 3.
- Simulate a restart operation.
- Test scaling replicas from 3 to 4.

**PR Readiness Checklist**:

### DIFF
--- a/controller/dataplane/owned_deployment.go
+++ b/controller/dataplane/owned_deployment.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"path/filepath"
+	"time"
 
 	"github.com/go-logr/logr"
 	"github.com/google/go-cmp/cmp"
@@ -17,6 +18,7 @@ import (
 	"github.com/kong/gateway-operator/controller/pkg/log"
 	"github.com/kong/gateway-operator/controller/pkg/op"
 	"github.com/kong/gateway-operator/controller/pkg/patch"
+	"github.com/kong/gateway-operator/controller/pkg/utils"
 	"github.com/kong/gateway-operator/internal/utils/config"
 	"github.com/kong/gateway-operator/internal/versions"
 	"github.com/kong/gateway-operator/pkg/consts"
@@ -26,6 +28,13 @@ import (
 
 	operatorv1beta1 "github.com/kong/kubernetes-configuration/api/gateway-operator/v1beta1"
 )
+
+// restartAnnotationKey is the annotation key used to mark a Deployment as restarted.
+// This is used to detect if a Deployment was restarted using `kubectl rollout restart`.
+// The value is a timestamp in RFC3339 format.
+// It's hardcoded here to match the annotation used by the kubectl command.
+// https://github.com/kubernetes/kubernetes/blob/82db38a23c7820b1924d89f458fd368023f3980c/staging/src/k8s.io/kubectl/pkg/polymorphichelpers/objectrestarter.go#L51
+const restartAnnotationKey = "kubectl.kubernetes.io/restartedAt"
 
 // DeploymentBuilder builds a Deployment for a DataPlane.
 type DeploymentBuilder struct {
@@ -244,6 +253,46 @@ func listOrReduceDataPlaneDeployments(
 	return false, &deployments[0], nil
 }
 
+func podTemplateSpecHasRestartAnnotation(template *corev1.PodTemplateSpec) (string, bool) {
+	if template == nil || template.Annotations == nil {
+		return "", false
+	}
+	v, ok := template.Annotations[restartAnnotationKey]
+	return v, ok && v != ""
+}
+
+// isRecentDeploymentRestart detects if a deployment is undergoing a recent restart operation.
+// It checks for the presence of the kubectl restart annotation and verifies if it's recent (within 5 minutes).
+// Returns the restart timestamp if present, and a boolean indicating if a recent restart is detected.
+func isRecentDeploymentRestart(template *corev1.PodTemplateSpec, logger logr.Logger) (string, bool) {
+	// Check if we have a fresh restart annotation (from the last 5 minutes)
+	// This prevents us from treating all deployments with old restart annotations
+	// as perpetually in a restart state, fixing issue #1390
+	restartTimeStr, hasRestartAnnotation := podTemplateSpecHasRestartAnnotation(template)
+	if !hasRestartAnnotation {
+		return "", false
+	}
+
+	// Only treat it as a restart if the timestamp is recent (within 5 minutes)
+	restartTime, err := time.Parse(time.RFC3339, restartTimeStr)
+	if err != nil {
+		log.Debug(logger, "detected restart with unparseable timestamp", "timestamp", restartTimeStr)
+		return restartTimeStr, true // If we can't parse time, assume it's a restart for safety
+	}
+
+	// Check if restart annotation is less than 5 minutes old
+	// We use 5m here as threshold as that correlates with constants used by Kubernetes itself.
+	// Restart annotations older than this duration shouldn't cause a reconciliation effect.
+	// See: https://github.com/kubernetes/kubernetes/blob/82db38a23c7820b1924d89f458fd368023f3980c/pkg/controller/namespace/config/v1alpha1/defaults.go#L41
+	if time.Since(restartTime) < 5*time.Minute {
+		log.Debug(logger, "detected recent restart operation", "isRestartOperation", true, "restartTime", restartTime)
+		return restartTimeStr, true
+	}
+
+	log.Debug(logger, "found old restart annotation, not treating as restart", "restartTime", restartTime)
+	return restartTimeStr, false
+}
+
 // reconcileDataPlaneDeployment takes any existing DataPlane Deployment and a desired DataPlane Deployment and
 // reconciles the existing state to the desired state by either updating an existing Deployment, creating a new one,
 // or doing nothing.
@@ -279,17 +328,30 @@ func reconcileDataPlaneDeployment(
 
 		k8sresources.SetDefaultsPodTemplateSpec(&desired.Spec.Template)
 
+		// Keep track of this for logging purposes
+		originalReplicaCount := existing.Spec.Replicas
+
 		// ensure that object metadata is up to date
 		updated, existing.ObjectMeta = k8sutils.EnsureObjectMetaIsUpdated(existing.ObjectMeta, desired.ObjectMeta)
 
-		// some custom comparison rules are needed for some PodTemplateSpec sub-attributes, in particular
-		// resources and affinity.
+		// some custom comparison rules are needed for some PodTemplateSpec sub-attributes
 		opts := []cmp.Option{
 			cmp.Comparer(k8sresources.ResourceRequirementsEqual),
+			utils.IgnoreAnnotationKeysComparer(restartAnnotationKey),
 		}
 
-		// ensure that PodTemplateSpec is up to date
 		if !cmp.Equal(existing.Spec.Template, desired.Spec.Template, opts...) {
+			restartTimeStr, isRestartOperation := isRecentDeploymentRestart(&existing.Spec.Template, logger)
+			if isRestartOperation {
+				log.Debug(logger, "found restart annotation", "timestamp", restartTimeStr)
+				// Preserve the restart annotation
+				if desired.Spec.Template.Annotations == nil {
+					desired.Spec.Template.Annotations = make(map[string]string)
+				}
+				// Use the restart timestamp we already retrieved
+				desired.Spec.Template.Annotations[restartAnnotationKey] = restartTimeStr
+			}
+
 			existing.Spec.Template = desired.Spec.Template
 			updated = true
 		}
@@ -300,6 +362,8 @@ func reconcileDataPlaneDeployment(
 			updated = true
 		}
 
+		// Always process replica updates, even during/after restart operations
+		// This ensures that if a user scales after a restart, the changes take effect
 		if scaling := dataplane.Spec.Deployment.Scaling; false ||
 			// If the scaling strategy is not specified, we compare the replicas.
 			(scaling == nil || scaling.HorizontalScaling == nil) ||
@@ -312,7 +376,11 @@ func reconcileDataPlaneDeployment(
 				scaling.HorizontalScaling.MinReplicas != nil &&
 				existing.Spec.Replicas != nil &&
 				*existing.Spec.Replicas < *scaling.HorizontalScaling.MinReplicas) {
+
+			// Always check for replica changes, regardless of restart status
+			// Fixed issue #1390: Ensure replicas update correctly after restart
 			if !cmp.Equal(existing.Spec.Replicas, desired.Spec.Replicas) {
+				log.Debug(logger, "updating replica count", "original", originalReplicaCount, "from", existing.Spec.Replicas, "to", desired.Spec.Replicas)
 				existing.Spec.Replicas = desired.Spec.Replicas
 				updated = true
 			}
--- a/test/integration/dataplane_scale_subresource_test.go
+++ b/test/integration/dataplane_scale_subresource_test.go
@@ -1,13 +1,17 @@
 package integration
 
 import (
+	"fmt"
+	"os/exec"
 	"testing"
+	"time"
 
 	"github.com/samber/lo"
 	"github.com/stretchr/testify/require"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/kong/gateway-operator/pkg/consts"
@@ -16,6 +20,18 @@ import (
 
 	operatorv1beta1 "github.com/kong/kubernetes-configuration/api/gateway-operator/v1beta1"
 )
+
+// runKubectlScaleDataPlane runs the kubectl scale command to test the scale subresource for the DataPlane CRD.
+// It scales the DataPlane resource to the specified number of replicas and returns the output and error.
+func runKubectlScaleDataPlane(t *testing.T, namespacedName types.NamespacedName, replicas int) (string, error) {
+	t.Helper()
+
+	// Execute kubectl scale command
+	cmd := "kubectl scale --namespace=" + namespacedName.Namespace + " dataplane/" + namespacedName.Name + " --replicas=" + fmt.Sprint(replicas)
+
+	out, err := exec.CommandContext(t.Context(), "sh", "-c", cmd).CombinedOutput()
+	return string(out), err
+}
 
 // TestDataPlaneScaleSubresource tests the scale subresource of the DataPlane CRD.
 // It verifies that when a deployment is restarted using kubectl rollout restart,
@@ -38,35 +54,12 @@ func TestDataPlaneScaleSubresource(t *testing.T) {
 					DeploymentOptions: operatorv1beta1.DeploymentOptions{
 						Replicas: lo.ToPtr(int32(2)), // Set initial replica count to 2
 						PodTemplateSpec: &corev1.PodTemplateSpec{
-							ObjectMeta: metav1.ObjectMeta{
-								Labels: map[string]string{
-									"label-a": "value-a",
-									"label-x": "value-x",
-								},
-							},
 							Spec: corev1.PodSpec{
 								Containers: []corev1.Container{
 									{
-										Env: []corev1.EnvVar{
-											{
-												Name:  "TEST_ENV",
-												Value: "test",
-											},
-										},
 										Name:  consts.DataPlaneProxyContainerName,
 										Image: helpers.GetDefaultDataPlaneImage(),
 									},
-								},
-							},
-						},
-					},
-				},
-				Network: operatorv1beta1.DataPlaneNetworkOptions{
-					Services: &operatorv1beta1.DataPlaneServices{
-						Ingress: &operatorv1beta1.DataPlaneServiceOptions{
-							ServiceOptions: operatorv1beta1.ServiceOptions{
-								Annotations: map[string]string{
-									"purpose": "scale",
 								},
 							},
 						},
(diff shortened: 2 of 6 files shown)


test/integration/test_dataplane_scale_subresource.go
```diff
@@ -105,4 +85,82 @@ func TestDataPlaneScaleSubresource(t *testing.T) {
...
+		// Find the new ReplicaSet (one that wasn't in the original list)
+		for i := range newReplicaSetList.Items {
+			rs := &newReplicaSetList.Items[i]
+			if _, exists := existingRSNames[rs.Name]; !exists {
+				newReplicaSet = rs
+				return true
+			}
+		}
+		return false
+	}, testutils.GatewayReadyTimeLimit, testutils.ObjectUpdateTick, "Failed to find new ReplicaSet after rollout restart")
```
pmalek: WDYT about adding a helper func which will for example return a `ReplicaSet` which is newer than the provided `time.Time`? This helper could fail the test (using the provided `*testing.T`) if there's more than 1.

I believe this could improve the readability of this test. This would also prevent the need for `existingRSNames`.

test/integration/test_dataplane_scale_subresource.go
```diff
@@ -105,4 +85,82 @@ func TestDataPlaneScaleSubresource(t *testing.T) {
 
 	t.Log("verifying dataplane scales to 3 replicas")
 	require.Eventually(t, testutils.DataPlaneHasNReadyPods(t, GetCtx(), dataplaneName, clients, 3), testutils.GatewayReadyTimeLimit, testutils.ObjectUpdateTick)
+
+	// Get the current ReplicaSets before restart
+	replicaSetList := &appsv1.ReplicaSetList{}
+	err = clients.MgrClient.List(GetCtx(), replicaSetList,
+		client.InNamespace(namespace.Name),
+		client.MatchingLabels(deployment.Spec.Selector.MatchLabels))
```
pmalek: Can we add a test helper like `testutils.GetDataPlaneReplicaSets(dp *operatorv1beta1.DataPlane)` (or similar)?

It could find the provided `DataPlane`'s `Deployment` and then find the `ReplicaSet` using the `ownerReferences` matching the `Deployment`.

(Accepting the proposal in the other comment would make us not need the `existingRSNames` anyway).

controller/dataplane/owned_deployment.go
```diff
@@ -306,17 +306,64 @@ func reconcileDataPlaneDeployment(
...
+		// This is different from our previous approach because the desired deployment
+		// is generated from the DataPlane resource and won't have the restart annotation
+		isRestartOperation := false
+		if existing.Spec.Template.Annotations != nil {
+			_, hasRestartAnnotation := existing.Spec.Template.Annotations["kubectl.kubernetes.io/restartedAt"]
+			if hasRestartAnnotation {
+				isRestartOperation = true
+				log.Debug(logger, "detected restart operation", "isRestartOperation", isRestartOperation)
+			}
+		}
```
pmalek: Can we extract this into a `func` like: `PodTemplateSpecHasRestartAnnotation()` which could return the annotation value and a `boolean` indicating if it was found?

controller/dataplane/owned_deployment.go
```diff
@@ -306,17 +306,64 @@ func reconcileDataPlaneDeployment(
...
+
+				// Ensure the replica count is maintained
+				if originalReplicas != nil {
+					log.Debug(logger, "preserving original replica count", "replicas", *originalReplicas)
+					existing.Spec.Replicas = originalReplicas
+				}
+
+				updated = true
+			}
+		} else if !cmp.Equal(existing.Spec.Template, desired.Spec.Template, opts...) {
```
pmalek: I think this could be made slightly more readable my using a comparer option to ignore the restart key (and in the future possibly extended with more keys to ignore).

This can be achieved using something like:

```
package dataplane

import (
	"testing"

	"github.com/google/go-cmp/cmp"
	"github.com/stretchr/testify/require"
	corev1 "k8s.io/api/core/v1"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
)

const (
	KeyNameRestart = "kubectl.kubernetes.io/restartedAt"
)

func TestX(t *testing.T) {
	a := corev1.PodTemplateSpec{
		ObjectMeta: metav1.ObjectMeta{
			Annotations: map[string]string{
				"obj":          "state",
				KeyNameRestart: "a",
			},
		},
	}
	b := corev1.PodTemplateSpec{
		ObjectMeta: metav1.ObjectMeta{
			Annotations: map[string]string{
				"obj":          "state",
				KeyNameRestart: "b",
			},
		},
	}
	opts := []cmp.Option{
		cmp.FilterPath(
			func(p cmp.Path) bool {
				return p.String() == "ObjectMeta.Annotations"
			},
			cmp.Comparer(
				func(a map[string]string, b map[string]string) bool {
					// NOTE: maps can be of different lengths because of
					// the restart annotation.

					for k, v := range a {
						if k == KeyNameRestart {
							continue
						}
						if b[k] != v {
							return false
						}
					}

					for k, v := range b {
						if k == KeyNameRestart {
							continue
						}
						if a[k] != v {
							return false
						}
					}

					return true

					// NOTE: Also possible to use maps.Equal but less efficient:
					// a = maps.Clone(a)
					// b = maps.Clone(b)
					// delete(a, KeyName)
					// delete(b, KeyName)
					// return maps.Equal(a, b)
				}),
		),
	}

	require.True(t, cmp.Equal(a, b, opts...))

	t.Logf("map a (should be unchanged): %+v", a.Annotations)
	t.Logf("map b (should be unchanged): %+v", b.Annotations)
}
```

This compare func could be placed in a named function and used here. WDYT?

pmalek:
@tao12345666333 Are you working on addressing the comments from [last review](https://github.com/Kong/gateway-operator/pull/1540#pullrequestreview-2825280163)? Looks like they have not been addressed yet.

tao12345666333:
Yes, I am updating it.

Last time, I found that I missed a situation that would cause it to not scale correctly after a restart.

test/integration/test_dataplane_scale_subresource.go
tao12345666333: Addressed and removed the `existingRSNames`

controller/dataplane/owned_deployment.go
tao12345666333: Yes, addressed.

test/integration/test_dataplane_scale_subresource.go
tao12345666333: Great!

controller/dataplane/owned_deployment.go
```diff
@@ -27,6 +29,8 @@ import (
 	operatorv1beta1 "github.com/kong/kubernetes-configuration/api/gateway-operator/v1beta1"
 )
 
+const restartAnnotationKey = "kubectl.kubernetes.io/restartedAt"
```
pmalek, line 37: Can we add a comment to this `const`? Possibly linking it to a relevant k8s doc page?

controller/dataplane/owned_deployment.go
```diff
@@ -271,6 +275,14 @@ func listOrReduceDataPlaneDeployments(
 	return false, &deployments[0], nil
 }
 
+func podTemplateSpecHasRestartAnnotation(template *corev1.PodTemplateSpec) (bool, string) {
+	if template.Annotations == nil {
+		return false, ""
+	}
+	restartTimeStr, hasRestartAnnotation := template.Annotations[restartAnnotationKey]
+	return hasRestartAnnotation, restartTimeStr
+}
```
pmalek: Idiomatically in Go, you'd return a value first and then a boolean indicating whether it was found/valid/etc.

```suggestion
func podTemplateSpecHasRestartAnnotation(template *corev1.PodTemplateSpec) (string, bool) {
	if template.Annotations == nil {
		return "", false
	}
	return template.Annotations[restartAnnotationKey]
}
```

controller/dataplane/owned_deployment.go
```diff
@@ -306,17 +318,55 @@ func reconcileDataPlaneDeployment(
...
+					isRestartOperation = true
+					log.Debug(logger, "detected recent restart operation", "isRestartOperation", isRestartOperation, "restartTime", restartTime)
+				} else {
+					log.Debug(logger, "found old restart annotation, not treating as restart", "restartTime", restartTime)
+				}
+			} else {
+				isRestartOperation = true // If we can't parse time, assume it's a restart for safety
+				log.Debug(logger, "detected restart with unparseable timestamp", "timestamp", restartTimeStr)
+			}
+		}
```
pmalek: Can we encapsulate this whole piece of logic in a separate function to prevent `reconcileDataPlaneDeployment` becoming any bigger and to make it more readable?

controller/dataplane/owned_deployment.go
```diff
@@ -306,17 +318,55 @@ func reconcileDataPlaneDeployment(
...
+		// Check if we have a fresh restart annotation (from the last 5 minutes)
+		// This prevents us from treating all deployments with old restart annotations
+		// as perpetually in a restart state, fixing issue #1390
+		hasRestartAnnotation, restartTimeStr := podTemplateSpecHasRestartAnnotation(&existing.Spec.Template)
+		if hasRestartAnnotation {
+			// Only treat it as a restart if the timestamp is recent (within 5 minutes)
+			restartTime, err := time.Parse(time.RFC3339, restartTimeStr)
+			if err == nil {
+				// Check if restart annotation is less than 5 minutes old
+				if time.Since(restartTime) < 5*time.Minute {
```
pmalek: How did we establish the 5 min as a threshold here? Was it suggested by the community (other operators do the same?) or it's a reasonable estimate that we believe would work well here?

test/integration/test_dataplane_scale_subresource.go
```diff
@@ -105,4 +85,80 @@ func TestDataPlaneScaleSubresource(t *testing.T) {
...
+		return newReplicaSet.Status.ReadyReplicas == *newReplicaSet.Spec.Replicas
+	}, testutils.GatewayReadyTimeLimit, testutils.ObjectUpdateTick, "New ReplicaSet should have all replicas ready")
+
+	t.Log("verifying dataplane still has 3 ready replicas after rollout")
+	require.Eventually(t, testutils.DataPlaneHasNReadyPods(t, GetCtx(), dataplaneName, clients, 3), testutils.GatewayReadyTimeLimit, testutils.ObjectUpdateTick)
+
+	// Test the specific issue where scaling after restart didn't work properly
+	t.Log("scaling DataPlane to 4 replicas after restart")
+	// Test scaling through the scale subresource
+	t.Log("testing scaling through scale subresource")
```
pmalek: This is not technically scaling through the subresource but rather through the `.replicas` spec field, right? If so I'd update the comment here.

test/integration/replicaset_helpers.go
```diff
@@ -0,0 +1,68 @@
...
+	operatorv1beta1 "github.com/kong/kubernetes-configuration/api/gateway-operator/v1beta1"
+)
+
+// FindReplicaSetNewerThan finds a ReplicaSet created after or at the specified time.
+// This helper logs detailed information about ReplicaSets found to help
+// with debugging timestamp issues.
+//
+// It will return nil if no ReplicaSet is found, and will log a warning if
+// multiple ReplicaSets are found that were created after or at the specified time.
+// In case of multiple matches, it returns the most recently created one.
```
pmalek: This doesn't seem accurate:

This `func` will fail the test outright when there's more than 1 matching `ReplicaSet` (lines 61-63). 

I think we need to either change the comment or update the implementation.

pmalek (changes requested):

controller/dataplane/owned_deployment.go
tao12345666333: @pmalek In the past, DeploymentController had a configuration to control the synchronization period of resources. However, it was later removed after restructuring and version updates. 

It is similar to the control parameters for namespaces that are still retained today, and is essentially an empirical value.   :smile:  (So I chose the same value here.)

https://github.com/kubernetes/kubernetes/blob/82db38a23c7820b1924d89f458fd368023f3980c/pkg/controller/namespace/config/v1alpha1/defaults.go#L41

test/integration/test_dataplane_scale_subresource.go
tao12345666333: Yes. I should conduct this test here using `kubectl scale` or through the `scale` API.

test/integration/replicaset_helpers.go
tao12345666333: Good catch! I forgot to update the comments after modifying its implementation based on your previous feedback.

controller/dataplane/owned_deployment.go
tao12345666333: Are you only referring to the selected part? Or does it include the overall logic related to restart or scale?

(In fact, some logs here can be removed.)

test/integration/test_dataplane_scale_subresource.go
pmalek: 👍 Let's do this then.

controller/dataplane/owned_deployment.go
pmalek: The selected part. If possible I think we could also pass the desired spec, explaining  why we need it ( to set the restart annotation if found, to prevent changing it).

controller/dataplane/owned_deployment.go
pmalek: Got it. So I believe we could leave a comment here about the chosen value: something like "We use 5m here as threshold as that correlates with const used by k/k (<PROVIDE LINK HERE>). Restart annotations older than this duration shouldn't cause a reconciliation effect". Or similar.

controller/dataplane/owned_deployment.go
```diff
@@ -271,6 +280,43 @@ func listOrReduceDataPlaneDeployments(
...
+				return true
+			} else {
+				log.Debug(logger, "found old restart annotation, not treating as restart", "restartTime", restartTime)
+			}
+		} else {
+			log.Debug(logger, "detected restart with unparseable timestamp", "timestamp", restartTimeStr)
+			return true // If we can't parse time, assume it's a restart for safety
+		}
+	}
+	return false
```
pmalek: Can we make the happy path the least indented?

So e.g. 

```
	if !hasRestartAnnotation {
		return false
	}
```

```
	restartTime, err := time.Parse(time.RFC3339, restartTimeStr)
	if err != nil {
		log.Debug(logger, "detected restart with unparseable timestamp", "timestamp", restartTimeStr)
		return true // If we can't parse time, assume it's a restart for safety
	}
	// NOTE: No else clause needed here
	...
```

related: https://maelvls.dev/go-happy-line-of-sight/

controller/dataplane/owned_deployment.go
```diff
@@ -271,6 +280,43 @@ func listOrReduceDataPlaneDeployments(
 	return false, &deployments[0], nil
 }
 
+func podTemplateSpecHasRestartAnnotation(template *corev1.PodTemplateSpec) (string, bool) {
+	if template == nil || template.Annotations == nil {
+		return "", false
+	}
+	return template.Annotations[restartAnnotationKey], template.Annotations[restartAnnotationKey] != ""
```
pmalek: We should be able to do:

```suggestion
	return template.Annotations[restartAnnotationKey]
```

right?

controller/dataplane/owned_deployment.go
```diff
@@ -306,17 +352,32 @@ func reconcileDataPlaneDeployment(
...
-		// ensure that PodTemplateSpec is up to date
 		if !cmp.Equal(existing.Spec.Template, desired.Spec.Template, opts...) {
+
+			if isRestartOperation {
+				// Preserve the restart annotation
+				if desired.Spec.Template.Annotations == nil {
+					desired.Spec.Template.Annotations = make(map[string]string)
+				}
+				// Get the restart timestamp from the existing template
+				restartTimeStr, _ := podTemplateSpecHasRestartAnnotation(&existing.Spec.Template)
```
pmalek: To prevent calling `podTemplateSpecHasRestartAnnotation` here again (it's already called inside `isRecentDeploymentRestart`) we could add a return value to `isRecentDeploymentRestart` to return the annotation value if present. WDYT?

test/integration/replicaset_helpers.go
```diff
@@ -0,0 +1,69 @@
...
+)
+
+// FindReplicaSetNewerThan finds a ReplicaSet created after or at the specified time.
+// This helper logs detailed information about ReplicaSets found to help
+// with debugging timestamp issues.
+//
+// It will return nil if no ReplicaSet is found or if there's an error listing ReplicaSets.
+// If multiple ReplicaSets are found that were created after or at the specified time,
+// it will fail the test immediately with t.FailNow().
+// A single matching ReplicaSet is required for the test to proceed.
```
pmalek: Mentioning test here, as in "for the test to proceed" seems a bit unclear as there's no test in this context. I suggest to reword this to:

```suggestion
// It succeed when exactly 1 matching ReplicaSet is found.
```

test/integration/replicaset_helpers.go
```diff
@@ -0,0 +1,69 @@
...
+func FindReplicaSetNewerThan(
```
pmalek:
```suggestion
func FindDataPlaneReplicaSetNewerThan(
```

test/integration/dataplane_scale_subresource_test.go
```diff
@@ -95,14 +89,90 @@ func TestDataPlaneScaleSubresource(t *testing.T) {
...
+		if rs != nil {
+			t.Logf("Found new ReplicaSet %s created at %v", rs.Name, rs.CreationTimestamp)
+			newReplicaSet = rs
+			return true
+		}
+		return false
+	}, testutils.GatewayReadyTimeLimit, testutils.ObjectUpdateTick, "Failed to find new ReplicaSet after rollout restart")
+
+	t.Log("verifying new ReplicaSet has correct replica count")
+	assert.NotNil(t, newReplicaSet, "New ReplicaSet should be created")
```
pmalek:
```suggestion
	require.NotNil(t, newReplicaSet, "New ReplicaSet should be created")
```

test/integration/dataplane_scale_subresource_test.go
```diff
@@ -17,6 +22,18 @@ import (
...
 
+// run_kubectl_scale runs the kubectl scale command to test the scale subresource for the DataPlane CRD.
+// It scales the DataPlane resource to the specified number of replicas and returns the output and error.
+func run_kubectl_scale(t *testing.T, namespace, name string, replicas int) (string, error) {
+	t.Helper()
+
+	// Execute kubectl scale command
+	cmd := "kubectl scale --namespace=" + namespace + " dataplane/" + name + " --replicas=" + fmt.Sprint(replicas)
+
+	out, err := exec.Command("sh", "-c", cmd).CombinedOutput()
```
pmalek:
```suggestion
	out, err := exec.CommandContext(t.Context(), "sh", "-c", cmd).CombinedOutput()
```

test/integration/dataplane_scale_subresource_test.go
```diff
@@ -17,6 +22,18 @@ import (
 	operatorv1beta1 "github.com/kong/kubernetes-configuration/api/gateway-operator/v1beta1"
 )
 
+// run_kubectl_scale runs the kubectl scale command to test the scale subresource for the DataPlane CRD.
+// It scales the DataPlane resource to the specified number of replicas and returns the output and error.
+func run_kubectl_scale(t *testing.T, namespace, name string, replicas int) (string, error) {
```
pmalek:
```suggestion
// runKubectlScale runs the kubectl scale command to test the scale subresource for the DataPlane CRD.
// It scales the DataPlane resource to the specified number of replicas and returns the output and error.
func runKubectlScaleDataPlane(t *testing.T, namespace, name string, replicas int) (string, error) {
```

Also: WDYT about using `types.NamespacedName` to prevent accidental errors from providing name first instead of the namespace?

controller/dataplane/owned_deployment.go
tao12345666333: No, we can't.

When we put it in the return, it can only return one value, which does not match the function's signature.
You can test here: https://go.dev/play/p/DYwwm-re340

pmalek: TIL. Thanks for checking this.

So it seems we can still do the following to save on the double lookup:

```suggestion
	v, ok := template.Annotations[restartAnnotationKey]
	return v, ok
```

right?

tao12345666333: @pmalek  I don't think so. Because this function is meant to check for the existence of a restart annotation. If its value is empty `""`, that is not expected,  `v, ok := template.Annotations[restartAnnotationKey]` ok will be `true`.

This will lead to logical anomalies later on (however, I have also added relevant error handling for it; in this case, I will assume it is a restart operation).

controller/dataplane/owned_deployment.go
tao12345666333: Yes, we can avoid the function call here by increasing the return value. Do you have any suggestions regarding the function name?   If the return value is increased, then the name `isRecentDeploymentRestart` will no longer be appropriate.

controller/dataplane/owned_deployment.go
pmalek: I see your concern I'm ok with adding this type of checks but OTOH I believe this is defensive programming. I'm just curious: is `""` possible under normal operating conditions?

pmalek: If we insist on leaving the check in, then this piece of code can become:

```suggestion
	v, ok := template.Annotations[restartAnnotationKey]
	return v, ok && v != ""
```

controller/dataplane/owned_deployment.go
pmalek: `isRecentDeploymentRestart` I think would do with an appropriate godoc added to it. WDYT?

controller/dataplane/owned_deployment.go
```diff
@@ -271,6 +280,46 @@ func listOrReduceDataPlaneDeployments(
...
+		return "", false
+	}
+	v, ok := template.Annotations[restartAnnotationKey]
+	return v, ok && v != ""
+}
+
+// isRecentDeploymentRestart detects if a deployment is undergoing a recent restart operation.
+// It checks for the presence of the kubectl restart annotation and verifies if it's recent (within 5 minutes).
+// Returns true if a recent restart is detected, false otherwise, along with the restart timestamp if present.
+func isRecentDeploymentRestart(template *corev1.PodTemplateSpec, logger logr.Logger) (bool, string) {
```
pmalek: Minor nit:

I'd keep the order of the return values to be the same as in other functions: so

```suggestion
func isRecentDeploymentRestart(template *corev1.PodTemplateSpec, logger logr.Logger) (string, bool) {
```

controller/dataplane/owned_deployment.go
```diff
@@ -306,17 +355,31 @@ func reconcileDataPlaneDeployment(
...
 		if !cmp.Equal(existing.Spec.Template, desired.Spec.Template, opts...) {
+
+			if isRestartOperation {
+				// Preserve the restart annotation
+				if desired.Spec.Template.Annotations == nil {
+					desired.Spec.Template.Annotations = make(map[string]string)
+				}
+				// Use the restart timestamp we already retrieved
+				desired.Spec.Template.Annotations[restartAnnotationKey] = restartTimeStr
+			}
```
pmalek: If we move the log about the found restart annotation in here then we'd be able to call `isRecentDeploymentRestart` in this scope and thus making scope of `isRestartOperation` and `restartTimeStr` that smaller. WDYT?

We'd have to remove logging of `isRestartOperation` from line 411 but I'm OK with that.

pmalek (commented):
Minor nits, otherwise this already looks good.

controller/dataplane/owned_deployment.go
tao12345666333: makes sense

pmalek (approved):

tao12345666333:
Got 502 error from konnect

tao12345666333:
Incident-508 has been resolved.

PR #1636, opened 2025-05-19
chore: simplify bootstrap of controllers and other EE cleanups

programmer04:

One of the subsequent PRs for removing machinery that was needed for importing the OSS repo to the EE one. This PR simplifies bootstrap of controllers, the implementation of the CLI, version management, etc.

**Which issue this PR fixes**

Part of https://github.com/Kong/gateway-operator/issues/1493

The mechanism implemented in `controller/pkg/ctxinjector/ctxinjector.go` will be removed in a separate PR. The same for moving the generation of CLI docs. Also, simplification of controllers to maybe get rid of the mechanism for callbacks will be considered. Also renagment where particular packages are


modules/manager/controller_setup.go
```diff
@@ -646,83 +578,44 @@ func SetupControllers(mgr manager.Manager, c *Config) (map[string]ControllerDef,
...
+			newKonnectEntityController[configurationv1alpha1.KongCredentialBasicAuth](controllerFactory),
+			newKonnectEntityController[configurationv1alpha1.KongCredentialAPIKey](controllerFactory),
+			newKonnectEntityController[configurationv1alpha1.KongCredentialACL](controllerFactory),
+			newKonnectEntityController[configurationv1alpha1.KongCredentialHMAC](controllerFactory),
+			newKonnectEntityController[configurationv1alpha1.KongCredentialJWT](controllerFactory),
+			newKonnectEntityController[configurationv1alpha1.KongKey](controllerFactory),
+			newKonnectEntityController[configurationv1alpha1.KongKeySet](controllerFactory),
+			newKonnectEntityController[configurationv1alpha1.KongDataPlaneClientCertificate](controllerFactory),
+			newKonnectEntityController[configurationv1alpha1.KongVault](controllerFactory),
+			newKonnectEntityController[configurationv1alpha1.KongSNI](controllerFactory),
```
pmalek, line 617: The only comment on this I'd have is that previously, we'd be safeguarded against e.g. adding the same controller twice by using a map, whereas here we use a simple slice where that could happen.

While not likely to happen it would be nice to have this guardrails here. Leaving up to you to decide whether we want to merge as is or retain the map behavior.

pmalek (approved):

modules/manager/controller_setup.go
programmer04: How does adding twice the same controller manifest? I think ideally, it would be to catch something like that with a test, e.g envtests. For now, I'm merging as it is, because it's much less code 🙂

PR #1646, opened 2025-05-20
ci: skip workflow runs when only markdown files are changed

tao12345666333:

During my last version release process, I found that when I only modified the CHANGELOG, it also triggered all the tests. Since we enabled the merge queue feature, this meant that even modifying the CHANGELOG (adding a release date) took over 20 minutes to merge.

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:

~- [ ] the `CHANGELOG.md` release notes have been updated to reflect significant changes~


.github/workflows/build.yaml
```diff
@@ -14,9 +14,19 @@ on:
   push:
     branches:
     - 'main'
+    paths-ignore:
+      - '**/*.md'
```
pmalek: I'd suggest that we're careful with this.

For example, someone can change https://github.com/Kong/gateway-operator/blob/cedb1b01f5a7edf1b3ec662d3361c7e26333efec/docs/api-reference.md without changing the sources (or can remove it) and CI should catch that and prevent the merge.

I think we can enlist the files that typically shouldn't require tests or code generation to run and pass, like readme, changelog, license files etc.

WDYT?

tao12345666333: Make sense. 

I will keep only a few certain files.

programmer04: I think `README.md` can be added and also everything under `.github/ISSUE_TEMPLATE`

tao12345666333: Good point :+1:

pmalek (dismissed):
The fact that these lists have to be hardcoded everywhere is being me 🙄  ( :octocat: I'm looking at you ).

Thanks @tao12345666333 for taking care of this.

pmalek:
BTW there are also https://github.com/Kong/gateway-operator/blob/main/SECURITY.md and https://github.com/Kong/gateway-operator/blob/main/FEATURES.md which could be added to these lists.

programmer04 (dismissed):

tao12345666333:
> The fact that these lists have to be hardcoded everywhere is being me 🙄 ( :octocat: I'm looking at you ).

I hope it can add YAML Anchors support.  https://github.com/actions/runner/issues/1182
But it seems that there hasn't been much progress yet.

pmalek (approved):

pmalek:
I totally forgot about 1 important thing that would prevent this from working correctly: required checks.

Since these are unconditional it basically means that now each PR that changes e.g. README.md will not get its workflows run and thus it won't pass the required checks. I **think** that it might be possible to check the changed files (e.g. using https://github.com/tj-actions/changed-files but that action had its problems recently so I'm not sure we want to use it) and then conditionally pass jobs in workflows. That might end up being a nightmare to maintain though.

Until we figure how to work around this I suggest we revert this change @tao12345666333, WDYT?

Example of a blocked PR: https://github.com/Kong/gateway-operator/pull/1649.

tao12345666333:
@pmalek Thank you for pointing this out. I opened a new PR for changing the conditions https://github.com/Kong/gateway-operator/pull/1651

tao12345666333:
**Before**  15m

https://github.com/Kong/gateway-operator/actions/runs/14883429390
image


**After**  5m

https://github.com/Kong/gateway-operator/actions/runs/15152497381?pr=1649

image

PR #1653, opened 2025-05-21
fix workflow

tao12345666333:

In my previous modifications to the GitHub Actions workflow, although they did skip some tasks as expected, they did not end with "passed" as originally planned.

image

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -34,22 +34,7 @@ jobs:
           fetch-depth: 0
       - id: check-files
         name: Check if only documentation files changed
-        run: |
-          # Get changed files compared to the base
-          CHANGED_FILES=$(git diff --name-only ${{ github.event.pull_request.base.sha || 'HEAD~1' }} ${{ github.sha }})
-
-          # Check if all changed files are documentation or license files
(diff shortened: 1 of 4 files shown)


tao12345666333:
image


It looks good. And we need to change the required job to real-passed.

tao12345666333:
image

It looks good.

pmalek (commented):
- I think we have missed adding the envtest-tests as the conditional job. We can save some CI time with that as well.
- As discussed on the sync, let's put [the script](https://github.com/Kong/gateway-operator/blob/14bb185f8f24c163606abcca3859ab1e0482cc9f/.github/workflows/tests.yaml#L41-L55) to a script file so that it can be reused.
  - I believe the regex in that script: `^(.*\.md|LICENSE|LICENSES|\.github/ISSUE_TEMPLATE/.*)$` might still catch the docs in https://github.com/Kong/gateway-operator/blob/cedb1b01f5a7edf1b3ec662d3361c7e26333efec/docs/api-reference.md

pmalek (approved):

Issue #1598, opened 2025-05-13
Use kong/kong-operator image in chart when available


pmalek:
### Problem statement

#1587 introduces the `kong-operator` chart which has a default image set to: `docker.io/kong/gateway-operator` until `docker.io/kong/kong-operator` becomes available.

This issue tracks the change of that image to `docker.io/kong/kong-operator` and any other necessary changes.

### Acceptance criteria

- [ ] `kong-operator` chart uses `docker.io/kong/kong-operator` image.

pmalek:
Blocked (part of #1630).

We need the first image to be available in https://hub.docker.com/r/kong/kong-operator/ to change this.

programmer04:
Release has been mad,e andthe  Chart with automation has been adjusted

Issue #1490, opened 2025-04-14
Release CI for kong-operator alpha


sandromodarelli:
## Problem Statement

Once we got the new docker registry available and accessible  for the same docker team operaring with gateway-opeartor registry, we need to setup the release branch for 2.0.0-alpha.1 and work on the CI to build and push images in the new registry

## Proposed Solution

we can modify the current gateway-operator CI to build and push to kong-operator registry. since we would like to use the same dockerhub team, we can use the same GH secrets to retrieve tokens

## Additional Information

...

## Acceptance Criteria
- [ ] having a release branch for KO 2.0.0-alpha.1
- [ ] having a CI to build and push to new kong-operator docker registry

programmer04:
There is already a registry [kong-operator](https://hub.docker.com/r/kong/kong-operator), which contains very old images - over 3 years old, built probably from [gateway-operator](https://github.com/Kong/gateway-operator) repo, which will be removed and replaced. So, probably it's sensible to ask about emptying this repo

pmalek:
> There is already a registry [kong-operator](https://hub.docker.com/r/kong/kong-operator), which contains very old images - over 3 years old, built probably from [gateway-operator](https://github.com/Kong/gateway-operator) repo, which will be removed and replaced. So, probably it's sensible to ask about emptying this repo

These are probably from https://github.com/Kong/kong-operator-archive which would align with tags present at https://operatorhub.io/operator/kong (newest is 0.9). I believe these could be removed (given how long this repo has already been in a deprecated state) but better ask I think (perhaps product's a good audience to chip in on this).

programmer04:
Yes, my mistake in the comment. I copied from the wrong tab in my browser. I exactly thought about the repo that you mentioned

pmalek:
Removing my assignment. @programmer04 currently has #1686 to cover part of this.

programmer04:
Relase happened, it's resolved

PR #1686, opened 2025-05-27
chore: ci, chart adjustments and some renames from KGO to KO

programmer04:

Enable and prepare the release of `2.0.0-alpha.0` as a step forward to

- #1630

Furthermore, use `kong-operator` Docker registry and adjust Helm Chart.

Two spin-off issues have been created

- https://github.com/Kong/gateway-operator/issues/1717 (can be done just after successful release)
- https://github.com/Kong/gateway-operator/issues/1716 (it'll be doable when at least two minor versions will exist)

**Which issue this PR fixes**

Part of #1490, #1598


VERSION
pmalek, line 1: This file is typically upgraded by the bot (example: https://github.com/Kong/gateway-operator/pull/1579).

Do we need to bump it now?

pmalek (approved):
1 question but I believe we want to bump change `VERSION` anyway at this point to be consistent with other changes in this PR so 👍

VERSION
programmer04: Let's do it to have everyhing in one shot and ensure that everything passes as expected

PR #1721, opened 2025-06-11
chore(ci): fix release for KO

programmer04:

Continuation of 

- https://github.com/Kong/gateway-operator/pull/1686

After reconnaissance in force - [triggering release for v2.0.0-alpha.0](https://github.com/Kong/gateway-operator/actions/runs/15586492873).

Respective secret/vars have been added to CI, they need to be duplicated for now to allow the release of KGO.

**Which issue this PR fixes**

Part of https://github.com/Kong/gateway-operator/issues/1490, https://github.com/Kong/gateway-operator/issues/1598


.github/workflows/build.yaml
```diff
@@ -13,10 +13,10 @@ concurrency:
 on:
   push:
     branches:
-    - 'main'
+      - "main"
```
pmalek, line 16: Any reason to change these now?

pmalek (approved):
Just one minor Q on changing " to '. Otherwise 👍

.github/workflows/build.yaml
programmer04: Oh, when I did find and replace it saved, and my editor format on-save automatically. But tbh I prefer single quotes when there is no interpolation. Let's leave it

PR #1724, opened 2025-06-11
chore(ci): re-enable after release of 2.0.0-alpha.1

programmer04:
**What this PR does / why we need it**:

Continuation of https://github.com/Kong/gateway-operator/pull/1686 and https://github.com/Kong/gateway-operator/pull/1721

Furthermore, it fixes the release-bot to update the version in `charts/kong-operator/Chart.yaml` and `charts/kong-operator/values.yaml` too, which was omitted in 
- https://github.com/Kong/gateway-operator/pull/1722

**Which issue this PR fixes**

Fixes #1717

Part of #1490 & #1598

PR #1725, opened 2025-06-12
chore: remove redundant EE mentions and field `.tag` in chart `values.yaml`

programmer04:

PR #1724 introduced by mistake redundant field `.tag` in chart `values.yaml`, this PR removes it.

Furthermore, remove mentions about EE that are not valid anymore, since everything is OSS.

### DIFF
--- a/.github/ISSUE_TEMPLATE/---release.md
+++ b/.github/ISSUE_TEMPLATE/---release.md
@@ -39,11 +39,6 @@ If the troubleshooting section does not contain the answer to the problem you en
     ${KUBERNETES_CONFIGURATION_REPO}/scripts/apidocs-gen/post-process-for-konghq.sh ${KUBERNETES_CONFIGURATION_REPO}/docs/gateway-operator-api-reference.md ${KONGHQ_DOCS_REPO}/app/_src/gateway-operator/reference/custom-resources/1.2.x.md
     ```
 
-  - NOTE: [CLI configuration options docs][cli_ref_docs] should be updated when releasing KGO EE as that's the source of truth for those.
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -358,6 +358,7 @@ manifests.crds: controller-gen manifests.versions ## Generate CustomResourceDefi
 manifests.versions: kustomize yq
 	$(YQ) eval '.appVersion = "$(VERSION)"' -i charts/kong-operator/Chart.yaml
 	$(YQ) eval '.image.tag = "$(VERSION)"' -i charts/kong-operator/values.yaml
+	$(YQ) eval '.tag = "$(VERSION)"' -i charts/kong-operator/values.yaml
```
pmalek: This is not really desired, right? I believe we shouldn't add this but we should remove https://github.com/Kong/gateway-operator/blob/8b56f338712cb6229ad1df89f5442fd41ac03ee8/charts/kong-operator/values.yaml#L99

```suggestion
```

pmalek (changes requested):

Makefile
programmer04: Why not? String with version in file `charts/kong-operator/values.yaml` is in lines `3` and `99` (`.tag`) so both should be set to the same value by automation. Or am I missing something?

programmer04: Or you mean to remove this field entirely, because it's redundant?

programmer04: I see, this file is redundant, I'm removing

pmalek: > Or you mean to remove this field entirely, because it's redundant?

Sorry for not making this more clear. Yes, `.tag` is redundant. Thanks for taking care of this.

pmalek (approved):

PR #1763, opened 2025-06-24
feat: reschedule ControlPlane manager when its spec changes

### LINKED ISSUE #1374: Reschedule KIC instance when ControlPlane specification changes
## Problem Statement

In the PoC we didn't handle the case when ControlPlane specification changes while the KIC instance is already running. We have to take that into account and restart the instance with the adjusted config once the spec changes.

## Proposed Solution

Extend the `multiinstance.Manager` with a method that will allow checking if an instance with a given ID and a config hash exists. Use that method in the ControlPlane reconciler and restart an instance with the new config if the config changes.

## Acceptance Criteria
- [ ] When ControlPlane controller has already started a KIC instance and the CP's spec changes, the instance is properly restarted with the updated config following spec

pmalek:

Prerequisite: https://github.com/Kong/kubernetes-ingress-controller/pull/7535.

**Which issue this PR fixes**

Fixes #1374

**PR Readiness Checklist**:


controller/controlplane/controller.go
```diff
@@ -273,6 +267,40 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+			if err := r.InstancesManager.StopInstance(mgrID); err != nil {
+				return ctrl.Result{}, err
+			}
+
+			if err := r.scheduleInstance(ctx, logger, mgrID, mgrCfg); err != nil {
+				return ctrl.Result{}, fmt.Errorf("failed to schedule instance: %w", err)
+			}
+			return ctrl.Result{RequeueAfter: requeueAfterBoot}, nil
+		}
+	}
```
programmer04, line 302: If statement returns, thus it seems redundant to have `else` block. WDYT?

pmalek: That's true but then `hashRunning` has to be defined out of scope of this `if` clause.

Do you believe this would be more readable then?

programmer04: Let's leave it

programmer04 (approved):

Issue #1728, opened 2025-06-12
Kong operator: support v2 GatewayConfiguration


pmalek:
### Problem statement

#1358 tracks the changes to the new `ControlPlane` and `GatewayConfiguration` APIs for KO.

This issue tracks the necessary changes in `Gateway` reconciler logic which will have to be done to make it work with new changes in the `v2` (`v2alpha` or newer) APIs.

### Acceptance criteria

- [ ] As a user I can use operator to work with `GatewayConfiguration` `v2` (`v2alpha` or newer) API

PR #1766, opened 2025-06-25
chore: adapt to v2alpha1 GatewayConfiguration API

pmalek:

This is a first PR that adapts the GatewayConfiguration API usage to v2alpha1.

**Which issue this PR fixes**

Part of #1728

**PR Readiness Checklist**:

### DIFF
--- a/controller/gateway/controller_reconciler_utils.go
+++ b/controller/gateway/controller_reconciler_utils.go
@@ -37,6 +37,7 @@ import (
 	kcfgdataplane "github.com/kong/kubernetes-configuration/api/gateway-operator/dataplane"
 	kcfggateway "github.com/kong/kubernetes-configuration/api/gateway-operator/gateway"
 	operatorv1beta1 "github.com/kong/kubernetes-configuration/api/gateway-operator/v1beta1"
+	operatorv2alpha1 "github.com/kong/kubernetes-configuration/api/gateway-operator/v2alpha1"
 )
 
 // -----------------------------------------------------------------------------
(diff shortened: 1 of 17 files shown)


controller/gateway/controller_reconciler_utils.go
```diff
@@ -139,27 +137,39 @@ func (r *Reconciler) getGatewayAddresses(
...
 				}
-				return pluginReference
+				return nn
 			},
 		)
 	}
 
 	if opts.Resources != nil {
-		dataPlaneOptions.Resources.PodDisruptionBudget = opts.Resources.PodDisruptionBudget
+		if opts.Resources.PodDisruptionBudget != nil {
```
programmer04, line 164: `if opts.Resources != nil && opts.Resources.PodDisruptionBudget != nil {` keeping code more to the left

controller/pkg/controlplane/controlplane.go
```diff
@@ -21,17 +21,5 @@ type DefaultsArgs struct {
...
-	if !reflect.DeepEqual(spec1.Extensions, spec2.Extensions) {
-		return false
-	}
-
-	if !reflect.DeepEqual(spec1.WatchNamespaces, spec2.WatchNamespaces) {
-		return false
-	}
-
-	return true
+	return reflect.DeepEqual(spec1, spec2)
```
programmer04, line 24: As I understand, you keep the parameter `env Vars Ignore ...string` so as not to break an API. It'll be clean up later, won't be?

So I recommend do now `envVarsToIgnore ...string` to `_ ...string`

pmalek: I missed that during the cleanup. There's no need to keep that argument around. f4e09cdc3c19445a02e06ed2efdfa5fd6059afd3

controller/gateway/controller_reconciler_utils.go
pmalek: The way it's nested here we try to signal that there can be more `.Resources` fields that can be handled here.

Do you think it'd be better to change this given the above?

programmer04: resonable, it can be left

programmer04 (approved):

PR #1811, opened 2025-07-03
fix(konnect): ignore sdkerrors.ForbiddenError in konnect entity reconciler

### LINKED ISSUE #1787: Konnect entities: Konnect returning 403s during KonnectCloudGatewayNetwork operations cause endless reconciliation
### Problem statement

When working with `KonnectCloudGatewayNetwork`s and causing provider related errors (e.g. exhausting the alloted of number of networks provisioned in an org) will cause the operator to endlessly reconcile the resource.

### Anything else?

Slack thread: https://kongstrong.slack.com/archives/C04KD8UEK8Q/p1750862231626689

@baumac stated that this will cause Konnect APIs to return response body like:

```
{
  "status": 403,
  "title": "Quota Exceeded",
  "instance": "kong:trace:0000000000000000000",
  "detail": "Maximum number of Active Networks exceeded. Max allowed: 0"
}
```

### Acceptance criteria

- [ ] Provider related errors (i.e. 403s) do not cause endless reconciliation.

randmonkey:
Ignore `sdkkonnecterrs.ForbiddenError` when running CRUD ops against Konnect APIs.

**Which issue this PR fixes**

Fixes #1787

**PR Readiness Checklist**:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,8 @@
 ### Fixed
 
 - Ignore the `ForbiddenError` in `sdk-konnect-go` returned from running CRUD
-  operations against Konnect APIs.
+  operations against Konnect APIs. This prevents endless reconciliation when an
+  operation is not allowed (due to e.g. exhausted quota).
   [#1811](https://github.com/Kong/kong-operator/pull/1811)
 
 ## [v2.0.0-alpha.1]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,14 @@
 - [v0.1.1](#v011)
 - [v0.1.0](#v010)
 
+## Unreleased
+
+### Fixed
+
+- Ignore the `ForbiddenError` in `sdk-konnect-go` returned from running CRUD
+  operations against Konnect APIs.
+  [#1811](https://github.com/Kong/kong-operator/pull/1811)
+
 ## [v2.0.0-alpha.1]
 
 > Release date: 2025-06-11
--- a/controller/konnect/ops/ops_errors.go
+++ b/controller/konnect/ops/ops_errors.go
@@ -129,7 +129,7 @@ func ParseSDKErrorBody(body string) (sdkErrorBody, error) {
 	return sdkErr, nil
 }
 
-// ErrorIsSDKError403 returns true if the provided error is a 403 Forbidden error.
+// ErrorIsForbiddenError returns true if the provided error is a 403 Forbidden error.
 // This can happen when the requested operation is not permitted.
 // Example SDKError body (SDKError message is a separate field from body message):
 //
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -29,6 +29,14 @@
 - [v0.1.1](#v011)
 - [v0.1.0](#v010)
 
+## Unreleased
+
+### Fixed
+
+- Ignore the `ForbiddenError` in `sdk-konnect-go` returned from running CRUD
+  operations against Konnect APIs.
```
```suggestion
- Ignore the `ForbiddenError` in `sdk-konnect-go` returned from running CRUD
  operations against Konnect APIs. This prevents endless reconciliation when an
  operation is not allowed (due to e.g. exhausted quota).
```

pmalek (commented):
Ideally, we'd extend the test in the envtest suite to check that indeed the expected error response body causes the status to be set in the object and (I'm not sure what's the best predicate here) the resource version stabilizes?

WDYT @randmonkey?

randmonkey:
> Ideally, we'd extend the test in the envtest suite to check that indeed the expected error response body causes the status to be set in the object and (I'm not sure what's the best predicate here) the resource version stabilizes?
> 
> WDYT @randmonkey?

I checked our `envtest` suite, the mock is acted in the SDK level (KO -> mock SDK instead of KO -> konnect SDK -> konnect API). So we can make a mock that returns a `ForbiddenError` to check if the reconciler can set the status of the object and stops reconciliation. 

But if we want to verify the behavior of receiving the response body, we need to run a mock on the konnect API level (KO -> konnect SDK -> mock konnect API). Then we need to make up a new mock system for the envtests.
So I think we can add a single envtest case for the `ForbiddenError` in this PR, and make a new mock system if we really want it.

pmalek (approved):
Thanks for working on this 🙇 

We'd ideally backport this to 1.6.

pmalek:
> > Ideally, we'd extend the test in the envtest suite to check that indeed the expected error response body causes the status to be set in the object and (I'm not sure what's the best predicate here) the resource version stabilizes?
> > WDYT @randmonkey?
> 
> I checked our `envtest` suite, the mock is acted in the SDK level (KO -> mock SDK instead of KO -> konnect SDK -> konnect API). So we can make a mock that returns a `ForbiddenError` to check if the reconciler can set the status of the object and stops reconciliation.
> 
> But if we want to verify the behavior of receiving the response body, we need to run a mock on the konnect API level (KO -> konnect SDK -> mock konnect API). Then we need to make up a new mock system for the envtests. So I think we can add a single envtest case for the `ForbiddenError` in this PR, and make a new mock system if we really want it.

That's correct. We'd need to run the tests against a locally running Konnect API mock rather than mocking the SDK methods themselves.

Let's postpone this until more use cases for these tests come along.

PR #1829, opened 2025-07-09
feat(gateway): Set network policy on DP pods

### LINKED ISSUE #1700: Limit communication to Kong Gatway admin API with `NetworkPolicy`
## Problem Statement

In the KGO instance of a KIC used to be a separate `Pod`, hence `NetworkPolicy` was set accordingly to allow only communication from KIC `Pod` to `DataPlane`'s `Pod`s.

Currently, this approach is problematic because the whole communication happens from KO `Pod` not a `DataPlane`'s Pod. Moreover, during tests run, this `Pod` is not deployed (KO is run outside a cluster), to move forward in the PR #1701 it's been temporarily relaxed. Targets `make run` has been fixed in #1714.

This issue points this out and tracks the resolution of it to bring back this policy.

## Proposed Solution

Reenable the proper `NetworkPolicy` that limits the communication. For running tests and development environments, the interception feature of telepresence may be useful.

## Acceptance Criteria
- [ ] `NetworkPolicy` is more specific and allows only required KO access to the admin API endpoint
- [ ] `TestGatewayDataPlaneNetworkPolicy` is adjusted
- [ ] all ways of running KO for development - `make skaffold`, `make run` work as they used to

randmonkey:

Set `NetworkPolicy` to limit only KO pod can access admin API of Kong gateway

**Which issue this PR fixes**

Fixes #1700

**PR Readiness Checklist**:

### DIFF
--- a/charts/kong-operator/templates/_helpers.tpl
+++ b/charts/kong-operator/templates/_helpers.tpl
@@ -94,7 +94,7 @@ Create a list of env vars based on the values of the `env` and `customEnv` maps.
 - name: {{ template "kong.fullname" . }}-certs-dir
   emptyDir:
     sizeLimit: {{ .Values.certsDir.sizeLimit }}
-- name: {{ template "kong.fullname" . }}-pod-info
+- name: {{ template "kong.fullname" . }}-pod-labels
   downwardAPI:
     items:
     - path: labels
@@ -105,7 +105,7 @@ Create a list of env vars based on the values of the `env` and `customEnv` maps.
 {{- define "kong.volumeMounts" -}}
 - name: {{ template "kong.fullname" . }}-certs-dir
   mountPath: /tmp/k8s-webhook-server/serving-certs
-- name: {{ template "kong.fullname" . }}-pod-info
+- name: {{ template "kong.fullname" . }}-pod-labels
   mountPath: /etc/podinfo
 {{- end }}
 
--- a/charts/kong-operator/templates/_helpers.tpl
+++ b/charts/kong-operator/templates/_helpers.tpl
@@ -94,11 +94,19 @@ Create a list of env vars based on the values of the `env` and `customEnv` maps.
 - name: {{ template "kong.fullname" . }}-certs-dir
   emptyDir:
     sizeLimit: {{ .Values.certsDir.sizeLimit }}
+- name: {{ template "kong.fullname" . }}-pod-info
+  downwardAPI:
+    items:
+    - path: labels
+      fieldRef:
+        fieldPath: metadata.labels
 {{- end }}
 
 {{- define "kong.volumeMounts" -}}
 - name: {{ template "kong.fullname" . }}-certs-dir
   mountPath: /tmp/k8s-webhook-server/serving-certs
+- name: {{ template "kong.fullname" . }}-pod-info
+  mountPath: /etc/podinfo
 {{- end }}
 
 {{/* effectiveVersion takes the image dict from values.yaml. */}}
--- a/controller/gateway/controller_reconciler_utils.go
+++ b/controller/gateway/controller_reconciler_utils.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"maps"
 	"net"
 	"net/http"
 	"strconv"
@@ -305,7 +304,7 @@ func (r *Reconciler) ensureDataPlaneHasNetworkPolicy(
 		return false, errors.New("number of networkPolicies reduced")
 	}
 
-	// generate the network policy that allow the KO pod to access dataplane pods.
+	// generate the network policy that allows the KO pod to access the admin APIs of dataplane pods.
 	generatedPolicy, err := generateDataPlaneNetworkPolicy(r.Namespace, dataplane, r.PodLabels)
 	if err != nil {
 		return false, fmt.Errorf("failed generating network policy for DataPlane %s: %w", dataplane.Name, err)
@@ -333,6 +332,8 @@ func (r *Reconciler) ensureDataPlaneHasNetworkPolicy(
 	return true, r.Create(ctx, generatedPolicy)
 }
 
+// generateDataPlaneNetworkPolicy generates the NetworkPolicy that allows the KO pod to access admin API of dataplane pods.
+// the params `namespace` and `podLabels` are namespace and labels of the KO pod itself, and `dataplane` is the target dataplane.
 func generateDataPlaneNetworkPolicy(
 	namespace string,
 	dataplane *operatorv1beta1.DataPlane,
@@ -344,6 +345,16 @@ func generateDataPlaneNetworkPolicy(
 		proxyPort       = intstr.FromInt(consts.DataPlaneProxyPort)
 		proxySSLPort    = intstr.FromInt(consts.DataPlaneProxySSLPort)
 		metricsPort     = intstr.FromInt(consts.DataPlaneMetricsPort)
+		// The label keys to match Kong operator pod.
+		// To not create new NetworkPolicy on upgrade of , we just keep the keys marking the application
+		// and remove the keys related to versions such as `version`,`pod-template-hash`,`helm.sh/chart`.
+		podLabelSelectorKeys = []string{
+			"app",
+			"app.kubernetes.io/component",
+			"app.kubernetes.io/instance",
+			"app.kubernetes.io/name",
+			"control-plane",
+		}
 	)
 
 	// Check if KONG_PROXY_LISTEN and/or KONG_ADMIN_LISTEN are set in
@@ -385,10 +396,17 @@ func generateDataPlaneNetworkPolicy(
 			},
 		},
 	}
+
 	if len(podLabels) > 0 {
-		podLabelsCopy := maps.Clone(podLabels)
+		matchPodLabels := map[string]string{}
+		for _, key := range podLabelSelectorKeys {
+			value, ok := podLabels[key]
+			if ok {
+				matchPodLabels[key] = value
+			}
+		}
 		policyPeerForControllerPod.PodSelector = &metav1.LabelSelector{
-			MatchLabels: podLabelsCopy,
+			MatchLabels: matchPodLabels,
 		}
 	}
 	limitAdminAPIIngress := networkingv1.NetworkPolicyIngressRule{
--- a/controller/gateway/controller_reconciler_utils.go
+++ b/controller/gateway/controller_reconciler_utils.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"maps"
 	"net"
 	"net/http"
 	"strconv"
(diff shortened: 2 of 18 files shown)


controller/gateway/controller_reconciler_utils.go
```diff
@@ -376,24 +377,27 @@ func generateDataPlaneNetworkPolicy(
...
+	// Construct the policy to allow the KO pod to access DataPlane admin APIs.
+	policyPeerForControllerPod := networkingv1.NetworkPolicyPeer{
+		NamespaceSelector: &metav1.LabelSelector{
+			MatchLabels: map[string]string{
+				"kubernetes.io/metadata.name": namespace,
+			},
+		},
+	}
+	if len(podLabels) > 0 {
+		podLabelsCopy := maps.Clone(podLabels)
```
randmonkey: REVIEW: Do we need to add all labels from KO pod to the network policy, or just choose some certain label(s) like `app`? The labels includes the pod hash and helm metadata (charts, version) like:
```
        matchLabels:
          app: ko-kong-operator
          app.kubernetes.io/component: ko
          app.kubernetes.io/instance: ko
          app.kubernetes.io/name: kong-operator
          app.kubernetes.io/version: 2.0.0-alpha.1
          control-plane: controller-manager
          helm.sh/chart: kong-operator-0.0.1
          pod-template-hash: 74b5fdd4b4
          version: 2.0.0-alpha.1
```

pkg/utils/kubernetes/self_metadata.go
```diff
@@ -0,0 +1,56 @@
...
+	serviceAccountNamespaceFile = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
+)
+
+// GetSelfNamespace gets the namespace in which KO runs.
+func GetSelfNamespace() (string, error) {
+	if ns := os.Getenv(podNamespaceEnvName); ns != "" {
+		return ns, nil
+	}
+	// This actually gets the namespace of the service account to run the pod.
+	buf, err := os.ReadFile(serviceAccountNamespaceFile)
```
randmonkey, line 25: REVIEW: Should we use the "backup" method to get the namespace or directly return error, since this is actually the namespace of the service account?

modules/manager/controller_setup.go
```diff
@@ -354,6 +354,11 @@ func SetupControllers(mgr manager.Manager, c *Config, cpsMgr *multiinstance.Mana
 	if err := mgr.Add(scrapersMgr); err != nil {
 		return nil, fmt.Errorf("failed to add scrapers manager to controller-runtime manager: %w", err)
 	}
+	podLabels, err := k8sutils.GetSelfPodLabels()
```
randmonkey, line 357: REVIEW: When we do not run KO code inside the cluster (in unit/envtest/integration tests), how should we fill the labels? The full labels can only be attached by volume so we cannot specify one by env. Here we just provide an empty one in the case when we cannot find the labels from the file.

charts/kong-operator/templates/_helpers.tpl
```diff
@@ -94,11 +94,19 @@ Create a list of env vars based on the values of the `env` and `customEnv` maps.
...
+    - path: labels
+      fieldRef:
+        fieldPath: metadata.labels
 {{- end }}
 
 {{- define "kong.volumeMounts" -}}
 - name: {{ template "kong.fullname" . }}-certs-dir
   mountPath: /tmp/k8s-webhook-server/serving-certs
+- name: {{ template "kong.fullname" . }}-pod-info
+  mountPath: /etc/podinfo
```
pmalek: Perhaps we call this explicitly as labels


```suggestion
- name: {{ template "kong.fullname" . }}-pod-labels
  downwardAPI:
    items:
    - path: labels
      fieldRef:
        fieldPath: metadata.labels
{{- end }}

{{- define "kong.volumeMounts" -}}
- name: {{ template "kong.fullname" . }}-certs-dir
  mountPath: /tmp/k8s-webhook-server/serving-certs
- name: {{ template "kong.fullname" . }}-pod-labels
  mountPath: /etc/podinfo
```

?

controller/gateway/controller_reconciler_utils.go
```diff
@@ -305,7 +305,8 @@ func (r *Reconciler) ensureDataPlaneHasNetworkPolicy(
 		return false, errors.New("number of networkPolicies reduced")
 	}
 
-	generatedPolicy, err := generateDataPlaneNetworkPolicy(gateway.Namespace, dataplane, controlplane)
+	// generate the network policy that allow the KO pod to access dataplane pods.
```
pmalek:
```suggestion
	// generate the network policy that allows the KO pod to access dataplane pods.
```

controller/gateway/controller_reconciler_utils.go
pmalek: Thinking out loud:

If we add all of these then only this particular Pod will be able to access the dataplane which is partially desired.

When users upgrade the helm release they'd like to have both of the operators (the one that's getting decommissioned will most likely not access the dataplane anymore so that should be fine) to access Kong.

On the other hand operator doesn't need to change the `NetworkPolicy` when an unrelated change happens on the Pod (e.g. resources/limits or envs are changes)

So I think these would do:

```
        matchLabels:
          app: ko-kong-operator
          app.kubernetes.io/component: ko
          app.kubernetes.io/instance: ko
          app.kubernetes.io/name: kong-operator
          control-plane: controller-manager
```

This will also prevent enforcing a new version of the `NetworkPolicy` when new chart version is applied.

controller/gateway/controller_reconciler_utils.go
```diff
@@ -335,7 +336,7 @@ func (r *Reconciler) ensureDataPlaneHasNetworkPolicy(
 func generateDataPlaneNetworkPolicy(
 	namespace string,
 	dataplane *operatorv1beta1.DataPlane,
-	_ *gwtypes.ControlPlane,
+	podLabels map[string]string,
```
pmalek, line 340: Let's make sure to add a godoc for this argument as it's not obvious which Pod these labels belong to.

modules/manager/controller_setup.go
```diff
@@ -354,6 +354,11 @@ func SetupControllers(mgr manager.Manager, c *Config, cpsMgr *multiinstance.Mana
...
+	if err != nil {
+		// TODO: output some log here?
+		podLabels = map[string]string{}
```
pmalek: I think that we could check if we're running in k8s (via a check on `KUBERNETES_SERVICE_HOST` env) and return an error then.

To allow running outside of the cluster I'd suggest to make sure failing to get self Pod labels is still supported (and does not create the network policy).

WDYT?

pkg/utils/kubernetes/self_metadata.go
pmalek: I believe this is fine as is. Getting the env should be enough (since it's us who control the envs injected in operator's container) but getting the service account namespace file is a nice fallback 👍

pkg/utils/kubernetes/self_metadata.go
```diff
@@ -0,0 +1,56 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("Cannot find pod labels from file %s: %w", podLabelsFile, err)
+	}
+
+	labelList := strings.Split(string(buf), "\n")
+	ret := make(map[string]string, len(labelList))
+	for _, label := range labelList {
+		labelKV := strings.SplitN(label, "=", 2)
+		if len(labelKV) != 2 {
+			// TODO: return error here?
```
pmalek: Since we're controlling which labels are passed down using downward API I say we return an error here because we're working and a known set.

controller/gateway/controller_reconciler_utils.go
randmonkey: So you think we should just keep the labels with these given keys? While all the labels specific to the version of charts/operator, and other custom labels (such as labels that are specified by `Values.extraLabels`) are not preserved?

pmalek: That's the approach I think make most sense at this moment. I can be persuaded otherwise though.

What are your thoughts on this?

randmonkey: Changed to only reserve partial labels.

charts/kong-operator/templates/_helpers.tpl
randmonkey: done

modules/manager/controller_setup.go
randmonkey: Updated the logic like this:
- When running in k8s, return error if failed to get pod labels
- When running outside of k8s, bypass the check and do not create network policies
WDYT?

pkg/utils/kubernetes/self_metadata.go
```diff
@@ -0,0 +1,61 @@
...
+// GetSelfPodLabels gets all the labels of the KO pod.
+func GetSelfPodLabels() (map[string]string, error) {
+	buf, err := os.ReadFile(podLabelsFile)
+	if err != nil {
+		return nil, fmt.Errorf("cannot find pod labels from file %s: %w", podLabelsFile, err)
+	}
+
+	labelList := strings.Split(string(buf), "\n")
+	ret := make(map[string]string, len(labelList))
+	for _, label := range labelList {
```
pmalek: We could split using the iterator (that would prevent us being able to use the map size hint but that's rarely useful, and doesn't impact performance much, especially on small sized maps like here).

```suggestion
	labels := strings.SplitSeq(string(buf), "\n")
	ret := make(map[string]string)
	for label := range labels {
```

pkg/utils/kubernetes/self_metadata.go
```diff
@@ -0,0 +1,61 @@
...
+		if err != nil {
+			continue
+		}
+		ret[key] = value
+	}
+	return ret, nil
+}
+
+// RunningOnKubernetes returns true if it is running in the kubernetes environment.
+// If the env KUBERNETES_SERVICE_HOST is configured to access the kubernetes API server, it is considered to running on k8s.
```
pmalek:
```suggestion
// If the env KUBERNETES_SERVICE_HOST is configured to access the kubernetes API server, // it is considered to be running on k8s.
```

pmalek (commented):
Minor nits, otherwise 👍

pkg/utils/kubernetes/self_metadata.go
```diff
@@ -0,0 +1,61 @@
...
+		if err != nil {
+			continue
+		}
+		ret[key] = value
+	}
+	return ret, nil
+}
+
+// RunningOnKubernetes returns true if it is running in the kubernetes environment.
+// If the env KUBERNETES_SERVICE_HOST is configured to access the kubernetes API server, // it is considered to be running on k8s.
```
pmalek:
```suggestion
// If the env KUBERNETES_SERVICE_HOST is configured to access the kubernetes API server,
// it is considered to be running on k8s.
```

pmalek (approved):

PR #1846, opened 2025-07-11
feat: add cache sync cli flag

### LINKED ISSUE #1845: Add cache sync-period configuration option
### Problem statement

For feature parity with KIC this issue tracks the effort of adding the API change and implementation for an equivalent of [`--sync-period` KIC CLI flag](https://github.com/Kong/kubernetes-ingress-controller/blob/f3d04cb0b1507041dc0dde408a28f468ee1ebfff/internal/cmd/rootcmd/config/cli.go#L75) in KO.

### Acceptance criteria

- [ ] As a user I can control the cache sync period using a `--cache-sync-period` CLI flag and a corresponding environment variable.

pmalek:

This adds the `--cache-sync-period` flag which controls the controller-runtime's cache sync period which determines the minimum frequency at which watched resources are reconciled.

**Which issue this PR fixes**

Fixes #1845

This setting could be a setting on a `ControlPlane` but this PR proposes a CLI flag instead to set this globally to the same value for operator manager and `ControlPlane`'s managers.

The demand for this option on the `ControlPlane` itself is rather unlikely to happen but if that happens we can still introduce this in the `ControlPlane` API.

**PR Readiness Checklist**:

### DIFF
--- a/docs/cli-arguments.md
+++ b/docs/cli-arguments.md
@@ -6,6 +6,7 @@
 | ---- | ---- | ----------- | ------- |
 | `--anonymous-reports` | `bool` | Send anonymized usage data to help improve Kong. | `true` |
 | `--apiserver-host` | `string` | The Kubernetes API server URL. If not set, the operator will use cluster config discovery. | `""` |
+| `--cache-sync-period` | `string` | Determine the minimum frequency at which watched resources are reconciled. By default falls back to controller-runtime's default. | `0s` |
 | `--cache-sync-timeout` | `string` | The time limit set to wait for syncing controllers' caches. Defaults to 0 to fall back to default from controller-runtime. | `0s` |
 | `--cluster-ca-key-size` | `string` | Size (in bits) of the key used for the cluster CA certificate. Only used for RSA keys. | `4096` |
 | `--cluster-ca-key-type` | `string` | Type of the key used for the cluster CA certificate (possible values: ecdsa, rsa). Default: ecdsa. | `ecdsa` |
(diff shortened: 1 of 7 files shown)


docs/cli-arguments.md
```diff
@@ -6,6 +6,7 @@
 | ---- | ---- | ----------- | ------- |
 | `--anonymous-reports` | `bool` | Send anonymized usage data to help improve Kong. | `true` |
 | `--apiserver-host` | `string` | The Kubernetes API server URL. If not set, the operator will use cluster config discovery. | `""` |
+| `--cache-sync-period` | `string` | Determine the minimum frequency at which watched resources are reconciled. By default falls back to controller-runtime's default. | `0s` |
```
programmer04: I suppose [this is it](https://github.com/kubernetes-sigs/controller-runtime/blob/252af6420feba970d0020ba4c029a0a1ce3f9688/pkg/cache/cache.go#L45-L47), thus default is `10h` not `0s`

pmalek: Yes, that's true but because we check the zero value in https://github.com/Kong/kong-operator/pull/1846/files#diff-a4f03a759edc33bd5b219e1699a2d0a32d4c367ef1263865097c64fedeedaf92R452-R454 we'll not set it in the config and thus fall back to the controller-runtime default (10h).

Do you think there's a better way to handle this?

programmer04: To make it simple, maybe adjust the sentence
> By default falls back to controller-runtime's default.

to something like that

> By default or for `0s` value, it falls back to controller-runtime's default.

It's bad that `controller-runtime` doesn't export a constant for this value, because otherwise we could use it here instead of zero to set it as the default. Inling `10h` may change in the future, and we'll probably forget about adjusting it

pmalek: 👍 Adjusted in 6033d759d4aa6a228ef730cb16e7527b670e373d. PTAL.

programmer04 (dismissed):

programmer04 (approved):

Issue #1847, opened 2025-07-11
Adjust golangci-lint configuration


programmer04:
## Problem Statement

Due to the merge of repositories (putting KIC into KO)

- https://github.com/Kong/kong-operator/issues/1567

Some rules were relaxed to keep PR manageable.

Furthermore, for KIC `statickcheck` used to be run as a separate, decide if it's worth to have it in KO too.

## Proposed Solution

See what has been relaxed and how the rules used to look in KIC](https://github.com/Kong/kubernetes-ingress-controller/blob/main/.golangci.yaml) to finish with the ultimate strict config and reported findings fixed

## Acceptance Criteria
- [ ] no more `TODO:` in `.golangci-lint.yaml`
- [ ] all findings fixed
- [ ] thoughtful configuration
- [ ] `statickcheck` decision

sandromodarelli:
punting due to priorities

PR #1853, opened 2025-07-14
chore(lint): ensure the same order of imports

programmer04:

Ensure that order of imports is the same everywhere, `gci` allows it.

**Which issue this PR fixes**

Part of 

- #1847

needed now to make working with repo easier


alacuku (approved):
/lgtm

Issue #1567, opened 2025-05-06
KIC source code is located under `ingress-controller` directory in the new repo, preserving its history


sandromodarelli:
KIC source code should be located under ingress-controller directory in the new repo, preserving its history.
KIC git tags perhaps should also be migrated and prefixed accordingly to avoid potential collisions.

For more context, see the [Repository and branding](https://docs.google.com/document/d/1CsD-KwBkN2UStdvCk9r5rhjXhjAiblUMIiH2yQdwBv0/edit?tab=t.0#heading=h.tl1464pma79w) section of the design doc.

sandromodarelli:
punting to KO 2.0.0 because we can still import the logic from KIC repository to have the alpha release

sandromodarelli:
blocked by https://github.com/Kong/kubernetes-ingress-controller/issues/7501

sandromodarelli:
unblocked since we released KIC 3.5.0

PR #1881, opened 2025-07-17
chore(ci): run all e2e tests

programmer04:
## What this PR does / why we need it:

Run e2e tests from `ingress-controller` subdir too

## Which issue this PR fixes

Part of

- https://github.com/Kong/kong-operator/issues/1567

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -605,7 +605,9 @@ _test.e2e: gotestsum
 		-timeout $(E2E_TEST_TIMEOUT) \
 		-ldflags "$(LDFLAGS_COMMON) $(LDFLAGS) $(LDFLAGS_METADATA)" \
 		-race \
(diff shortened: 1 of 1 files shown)


pmalek (commented):
I'm wondering if these e2e tests make much sense now: they execute in the context of installing KIC in the cluster which is not what we'll leverage at any point in time now ( we're going to be using KIC in-process as a `ControlPlane` ).

I think ( but I can be persuaded ) that we should rather get rid of these or rework them ( but I doubt this would make sense.

I'm curious to your thoughts on this. LMK.

programmer04:
You're right @pmalek, please check

- https://github.com/Kong/kong-operator/issues/1903

and I'm closing this PR

PR #1882, opened 2025-07-17
fix(test/helpers): handle telepresence traffic manager upgrade if aleady installed

alacuku:

Update SetupTelepresence to run `telepresence helm upgrade` if `telepresence helm install` fails with a message indicating the manager is already installed. This improves idempotency and error handling, making it handy when running tests multiple times and reusing the same environment.

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:

### DIFF
--- a/test/helpers/telepresence.go
+++ b/test/helpers/telepresence.go
@@ -1,11 +1,11 @@
 package helpers
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"os"
 	"os/exec"
-	"strings"
 
 	"github.com/kong/kong-operator/test"
 )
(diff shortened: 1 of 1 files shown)


test/helpers/telepresence.go
```diff
@@ -28,7 +29,11 @@ func SetupTelepresence(ctx context.Context) (func(), error) {
 	}
 
 	out, err := exec.CommandContext(ctx, telepresenceExecutable, "helm", "install").CombinedOutput()
-	if err != nil {
+	if err != nil && strings.Contains(string(out), "use 'telepresence helm upgrade' instead to replace it") {
```
programmer04:
```suggestion
	if err != nil && bytes.Contains(out, "use 'telepresence helm upgrade' instead to replace it") {
```
and one conversion less

programmer04 (commented):
🆒

test/helpers/telepresence.go
alacuku: Well, we still need to convert the string to [] byte.

programmer04: yo're right heh

programmer04 (approved):

PR #1887, opened 2025-07-21
feat: introduce flags `--apiserver-host`, `--apiserver-qps` and `--apiserver-burst`

### LINKED ISSUE #1794: Add API server related configuration options
### Problem statement

For feature parity with KIC we need to add the following flags for controlling API server request queues:

- `--apiserver-burst`
- `--apiserver-host`
- `--apiserver-qps`

### Acceptance criteria 

- As a user I can control API server request queues via KO CLI flags

programmer04:

This PR uses a value set by `--apiserver-host` flag in a proper way and introduces two flags related to K8s client rate-limiting `--api-server-qps` and `--api-server-burst`.

**Which issue this PR fixes**

Closes #1794

**PR Readiness Checklist**:


CHANGELOG.md
```diff
@@ -39,6 +39,10 @@
   - Specifying the delay to wait for Kubernetes object caches sync before
     updating dataplanes by `spec.cache.initSyncDuration`
     [#1858](https://github.com/Kong/kong-operator/pull/1858)
+- Introduce flags `--apiserver-host` for API, `--api-server-qps` and
+  `--api-server-burst` to control the QPS and burst (rate-limiting) for the
```
```suggestion
- Introduce flags `--apiserver-host` for API, `--apiserver-qps` and
  `--apiserver-burst` to control the QPS and burst (rate-limiting) for the
```

controller/controlplane/manager_options.go
```diff
@@ -133,6 +133,14 @@ func WithFeatureGates(logger logr.Logger, featureGates []gwtypes.ControlPlaneFea
 	}
 }
 
+// WithQPSAndBurst sets the QPS and burst for the API server.
+func WithQPSAndBurst(qps float32, burst int) managercfg.Opt {
+	return func(c *managercfg.Config) {
+		c.APIServerQPS = int(qps)
```
pmalek, line 148: Any particular reason to accept a float32 when we need an int?

```suggestion
func WithQPSAndBurst(qps int, burst int) managercfg.Opt {
	return func(c *managercfg.Config) {
		c.APIServerQPS = qps
```

programmer04: because on the caller site we have `float32`, and having it like that prevents from an ordering mistake. But I agree, one way or another it's a little bit ugly at the end of the day

pmalek: Perhaps adding new types QPS and Burst for these would solve this more elegantly and prevent mistakes? WDYT?

programmer04: But still, from CLI we take `int`, but here we need to pass `float32`, and in KIC subsystem we have the same; thus, forth and back conversion is needed. When we get rid of cli portion of KIC config will be able to be adjusted.

Type system in Go still doesn't prevent you from making a mistake. You can pass untyped `const` as whatever type that wraps a numeric type, and it will compile. 

So tbh, from my perspective, it doesn't matter much where conversion happens in this case. I can move it upper if you prefer that way. WDYT?

pmalek: > Type system in Go still doesn't prevent you from making a mistake. You can pass untyped const as whatever type that wraps a numeric type, and it will compile.

Sadly that a reality we have to live in. Adding new types for arguments would at least help in suggesting to user which argument should use which type.

I see your point about using KIC's config fields. I'm OK leaving this as is.

pmalek (dismissed):
👍 

Please make sure to update the PR title and description so that when merged it contains the correct flags.

.gitignore
```diff
@@ -6,6 +6,7 @@
 *.dylib
 bin/*
 *_debug_bin*
+main
```
tao12345666333, line 9: Nit: why we add this one here?

programmer04: because when you compile directly, you'll get this artifact, so it's better to ignore it imho

pmalek (approved):

PR #1918, opened 2025-07-24
chore(controlplane): Use time.Duration to specify interval and timeout for syncing config with dataplane

### LINKED ISSUE #1904: Use `time.Duration` to specify the interval and timeout of control planes to sync configuration to data planes
## Problem Statement

We migrated the `ProxySyncSeconds` and `ProxyTimeoutSeconds` in the KIC repo to specify intervals and timeouts for control planes to sync configuration to data planes. They are in `float32` types to present the durations in seconds. 

We now control the configuration by CRDs instead of CLI flags, so we can change the config items to `time.Duration` type to make it clearer and consistent with CRD.

## Proposed Solution

- Change the `ProxySyncSeconds` and `ProxyTimeoutSeconds` into `time.Duration` type for specifying intervals and timeouts
- Update the code to create `dataplaneSynchronizer` to use `time.Duration` instead of `float32`

## Additional Information

...

## Acceptance Criteria
- [ ] The control plane manager uses `time.Duration` to specify the interval and timeout of control planes to sync configuration to data planes

randmonkey:

Use `time.Duration` instead of `float32` to specify interval and timeout of  syncing config with dataplane.

**Which issue this PR fixes**

Fixes #1904

Since we will not run a standalone KIC from the `ingress-controller` package, we do not need to care about compatibility in the  CLI flags of KIC. Maybe later we should remove the CLI flags in `ingress-controller` package and use new approach to generate default config.
**PR Readiness Checklist**:

Does this chagne requires a changelog? I think this is a non-userfacing chagne.


ingress-controller/internal/manager/setup.go
```diff
@@ -153,21 +153,21 @@ func setupDataplaneSynchronizer(
...
+	proxySyncInterval time.Duration,
 	initCacheSyncWait time.Duration,
 ) (*dataplane.Synchronizer, error) {
-	if proxySyncSeconds < dataplane.DefaultSyncSeconds {
+	if proxySyncInterval < dataplane.DefaultSyncInterval {
 		logger.Info(fmt.Sprintf(
-			"WARNING: --proxy-sync-seconds is configured for %fs, in DBLESS mode this may result in"+
-				" problems of inconsistency in the proxy state. For DBLESS mode %fs+ is recommended (3s is the default).",
-			proxySyncSeconds, dataplane.DefaultSyncSeconds,
+			"WARNING: --proxy-sync-seconds is configured for %v, in DBLESS mode this may result in"+
```
pmalek: This is not really relevant but for consistency let's keep this in line with the flag name:


```suggestion
			"WARNING: --proxy-sync-timeout is configured for %v, in DBLESS mode this may result in"+
```

ingress-controller/pkg/manager/config/config.go
```diff
@@ -62,9 +62,10 @@ type Config struct {
 	GatewayDiscoveryReadinessCheckInterval time.Duration
 	GatewayDiscoveryReadinessCheckTimeout  time.Duration
 	KongAdminSvcPortNames                  []string
-	ProxySyncSeconds                       float32
-	InitCacheSyncDuration                  time.Duration
-	ProxyTimeoutSeconds                    float32
+	// REVIEW: totally remove it as we do not support running ingress-controller by CLI in KO repo?
```
pmalek: I believe we can create a separate tracking issue for getting rid of all the CLI machinery in KIC's subdirectory. WDYT?

randmonkey: Created https://github.com/Kong/kong-operator/issues/1921.

ingress-controller/internal/manager/setup.go
randmonkey: Done. It should be `--proxy-sync-interval`.

pmalek (approved):

PR #1924, opened 2025-07-24
fix(ci): sync of KO chart to charts repo

programmer04:

See for the description

- https://github.com/Kong/charts/pull/1358

This change ensures that the removal of the file is replicated too, see an example PR (to close without merging) created here

- https://github.com/Kong/charts/pull/1359

by test commit 997be67ff0209e9d1f214260fcaf09a95fc4e36a from this PR, which is now reverted.


pmalek (commented):
Why do we remove the readme and notes files 🤔 ?

programmer04:
Fixed it was a leftover from test

pmalek (approved):

PR #1925, opened 2025-07-24
feat(makefile): add telepresence targets for IDE debugging

alacuku:

- Add install.telepresence target for connecting to Kubernetes clusters
- Add uninstall.telepresence target for cleaning up telepresence resources

These new targets provide essential functionality for debugging the operator from an IDE by establishing connectivity between the local development environment and the Kubernetes cluster. The implementation handles common edge cases like already-installed or connected instances, ensuring a smoother development experience when testing operator functionality locally.

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -882,3 +882,20 @@ deploy: manifests kustomize
 .PHONY: undeploy
 undeploy:
 	$(KUSTOMIZE) build config/default | kubectl delete --wait=false --ignore-not-found=$(ignore-not-found) -f -
+
+# Install and connect telepresence to the cluster.
+# This target is essential for debugging the operator from an IDE as it establishes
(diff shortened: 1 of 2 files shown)


mlavacca (commented):
This is a great improvement. However, I think we need a dedicated script under the hack folder for it, as this Makefile target is definitely bigger than we want. What do you think?

alacuku:
> This is a great improvement. However, I think we need a dedicated script under the hack folder for it, as this Makefile target is definitely bigger than we want. What do you think?

The core logic lives now in `scripts/telepresence-manager.sh`.

mlavacca (approved):

PR #1933, opened 2025-07-24
chore(ci): run integration and envtest tests for `ingress-controller` too

### LINKED ISSUE #1567: KIC source code is located under `ingress-controller` directory in the new repo, preserving its history

programmer04:

PR https://github.com/Kong/kong-operator/pull/1931 adjusted wrong workflow for `envtest` tests, here it's fixed. 

The same method ensures that all integration tests are run too.

**Which issue this PR fixes**

Closes #1567

will create a spin-off issues that describe the next steps for test suites

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -367,11 +367,18 @@ jobs:
       with:
         install: false
 
+    - uses: Kong/kong-license@c4decf08584f84ff8fe8e7cd3c463e0192f6111b # master @ 20250107
+      id: license
+      with:
+        op-token: ${{ secrets.OP_SERVICE_ACCOUNT_TOKEN }}
+        
+
     - name: run envtest tests
       working-directory: ${{ matrix.directory }}
       run: make test.envtest
       env:
-        GOTESTSUM_JUNITFILE: "envtest-tests.xml"
+        KONG_LICENSE_DATA: ${{ steps.license.outputs.license }}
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yaml
```diff
@@ -486,15 +498,15 @@ jobs:
     - name: collect test coverage
       uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
       with:
-        name: coverage-integration
-        path: coverage.integration.out
+        name: ${{ matrix.directory == '.' && 'ko' || matrix.directory }}-coverage-integration
```
pmalek, line 508: This looks a little bit confusing, does that actually work? It shouldn't really be that important given that we'll eventually merge all the tests and we won't run them separately like here.

programmer04: It works (you can see by examining logs), I took it from here https://github.com/actions/runner/issues/409
I have to do it to have a valid and non-duplicated name for an artifact. And you're right, we got rid of this

pmalek (approved):

PR #1942, opened 2025-07-28
feat(controller): handle transient errors during ControlPlane scheduling

### LINKED ISSUE #1375: Improve logging when creating a new KIC instance in CP controller
## Problem Statement

Currently, when creating a new KIC instance in the CP controller it's likely to hit a case where DP instances are not ready yet. That makes KIC instances not boot successfully and log errors that might be worrying to the users.

```text
[manager] 2025-03-19T08:11:31.382Z - ERROR - Reconciler error - {"controller": "controlplane", "controllerGroup": "gateway-operator.konghq.com", "controllerKind": "ControlPlane", "ControlPlane": {"name":"kong-xjw9p","namespace":"default"}, "namespace": "default", "name": "kong-xjw9p", "reconcileID": "ca7e9913-6b35-428c-bde1-f61638d275a8", "error": "failed to create manager: failed to create manager: unable to build kong api client(s): All attempts fail
```

## Proposed Solution

Make the KIC instance constructor not log errors when it cannot connect to DP instances. Instead, return a specific error that can be handled on the CP controller side. Make the CP controller requeue the request for a given CP if it encounters such an error.

## Acceptance Criteria
- [ ] No `Failed to create kong client(s)` error logs are emitted when CP controller cannot create a KIC instance because of not ready DPs

alacuku:

- Added handling for `NoAvailableEndpointsError` and `io.EOF` as transient errors in `handleScheduleInstanceOutcome`.
- Updated the status of `ControlPlane` to reflect retry conditions when transient errors occur.
- Improved logging for better traceability during error handling and retries.

**Which issue this PR fixes**

Fixes #1375 

**PR Readiness Checklist**:

### DIFF
--- a/controller/controlplane/controller.go
+++ b/controller/controlplane/controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"reflect"
 	"strings"
 	"time"
@@ -32,6 +33,7 @@ import (
 	"github.com/kong/kong-operator/controller/pkg/log"
 	"github.com/kong/kong-operator/controller/pkg/op"
 	"github.com/kong/kong-operator/controller/pkg/secrets"
+	ingresserrors "github.com/kong/kong-operator/ingress-controller/pkg/errors"
 	"github.com/kong/kong-operator/ingress-controller/pkg/manager"
 	managercfg "github.com/kong/kong-operator/ingress-controller/pkg/manager/config"
 	"github.com/kong/kong-operator/ingress-controller/pkg/manager/multiinstance"
@@ -263,7 +265,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 			}
 			mgrCfg.DisableRunningDiagnosticsServer = true
 			if err := r.scheduleInstance(ctx, logger, mgrID, mgrCfg); err != nil {
-				return ctrl.Result{}, err
+				return r.handleScheduleInstanceOutcome(ctx, logger, cp, err)
 			}
 			r.ensureControlPlaneStatus(cp, mgrCfg)
 		}
@@ -299,7 +301,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 			}
 
 			if err := r.scheduleInstance(ctx, logger, mgrID, mgrCfg); err != nil {
-				return ctrl.Result{}, fmt.Errorf("failed to schedule instance: %w", err)
+				return r.handleScheduleInstanceOutcome(ctx, logger, cp, err)
 			}
 			r.ensureControlPlaneStatus(cp, mgrCfg)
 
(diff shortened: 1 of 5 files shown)


ingress-controller/internal/manager/setup.go
```diff
@@ -374,10 +368,10 @@ func AdminAPIClientFromServiceDiscovery(
 		retry.Delay(delay),
 		retry.OnRetry(func(_ uint, err error) {
 			// log the error if the error is NOT caused by 0 available gateway endpoints.
-			if !errors.As(err, &NoAvailableEndpointsError{}) {
-				logger.Error(err, "Failed to create kong client(s)")
+			if !errors.As(err, &ingresserrors.NoAvailableEndpointsError{}) {
+				logger.V(logging.DebugLevel).Info("Failed to create kong client(s)", "error", err)
```
pmalek, line 374: I think this way this will be logged twice, how about adding a return here to prevent that?

```suggestion
				logger.V(logging.DebugLevel).Info("Failed to create kong client(s)", "error", err)
				return
```

controller/controlplane/controller.go
pmalek, line 1: Can we add an integration test that will check the condition set on the `ControlPlane`?

We can extend the `TestControlPlaneEssentials` test for this with a subtest. WDYT?

controller/controlplane/controller.go
```diff
@@ -499,3 +501,39 @@ func (r *Reconciler) initStatusToWaitingToBecomeReady(
...
+	// - io.EOF: indicates that the connection to the dataplane service was closed unexpectedly.
+	// These errors are considered transient and will be retried after a delay.
+	if errors.As(err, &ingresserrors.NoAvailableEndpointsError{}) || errors.Is(err, io.EOF) {
+		logger.Info("Transient error encountered while creating kong api clients, retrying after delay", "error", err, "retryDelay", requeueAfterBoot)
+		k8sutils.SetCondition(
+			k8sutils.NewCondition(
+				kcfgdataplane.ReadyType,
+				metav1.ConditionFalse,
+				kcfgdataplane.WaitingToBecomeReadyReason,
+				kcfgdataplane.WaitingToBecomeReadyMessage,
```
pmalek: How about we're more specific here and add the message about not available endpoints (and also set an appropriate reason)? Perhaps `NoAvailableKongEndpoints` would be a good fit here?

alacuku: There could be other reasons besides `NoAvailableKongEndpoints` for the failure. We could consider setting the error message as the message.

Regarding the reason, `WaitingToBecomeReadyReason` seems appropriate because it accurately reflects the state of waiting for the dataplane endpoints to become ready.

ingress-controller/internal/manager/setup.go
alacuku: I tried to not modify the existing code, but yeah, I can do that.

controller/controlplane/controller.go
pmalek: > We could consider setting the error message as the message.

👍 

> There could be other reasons besides NoAvailableKongEndpoints for the failure. We could consider setting the error message as the message.

> Regarding the reason, WaitingToBecomeReadyReason seems appropriate because it accurately reflects the state of waiting for the dataplane endpoints to become ready.

👍

controller/controlplane/controller.go
alacuku: Adding an integration test to check the condition set on the `ControlPlane` is not a straightforward task due to the nature of the transient error being tested. The error occurs when no endpoints are available or the client cannot connect to the endpoints. This scenario is difficult to reproduce in integration tests because:
* Dependency on kubernetes controllers: the service and its endpoints are managed by the Kubernetes controller-manager and the dataplane controller. 
* Transient Nature of the Error: it is subject to race conditions. This makes it unreliable to reproduce consistently in tests.

The most viable approach would be to modify the `ControlPlane` controller to use a configurable `serviceGetter`. This way we can mock it during tests to return a crafted service with no endpoints allowing us to simulate the error condition is sett correctly.
However, introducing changes to the `ControlPlane` controller solely for testing purposes may no be ideal. We need to carefully evaluate whether they are justified for testing this specific case.

pmalek: That's a valid point 👍 I missed the fact that we're only adding a negative status condition here.

Let's leave it at that then. Thanks for looking into this 🙇

controller/controlplane/controller.go
alacuku: This is a snippet of how the conditions are updated based on the error type:
```json
[
  {
    "lastTransitionTime": "2025-07-29T12:50:53Z",
    "message": "Unable to connect to data plane: client not ready: making HTTP request: Get \"https://10.244.0.7:8444/status\": dial tcp 10.244.0.7:8444: i/o timeout",
    "reason": "WaitingToBecomeReady",
    "status": "False",
    "type": "Ready"
  },
   {
    "lastTransitionTime": "2025-07-29T13:04:21Z",
    "message": "Unable to connect to data plane: no endpoints for service: \"default/dataplane-admin-dataplane-v2-vnqm6\"",
    "reason": "WaitingToBecomeReady",
    "status": "False",
    "type": "Ready"
  },
  {
    "lastTransitionTime": "2025-07-29T13:04:23Z",
    "message": "Unable to connect to data plane: client not ready: making HTTP request: Get \"https://10.244.0.9:8444/status\": EOF",
    "reason": "WaitingToBecomeReady",
    "status": "False",
    "type": "Ready"
  },
  {
    "lastTransitionTime": "2025-07-29T13:04:27Z",
    "message": "Waiting for the resource to become ready",
    "reason": "WaitingToBecomeReady",
    "status": "False",
    "type": "Ready"
  }
]
```

WDYT?

pmalek: I think these will do for now 👍 ( even though 1st and 3rd do seem to be almost identical for the end user)

pmalek (approved):

Issue #1944, opened 2025-07-28
`GatewayConfiguration` `v1beta1` conversion to `v2beta1`


mlavacca:
## Problem Statement

Similarly as what we are doing in https://github.com/Kong/kong-operator/issues/1368 for the CP, we need to support conversion between `v1beta1` and `v2alpha1` versions of the `GatewayConfiguration` CRD.

## Acceptance Criteria
- [ ] as a user, I can migrate from `GatewayConfiguration` `v1beta1` to `GatewayConfiguration` `v2alpha1`.

mlavacca:
Blocked by #1728

randmonkey:
Removed the "blocked" tag as #1728 closed.

Issue #1368, opened 2025-03-14
ControlPlane v1beta1 conversion to v2beta1


czeslavo:
### Problem statement

To make sure our migration path from the old v1beta1 ControlPlane API is as smooth as possible, we should provide a conversion webhook that will map the old v1beta1 objects onto the new v2alpha1 API.

We should only support the conversion of the environment variables that are already exposed as strongly typed parameters in the v2alpha1 API.

See https://book.kubebuilder.io/multiversion-tutorial/conversion for reference on how it can be achieved.

### Acceptance criteria
- [ ] As a user that has a `ControlPlane` v1beta1 installed in the cluster, KO reconciles it properly (converts it to the new v2alpha1, handles tearing down the old resources and spins up a KIC instance)
- [ ] GatewayConfiguration is covered similarly

programmer04:
#1733 should be resolved or at least updated when this issue will be resolved

PR #1959, opened 2025-07-30
feat: introduce conversion webhook machinery, register `ControlPlane`, `GatewayConfiguration` CRDs

programmer04:

This PR introduces conversion webhook machinery.

A single place to register types, based on which the autogeneration of respective YAMLs happens. Actual conversion logic is supposed to be implemented on a particular type, thus, for now it's `kubernetes-configuration` repo.

Structure of kustomize related YAMLs is inherited from [kubebuilder machinery](https://book.kubebuilder.io/multiversion-tutorial/conversion), see also [their example](https://github.com/kubernetes-sigs/kubebuilder/tree/98bc8f4b45463f599155270fb6548a1ede817791/docs/book/src/multiversion-tutorial/testdata/project/config). It means that files under `config/` are autogenerated by kubebuilder machinery and they won't change (it's one time when you're adding a webhook), only the file `config/crd/patches/zz_generated_conversion_webhook.yaml` is generated by a custom script `hack/generators/conversion-webhook/main.go` to include registered types out of the box (instead of adding them manually).

Dependency on `cert-manager` is presented in the official docs and seems to be the best solution supported out of the box.

Spin-off issues created based on this PR to handle them later

- #1986
- #1987
- #1989

**Which issue this PR fixes**

Prerequisite for e.g.
- #1368
- #1944

It relies on changes introduced in https://github.com/Kong/kubernetes-configuration/pull/488. To prove that the machinery works, the actual conversion attempt ends with a panic

To ensure that `cert-manager` is in place 

```
make install.all
```

and next

```
make run.skaffold
```

next 

```
kubectl apply -f config/samples/controlplane.yaml
```

ends with 

```log
Warning: ControlPlane v1beta1 has been deprecated in favor of v2alpha1 and it will be removed in the future.
Error from server: error when creating "config/samples/controlplane.yaml": conversion webhook for gateway-operator.konghq.com/v1beta1, Kind=ControlPlane failed: Post "https://gateway-operator-webhook-service.kong-system.svc:443/convert?timeout=30s": EOF
```

And in logs, you can observe

```log
v1beta1 ControlPlane ConvertFrom(...) is not implemented yet
```

Something similar will happen to `GatewayConfiguration`, just apply `config/samples/gateway-with-gatewayconfiguration.yaml` and look for `v1beta1 GatewayConfiguration ConvertTo(...) is not implemented yet`


mlavacca (commented):
I have a broader question before getting into the code details: are we posing the requirement to have cert-manager deployed in the cluster to the users? It seems so from the code, but want to double check here.

programmer04:
Yes, @mlavacca, I don't see other way to do it easily, but I'm open to the discussion. As long as we are not releasing 2.0, we can adjust/change it (I'd like to raise this topic on team-k8s sync).

For this PR, I want to proceed to allow the implementation of conversion logic for types, to unblock others. And the aforementioned can be adjusted later

pmalek:
(before going into the code review):

We discussed this with @programmer04 and the alternative seems to be handling this in the Helm chart.

The advantage of using the cert-manager is that
- we can safely assume that many of our users are either already familiar with it or even already using it
- we have the cert rotation solved and delegated

hack/generators/conversion-webhook/main.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/kong/kong-operator/internal/webhook/conversion"
+	"github.com/samber/lo"
+)
+
+var template = `---
```
mlavacca, line 11: Can we add a note along the lines of "this file is generated by X"?

hack/generators/conversion-webhook/main.go
```diff
@@ -0,0 +1,53 @@
...
+          name: gateway-operator-webhook-service
+          path: /convert
+      conversionReviewVersions:
+      - v1
+`
+
+// Update conversion_webhook.yaml with types for which conversion webhook
+// is registered in KO, to make sure that the CRD patch is always up to date.
+func main() {
+	file, err := os.Create("config/crd/patches/conversion_webhook.yaml")
```
mlavacca: we should follow the naming convention for generated files `zz_generated_<filename>.go`

mlavacca (commented):
This change looks mostly good to me, I've just jeft a couple of comments

hack/generators/conversion-webhook/main.go
tao12345666333: +1

programmer04: It's been done in line 43

config/certmanager/issuer.yaml
```diff
@@ -0,0 +1,13 @@
+# The following manifest contains a self-signed issuer CR.
+# More information can be found at https://docs.cert-manager.io
+# WARNING: Targets CertManager v1.0. Check https://cert-manager.io/docs/installation/upgrading/ for breaking changes.
```
tao12345666333:
```suggestion
# WARNING: Targets CertManager v1.18. Check https://cert-manager.io/docs/installation/upgrading/ for breaking changes.
```

we use the v1.18, right?

tao12345666333 (commented):
:+1:  LGTM.

config/certmanager/issuer.yaml
programmer04: Good catch. I think that those comments are even redundant, since v1.0 was released ~5 years ago. It's generated by kubebuilder when you bootstrap a skafolding for webhook. I'm removed them

mlavacca (approved):

PR #1972, opened 2025-08-01
feat: GatewayConfiguration v2alpha1 support

### LINKED ISSUE #1728: Kong operator: support v2 GatewayConfiguration

mlavacca:

**Which issue this PR fixes**

Fixes #1728 

**PR Readiness Checklist**:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,7 +43,10 @@
   `konghq.com/configmap`.
   [#1922](https://github.com/Kong/kong-operator/pull/1922)
 - `GatewayConfiguration` `v1beta1` has been replaced by the new API version `v2alpha1`.
+  The `GatewayConfiguration` `v1beta1` is still available but has been marked as
+  deprecated.
   [#1792](https://github.com/Kong/kong-operator/pull/1972)
+  <!-- https://github.com/Kong/kong-operator/issues/1944 -->
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,8 @@
   The default value of the two labels are set to `konghq.com/secret` and
   `konghq.com/configmap`.
   [#1922](https://github.com/Kong/kong-operator/pull/1922)
+- `GatewayConfiguration` `v1beta1` has been replaced by the new API version `v2alpha1`.
+  [#1792](https://github.com/Kong/kong-operator/pull/1972)
 
 ### Added
 
--- a/controller/gateway/controller_watch.go
+++ b/controller/gateway/controller_watch.go
@@ -17,6 +17,7 @@ import (
 	gatewayv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
 
 	operatorv1beta1 "github.com/kong/kubernetes-configuration/v2/api/gateway-operator/v1beta1"
+	operatorv2alpha1 "github.com/kong/kubernetes-configuration/v2/api/gateway-operator/v2alpha1"
 	konnectv1alpha2 "github.com/kong/kubernetes-configuration/v2/api/konnect/v1alpha2"
 
 	"github.com/kong/kong-operator/controller/pkg/log"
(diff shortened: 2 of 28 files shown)


CHANGELOG.md
```diff
@@ -42,6 +42,8 @@
   The default value of the two labels are set to `konghq.com/secret` and
   `konghq.com/configmap`.
   [#1922](https://github.com/Kong/kong-operator/pull/1922)
+- `GatewayConfiguration` `v1beta1` has been replaced by the new API version `v2alpha1`.
```
pmalek, line 45: Can we expand this to describe 
- If any user action is required?
- If we're still going to support `v1beta1`? If so are we going to provide a migration webhook?
- Should users first install the new CRD and then upgrade the operator?

controller/gateway/controller_watch.go
```diff
@@ -17,6 +17,7 @@ import (
 	gatewayv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
 
 	operatorv1beta1 "github.com/kong/kubernetes-configuration/v2/api/gateway-operator/v1beta1"
+	operatorv2alpha1 "github.com/kong/kubernetes-configuration/v2/api/gateway-operator/v2alpha1"
```
pmalek, line 20: WDYT about using https://github.com/Kong/kong-operator/blob/57307f0d56c4c5e109baab66696d3b957c1ab5a5/controller/gateway/types.go#L15 (or https://github.com/Kong/kong-operator/blob/0356fc46538454b8d856662de5f979d1b92d6abd/internal/types/operatortypes.go#L46 if you prefer, the former is an alias to this).

This is already in place to drive the `GatewayConfiguration` used in this repo. WDYT?

Whatever's our decision on this I believe we should be consistent, so that we either use it or drop these alias types entirely.

mlavacca: I am on the fence on these aliases, honestly. 
We have never been too consistent on using them, and I can find several places where we reference the APIs by their original package. When navigating the code, I find it more useful to have the package available at a first glance rather than navigating the code to figure it out.

CHANGELOG.md
mlavacca: I'd lean toward not having guides on how to migrate in the CHANGELOG. The conversion webhook is something we have planned in https://github.com/Kong/kong-operator/issues/1944, and I wouldn't mention it until we address it. Let me add a TODO in the CHANGELOG with that issue.

mlavacca: f9245ba88024de0448fe83d1ec35fc93d882b106

controller/gateway/controller_watch.go
pmalek: I do see your point but OTOH changes like the one in this PR would require only 1 line to be changed.

Do you feel that chasing all the references of a package that's being changed is worth throwing away all the aliases?

mlavacca: > I do see your point but OTOH changes like the one in this PR would require only 1 line to be changed.
>
> Do you feel that chasing all the references of a package that's being changed is worth throwing away all the aliases?

It's not the point of changing this single line. My opinion is that having these aliases in place is not worth it at all. From my point of view is a form of over-engineering with very little benefit. Bumping an API from versions X to version X+1 is a matter of `replace all X->X+1` in the whole codebase. A not complex task. While maintaining this aliases package for all our APIs has been proven to be not easy, as we are not consistent at all. Moreover, I don't see the advantage of it when navigating the code: this way, the API version is hidden by the aliases package and requires some code navigation to be figured out.

mlavacca: Do you strongly feel that the aliases package is something we need to have?

pmalek: I see your point. I have 2 comments to make:

- > Bumping an API from versions X to version X+1 is a matter of `replace all X->X+1` in the whole codebase. A not complex task.

  This is not always (or rather rarely is) the case. We often use more than 1 API from a package so it's not always "find & replace" but it involves a bit more than that.

- Code reviews are way easier to conduct when we change 1 line vs when we change 10s or 100s of files.

All in all, if consistency's a problem here we can always make linter rules a bit more strict to ensure that only the "types" package imports the API packages.

pmalek: > Do you strongly feel that the aliases package is something we need to have?

I don't but I think there's a benefit to be had there when they are properly incorporated in the repo. We can take this discussion to a separate issue or discuss it at a team sync. WDYT?

mlavacca: sure, let's discuss it in the team sync

pmalek: [Here's a PR](https://github.com/Kong/kong-operator/pull/1977) that takes care of enforcing the `ControlPlane` related types (incomplete yet but a good starter).

I'll close this discussion as we do not want to handle that in you PR here.

pmalek (approved):

controller/gateway/controller_watch.go
mlavacca: cool, thanks for creating that!

mlavacca:
Blocking as I want to perform a last check on the logic before merging it.

Issue #1693, opened 2025-05-30
Kong operator: multi tenancy - watch namespaces


pmalek:
### Problem statement

KGO 1.6 introduced a concept of watch namespaces for `ControlPlane` which controls which namespaces are being watched by it.

This task tracks the design and implementation of the same mechanism for KO to allow multi tenancy to work.

One has to note that with KO, the control plane will be an in process entity and thus will share the RBAC permissions with the operator.

### Acceptance criteria

- [ ] As a user I can set a namespace (or a list of namespaces?) in operator config to be watched and I expect only resources in those namespaces to be watched.

pmalek:
@justindavies 👋 Do we have feedback / decision on this yet?

Punting until we have a decision on this.

justindavies:
Ideally, to satisfy customer requirements, we have the ability for  KO that is namespaced scoped (multiple operators), OR a cluster scoped operator that is able to watch specific namespaces.  Eventually we would need to be able to have both scenarios based on the operator persona of how customers are handling configuration and separation of concerns.  I am ok with either of thos landing in a > 2.0 release as most customers will be still on KGO or KIC due to the newness of KO if it is going to significantly impact the KO 2.0 release timelines

pmalek:
👍 

Just to make sure: this is all doable and relatively easy to add. One thing we need to make sure we get right is making this compatible with `ControlPlane`'s own watch namespaces setting (introduced in 1.6).

One thought I just had would be to have the `ControlPlane` by default only watch its own namespace, and allow users to set `all` as an option. This would still leave some place for misconfiguration (namespaced KO vs `ControlPlane` watching all namespaces) but OOTB experience would be predictable, more contained and arguably more secure.

pmalek:
Thread started: https://kongstrong.slack.com/archives/C011RQPHDC7/p1753819507528729

pmalek:
https://github.com/Kong/developer.konghq.com/pull/2523 contains the guide for KO 2.0 (how to use the new watch namespaces).

`ControlPlane`'s guide is already available (for the feature from 1.6) at https://developer.konghq.com/operator/konnect/control-plane-watch-namespaces/.

pmalek:
Guide mentioned above has been merged to https://github.com/Kong/developer.konghq.com/tree/release/operator/2.0.0 branch which will be used for KO 2.0 release.

Closing.

PR #1974, opened 2025-08-01
feat(controlplane): check watch namespaces set on ControlPlane with operator flags

pmalek:

During the development of KO we've removed the watch namespaces related code from `ControlPlane` controller.

This PR brings it back and adds checks which ensure that watch namespace set on `ControlPlane` spec are compatible with what's set through operator's flags/envs.

Most notably:

- When `-watch-namespaces` flag is unset, `ControlPlane`'s `spec.watchNamespaces` can be set to anything. Operator watches all namespaces.
- When `-watch-namespaces` flag is set to a list:
  - `ControlPlane`'s `spec.watchNamespaces` can be set to a list which is a subset of that list or
  - `ControlPlane`'s `spec.watchNamespaces` can be set to `own` option and in that case operator has to include CP's namespace in its `-watch-namespaces`

**Which issue this PR fixes**

Part of #1693

**PR Readiness Checklist**:

### DIFF
--- a/controller/controlplane/controller.go
+++ b/controller/controlplane/controller.go
@@ -246,6 +246,9 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	log.Trace(logger, "validating WatchNamespaceGrants exist for the ControlPlane")
 	validatedWatchNamespaces, err := r.validateWatchNamespaceGrants(ctx, cp)
 	if err != nil { //nolint:gocritic
+		// If there was an error validating the WatchNamespaceGrants, we set the condition
+		// to false indicating that the WatchNamespaceGrants are invalid or missing.
+
 		k8sutils.SetCondition(
 			k8sutils.NewConditionWithGeneration(
 				kcfgcontrolplane.ConditionTypeWatchNamespaceGrantValid,
@@ -260,24 +263,21 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		// This will prevent users using the ControlPlane with previous
 		// WatchNamespaces spec.
 		// We do not patch the status here either because that's done below.
-	} else if cp.Spec.WatchNamespaces != nil {
-		k8sutils.SetCondition(
-			k8sutils.NewConditionWithGeneration(
-				kcfgcontrolplane.ConditionTypeWatchNamespaceGrantValid,
-				metav1.ConditionTrue,
-				kcfgcontrolplane.ConditionReasonWatchNamespaceGrantValid,
-				"WatchNamespaceGrant(s) are present and valid",
-				cp.GetGeneration(),
-			),
-			cp,
-		)
 	} else {
+		// If the WatchNamespaceGrants are present and valid, we set the condition to true.
+		// We note that grants are not expected if the watch namespaces are not
+		// specified in the ControlPlane spec.
+
+		msg := "WatchNamespaceGrant(s) are present and valid"
+		if cp.Spec.WatchNamespaces != nil {
+			msg = "WatchNamespaceGrant(s) not required"
+		}
 		k8sutils.SetCondition(
 			k8sutils.NewConditionWithGeneration(
 				kcfgcontrolplane.ConditionTypeWatchNamespaceGrantValid,
 				metav1.ConditionTrue,
 				kcfgcontrolplane.ConditionReasonWatchNamespaceGrantValid,
-				"WatchNamespaceGrant(s) not required",
+				msg,
 				cp.GetGeneration(),
 			),
 			cp,
--- a/controller/controlplane/controller.go
+++ b/controller/controlplane/controller.go
@@ -245,7 +245,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	log.Trace(logger, "validating WatchNamespaceGrants exist for the ControlPlane")
 	validatedWatchNamespaces, err := r.validateWatchNamespaceGrants(ctx, cp)
-	if err != nil { //nolint:gocritic
+	if err != nil {
 		// If there was an error validating the WatchNamespaceGrants, we set the condition
 		// to false indicating that the WatchNamespaceGrants are invalid or missing.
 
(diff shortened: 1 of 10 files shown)


controller/controlplane/controller_reconciler_utils.go
```diff
@@ -120,3 +123,126 @@ func (r *Reconciler) ensureAdminMTLSCertificateSecret(
...
+				watchNamespaces,
+			)
+		}
+
+	case operatorv2alpha1.WatchNamespacesTypeList:
+		if len(watchNamespaces) == 0 {
+			return nil
+		}
+
+		for _, ns := range cp.Spec.WatchNamespaces.List {
```
randmonkey: There is a `lo.Every` that checks if every elements is in another slice.

controller/controlplane/controller_reconciler_utils.go
```diff
@@ -120,3 +123,126 @@ func (r *Reconciler) ensureAdminMTLSCertificateSecret(
...
+	case operatorv2alpha1.WatchNamespacesTypeAll:
+		if len(watchNamespaces) > 0 {
+			return fmt.Errorf(
+				"ControlPlane's watchNamespaces is set to 'All', but operator is only allowed on: %v",
+				watchNamespaces,
+			)
+		}
+
+	case operatorv2alpha1.WatchNamespacesTypeOwn:
+		if len(watchNamespaces) > 0 && !lo.Contains(watchNamespaces, cp.Namespace) {
```
randmonkey, line 227: I think it is unlikely that a `ControlPlane`'s own namespace is not included in operator's watched namespaces but it is reconciled. It is OK to include the unusual case but I think we need to add some comments here and pay more attention to it.

modules/manager/watchnamespaces.go
```diff
@@ -0,0 +1,25 @@
...
+// NewWatchNamespaces parses the watchNamespaces string and returns a slice
+// of trimmed namespace names.
+func NewWatchNamespaces(watchNamespaces string) ([]string, error) {
+	if watchNamespaces == "" {
+		return nil, nil
+	}
+
+	// Split the namespaces by comma and trim whitespace
+	namespaces := make([]string, 0)
+	for ns := range strings.SplitSeq(watchNamespaces, ",") {
```
randmonkey, line 17: :+1: for using `SplitSeq` that can return early when error happens without making the whole string slice.

controller/controlplane/controller_reconciler_utils.go
pmalek: 👍 PTAL 12f070c41e9ca29f8f2cf52f8b8aa83211ffafcb

controller/controlplane/controller_reconciler_utils.go
pmalek: 👍 Added some comments and extended the error message in 1b5721da57cf33581511431bc7b451df2deecddd

PTAL

controller/controlplane/controller.go
```diff
@@ -227,6 +235,55 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+				fmt.Sprintf("WatchNamespaceGrant(s) are missing or invalid for the ControlPlane: %v", err),
+				cp.GetGeneration(),
+			),
+			cp,
+		)
+		// We do not return here as we want to proceed with reconciling the Deployment.
+		// This will prevent users using the ControlPlane with previous
+		// WatchNamespaces spec.
+		// We do not patch the status here either because that's done below.
+	} else if cp.Spec.WatchNamespaces != nil {
```
randmonkey: The `if - else if - else` is a bit confusing. I think is may be clearer to restructure like this and add comments to talk about which situation is each branch processing
```go
if err != nil {
// Set conditions for the case validation of watched namespaces failed
} else {
// Set conditions for the case validation of watched namespaces passed
    if cp.Spec.WatchNamespaces != nil {
    // Set condition when CP has watched namespaces
    } else {
    // Set condition when CP does not have watchNamespaces, indicating that the CP watches all the namespaces
    }

}
```

controller/controlplane/controller_reconciler_utils.go
```diff
@@ -120,3 +123,127 @@ func (r *Reconciler) ensureAdminMTLSCertificateSecret(
...
+	cl client.Client,
+	cp *ControlPlane,
+	ns string,
+) error {
+	var grants operatorv1alpha1.WatchNamespaceGrantList
+	if err := cl.List(ctx, &grants, client.InNamespace(ns)); err != nil {
+		return fmt.Errorf("failed listing WatchNamespaceGrants in namespace %s: %w", ns, err)
+	}
+	for _, grant := range grants.Items {
+		if !watchNamespaceGrantContainsControlPlaneFrom(grant, cp) {
```
randmonkey: Does this mean "If one item contains the grant to watch the namespace, return nil?" If so I think structure like this should be clearer

```suggestion
		if watchNamespaceGrantContainsControlPlaneFrom(grant, cp) {
		// return nil if there is one grant allows the CP to watch the namespace.
		    return nil
		}
```

controller/controlplane/controller.go
pmalek: WDYT about fe3a7e3b86bedf70df316d1e899a896ca38b57f0 ?

randmonkey (dismissed):

controller/controlplane/controller.go
```diff
@@ -227,6 +235,55 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	if err := validateWatchNamespaces(cp, r.WatchNamespaces); err != nil {
+		// TODO: Set status condition on the ControlPlane.
+		// https://github.com/Kong/kong-operator/issues/1975
+		log.Debug(logger, "watch namespaces validation failed", "error", err)
+		return ctrl.Result{}, err
+	}
+
+	log.Trace(logger, "validating WatchNamespaceGrants exist for the ControlPlane")
+	validatedWatchNamespaces, err := r.validateWatchNamespaceGrants(ctx, cp)
+	if err != nil { //nolint:gocritic
```
randmonkey:
```suggestion
	if err != nil {
```

randmonkey (approved):

PR #1976, opened 2025-08-04
fix(controlplane): Fix the invalid label causing ingress controller fails to store license to secret

randmonkey:
Fix the issue where control plane (ingress-controller package) fails to store license to local `Secret`. Migrated from https://github.com/Kong/kubernetes-ingress-controller/pull/7648.
**Which issue this PR fixes**

Fixes #

How to write changelog here? I think we should put it in the KO changelog but not the embedded `CHANGELOG.md` file in the `ingress-controller` package. That one should be only for archive.
**PR Readiness Checklist**:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,12 @@
   The default value is `""` which makes the operator watch all namespaces.
   [#1958](https://github.com/Kong/kong-operator/pull/1958)
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


pmalek (approved):

pmalek:
> How to write changelog here? I think we should put it in the KO changelog but not the embedded `CHANGELOG.md` file in the `ingress-controller` package. That one should be only for archive.

That's exactly what I'd expect here 👍

Issue #1730, opened 2025-06-12
Kong operator: implement KonnectExtension handling for `ControlPlane` `v2alpha1`


pmalek:
### Problem statement
 
#1358 tracks the changes in `ControlPlane` v2 API (and `GatewayConfiguration` API) so that ingress controller functionality is embedded within the operator.

This issue tracks the effort of making sure that `KonnectExtension` works with that API.

More specifically we need to change the handling of Konnect related configuration in "KIC" via envs to set the appropriate configuration fields when spinning up an in process "KIC" instance.

### Acceptance criteria

- [ ] As a User I can use `KonnectExtension` API with `ControlPlane` v2 (`v2alpha1` or newer)

pmalek:
Duplicate of #1361

PR #1978, opened 2025-08-04
feat(extensions): refactor Konnect extension processing for ControlPlane and DataPlane

### LINKED ISSUE #1361: Kong operator: implement KonnectExtension handling for `ControlPlane` `v2alpha1`
### Problem statement
 
#1358 tracks the changes in `ControlPlane` v2 API (and `GatewayConfiguration` API) so that ingress controller functionality is embedded within the operator.

This issue tracks the effort of making sure that `KonnectExtension` works with that API.

More specifically we need to change the handling of Konnect related configuration in "KIC" via envs to set the appropriate configuration fields when spinning up an in process "KIC" instance.

### Acceptance criteria

- [ ] As a User I can use `KonnectExtension` API with `ControlPlane` v2 (`v2alpha1` or newer)
- [ ] ControlPlane controller handles KonnectExtension attached to ControlPlane and configures KIC instances following the extension's spec

alacuku:
This PR introduces a standardized approach to processing extensions for ControlPlane and DataPlane resources by implementing the ExtensionProcessor interface. Key changes include:

* New Interface: Added `ExtensionProcessor` to unify extension processing logic.
* Processor Implementations: Introduced `ControlPlaneKonnectExtensionProcessor` and `DataPlaneKonnectExtensionProcessor` for resource-specific logic.
* Refactored Logic: Updated `ApplyExtensions` to leverage the new interface.
* Controller Updates: Modified `ControlPlane` and `DataPlane` controllers to use the new processors.
* Testing: Added comprehensive unit tests for the new processors to ensure reliability and coverage.

The processor-based design enhances flexibility and maintainability by encapsulating resource-specific logic within dedicated processors. This approach:

* Keeps the core `ApplyExtensions` function reusable.
* Simplifies customization for specific resources like `ControlPlane` and `DataPlane`.
* Promotes scalability, allowing new resource types or extension behaviors to be added with minimal disruption to existing code.

**Which issue this PR fixes**

Fixes #1361 

There is a commit that reverts recent changes to the `ControlPlane ID` handling logic. The previous implementation
is restored to address issues caused by the updated logic, which disrupted functionality. This ensures compatibility and correct behavior with the original ID usage in the` ControlPlane`.

**PR Readiness Checklist**:


controller/controlplane/controller.go
```diff
@@ -117,7 +120,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
-	mgrID, err := manager.NewID(builControlPlaneInstanceID(cp))
+	mgrID, err := manager.NewID(string(cp.GetUID()))
```
randmonkey, line 126: Is there any reason to change the `mgrID` to include only UID of `ControlPlane` but not namespace and name?

controller/controlplane/controller.go
```diff
@@ -31,6 +31,9 @@ import (
 	konnectv1alpha2 "github.com/kong/kubernetes-configuration/v2/api/konnect/v1alpha2"
 
 	ctrlconsts "github.com/kong/kong-operator/controller/consts"
+	"github.com/kong/kong-operator/controller/pkg/extensions"
+	extensionserrors "github.com/kong/kong-operator/controller/pkg/extensions/errors"
+	extensionsKonnect "github.com/kong/kong-operator/controller/pkg/extensions/konnect"
```
randmonkey: Should we use `extensionkonnect` instead of camelCase here for renaming packages?

controller/pkg/extensions/konnect/controlplane.go
```diff
@@ -28,7 +60,132 @@ func ApplyControlPlaneKonnectExtension(ctx context.Context, cl client.Client, co
...
+		return true
+	})
+	return fmt.Sprintf("https://%s.kic.api.%s", addressSlice[0], strings.Join(addressSlice[1:], "."))
+}
+
+// getTLSClientCertAndKey retrieves the TLS client certificate and key from the specified secret.
+// Returns the certificate and key as strings, or an error if the secret is not found or if the data is missing.
+// The secret must contain "tls.crt" and "tls.key" keys.
+// This function is used to extract the TLS client cert and key for Konnect configuration.
+func getTLSClientCertAndKey(ctx context.Context, cl client.Client, secretName, secretNamespace string) (tlsClientCert, tlsClientKey string, err error) {
```
randmonkey, line 143: Looks like we can move this function to `pkg/utils/kubernetes` package to reuse it for extracting certificates from secrets -- or do we already have the implementation elsewhere?

controller/pkg/extensions/konnect/controlplane.go
```diff
@@ -28,7 +60,132 @@ func ApplyControlPlaneKonnectExtension(ctx context.Context, cl client.Client, co
...
+	}
+}
+
+// buildKonnectAddress builds the Konnect address out of the control plane endpoint.
+// input: "https://7b46471d3b.us.tp0.konghq.tech:443"
+// output: "https://us.kic.api.konghq.tech"
+func buildKonnectAddress(endpoint string) string {
+	portlessEndpoint := strings.TrimSuffix(endpoint, ":443")
+	addressSlice := strings.Split(portlessEndpoint, ".")
+	addressSlice = lo.Filter(addressSlice, func(_ string, i int) bool {
```
randmonkey, line 129: How about `lo.DropByIndex(addressSlice,0,2)`? 
BTW, The `addressSlice[0]` actually includes `https://` part but it is dropped in this step.

controller/pkg/extensions/konnect/controlplane.go
```diff
@@ -28,7 +60,132 @@ func ApplyControlPlaneKonnectExtension(ctx context.Context, cl client.Client, co
...
+		// default never happens as the validation is at the CRD level
+		panic(fmt.Sprintf("unsupported Konnect cluster type: %s", konnectExtension.Status.Konnect.ClusterType))
+	}
+}
+
+// buildKonnectAddress builds the Konnect address out of the control plane endpoint.
+// input: "https://7b46471d3b.us.tp0.konghq.tech:443"
+// output: "https://us.kic.api.konghq.tech"
+func buildKonnectAddress(endpoint string) string {
+	portlessEndpoint := strings.TrimSuffix(endpoint, ":443")
```
randmonkey, line 127: Is there guarantee that the CP endpoint always end with `:443`? If  there is such guarantee, please use comments to tell us.

controller/pkg/extensions/konnect/controlplane.go
```diff
@@ -28,7 +60,132 @@ func ApplyControlPlaneKonnectExtension(ctx context.Context, cl client.Client, co
...
+	addressSlice := strings.Split(portlessEndpoint, ".")
+	addressSlice = lo.Filter(addressSlice, func(_ string, i int) bool {
+		// remove 7b46471d3b and tp0
+		if i == 0 || i == 2 {
+			return false
+		}
+		return true
+	})
+	return fmt.Sprintf("https://%s.kic.api.%s", addressSlice[0], strings.Join(addressSlice[1:], "."))
```
randmonkey, line 136: Looks like we need some check on length of `addressSlice` (or check on the input address itself) to prevent panics when exceptional address is in the `status`.

controller/pkg/extensions/konnect/controlplane.go
```diff
@@ -28,7 +60,132 @@ func ApplyControlPlaneKonnectExtension(ctx context.Context, cl client.Client, co
...
+// If the FeatureGates field is nil, it initializes it with the FillIDs feature gate.
+// If the FillIDs feature gate is already present, it ensures its value is set to enabled.
+// If the FillIDs feature gate is not present, it appends it to the FeatureGates slice.
+func applyFeatureGatesToControlPlane(cp *gwtypes.ControlPlane) {
+	fillIDsFeatureGate := gwtypes.ControlPlaneFeatureGate{
+		Name:  managercfg.FillIDsFeature,
+		State: gwtypes.FeatureGateStateEnabled,
+	}
+
+	if cp.Spec.FeatureGates == nil {
```
randmonkey: We do not need to distinguish the `nil` slices. You can also traverse a `nil` slice and append elements to it.

controller/controlplane/controller.go
alacuku: Please check the `Special notes for your reviewer` section in the PR body.

controller/controlplane/controller.go
alacuku: Nice catch!

controller/pkg/extensions/konnect/controlplane.go
alacuku: Not sure if we use the same format for the certificates secrets elsewhere in the code. I encapsulated the needed logic in a function for better  testability.

controller/pkg/extensions/konnect/controlplane.go
alacuku: This is existing code that I moved around. I think it's out of scope for this PR to modify existing code.

controller/pkg/extensions/konnect/controlplane.go
alacuku: As mentioned earlier, this code has been in place for some time, is critical, and has proven to work reliably in production. Given its importance and stability, I would prefer not to modify it unless there is a reason or evidence of an issue.

controller/pkg/extensions/konnect/controlplane.go
alacuku: Same as above.

controller/controlplane/controller.go
mlavacca, line 126: Can we add a comment in the code about why we are using the UID only to be sure we do not change this again in the future?

randmonkey: OK. But this is better to be submitted in a standalone PR. If the revert can be done by reverting another PR it is even better.

controller/pkg/extensions/konnect/controlplane.go
randmonkey: I mean getting the TLS cert and key from the secret may be a common used function so we can extract it to some common packages. But this is not necessary in the PR and maybe we can do it if there is another place requires it.

controller/pkg/extensions/konnect/controlplane.go
randmonkey: OK, found where the code come from. Since we have been using the code without issues (merged since March 05), I agree that we should not modify it now.

controller/pkg/extensions/konnect/controlplane.go
```diff
@@ -28,7 +60,126 @@ func ApplyControlPlaneKonnectExtension(ctx context.Context, cl client.Client, co
...
+// GetKonnectConfig returns the KonnectConfig from the KonnectExtensionConfig.
+func (p *ControlPlaneKonnectExtensionProcessor) GetKonnectConfig() *managercfg.KonnectConfig {
+	if p.KonnectExtensionConfig == nil {
+		return nil
+	}
+	return p.KonnectExtensionConfig.KonnectConfig
+}
+
+// kicInKonnectDefaults generates the KonnectExtensionConfig for a KIC instance based on the KonnectExtension resource.
+func kicInKonnectDefaults(ctx context.Context, cl client.Client, konnectExtension *konnectv1alpha2.KonnectExtension) (*KonnectExtensionConfig, error) {
```
randmonkey, line 85: I know this function is also moved from elsewhere but I think the function name `kicInKonnectConfig()` or `kicInKonnectConfigFromExtension` should describe the usage of the function better. WDYT?

controller/pkg/extensions/konnect/controlplane.go
```diff
@@ -28,7 +60,126 @@ func ApplyControlPlaneKonnectExtension(ctx context.Context, cl client.Client, co
...
+	}
+
+	return string(tlsCert), string(tlsKey), nil
+}
+
+// applyFeatureGatesToControlPlane applies the FillIDs feature gate to the ControlPlane's FeatureGates.
+// If the FeatureGates field is nil, it initializes it with the FillIDs feature gate.
+// If the FillIDs feature gate is already present, it ensures its value is set to enabled.
+// If the FillIDs feature gate is not present, it appends it to the FeatureGates slice.
+func applyFeatureGatesToControlPlane(cp *gwtypes.ControlPlane) {
```
randmonkey, line 163: Looks like the function only adds the `fillIDsFeatureGate` to the CP but not other feature gates. Could the name `applyFillIDsFeatureGatesToControlPlane` describe it better?

randmonkey (commented):
Generally looks good bit some minor comments.

controller/pkg/extensions/processor.go
```diff
@@ -0,0 +1,13 @@
+package extensions
+
+import (
+	"context"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// ExtensionProcessor defines how to process extensions for a specific resource type.
+type ExtensionProcessor interface {
```
pmalek: Since this exists in the context of `extensions` package the name `ExtensionProcessor` stutters (This typically is detected by the linter, not sure why it wasn't).

I suggest to call this simply `Processor`, this way at the call site it's `extensions.Processor`. WDYT?

(not suggesting a change here through GH UI as that will require a change in other files as well).

controller/pkg/extensions/konnect/controlplane.go
```diff
@@ -28,7 +60,126 @@ func ApplyControlPlaneKonnectExtension(ctx context.Context, cl client.Client, co
 		return false, nil
 	}
 
-	// TODO: implement KonnectExtension for ControlPlane v2alpha1: https://github.com/kong/kong-operator/issues/1730
+	config, err := kicInKonnectDefaults(ctx, cl, konnectExtension)
+	if err != nil {
+		return false, fmt.Errorf("failed to generate configuration from KonnectExtension %s/%s for ControlPlane %s/%s: %w",
+			konnectExtension.Namespace, konnectExtension.Name, cp.Namespace, cp.Name, err)
```
pmalek: Nit:

```suggestion
		return false, fmt.Errorf("failed to generate configuration from KonnectExtension %s for ControlPlane %s: %w",
			client.ObjectKeyFromObject(konnectExtension), client.ObjectKeyFromObject(cp), err,
		)
```

controller/pkg/extensions/konnect/controlplane_test.go
```diff
@@ -0,0 +1,729 @@
...
+		})
+	}
+}
+
+func TestControlPlaneKonnectExtensionProcessor_Process(t *testing.T) {
+	// Create a scheme for the fake client.
+	s := scheme.Scheme
+	require.NoError(t, konnectv1alpha1.AddToScheme(s))
+	require.NoError(t, konnectv1alpha2.AddToScheme(s))
+	require.NoError(t, corev1.AddToScheme(s))
```
pmalek: For this you can just use `scheme.Get()` from https://github.com/Kong/kong-operator/blob/0356fc46538454b8d856662de5f979d1b92d6abd/modules/manager/scheme/scheme.go#L21-L43 which will always contain all the necessary schemas (as otherwise all the other code will fail).

This will also prevent the need for updating code here in case the required API versions change.

pmalek (commented):
This looks good, just a couple of minor nits.

Additionally I think we have to either update the PR description or add the integration tests update (as described in the #1730 linked at https://github.com/Kong/kong-operator/blob/b264059b1c854b85f96ad8755b8c16cbf2a459ba/test/integration/konnect_extension_test.go#L68). That issues has been marked as a duplicate of the #1361 which is going to be closed by this PR.

controller/pkg/extensions/processor.go
mlavacca: +1. I think `Processor` under the `extensions` package is best.

controller/pkg/extensions/konnect/controlplane.go
```diff
@@ -2,20 +2,52 @@ package konnect
...
+	KonnectConfig *managercfg.KonnectConfig
+}
+
+// ControlPlaneKonnectExtensionProcessor processes Konnect extensions for ControlPlane resources.
+type ControlPlaneKonnectExtensionProcessor struct {
+	KonnectExtensionConfig *KonnectExtensionConfig
+}
+
+// Compile-time check to ensure ControlPlaneKonnectExtensionProcessor implements the extensions.ExtensionProcessor interface.
+var _ extensions.ExtensionProcessor = (*ControlPlaneKonnectExtensionProcessor)(nil)
```
mlavacca: 👍

controller/pkg/extensions/konnect/controlplane.go
```diff
@@ -2,20 +2,52 @@ package konnect
...
+
+// Process extracts the KonnectExtension from the ControlPlane and generates the KonnectExtensionConfig.
+// It returns true if a KonnectExtension was found and processed, false otherwise.
+// If an error occurs during processing, it returns false and the error.
+// It captures configuration from the KonnectExtension resource and prepares it for use in KIC instances.
+// This implementation of the ExtensionProcessor interface has side effects, as it modifies the KonnectExtensionConfig field of the processor.
+func (p *ControlPlaneKonnectExtensionProcessor) Process(ctx context.Context, cl client.Client, object client.Object) (bool, error) {
```
mlavacca, line 40: I believe we can use the `ExtendableT` interface rather than `client.Object` here, to make it more robust.

controller/pkg/extensions/konnect/controlplane.go
mlavacca: Being under the `extension/konnect` package, I think `kicInKonnectConfigFromExtension` is redundant here.

controller/pkg/extensions/konnect/controlplane.go
mlavacca: I am on the fence here. `applyFeatureGatesToControlPlane` is more future-proof in case we need to set additional Features Gates, but OTOH the godoc is focused on the `FillIDs` FG. We could instead change the comment to something not `FIllIDs`-specific and keep this name I think.

controller/pkg/extensions/konnect/controlplane.go
```diff
@@ -28,7 +60,126 @@ func ApplyControlPlaneKonnectExtension(ctx context.Context, cl client.Client, co
...
+	found := false
+	for i, fg := range cp.Spec.FeatureGates {
+		if fg.Name == fillIDsFeatureGate.Name {
+			found = true
+			if fg.State != gwtypes.FeatureGateStateEnabled {
+				cp.Spec.FeatureGates[i].State = gwtypes.FeatureGateStateEnabled
+			}
+			break
+		}
+	}
```
mlavacca, line 178: This is an implementation detail, but `lo.FindKeyBy` could help here to make this section clearer. Up to you, though.

mlavacca (commented):
This PR looks mostly good to me. Something I'd love to see, though, is the addition of an integration test for KIC in Konnect, I don't think we have any. Ideally, this would be similar to what we already have in https://github.com/Kong/kong-operator/blob/kcl%2Fkonnectextensions-controlplane/test/integration/konnect_extension_test.go. 
However, I don't think it necessarily need to be part of this PR, you can open a follow-up isse related to it, if you prefer.

mlavacca:
> This looks good, just a couple of minor nits.
> 
> Additionally I think we have to either update the PR description or add the integration tests update (as described in the #1730 linked at
> 
> https://github.com/Kong/kong-operator/blob/b264059b1c854b85f96ad8755b8c16cbf2a459ba/test/integration/konnect_extension_test.go#L68
> ). That issues has been marked as a duplicate of the #1361 which is going to be closed by this PR.

The issue #1730, superseded by #1361 has nothing in common with the change in this file. The reason why they were broken at a certain point in time is unrelated to the `ControlPlane` API, as they have no interaction at all with that API. We should create an issue to re-enable them and figure out the root cause of why they stopped working. As I wrote in a comment here, we should create a KIC-in-Konnect integration test (it's ok to do it in a follow-up), but all the tests in this file are related to the hybrid Konnect CPs.

pmalek:
> > This looks good, just a couple of minor nits.
> > Additionally I think we have to either update the PR description or add the integration tests update (as described in the #1730 linked at
> > https://github.com/Kong/kong-operator/blob/b264059b1c854b85f96ad8755b8c16cbf2a459ba/test/integration/konnect_extension_test.go#L68
> > 
> > ). That issues has been marked as a duplicate of the #1361 which is going to be closed by this PR.
> 
> The issue #1730, superseded by #1361 has nothing in common with the change in this PR. The reason why they were broken at a certain point in time is unrelated to the `ControlPlane` API, as they have no interaction at all with that API. We should create an issue to re-enable them and figure out the root cause of why they stopped working. As I wrote in a comment here, we should create a KIC-in-Konnect integration test (it's ok to do it in a follow-up), but all the tests in this file are related to the hybrid Konnect CPs.

Ah, you're right. I jumped the gun on adding the association to that file not checking thoroughly which tests it contains.

I agree with the statement

>  we should create a KIC-in-Konnect integration test (it's ok to do it in a follow-up)

as long as either it's done here or added an issue as a follow up (which is handled for 2.0 release).

Thanks for staying vigilant 🙇

controller/pkg/extensions/konnect/controlplane.go
alacuku: The Processors is used in `ApplyExtensions` function that already requires a parameterized type that satisfies the `ExtanableT` intreface. Furthermore, the `ExtandableT` is an interface with type constraints and cannot be used as a type for function parameters.

controller/pkg/extensions/konnect/controlplane.go
alacuku: Updated the go doc to reflect the generic nature of the function.

alacuku:
> This PR looks mostly good to me. Something I'd love to see, though, is the addition of an integration test for KIC in Konnect, I don't think we have any. Ideally, this would be similar to what we already have in https://github.com/Kong/kong-operator/blob/kcl%2Fkonnectextensions-controlplane/test/integration/konnect_extension_test.go. However, I don't think it necessarily need to be part of this PR, you can open a follow-up isse related to it, if you prefer.

Tracking issue: https://github.com/Kong/kong-operator/issues/1991

mlavacca:
> > This looks good, just a couple of minor nits.
> > Additionally I think we have to either update the PR description or add the integration tests update (as described in the #1730 linked at
> > https://github.com/Kong/kong-operator/blob/b264059b1c854b85f96ad8755b8c16cbf2a459ba/test/integration/konnect_extension_test.go#L68
> > 
> > ). That issues has been marked as a duplicate of the #1361 which is going to be closed by this PR.
> 
> The issue #1730, superseded by #1361 has nothing in common with the change in this file. The reason why they were broken at a certain point in time is unrelated to the `ControlPlane` API, as they have no interaction at all with that API. We should create an issue to re-enable them and figure out the root cause of why they stopped working. As I wrote in a comment here, we should create a KIC-in-Konnect integration test (it's ok to do it in a follow-up), but all the tests in this file are related to the hybrid Konnect CPs.

Created #1992 and #1993

mlavacca (approved):

randmonkey (approved):

PR #1982, opened 2025-08-05
feat(controlplane): Support `ControlPlane` level label selectors for `Secret`s and `ConfigMap`s

### LINKED ISSUE #1863: Support configuring label selector for watched secrets and configmaps per control plane
## Problem Statement

For the feature parity with KIC, we need to implement the equivalent  of `--configmap-label-selector` and `--secret-label-selector` CLI flags In KIC including both API change and implementation in KO.

## Proposed Solution

- Add `spec.objetFilters` in `ControlPlane` CRD including the fields:
  - `secrets.labelSelector` for `Secret`s
  - `configMaps.labelSelector` for `ConfigMap`s
- KO's `ControlPlane` reconciler supports configuring label selectors for each spawned controller based on the spec in `ControlPlane` 
- When KO configures label selectors of secrets and configmaps and `ControlPlane` also sets the label selecors, the controllers for `ControlPlane` should only watch `Secret`s and `ConfigMap`s with BOTH labels set to `true`.

## Additional Information

...

## Acceptance Criteria
- [x] Can specify the label selectors to filter watched `Secret`s and `ConfigMap`s of each controller run for a `ControlPlane` based on the spec in the CRD.
- [x] If both KO level and control plane level label selectors are configured, the manager for the control plane only reconciles the `Secret`s and `ConfigMap`s with both labels set to `true`.

randmonkey:

Support applying label selectors for `Secret`s and `ConfigMap`s for ControlPlanes based on their `spec.objectFilters` and also operator level selector. 

This adds support in CP controllers to set multiple matching labels and also sets label selectors on the manager level. 

**Which issue this PR fixes**

Fixes #1863

**PR Readiness Checklist**:

NOTE: This PR chanaged the `SecretLabelSelector` and `ConfigMapLabelSelector` in ingress-controller's `Config` to `map[string]string` to support multiple selectors. Also it deleted the related CLI flags in `ingress-controller` package.


ingress-controller/internal/controllers/configuration/secret_controller.go
```diff
@@ -42,8 +43,8 @@ type CoreV1SecretReconciler struct {
 	DataplaneClient  controllers.DataPlane
 	CacheSyncTimeout time.Duration
 
-	ReferenceIndexers ctrlref.CacheIndexers
-	LabelSelector     string
+	ReferenceIndexers        ctrlref.CacheIndexers
+	LabelSelectorMatchLabels map[string]string
```
randmonkey: REVIEW: Should we remove the label selector in controllers for the dedicated types as we have the manager level label selectors?

controller/controlplane/controller.go
```diff
@@ -499,10 +499,27 @@ func (r *Reconciler) constructControlPlaneManagerConfigOptions(
 	}
 
 	if r.SecretLabelSelector != "" {
-		cfgOpts = append(cfgOpts, WithSecretLabelSelector(r.SecretLabelSelector))
+		cfgOpts = append(cfgOpts, WithSecretLabelSelectorMatchLabel(r.SecretLabelSelector, "true"))
 	}
+	if cp.Spec.ObjectFilters != nil && cp.Spec.ObjectFilters.Secrets != nil {
+		for k, v := range cp.Spec.ObjectFilters.Secrets.MatchLabels {
+			if k == r.SecretLabelSelector {
```
randmonkey, line 520: REVIEW: How should we deal with conflicts between CP level and KO level label selectors?

- Return error in the reconciliation stage
- Ignore one of them and tell the behavior explicitly in docs
- Use admission webhooks (Not recommended, because we do not have a validating webhook for `ControlPlane` now)

ingress-controller/internal/manager/setup.go
```diff
@@ -57,6 +59,37 @@ func setupManagerOptions(ctx context.Context, logger logr.Logger, c *managercfg.
 
 	cacheOptions := cache.Options{
 		SyncPeriod: &c.SyncPeriod,
+		ByObject:   map[client.Object]cache.ByObject{},
+	}
+
+	// Add label selector for Secrets and ConfigMaps on the manager level.
+	if c.SecretLabelSelector != nil {
+		// TODO: move the code to build cache's ByObject options to common utils packages.
```
randmonkey: Did not do this in the PR because moving it to common packages will make packages in `ingress-controller` to reference packages out of `ingress-controller` but in the whole project. Not sure if it will cause dependency cycles and are we ready to break the "internal wall".

controller/controlplane/controller.go
pmalek: Use a status condition and set its status to `False` to signal this.

ingress-controller/internal/controllers/configuration/secret_controller.go
pmalek: I wonder if watch predicates actually work with label selectors set at the manager level 🤔 

If they do I think we can remove them here (ideally making a note that we "inherit" the label selectors from the manager).

randmonkey: Removed them.

controller/controlplane/controller.go
randmonkey: Added a validation step to check and set conditions on conflict.
Also created an issue for adding dedicated condition types/reasons as it requires changes on `kubernetes-configuration` package: https://github.com/Kong/kong-operator/issues/1999.

controller/controlplane/controller.go
```diff
@@ -499,10 +513,27 @@ func (r *Reconciler) constructControlPlaneManagerConfigOptions(
 	}
 
 	if r.SecretLabelSelector != "" {
-		cfgOpts = append(cfgOpts, WithSecretLabelSelector(r.SecretLabelSelector))
+		cfgOpts = append(cfgOpts, WithSecretLabelSelectorMatchLabel(r.SecretLabelSelector, "true"))
+	}
+	if cp.Spec.ObjectFilters != nil && cp.Spec.ObjectFilters.Secrets != nil {
+		for k, v := range cp.Spec.ObjectFilters.Secrets.MatchLabels {
+			if k == r.SecretLabelSelector {
+				return nil, fmt.Errorf("ControlPlane's secret label selector confictts with the operator's secret label selector: %s", k)
```
pmalek:
```suggestion
				return nil, fmt.Errorf("ControlPlane's secret label selector conflicts with operator's secret label selector: %s", k)
```

controller/controlplane/controller.go
```diff
@@ -499,10 +513,27 @@ func (r *Reconciler) constructControlPlaneManagerConfigOptions(
...
 	}
+
 	if r.ConfigMapLabelSelector != "" {
-		cfgOpts = append(cfgOpts, WithConfigMapLabelSelector(r.ConfigMapLabelSelector))
+		cfgOpts = append(cfgOpts, WithConfigMapLabelSelectorMatchLabel(r.ConfigMapLabelSelector, "true"))
+	}
+	if cp.Spec.ObjectFilters != nil && cp.Spec.ObjectFilters.ConfigMaps != nil {
+		for k, v := range cp.Spec.ObjectFilters.ConfigMaps.MatchLabels {
+			if k == r.ConfigMapLabelSelector {
+				return nil, fmt.Errorf("ControlPlane's configMap label selector confictts with the operator's configMap label selector: %s", k)
```
pmalek:
```suggestion
				return nil, fmt.Errorf("ControlPlane's configMap label selector conflicts with operator's configMap label selector: %s", k)
```

controller/controlplane/controller.go
```diff
@@ -655,3 +686,31 @@ func (r *Reconciler) handleScheduleInstanceOutcome(
 	}
 	return ctrl.Result{}, err
 }
+
+// validateControlPlaneOptions checks if the ControlPlane's options is valid with the operator specific configuration.
```
pmalek:
```suggestion
// validateControlPlaneOptions checks if the ControlPlane's options are valid with the operator specific configuration.
```

controller/controlplane/controller.go
```diff
@@ -655,3 +686,31 @@ func (r *Reconciler) handleScheduleInstanceOutcome(
...
+// It returns a boolean as the validation results and a string as the error message when the validation fails.
```
pmalek:
```suggestion
// It returns a boolean as the validation result and a string as the error message when the validation fails.
```

controller/controlplane/controller_test.go
```diff
@@ -221,6 +221,9 @@ func TestReconciler_Reconcile(t *testing.T) {
 				_, err := reconciler.Reconcile(ctx, controlplaneReq)
 				require.NoError(t, err)
 			},
+			// TODO: add a test case for invalid ControlPlane options after the error
```
pmalek, line 225: Can we add a tracking issue for this?

ingress-controller/internal/manager/setup.go
pmalek: It should cause any import cycles if there's nothing from that package reaching back.

I _think_ we could put this code under `ingress-controller/` dir for now to not make a precedent for importing operator code from within the `ingress-controller`.

randmonkey: Extracted to a dedicated function in this file.

controller/controlplane/controller_test.go
randmonkey: https://github.com/Kong/kong-operator/issues/2024

ingress-controller/internal/manager/setup.go
```diff
@@ -533,3 +549,18 @@ func setupKonnectConfigSynchronizerWithMgr(
...
+// buildLabelSelector creates a `label.Selector` from specified match tabels
+// to build the cache options used in the manager's `ByObject` to filter objects (Secrets, ConfigMaps) by labels.
+// TODO: move the code to build cache's ByObject options to common utils packages.
+func buildLabelSelector(matchLabels map[string]string) labels.Selector {
+	s := labels.NewSelector()
+	for k, v := range matchLabels {
+		req, err := labels.NewRequirement(k, selection.Equals, []string{v})
+		if err != nil {
+			continue
+		}
```
pmalek: I'm not sure if silently ignoring this error is what we want. We're working here on user provided values and these are validated in https://github.com/kubernetes/apimachinery/blob/173776a0582da70432e19d16eca025c451600ae8/pkg/labels/selector.go#L209 via https://github.com/kubernetes/apimachinery/blob/v0.33.3/pkg/labels/selector.go#L923-L928.

WDYT about adding error propagation here?

randmonkey: Added error logs here.

ingress-controller/internal/manager/setup.go
```diff
@@ -57,6 +59,28 @@ func setupManagerOptions(ctx context.Context, logger logr.Logger, c *managercfg.
...
 		SyncPeriod: &c.SyncPeriod,
+		ByObject:   map[client.Object]cache.ByObject{},
+	}
+
+	// Add label selector for Secrets and ConfigMaps on the manager level.
+	if len(c.SecretLabelSelector) > 0 {
+		labelSelector, err := buildLabelSelector(c.SecretLabelSelector)
+		if err != nil {
+			logger.Error(err, "error happened in building label selectors for secrets")
+		}
```
pmalek: We shouldn't add the `labelSelector` if `err != nil` right? 

Same below for configMaps

randmonkey: Actually I think we should return error on the failures. Updated in https://github.com/Kong/kong-operator/pull/1982/commits/a9a50492db6f24edbfdc5aefa47bf0be4f33547d

pmalek (approved):

PR #2004, opened 2025-08-06
chore: rename env var prefix to KONG_OPERATOR

### LINKED ISSUE #1941: Change prefix of envs to specify CLI flags
## Problem Statement

Currently KO is using `GATEWAY_OPERATOR_` prefixed environment to specify CLI flags. As we are renaming the product to Kong operator, we should use a new prefix like `KONG_OPERATOR`.

## Proposed Solution

- Use a new environment prefix like `KONG_OPERATOR_` to parse CLI flags
- Mark the `GATEWAY_OPERATOR_` prefix as deprecated
- When there are no cli flags configured by `KONG_OPRATOR_` prefixed env, parse from `GATEWAY_OPERATOR_` prefixed envs
- Update charts to use `KONG_OPREATOR_` prefix 

## Additional Information

Should we also remove the support of `GATEWAY_OPERATOR_` prefix?
## Acceptance Criteria
- [ ] A new env prefix like `KONG_OPREATOR_` is used instead of `GATEWAY_OPERATOR_` to parse CLI flags
- [ ] `GATEWAY_OPERATOR_` prefixed envs are still supported but deprecated
- [ ] Helm charts `kong-operator` uses new prefix to generate envs from variables and `.Values.env` instead of `GATEWAY_OPERATOR`

fgiudici:

**Which issue this PR fixes**

Fixes #1941 

**PR Readiness Checklist**:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,9 @@
   For migration guidance from these resources to Gateway API, please refer to the
   [migration documentation](https://developer.konghq.com/kubernetes-ingress-controller/migrate/ingress-to-gateway/).
   [#1971](https://github.com/Kong/kong-operator/pull/1971)
+- Change env vars prefix from `GATEWAY_OPERATOR_` to `KONG_OPERATOR`.
+  `GATEWAY_OPERATOR_` prefixed env vars are still accepted but reported as deprecated.
+  [#2004](https://github.com/Kong/kong-operator/pull/2004)
 
 ### Added
 
--- a/modules/cli/cli.go
+++ b/modules/cli/cli.go
@@ -119,29 +119,42 @@ type flagsForFurtherEvaluation struct {
 }
 
 const (
-	envVarFlagPrefix = "GATEWAY_OPERATOR_"
+	envDeprecatedVarFlagPrefix = "GATEWAY_OPERATOR_"
+	envVarFlagPrefix           = "KONG_OPERATOR_"
 )
+
+// setFlagFromEnvVar looks for the env var name corresponding to `f`: if the env val is found its value is set
+// to the flag. In case of error it panics.
+func setFlagFromEnvVar(f *flag.Flag) {
+	envVarFlagName := strings.ToUpper(strings.ReplaceAll(f.Name, "-", "_"))
+
+	deprecatedEnvKey := envDeprecatedVarFlagPrefix + envVarFlagName
+	if envValue, envSet := os.LookupEnv(deprecatedEnvKey); envSet {
+		fmt.Printf("WARN: %q env variable is deprecated, please use %q instead\n", deprecatedEnvKey, deprecatedEnvKey)
+		if err := f.Value.Set(envValue); err != nil {
+			panic(fmt.Errorf("environment binding failed for variable %s: %w", deprecatedEnvKey, err))
+		}
+	}
+
+	envKey := envVarFlagPrefix + envVarFlagName
+	if envValue, envSet := os.LookupEnv(envKey); envSet {
+		if err := f.Value.Set(envValue); err != nil {
+			panic(fmt.Errorf("environment binding failed for variable %s: %w", envKey, err))
+		}
+	}
+}
 
 // bindEnvVarsToFlags, for each flag defined on `cmd` (local or parent persistent), looks up the corresponding environment
 // variable and (if the flag is unset) takes that environment variable value as the flag value.
-
 func (c *CLI) bindEnvVarsToFlags() (err error) {
-	var envKey string
+	// setFlagFromEnvVar panics: that way we can recover and extract the variable name that failed.
 	defer func() {
 		if r := recover(); r != nil {
-			err = fmt.Errorf("environment binding failed for variable %s: %v", envKey, r)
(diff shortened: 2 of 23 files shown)


CHANGELOG.md
```diff
@@ -47,6 +47,8 @@
   deprecated.
   [#1792](https://github.com/Kong/kong-operator/pull/1972)
   <!-- https://github.com/Kong/kong-operator/issues/1944 -->
+- Change env vars prefix from `GATEWAY_OPERATOR_` to `KONG_OPERATOR`.
+  `GATEWAY_OPERATOR_` prefixed env vars are still accepted but reported as deprecated.
```
pmalek, line 80:
```suggestion
  `GATEWAY_OPERATOR_` prefixed env vars are still accepted but reported as deprecated.
  [#2004](https://github.com/Kong/kong-operator/pull/2004)
```

modules/cli/cli.go
```diff
@@ -145,6 +146,19 @@ func (c *CLI) bindEnvVarsToFlags() (err error) {
 			if err := f.Value.Set(envValue); err != nil {
 				panic(err)
 			}
+			// ignore deprecated env var if newer is set
+			return
+		}
+
+		deprecatedEnvKey := fmt.Sprintf("%s%s", envDeprecatedVarFlagPrefix, strings.ToUpper(strings.ReplaceAll(f.Name, "-", "_")))
```
pmalek: This `func` had gotten a little bit long and I think it's time that it gets extracted and gets its own UT. WDYT?

We could then test this new behavior in a UT.

pmalek (commented):
👍 Just some minor nits about adding a test and missing changelog link.

modules/cli/cli.go
fgiudici: Yep, makes sense.
We will lose the closure side effect, i.e., the envKey in the main function cannot be filled anymore, but we can include directly the value of the failing env variable set in the error passed as argument to panic().

randmonkey (dismissed):

modules/cli/cli.go
```diff
@@ -119,29 +119,45 @@ type flagsForFurtherEvaluation struct {
...
 const (
-	envVarFlagPrefix = "GATEWAY_OPERATOR_"
+	envDeprecatedVarFlagPrefix = "GATEWAY_OPERATOR_"
+	envVarFlagPrefix           = "KONG_OPERATOR_"
 )
 
+// setFlagFromEnvVar looks for the env var name corresponding to `f`: if the env val is found its value is set
+// to the flag. In case of error it panics.
+func setFlagFromEnvVar(f *flag.Flag) {
+	envKey := fmt.Sprintf("%s%s", envVarFlagPrefix, strings.ToUpper(strings.ReplaceAll(f.Name, "-", "_")))
```
pmalek: Can we extract `strings.ToUpper(strings.ReplaceAll(f.Name, "-", "_"))` to a variable? 

We're doing this twice here and in line 138 which makes them a bit long and hard to comprehend. WDYT?

modules/cli/cli.go
```diff
@@ -119,29 +119,45 @@ type flagsForFurtherEvaluation struct {
...
+	deprecatedEnvKey := fmt.Sprintf("%s%s", envDeprecatedVarFlagPrefix, strings.ToUpper(strings.ReplaceAll(f.Name, "-", "_")))
+
+	if envValue, envSet := os.LookupEnv(deprecatedEnvKey); envSet {
+		fmt.Printf("WARN: %q env variable is deprecated, please use %q instead\n", deprecatedEnvKey, envKey)
+		// move to envKey var as that is what we look at in case of Set error.
+		envKey = deprecatedEnvKey
+		if err := f.Value.Set(envValue); err != nil {
+			panic(fmt.Errorf("environment binding failed for variable %s: %w", envKey, err))
+		}
+	}
```
pmalek: Nit: I'd suggest to consider making [the "happy path" the least indented ](https://maelvls.dev/go-happy-line-of-sight/)

Something along the lines of: (might not compile)

```suggestion
	deprecatedEnvKey := fmt.Sprintf("%s%s", envDeprecatedVarFlagPrefix, strings.ToUpper(strings.ReplaceAll(f.Name, "-", "_")))
	if envValue, envSet := os.LookupEnv(deprecatedEnvKey); envSet {
		fmt.Printf("WARN: %q env variable is deprecated, please use %q instead\n", deprecatedEnvKey, envKey)
		// move to envKey var as that is what we look at in case of Set error.
		if err := f.Value.Set(envValue); err != nil {
			panic(fmt.Errorf("environment binding failed for variable %s: %w", deprecatedEnvKey, err))
		}
	}
	if envValue, envSet := os.LookupEnv(envKey); envSet {
		if err := f.Value.Set(envValue); err != nil {
			panic(fmt.Errorf("environment binding failed for variable %s: %w", envKey, err))
		}
	}
```

WDYT?

fgiudici: uh, right, for sure we don't need anymore the `envKey = deprecatedEnvKey` ti should be dropped... good catch!
What I like from your proposal is that in case we have both the deprecated and the new variable we still print the WARN message.

pmalek (approved):

Issue #1608, opened 2025-05-14
v2alpha1 GatewayConfiguration CRD API is added as well (as there are common parts between CP and GC)


sandromodarelli:
### Acceptance criteria

- [ ] All Gateway and GatewayConfiguration tests are re-enabled (search for https://github.com/kong/kong-operator/issues/1608 in code).

pmalek:
`GatewayConfiguration` v2alpha1 has already been added to kubernetes-configuration: https://github.com/Kong/kubernetes-configuration/blob/4e14e2a62e23e63797eda5d085da00c0b43603e3/api/gateway-operator/v2alpha1/gatewayconfiguration_types.go

https://github.com/Kong/kong-operator/issues/1358 with its more focused sub issues tracks the API changes and implementation of remaining options.

Closing this as done.

pmalek:
Re-opening as we have not re-enabled the Gateway and GatewayConfiguration related tests.

Added acceptance criteria to reflect that.

PR #2045, opened 2025-08-12
fix(gateway): fix Gateway reconciliation and set dataplane in ControlPlane status

pmalek:

This PR:

- re-enabled `TestGatewayEssentials` integration test
- updates kubernetes-configuration to include https://github.com/Kong/kubernetes-configuration/pull/556 and https://github.com/Kong/kubernetes-configuration/pull/555
- make sure that `ControlPlane` controller sets the dataplane name in `status.dataplane` (which is useful in reconciliation, especially when deploying `ControlPlane` for `Gateway` which doesn't have a field for dataplane name in `spec`)
- makes sure that `ControlPlane` controllers disabled Gateway API controllers for those `ControlPlane`s that pair up with user deployed `DataPlane`s (for these we only allow ingress APIs since there's no Gateway to speak of)

**Which issue this PR fixes**

Part of #1608 (re-enable test).

The re-enabled test `TestGatewayEssentials` does not check for the network policies anymore since we've changed the code in https://github.com/Kong/kong-operator/blob/27e3c46cd201bf3d03d2e81000239b047da2b2ce/controller/gateway/controller.go#L397-L410 to not create them when running outside of cluster (and our integration tests run this way).

**PR Readiness Checklist**:


controller/controlplane_extensions/controller_enqueue_funcs.go
```diff
@@ -283,8 +283,23 @@ func enqueueControlPlaneForDataPlane(
 				}
 			}
 
-			if controlplane.Spec.DataPlane.Type != gwtypes.ControlPlaneDataPlaneTargetRefType ||
-				controlplane.Spec.DataPlane.Ref == nil {
+			switch controlplane.Spec.DataPlane.Type {
+			case gwtypes.ControlPlaneDataPlaneTargetRefType:
+				if controlplane.Spec.DataPlane.Ref == nil {
```
randmonkey: Could we merge the `if`s like 
```go
if controlplane.Spec.DataPlane.Ref == nil || controlplane.Spec.DataPlane.Ref.Name != dp.Name  {
    continue
}
```

controller/controlplane_extensions/controller_enqueue_funcs.go
```diff
@@ -283,8 +283,23 @@ func enqueueControlPlaneForDataPlane(
...
+					continue
+				}
+				if controlplane.Spec.DataPlane.Ref.Name != dp.Name {
+					continue
+				}
+
+			case gwtypes.ControlPlaneDataPlaneTargetManagedByType:
+				if controlplane.Status.DataPlane == nil {
```
randmonkey: Ditto here.

controller/controlplane_extensions/controller_enqueue_funcs.go
```diff
@@ -283,8 +283,23 @@ func enqueueControlPlaneForDataPlane(
 				}
```
randmonkey, line 283: Some additional comments for the `for controlplane := range controlplanes.Items` loop (the part of code not touched in this PR):
- The usage of the loop looks like to find the `controlPlane` that manages the input `DataPlane`. Do you think using `lo.Find(controlplanes.Items, func(cp ControlPlane) bool {...})` to find the satisfying `ControlPlane` could make the code easier to read?
- The `continue` in the inner loop `for _, ref := range controlplane.OwnerReferences` under `if owner != nil` just jumps out of the iteration of the inner loop but not out of the outer loop. Also, the `ref` defined in the loop is not used elsewhere. Should it jump out of the outer loop?

I don't mean that the comments are required to be resolves in this PR, but I'd like to see some explanation here.

controller/controlplane_extensions/metricsscraper/manager.go
```diff
@@ -291,9 +291,9 @@ func (msm *Manager) NotifyRemove(ctx context.Context, cp types.NamespacedName) {
 // that is associated with the given ControlPlane.
 // It returns true if the scraper was added.
 func (msm *Manager) Add(cp *gwtypes.ControlPlane, pipeline MetricsScrapePipeline) bool {
-	// TODO(pmalek): implement DataPlane external URL type
-	// ref: https://github.com/kong/kong-operator/issues/1366
-	if cp == nil || cp.Spec.DataPlane.Type != gwtypes.ControlPlaneDataPlaneTargetRefType || cp.Spec.DataPlane.Ref == nil {
+	if cp == nil ||
+		((cp.Spec.DataPlane.Type != gwtypes.ControlPlaneDataPlaneTargetRefType || cp.Spec.DataPlane.Ref == nil) &&
```
randmonkey: The compound of `||` s and `&&`s looks complex here (3 layers: `() || ( ( || ) && () ) `). Should we add comments or use separate `if`s here?

controller/controlplane_extensions/metricsscraper/manager.go
```diff
@@ -342,19 +342,30 @@ func (msm *Manager) enableMetricsScraperForControlPlanesDataPlane(
 	ctx context.Context,
 	controlplane *gwtypes.ControlPlane,
 ) error {
-	// TODO(pmalek): implement DataPlane external URL type
-	// ref: https://github.com/kong/kong-operator/issues/1366
-	if controlplane == nil || controlplane.Spec.DataPlane.Type != gwtypes.ControlPlaneDataPlaneTargetRefType || controlplane.Spec.DataPlane.Ref == nil {
-		return fmt.Errorf("ControlPlane %s does not have a valid, supported DataPlane target", controlplane.Name)
+	if controlplane == nil ||
```
randmonkey: Ditto here.

controller/controlplane_extensions/controller_enqueue_funcs.go
pmalek: Sure, b589f453d6d38a521c15e589a5cd003221ea835b

controller/controlplane_extensions/controller_enqueue_funcs.go
pmalek: b589f453d6d38a521c15e589a5cd003221ea835b

controller/controlplane_extensions/metricsscraper/manager.go
pmalek: WDYT about using a `switch` d7c3e2283f234728886c1694ebb677792abec5d7 ?

controller/controlplane_extensions/controller_enqueue_funcs.go
pmalek: It might refactored as you're proposing. We should just keep in mind when working on this to not overuse the closures (like it's done when utilizing `lo` functions).

Let's do this in a separate PR though.

controller/controlplane_extensions/metricsscraper/manager.go
randmonkey: Should we also change here to `switch`?

randmonkey:
The integration tests fails here:
```
	t.Log("verifying Gateway gets marked as not Programmed")
	require.Eventually(t, testutils.Not(testutils.GatewayIsProgrammed(t, GetCtx(), gatewayNN, clients)), testutils.GatewayReadyTimeLimit, time.Second)
	require.Eventually(t, testutils.GatewayListenersAreProgrammed(t, GetCtx(), gatewayNN, clients), testutils.GatewayReadyTimeLimit, time.Second) // <-- This is where the `eventually` fails
``` 
So the check's intention is to wait for the listeners to be "programmed" again after the gateway's "programmed" changed to `False`?

controller/controlplane_extensions/metricsscraper/manager.go
pmalek: 👍 9f60c986c18692703118b1359af6315e15f99d51

test/integration/gateway_test.go
```diff
@@ -80,14 +78,15 @@ func TestGatewayEssentials(t *testing.T) {
...
+	// NOTE: We're not verifying if the NetworkPolicies are created
+	// in integration tests.
+	// Code ref: https://github.com/Kong/kong-operator/blob/27e3c46cd201bf3d03d2e81000239b047da2b2ce/controller/gateway/controller.go#L397-L410
 
 	t.Log("verifying connectivity to the Gateway")
 	require.Eventually(t, Expect404WithNoRouteFunc(t, GetCtx(), "http://"+gatewayIPAddress), testutils.SubresourceReadinessWait, time.Second)
 
 	t.Log("verifying GatewayClass has supportedFeatures set")
-	requiredFeatures, err := gatewayapi.GetSupportedFeatures(consts.RouterFlavorExpressions)
+	requiredFeatures, err := gatewayapi.GetSupportedFeatures(consts.RouterFlavorTraditionalCompatible)
```
tao12345666333, line 89:
```suggestion
	requiredFeatures, err := gatewayapi.GetSupportedFeatures(consts.DefaultRouterFlavor)
```

I think it would be more convenient to use the DefaultRouterFlavor value directly here.

pmalek:
> The integration tests fails here:
> 
> ```
> 	t.Log("verifying Gateway gets marked as not Programmed")
> 	require.Eventually(t, testutils.Not(testutils.GatewayIsProgrammed(t, GetCtx(), gatewayNN, clients)), testutils.GatewayReadyTimeLimit, time.Second)
> 	require.Eventually(t, testutils.GatewayListenersAreProgrammed(t, GetCtx(), gatewayNN, clients), testutils.GatewayReadyTimeLimit, time.Second) // <-- This is where the `eventually` fails
> ```
> 
> So the check's intention is to wait for the listeners to be "programmed" again after the gateway's "programmed" changed to `False`?

I've changed this check to check if listeners are not programmed there (and decreased the tick to not cause flakiness when we'd skip the time when they are not programmed).

This PR is not ready for review.

controller/controlplane/controller.go
```diff
@@ -714,3 +743,70 @@ func (r *Reconciler) validateControlPlaneOptions(cp *ControlPlane) (string, bool
...
+			return "", op.Noop, fmt.Errorf(
+				"no DataPlanes owned by Gateway %s/%s",
+				cp.GetNamespace(), owner.Name,
+			)
+		}
+		dataplaneName = dataplaneList.Items[0].Name
+	}
+
+	if (cp.Status.DataPlane != nil && dataplaneName == cp.Status.DataPlane.Name) ||
+		(cp.Status.DataPlane == nil || dataplaneName == "") {
```
randmonkey: Should it be `&&` here? This condition returns `true` when `cp.Status.DataPlane == nil` OR `dataplaneName` is empty. So that the `cp.Status.DataPlane` can never be updated because it is `nil` in initialization.

controller/controlplane/controller.go
```diff
@@ -714,3 +743,70 @@ func (r *Reconciler) validateControlPlaneOptions(cp *ControlPlane) (string, bool
...
+		})
+		if err != nil {
+			return "", op.Noop, fmt.Errorf(
+				"failed to find DataPlanes owned by Gateway %s/%s: %w",
+				cp.GetNamespace(), owner.Name, err,
+			)
+		}
+		if len(dataplaneList.Items) != 1 {
+			return "", op.Noop, fmt.Errorf(
+				"no DataPlanes owned by Gateway %s/%s",
```
randmonkey: Is it possible to get more than 1 `DataPlane`s owned by `Gateway`? If so the error message may not be accurate.

pmalek: 👍 19786f7557f785f6764161515d5dbdf9b82a76c9

controller/controlplane/controller.go
pmalek: Good catch. I've fixed that and added a test for that in 41eda5b8099567a00a2f62c8cbd267c437fb8274.

PTAL

controller/controlplane/controller.go
```diff
@@ -94,7 +95,16 @@ func (r *Reconciler) SetupWithManager(_ context.Context, mgr ctrl.Manager) error
...
 		Watches(
 			&operatorv1alpha1.WatchNamespaceGrant{},
-			handler.EnqueueRequestsFromMapFunc(r.listControlPlanesForWatchNamespaceGrants))
+			handler.EnqueueRequestsFromMapFunc(r.listControlPlanesForWatchNamespaceGrants)).
+		Watches(
+			&operatorv1beta1.DataPlane{},
+			handler.EnqueueRequestsFromMapFunc(r.getControlPlanesFromDataPlane)).
+		// watch for changes in the DataPlane deployments, as we want to be aware of all
+		// the DataPlane pod changes (every time a new pod gets ready, the deployment
+		// status gets updated accordingly, leading to a reconciliation loop trigger)
```
programmer04:
```suggestion
		// Watch for changes in the DataPlane deployments, as we want to be aware of all
		// the DataPlane pod changes (every time a new pod gets ready, the deployment
		// status gets updated accordingly, leading to a reconciliation loop trigger).
```

controller/controlplane/controller.go
```diff
@@ -206,6 +216,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{}, nil // no need to requeue, status update will requeue
 	}
 
+	// Set DataPlane in ControlPlane's status
```
programmer04:
```suggestion
	// Set DataPlane in ControlPlane's status.
```

controller/controlplane/controller.go
```diff
@@ -714,3 +743,70 @@ func (r *Reconciler) validateControlPlaneOptions(cp *ControlPlane) (string, bool
...
+	if (cp.Status.DataPlane != nil && dataplaneName == cp.Status.DataPlane.Name) ||
+		(cp.Status.DataPlane == nil && dataplaneName == "") {
+		return dataplaneName, op.Noop, nil
+	}
+	oldControlPlane := cp.DeepCopy()
+	cp.Status.DataPlane = &gwtypes.ControlPlaneDataPlaneStatus{
+		Name: dataplaneName,
+	}
+	err := r.Status().Patch(ctx, cp, client.MergeFrom(oldControlPlane))
+	if err != nil {
```
programmer04:
```suggestion
	if err := r.Status().Patch(ctx, cp, client.MergeFrom(oldControlPlane)); err != nil {
```

controller/controlplane_extensions/metricsscraper/manager.go
```diff
@@ -342,19 +350,34 @@ func (msm *Manager) enableMetricsScraperForControlPlanesDataPlane(
...
+	if controlplane == nil {
+		return fmt.Errorf("ControlPlane does not have a valid, supported DataPlane target")
 	}
 
-	var (
+	var dpNN types.NamespacedName
+	switch controlplane.Spec.DataPlane.Type {
+	case gwtypes.ControlPlaneDataPlaneTargetRefType:
+		if controlplane.Spec.DataPlane.Ref == nil {
+			return fmt.Errorf("ControlPlane with spec.dataplane.type=ref does not have a valid, does not have the DataPlane set")
```
programmer04:
```suggestion
			return fmt.Errorf("ControlPlane with spec.dataplane.type=ref does not have a valid DataPlane set")
```

controller/gateway/controller.go
```diff
@@ -592,7 +592,7 @@ func (r *Reconciler) provisionControlPlane(
 		}
 		return nil
 	case count > 1:
-		err := fmt.Errorf("control plane deployments found: %d, expected: 1, requeing", count)
+		err := fmt.Errorf("control planes found: %d, expected: 1, requeing", count)
```
programmer04:
```suggestion
		err := fmt.Errorf("ControlPlanes found: %d, expected: 1, requeing", count)
```
like everywhere

programmer04 (approved):

PR #2061, opened 2025-08-14
chore: move mocks under test/mocks/ 

### LINKED ISSUE #955: Move mocks to a separate package
## Problem Statement
As we did in KIC: https://github.com/Kong/kubernetes-ingress-controller/pull/6868, we want to move the mocks to a separate package inside `/test`
...

## Proposed Solution
- Move all the mock code to a package or multiple packaged inside `/test`
- Update the script to generate mock SDK to generate mock code in the dedicated package
## Additional Information

...

## Acceptance Criteria
- [ ] All mocks are moved to a place inside `/test`

fgiudici:

**Which issue this PR fixes**

Fixes #955 

**PR Readiness Checklist**:


Dockerfile
```diff
@@ -22,6 +22,7 @@ COPY modules/ modules/
 COPY controller/ controller/
 COPY pkg/ pkg/
 COPY internal/ internal/
+COPY test/mocks test/mocks
```
pmalek: Why do we need to include the mocks in the final Dockerfile? Non test code shouldn't rely on these 🤔

pmalek (commented):
It looks like GitHub has figured it out on its own but we might as well update the .gitattributes in https://github.com/Kong/kong-operator/blob/e0a760f4182060eb7750dfc0a5fdc2c06b44b861/.gitattributes#L3.

Other than that just one comment regarding including mocks in Dockerfile.

Dockerfile
fgiudici: It is related to the metricsmock.MockRecorder implementation of metrics.Recorder : MockRecorder has been moved under test/mocks. It is used in the tests but also as the default in the  konnect.NewKonnectEntityReconciler() func, as it is an implementation with methods that do nothing.
It looked more on the safe side to leave that initialization in.
Opening a comment on the related part of the code, let's move the discussion there.

controller/konnect/reconciler_generic.go
```diff
@@ -101,7 +102,7 @@ func NewKonnectEntityReconciler[
 		Client:                  client,
 		SyncPeriod:              consts.DefaultKonnectSyncPeriod,
 		MaxConcurrentReconciles: consts.DefaultKonnectMaxConcurrentReconciles,
-		MetricRecorder:          &metrics.MockRecorder{},
+		MetricRecorder:          &metricsmocks.MockRecorder{},
```
fgiudici: in NewKonnectEntityReconciler() we use MockRecorder as the default initialization for the metrics.Recorder interface: MockRecorder is an implementation with methods that do nothing, so more safe than a nil interface.
This creates anyway a dependency on the metricsmocks package in main code (not test packages).
We either do not initialize MetricRecorder (leaving it to nil) or rename the implementation dropping "Mock" and move it out of test code/package.
Since NewKonnectEntityReconciler is called only once in non-test code and with the .WithMetricRecorder() option (which overwrites anyway the MetricRecorder parameter) I would go with the former proposal.
Will push an on-top commit with the proposal.

pmalek: Thanks for the detailed explanation. Now I recall all of this and I agree with your view.

Looking at the tests failures it would seem that we need to add the option in each like https://github.com/Kong/kong-operator/blob/85e364f9bcd34560feb9e966f77b14cc439e2c1e/test/envtest/konnect_entities_kongcacertificate_test.go#L44-L46 to set the metrics recorder in tests and we should be good.

fgiudici:
> It looks like GitHub has figured it out on its own but we might as well update the .gitattributes in
> 
> https://github.com/Kong/kong-operator/blob/e0a760f4182060eb7750dfc0a5fdc2c06b44b861/.gitattributes#L3
> 

Dropped `pkg/clientset/** linguist-generated=true` from the .gitattributes file as it looked a left over.
Changed the files generated from mockery to follow the zz_generated.<FILENAME> template already included in .gitattributes.

pmalek (approved):

PR #2070, opened 2025-08-15
chore(controlplane): Set OptionsValid condition and add tests

### LINKED ISSUE #1999: Use dedicated condition in status when CP level label selectors conflicts with operator level label selectors
When KO specifies `--secret-label-selector` or `--config-map-label-selector` to configure operator level label selectors to filter reconciled secrets or configMaps, the label selectors for secrets and configMaps in `ControlPlane.Spec` may conflict with the operator level selectors.
If the case happens, we should mark the `ControlPlane` as invalid and reflect the status in its `status.conditions`. We should define a new condition reason in the `kubernetes-configuration` repo and may define a dedicated condition type or reuse the `provisioned` type.

### Acceptance Criteria

- [x] When CP's secret/configMap label selectors conflict with KO's label selectors, a condition with `False` status and dedicated reason is updated in CP's status
- [x] No reconciliation errors should be returned as the error causes endless reconciliation loop

### LINKED ISSUE #2024: Add unit tests for invalid control plane options and other cases in reconciling `ControlPlane`
## Problem Statement

Now we only include one case for setting control plane defaults in unit tests inside the file `controller/controlplane/controller_test.go`. We need to add more test cases including that invalid control plane options should cause the `ControlPlane`'s conditions in status updated to reflect the error.

But when the reconciliation process runs to the step to update the `ControlPlane`'s status, error happens
```
failed updating ControlPlane's status : controlplanes.gateway-operator.konghq.com "test-controlplane" not found
```
but we have added the controlplane to the fake runtime client. This should be caused by the precedence of `v1beta1 over `v2alpha1`.

## Proposed Solution

Add unit test case to verify that `ControlPlane`'s condition is updated to reflect that the control plane options are invalid when the error gets fixed

## Additional Information

...

## Acceptance Criteria

- [x] Unit test in `ControlPlane` controller's `Reconcile` method for invalid control plane options is addedd

randmonkey:

Set `OptionsValid` condition in `ControlPlane`s' status. The reconcilier sets the status to `True` when options are valid and to `False` with messages when it is invalid. Also added unit test case for it.

**Which issue this PR fixes**

Fixes #1999 and also fixes #2024 

**PR Readiness Checklist**:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,10 @@
   control over Konnect integration settings including consumer synchronization, licensing
   configuration, node refresh periods, and config upload periods.
   [#2009](https://github.com/Kong/kong-operator/pull/2009)
+- Added `OptionsValid` condition to `ControlPlane`s' status. The status is set to
+  `True` if the `ControlPlane`'s options in its `spec` is valid and set to `False`
+  if the options are invalid against the operator's configuration.
+  [#2070](https://github.com/Kong/kong-operator/pull/2070)
 
 ### Fixed
 
--- a/controller/controlplane/controller.go
+++ b/controller/controlplane/controller.go
@@ -24,7 +24,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
 
-	kcfgconsts "github.com/kong/kubernetes-configuration/v2/api/common/consts"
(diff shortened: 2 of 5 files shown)


pmalek (commented):
Looks good to me. Do you think it would be beneficial to add an assertion in integration tests for this? e.g. in `TestControlPlaneEssentials`?

randmonkey:
> Looks good to me. Do you think it would be beneficial to add an assertion in integration tests for this? e.g. in `TestControlPlaneEssentials`?

Done.

pmalek (approved):

Issue #1367, opened 2025-03-13
Expose KIC validating webhook server from Kong Operator


czeslavo:
### Problem statement 

To not lose the capabilities we had in KIC that allowed users to validate their configuration, we will need to port it to KO.

### Proposed solution

Dynamically route to handlers exposed by individual ControlPlanes that match a given object (e.g., by matching an ingress class or gateway).

### Acceptance criteria

- [ ] All validation rules that KIC supports in its validating webhook are exposed in KO 
- [ ] If there are validation rules that require communication with DPs, the requests are properly routed to every KIC instance handler

pmalek:
#1793 tracks the changes in API.

Both of these tasks can be worked on independently as the setup can use the defaults initially and API changes are to be done in https://github.com/Kong/kubernetes-configuration/.

programmer04:
There is dedicated document that describes the design - [Exposing KIC validation webhook from Kong Operator](https://docs.google.com/document/d/13COgVoCv_Fa1cQLHp9z9UJ1Ghq1czzrUDlunpzSLgMo/edit?tab=t.0#heading=h.9ikojn8u15x)

Issue #2029, opened 2025-08-08
☔ Post-Release Activities for 2.x Major Release


sandromodarelli:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Problem Statement

This issue tracks all necessary activities and tasks after the 2.x major release.
The work is grouped into sub-issues and includes tasks like completing integration tests, end-to-end (E2E) tests, codebase rearrangement, and general maintenance.

### Acceptance Criteria

- [ ] Complete Integration Tests: ensure all components interact correctly and meet the expected integration criteria.
- [ ] Complete E2E Tests: run full system tests to validate user workflows and end-to-end functionality.
- [ ] Rearrange Codebase: refactor the codebase for better maintainability, structure, and performance.
- [ ] General Maintenance: address any bugs, performance improvements, or minor fixes.
- [ ] Documentation Updates: update documentation to reflect any breaking changes or new features introduced in the 2.x release.

Issue #1986, opened 2025-08-05
Enable conversion webhook for test suites


programmer04:
## Problem Statement

Spinoff of

- https://github.com/Kong/kong-operator/pull/1959

Read the description for more context.

Conversion webhook is enabled by default for the controller, but in test suites e.g. `integration`, it's disabled. 

The missing piece is ensuring that cert-manager is in place to allow TLS connections between the K8s cluster's control plane and the KO controller. 

Enhance the machinery of running those suites to include all the stuff required for the webhook to run, to be able to cover it with some integration tests too.

> It probably makes sense to do it when KO will have well-defined test suites 

## Acceptance Criteria

- [ ] identify test suites that should be run with webhook enabled
- [ ] allow running them with webhook enabled

programmer04:
Conversion webhook is triggered only for specific types (which are deprecated), so having it enabled for all of the test suites is pointless, because other resources than those registered for conversion, currently
- `ControlPlane` from `gateway-operator.konghq.com/v1beta1`
- `KonnectGatewayControlPlane` from `konnect.konghq.com/v1alpha1`

are not triggering the conversion webhook. Thus, only two e2e tests are enough to test it end-to-end with configurations

1. Scenario to run for `ControlPlane` from `gateway-operator.konghq.com/v1beta1`
    ```yml
    apiVersion: gateway-operator.konghq.com/v1beta1
    kind: DataPlane
    metadata:
      name: dataplane-cp-watchnamespace
      namespace: default
    spec:
      deployment:
        podTemplateSpec:
          spec:
            containers:
            - name: proxy
              image: kong:3.9
              env:
              - name: KONG_LOG_LEVEL
                value: debug
              resources:
                requests:
                  memory: "64Mi"
                  cpu: "250m"
                limits:
                  memory: "1024Mi"
                  cpu: "1000m"
              readinessProbe:
                initialDelaySeconds: 1
                periodSeconds: 1
    ---
    apiVersion: v1
    kind: Namespace
    metadata:
      name: namespace-a
    ---
    apiVersion: gateway-operator.konghq.com/v1alpha1
    kind: WatchNamespaceGrant
    metadata:
      name: watchnamespacegrant
      namespace: namespace-a
    spec:
      from:
      - group: gateway-operator.konghq.com
        kind: ControlPlane
        namespace: default
    ---
    apiVersion: gateway-operator.konghq.com/v1beta1
    kind: ControlPlane
    metadata:
      name: controlplane-example
[...]

PR #2084, opened 2025-08-19
feat: migrate KIC validation webhook

programmer04:

This PR brings the KIC validation webhook as part of KO in the way described in the dedicated document - [Exposing KIC validation webhook from Kong Operator](https://docs.google.com/document/d/13COgVoCv_Fa1cQLHp9z9UJ1Ghq1czzrUDlunpzSLgMo/edit?tab=t.0#heading=h.9ikojn8u15x). It boils down to making a webhook server part of KO and dynamically registering/unregistering specific validators based on KIC instances run inside.

Known limitations/characteristics

KIC has a webhook for validation. KGO only relies on CEL rules. KIC webhook will now be run on the KO level. Its validation has always been a best effort. KO will use a combination of CEL and webhook.

Sometimes it's hard to know where an object belongs (lack of ingressClass or gatewayClass or annotation) and to which DataPlane some validation requests e.g., regexes should be routed. and based on this trigger, a more detailed validation. Read more about this issue and possible resolution in RFC Deprecating ingressClass Annotation in Favor of a Unified Controller Reference.

To summarize, in KO 2.0.0, the behavior will be similar to what it was in KGO, where validation for a particular ControlPlane and DataPlane combination might allow an invalid configuration under certain circumstances. We've always aimed to never be falsely positive, to not block users, and we'll maintain this semantic.

**Which issue this PR fixes**

Part of #1367

Currently, it works here with `make run.skaffold`.

Unit tests are adjusted (since the check of ingressclass has been moved, some test cases became obsolete). 
For integration tests and envtests, due to a change in the architecture, they need to be ported. Currently, we live in the duality of KIC and KO suits. When they become one, it will be feasible to have proper webhook tests, see a dedicated issue https://github.com/Kong/kong-operator/issues/2176 that is also mentioned in specific skipped test cases. Some work has already been done in this PR - ensuring certs in proper paths.


pkg/utils/test/config.go
```diff
@@ -42,6 +42,7 @@ func DefaultControllerConfigForTests(opts ...ControllerConfigOption) manager.Con
 	cfg.ServiceAccountToImpersonate = ServiceAccountToImpersonate
 	// TODO: https://github.com/Kong/kong-operator/issues/1986
 	cfg.ConversionWebhookEnabled = false
+	cfg.ValidationWebhookEnabled = false
```
pmalek, line 46: Can we have a similar issue for validation webhook like we have https://github.com/Kong/kong-operator/issues/1986 for conversion?

ingress-controller/validation/webhook.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+// SetupAdmissionServer sets up the admission webhook server.
+func SetupAdmissionServer(
+	ctx context.Context,
+	m ctrl.Manager,
+	// clientsManager admission.GatewayClientsProvider,
+	// referenceIndexers ctrlref.CacheIndexers,
+	// translatorFeatures translator.FeatureFlags,
+	// storer store.Storer,
```
pmalek: ?

ingress-controller/validation/webhook.go
```diff
@@ -0,0 +1,38 @@
...
+) (*admission.RequestHandler, error) {
+	admissionLogger := ctrl.LoggerFrom(ctx).WithName("admission-server")
+
+	admissionReqHandler := &admission.RequestHandler{
+		// ReferenceIndexers: ctrlref.CacheIndexers{}, ????
```
pmalek: ?

ingress-controller/test/envtest/admission_webhook_envtest_test.go
pmalek, line 1: This I understand needs to be adapted, right?

ingress-controller/internal/admission/handler.go
```diff
@@ -31,17 +32,105 @@ const (
...
 	Logger logr.Logger
 }
 
+// dispatchValidationNoMatcher returns the first validator if exists, without any matching.
+// There is no specific matching further so it's safe to do it this way, see
+// - h.handleKongPlugin -> ValidatePlugin
+// - h.handleKongClusterPlugin -> ValidateClusterPlugin
+// - h.handleSecret -> ValidateCredential
+// - h.handleGateway -> ValidateGateway
+// - h.handleHTTPRoute -> ValidateHTTPRoute
```
pmalek: Can we please describe here what the return values mean? Specifically the `bool`

ingress-controller/internal/admission/handler.go
```diff
@@ -31,17 +32,105 @@ const (
...
+			validator = v
+			ok = true
+			break
+		}
+	}
+	return validator, ok
+}
+
+// ---
+
```
pmalek:
```suggestion
```

ingress-controller/internal/admission/handler.go
```diff
@@ -31,17 +32,105 @@ const (
...
+		if v.IngressClassMatcher(&obj) {
+			validator = v
+			ok = true
+			break
+		}
+	}
+	return validator, ok
+}
+
+// h.handleIngress-> ValidateIngress (2 matchers)
```
pmalek: Update godoc

ingress-controller/internal/admission/handler.go
```diff
@@ -31,17 +32,105 @@ const (
...
+			v.IngressV1ClassMatcher(&ing) {
+			validator = v
+			ok = true
+			break
+		}
+	}
+	return validator, ok
+}
+
+// h.handleKongCustomEntity -> ValidateCustomEntity
```
pmalek: Update godoc

pkg/utils/test/config.go
programmer04: https://github.com/Kong/kong-operator/issues/2029

pmalek (approved):

config/default/validating_policies/kustomization.yaml
```diff
@@ -3,3 +3,4 @@ kind: Kustomization
 
 resources:
 - dataplane_ports_validating_admission_policy.yaml
+- validating_webhook_configuration.yaml
```
pmalek, line 6: This shouldn't be under the `validating_policies` directory but https://github.com/Kong/kong-operator/pull/2190 moves that into a dedicated admission webhook directory so that's fine.

PR #2091, opened 2025-08-20
Chore(metadata): Use metadata from `modules/manager/metadata` in `ingress-controller` packages 

### LINKED ISSUE #2088: Set UserAgent of ingress controller's UserAgent
## Problem Statement

Since we do not run release job of KIC to set the `Release` metadata, the values are `NOT_SET`. The UserAgent used to call Kong gateway admin APIs are created from the metadata so the UserAgent becomes `kong-ingress-controller/NOTSET` when the ingress controller is running in KO. It could be confusing in observability  and statistics of Kong admin APIs. So we need to setup the release metadata to make the UserAgent with the "Kong-Operator" project name and the version of KO.
...

## Proposed Solution

- Change the prefix of the UserAgent in `ingress-controller/pkg/metadata` to `kong-operator`
- Fill variable `ingress-controller/pkg/metadata.Release` to the release version of KO in KO build scripts and release workflows
- Fill also the variables `Repo`, `ProjectURL`, `Commit` in the package correctly in build scripts
...
## Additional Information

...

## Acceptance Criteria

- [x] UserAgent of embedded ingress controller is filled with `kong-operator/x.x.x` where x.x.x is the release of KO
- [ ] There is document to introduce the meaning of UserAgent correctly

randmonkey:

Use metadata from `modules/manager/metadata` in `ingress-controller` packages which was using release metadata in KIC.

**Which issue this PR fixes**

Fixes #2088

**PR Readiness Checklist**:


pmalek (commented):
I think that since we already have the metadata package in operator: https://github.com/Kong/kong-operator/blob/b7b9b7ea79ffc7b7d63723c279ea46349dbe66a2/modules/manager/metadata/metadata.go we can use it for the "ingress-controller" user agent as well.

It will make ingress-controller code import package from ko but I believe it's about time. We use a single module in this repository so this shouldn't be breaking in any way. WDYT?

randmonkey:
> I think that since we already have the metadata package in operator: https://github.com/Kong/kong-operator/blob/b7b9b7ea79ffc7b7d63723c279ea46349dbe66a2/modules/manager/metadata/metadata.go we can use it for the "ingress-controller" user agent as well.
> 
> It will make ingress-controller code import package from ko but I believe it's about time. We use a single module in this repository so this shouldn't be breaking in any way. WDYT?

Our integration tests and env tests still directly run the tests from `ingress-controller` directory so the metadata are still used. Updated all the usage of `metadata` to use `modules/manager/metadata` and the tests seems OK. Let's see if there are any other affects.

ingress-controller/pkg/telemetry/reports.go
```diff
@@ -82,7 +82,7 @@ func SetupAnonymousReports(
 	}
 
 	fixedPayload := types.Payload{
-		"v":  metadata.Release,
+		"v":  metadata.Metadata().Release,
```
randmonkey, line 85: REVIEW: should we add a field to mark that the message is sent from KO to distinguish between the message sent from existing KIC instances?

pmalek: We should adjust the signal names here https://github.com/Kong/kong-operator/blob/b264059b1c854b85f96ad8755b8c16cbf2a459ba/internal/telemetry/telemetry.go#L38-L39 to distinguish that IMHO.

pmalek:
> > I think that since we already have the metadata package in operator: https://github.com/Kong/kong-operator/blob/b7b9b7ea79ffc7b7d63723c279ea46349dbe66a2/modules/manager/metadata/metadata.go we can use it for the "ingress-controller" user agent as well.
> > It will make ingress-controller code import package from ko but I believe it's about time. We use a single module in this repository so this shouldn't be breaking in any way. WDYT?
> 
> Our integration tests and env tests still directly run the tests from `ingress-controller` directory so the metadata are still used. Updated all the usage of `metadata` to use `modules/manager/metadata` and the tests seems OK. Let's see if there are any other affects.

I can see that all tests pass to I believe this should be good to go. WDYT @randmonkey ?

pmalek (commented):
👍 

Can we also remove the ingress controller metadata package in this PR https://github.com/Kong/kong-operator/tree/55ca3c63083d67751f1e9467b292cd47f0beb14b/ingress-controller/pkg/metadata?

randmonkey:
> 👍
> 
> Can we also remove the ingress controller metadata package in this PR https://github.com/Kong/kong-operator/tree/55ca3c63083d67751f1e9467b292cd47f0beb14b/ingress-controller/pkg/metadata?

2ce305bece342947d0f3625ebb5fbeca191b8731

ingress-controller/pkg/telemetry/reports.go
randmonkey: renamed in 49da05fa2f9d866a31c3bc3319c7e2319267206d.

pmalek (approved):

PR #2122, opened 2025-08-25
feat: proper support of conversion webhook (with cert-manager) in Helm Chart

### LINKED ISSUE #1987: Update Helm chart to work with conversion webhook (with cert-manager)
## Problem Statement

PR 

- https://github.com/Kong/kong-operator/pull/1959

introduced kustomize configuration YAMLs for the conversion webhook. It's done with the use of kubebuilder machinery (the same as they present in examples). The way for certificate management is `cert-manager`, which needs to be incorporated in the Helm chart and have a knob to allow installing with webhook disabled (by default it's enabled).

## Proposed Solution

Document `cert-manager` as a required dependency before installing the KO chart. If someone does not have it in a cluster yet, document how to install it using the Helm chart. Also, document why it's needed. Certificate management for webhook is not only needed for conversion webhook, but for validation too. So both for upgrading an existing KGO and a fresh installation, cert management is needed.

 As an alternative, show how to install KO with both (conversion and validation) webhooks disabled - in such case it will work without `cert-manager`.

## Acceptance Criteria

- [ ] Helm chart allows using KO with webhook enabled/disabled
- [ ] ~documentation is updated (mentions dependency on `cert-manager`)~ -> will be covered by https://github.com/Kong/charts/issues/1379

programmer04:

The biggest problem that this PR resolves is configuring based on condition (enabled/disabled) conversion webhook for CRDs. They are configured on the CRD level. When CRDs in Helm are put in the designed directory `‎charts/kong-operator/crds/` they can't be templated. Templating is required because `name` and `namespace` are required for configuring the K8s Service for webhook on the CRD level (and namespace can't be hardcoded). Related issues in Helm repository:
- https://github.com/helm/helm/issues/5871
- https://github.com/helm/helm/issues/8171
- https://github.com/helm/helm/issues/10043

To sum up, they propose a dedicated Chart only with CRDs that are in the template directory, and this PR implements it with automation to generate the CRDs config. Furthermore implemented solution follows what is done in `cert-manager` and utilizes annotation [helm.sh/resource-policy: keep](https://helm.sh/docs/howto/charts_tips_and_tricks/#tell-helm-not-to-uninstall-a-resource). See [cert-manager reationale](https://github.com/cert-manager/cert-manager/blob/a8829d041bb809dd477a5bd5daf7a25320ed4a5d/design/20240206.helm-resource-policy.md) and [api](https://artifacthub.io/packages/helm/cert-manager/cert-manager). The same Chart API implements this pr with `ko-crds.enabled` and `ko-crds.keep`.

Furthermore, it includes adjustments needed to make Chart work with both enabled and disabled conversion webhook and cert-manager. When the conversion webhook is disabled, deprecated versions of CRDs are not installed at all, such a try ends with
```
error: resource mapping not found for name: "controlplane-example" namespace: "default" from "config/samples/controlplane-dataplane-watchnamespaces.yaml": no matches for kind "ControlPlane" in version "gateway-operator.konghq.com/v1beta1"
```
When the webhook is enabled, conversion happens, and a derpecation message is displayed
```
Warning: ControlPlane v1beta1 has been deprecated in favor of v2beta1 and it will be removed in the future.
```

 

**Which issue this PR fixes**

Closes https://github.com/Kong/kong-operator/issues/1987

Configuring the conversion webhook without cert-manager will be handled in subsequent issues/PRs. So to test what is here prerequisite is 

```sh
helm install \
  cert-manager jetstack/cert-manager \
  --namespace cert-manager \
  --create-namespace \
  --version v1.18.2 \
  --set crds.enabled=true
```

**PR Readiness Checklist**:


charts/kong-operator/Chart.yaml
```diff
@@ -13,6 +13,8 @@ appVersion: "2.0.0-alpha.3"
 annotations:
   artifacthub.io/prerelease: "true"
 dependencies:
+  - name: ko-crds
+    version: 1.0.0
```
pmalek, line 17: Does this mean that we install this unconditionally (note no `condition:`)? Will this mean that CRDs are updated on upgrades as well?

How does this relate to the https://github.com/Kong/kong-operator/tree/87ec4996a08c9b1b99ab9fd700d1f526c7ff9a74/charts/kong-operator/crds which contains the CRDs? It feels like this will duplicate that so I'm wondering if we shouldn't remove that 🤔

programmer04: I thought that without those KO-specific CRDs operator is unusable, and only our operator uses them. But you're right condition should be added also for the case when multiple instances of KO are installed in a cluster - you only need one set of CRDs. So I'm going to add a condition similar to others. On the other hand there is something like an annotation [helm.sh/resource-policy: keep](https://helm.sh/docs/howto/charts_tips_and_tricks/#tell-helm-not-to-uninstall-a-resource), maybe it should be included in a similar way how [cert-manager](https://github.com/cert-manager/cert-manager/blob/a8829d041bb809dd477a5bd5daf7a25320ed4a5d/design/20240206.helm-resource-policy.md) does it (see also [here](https://artifacthub.io/packages/helm/cert-manager/cert-manager). WDYT? 

It's removed (moved) I mean replaced see the diff

image

😉

pmalek:
Just tested this and I helm install failed with:

```
helm install --create-namespace -n kong kong ./charts/kong-operator
Error: INSTALLATION FAILED: Unable to continue with install: CustomResourceDefinition "kongconsumergroups.configuration.konghq.com" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "kong"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "kong"
```

charts/kong-operator/Chart.yaml
pmalek: > On the other hand there is something like an annotation [helm.sh/resource-policy: keep](https://helm.sh/docs/howto/charts_tips_and_tricks/#tell-helm-not-to-uninstall-a-resource), maybe it should be included in a similar way how [cert-manager](https://github.com/cert-manager/cert-manager/blob/a8829d041bb809dd477a5bd5daf7a25320ed4a5d/design/20240206.helm-resource-policy.md) does it (see also [here](https://artifacthub.io/packages/helm/cert-manager/cert-manager). WDYT?

That does make sense to me. We'd like to achieve similar UX and that (even though it might have corner cases like outlined in the linked doc) seems to provide what we need.

programmer04: I did it accordingly, PTAL updated description

charts/kong-operator/charts/ko-crds/templates/ko-crds.yaml
```diff
@@ -54471,6 +54605,7 @@ spec:
     singular: konnectextension
   scope: Namespaced
   versions:
+{{ if .Values.global.conversionWebhook.enabled }}
```
pmalek, line 54608: Because of this, users that had previous versions installed will need to either set `global.conversionWebhook.enabled` to `true` or remove the `KonnectExtension` CRD (together with all the custom resources of that type).

```
helm upgrade --install --create-namespace -n kong kong ./charts/kong-operator
Release "kong" does not exist. Installing it now.
Error: cannot patch "konnectextensions.konnect.konghq.com" with kind CustomResourceDefinition: CustomResourceDefinition.apiextensions.k8s.io "konnectextensions.konnect.konghq.com" is invalid: status.storedVersions[0]: Invalid value: "v1alpha1": missing from spec.versions; v1alpha1 was previously a storage version, and must remain in spec.versions until a storage migration ensures no data remains persisted in v1alpha1 and removes v1alpha1 from status.storedVersions
```

I believe we should make a note in ( _I think_ ) chart's changelog to let people know about this.

Ideally we'd also expand the comment on the `global.conversionWebhook.enabled` in `values.yaml`.

programmer04: Yup, you're right. What do you propose to put in `values.yaml` to make a hint?

In terms of actual steps, etc. I'm thinking that in any changelog for `2.0.0` I mean specific for Chart or KO, we should make a big announcement about manual steps and link to the details in the dedicated e2e doc guide to not scatter pieces of advice everywhere. WDYT?

pmalek: I'm fine either way. We can add a link in the changelog to docs. Let's just make sure that we don't miss this.

We can add link to the issue that will add it separately when the migration guide is ready.

pmalek: Since that is a breaking change for users that used previous CRDs I'd make a note of that in the comment on the `global.conversionWebhook` to describe the behavior that users get.

programmer04: Added more details to `values.yaml`. Notified @alacuku about this thread and approach since he is responsible for preparing migration guides

pmalek (approved):

Issue #2136, opened 2025-08-26
Add self signed certs for conversion webhook and CP validating webhook


pmalek:
### Problem statement

https://github.com/Kong/kong-operator/issues/1987 (through https://github.com/Kong/kong-operator/pull/2122) will introduce support for conversion webhook using cert-manager for certificates management.

This will be provided to users as an **opt in** mechanism.

By default the chart will generate self signed certs and use those for nice out of the box experience.

This issue tracks introduction of this mechanism

### Acceptance criteria

- [ ] As a user I can use the `kong-operator` chart without cert-manager and I'll get certificates provisioned for webhooks

programmer04:
waits for 
- https://github.com/Kong/kong-operator/issues/1987

PR #2141, opened 2025-08-27
chart: add self signed cert for webhook by default

pmalek:

This PR adds helm generated self signed certs for operator's conversion webhook.

This option is now the default, controlled via `global.conversionWebhook.certManager.enabled` (default: `false`).

It is recommended to enable the cert-manager integration in production environments.

**Which issue this PR fixes**

Part of https://github.com/Kong/kong-operator/issues/2136

In order to implement this, the certificate `Secret` template has been added in the KO CRDs template. This looks up the `Secret` by name in the cluster and if it doesn't exist it generates the CA and cert using `genCA` and `genSignedCert`.

This needs to be placed in the CRD manifest file as there is no way to use that logic in a separate template and e.g. use helm hooks to "wait" for the Secret to be deployed and only then use its data.

**PR Readiness Checklist**:


charts/kong-operator/values.yaml
```diff
@@ -116,3 +116,10 @@ global:
     # compatibility for those old ones - conversion is handled automatically.
     # If it's disabled, deprecated CRDs won't be installed and can't be applied.
     enabled: true
+    certManager:
+      # Enable cert-manager integration. Setting this to true will generate
+      # cert-manager's Certificate and Issuer objects for the conversion webhook Service.
+      # If disabled, this chart will generate a self-signed certificate and inject
+      # it into CRDs' conversion webhook caBundle.
+      # It is recommended to enable cert-manager integration for production use.
```
programmer04, line 129: Maybe mention that cert-manager has to be installed on the cluster when you're enabling it

hack/generators/conversion-webhook/main.go
```diff
@@ -92,7 +147,7 @@ const (
         service:
           name: %s
           namespace: %s
-          path: /convert
+          path: /convert%s
```
programmer04, line 155: Is this templating actually used anywhere?

hack/generators/conversion-webhook/main.go
```diff
@@ -146,25 +206,32 @@ func wrapCertAnnotations(content string) string {
...
 				`{{ template "kong.namespace" . }}`,
+				`
+{{if not .Values.global.conversionWebhook.certManager.enabled }}
+        caBundle: |
+          {{ $caCert | b64enc }}
+{{ end }}`,
 			),
 		),
 	)
+	os.WriteFile("/tmp/webhook-config.yaml", []byte(s), 0o644)
```
programmer04:
```suggestion
```

charts/kong-operator/values.yaml
pmalek: Done in 288992f0a102f6eb05a0e84dc286057257656988

hack/generators/conversion-webhook/main.go
pmalek: It's used here: https://github.com/Kong/kong-operator/blob/a2d25c389e06f49c09eeccb627cc4d3f9ea2a4c6/hack/generators/conversion-webhook/main.go#L221-L230

programmer04: I see, ok

programmer04 (commented):
Since 

```
ko-crds:
  # Installation of CRDs is performed. For multiple installations
  # of KO in cluster only for one it should be enabled.
  enabled: false
```

disables the whole `ko-crds` chart, it disables certification configuration too. Let's consider the situation. 

Firstly, install the operator with CRDs and keep them.
Next, uninstall the operator, keep CRDs.
Second installation without CRDs to avoid conflicts. 
That second installation will have a broken webhook
```log
{"level":"error","ts":"2025-08-28T14:35:01Z","msg":"failed to run manager","error":"problem running manager: open /tmp/k8s-webhook-server/serving-certs/tls.crt: no such file or directory","stacktrace":"main.main\n\t/workspace/cmd/main.go:42\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:285"}
```


As a solution, I may propose not to evaluate `ko-crds.enabled` on a Chart level, but add an if statement inside `charts/kong-operator/charts/ko-crds/templates/ko-crds.yaml` to exclude CRDs independently (basically wrap them in a huge `if` statement) from certs

pmalek:
> ko-crds/templates/ko-crds.yaml

Valid point. I adjusted the chart to that in f88343ade13e1e8c8ef57e76aa964252c12a1cc4.

I've also noticed the same issue with https://github.com/Kong/kong-operator/issues/2154 which covers the `ValidatingAdmissionPolicy` and `ValidatingAdmissionPolicyBinding`.

hack/generators/conversion-webhook/main.go
```diff
@@ -84,6 +85,65 @@ func main() {
 	fmt.Println("Successfully finished")
 }
 
+func wrapInIfEnabled(v string) string {
+	return fmt.Sprintf("{{- if .Values.enabled }}\n%s\n{{- end }}", v)
```
programmer04:
```suggestion
	return fmt.Sprintf("{{- if .Values.enabled }}\n%s\n{{- end }}\n", v)
```

to not have this mark at the end of diff 🙂 

Image

programmer04 (commented):
Furthermore, test a scenario

1. Fresh cluster
2. Run `helm install test-ko-1 charts/kong-operator --namespace test-infra --create-namespace` - everything works (also secret `kong-operator-ca` is created)
3. Run `helm uninstall test-ko-1 -n test-infra` CRDs retain (as expected) and `Secret` `kong-operator-ca` in `test-infra` too
4. In `values.yaml` set `ko-crds.enabled: false`
5. `helm install test-ko-2 charts/kong-operator --namespace test-infra --create-namespace`
6. KO doesn't boot in logs, you can see
```json
{"level":"error","ts":"2025-08-31T10:45:54Z","msg":"failed to run manager","error":"problem running manager: open /tmp/k8s-webhook-server/serving-certs/tls.crt: no such file or directory","stacktrace":"main.main\n\t/workspace/cmd/main.go:42\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:285"}
```

The above with `certManager.enabled: true` behaves in the same way.

Why does this happen? The expected `Secret` is not mounted due to `charts/kong-operator/templates/_helpers.tpl:132`.

Furthermore, even in the case of fixing the above on the CRD

```
    Webhook:
      Client Config:
        Ca Bundle:  ...
        Service:
          Name:       test-ko-1-kong-operator-webhook
          Namespace:  test-infra
          Path:       /convert
          Port:       443
```

svc name is created with `kong.kong.webhookServiceName` that does `{.Release.Name}-kong-operator-webhook`.

Do you think it should be hardcoded entirely here or maybe just documented?

---

Btw. other observation, but probably a little bit unrelated 

```
Name:         kong-operator-ca
Namespace:    test-infra
Labels:       konghq.com/secret=true
Annotations:  <none>

Type:  kubernetes.io/tls

Data
====
tls.crt:  648 bytes
tls.key:  227 bytes
```

The above `Secret` is retained even after uninstalling the chart, as you can see it lacks any Helm-related annotations. Do you know if this is expected?

pmalek:
> Furthermore, test a scenario
> 
> 1. Fresh cluster
> 2. Run `helm install test-ko-1 charts/kong-operator --namespace test-infra --create-namespace` - everything works (also secret `kong-operator-ca` is created)
> 3. Run `helm uninstall test-ko-1 -n test-infra` CRDs retain (as expected) and `Secret` `kong-operator-ca` in `test-infra` too
> 4. In `values.yaml` set `ko-crds.enabled: false`
> 5. `helm install test-ko-2 charts/kong-operator --namespace test-infra --create-namespace`
> 6. KO doesn't boot in logs, you can see
> 
> ```json
> {"level":"error","ts":"2025-08-31T10:45:54Z","msg":"failed to run manager","error":"problem running manager: open /tmp/k8s-webhook-server/serving-certs/tls.crt: no such file or directory","stacktrace":"main.main\n\t/workspace/cmd/main.go:42\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:285"}
> ```
> 
> The above with `certManager.enabled: true` behaves in the same way.
> 
> Why does this happen? The expected `Secret` is not mounted due to `charts/kong-operator/templates/_helpers.tpl:132`.
> 
> Furthermore, even in the case of fixing the above on the CRD
> 
> ```
>     Webhook:
>       Client Config:
>         Ca Bundle:  ...
>         Service:
>           Name:       test-ko-1-kong-operator-webhook
>           Namespace:  test-infra
>           Path:       /convert
>           Port:       443
> ```
> 
> svc name is created with `kong.kong.webhookServiceName` that does `{.Release.Name}-kong-operator-webhook`.
> 
> Do you think it should be hardcoded entirely here or maybe just documented?

That's well spotted. I think that we can fix it but this won't solve the problem of conversion webhook being embedded in CRDs. Instead of using a different helm release name, users could use a different namespace (which I believe is a more valid use case) and then the same failure  would happen: even if we hardcode the webhook name.

Do you think we should attempt it fix it for multiple helm release in a namespace scenario or rather document this together with the multi release in separate namespaces? I'm leaning towards treating these both the same way and just documenting the limitations (of conversion webhook being embedded in the CRD).

IMHO the conversion webhook will go away in a version or 2 which will make this whole problem go away. WDYT?

pmalek:
> Btw. other observation, but probably a little bit unrelated
> 
> ```
> Name:         kong-operator-ca
> Namespace:    test-infra
> Labels:       konghq.com/secret=true
> Annotations:  <none>
> 
> Type:  kubernetes.io/tls
> 
> Data
> ====
> tls.crt:  648 bytes
> tls.key:  227 bytes
> ```
> 
> The above `Secret` is retained even after uninstalling the chart, as you can see it lacks any Helm-related annotations. Do you know if this is expected?

This is indeed unrelated. This is [handled by the `caManager`](https://github.com/Kong/kong-operator/blob/cde9d5c8545d8ab08bee314aee08aacf64f0c6d4/modules/manager/run.go#L398-L439). Since we're already introducing certs management at this point we could think about removing `caManager` in favor of helm self signed certs and cert-manager managed certificates.

If you believe that would be a good direction I can create an issue for that.

programmer04:
For the first comment 
- https://github.com/Kong/kong-operator/pull/2141#issuecomment-3241327750

I think this rough edge is justifiable. We can't make it robust for every case in terms of service and namespace name. But what about such a workflow (KO per namespace)

1. `helm install test-ko-1 charts/kong-operator --namespace test-infra --create-namespace`
2. In `values.yaml` set `ko-crds.enabled: false`
3. `helm install test-ko-2 charts/kong-operator --namespace test-infra-2 --create-namespace`

The result is 

```
Error: INSTALLATION FAILED: Unable to continue with install: ValidatingAdmissionPolicy "ports.dataplane.gateway-operator.konghq.com" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-name" must equal "test-ko-2": current value is "test-ko-1"; annotation validation error: key "meta.helm.sh/release-namespace" must equal "test-infra-2": current value is "test-infra"
```

---

Yes, you're right that observation https://github.com/Kong/kong-operator/pull/2141#issuecomment-3241338409 is 💯 unrelated, but I think it's good to have an issue to investigate it further. We shouldn't leave any leftovers

pmalek:
> For the first comment
> 
> * [chart: add self signed cert for webhook by default #2141 (comment)](https://github.com/Kong/kong-operator/pull/2141#issuecomment-3241327750)
> 
> I think this rough edge is justifiable. We can't make it robust for every case in terms of service and namespace name. But what about such a workflow (KO per namespace)
> 
> 1. `helm install test-ko-1 charts/kong-operator --namespace test-infra --create-namespace`
> 2. In `values.yaml` set `ko-crds.enabled: false`
> 3. `helm install test-ko-2 charts/kong-operator --namespace test-infra-2 --create-namespace`
> 
> The result is
> 
> ```
> Error: INSTALLATION FAILED: Unable to continue with install: ValidatingAdmissionPolicy "ports.dataplane.gateway-operator.konghq.com" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-name" must equal "test-ko-2": current value is "test-ko-1"; annotation validation error: key "meta.helm.sh/release-namespace" must equal "test-infra-2": current value is "test-infra"
> ```

For this we already have https://github.com/Kong/kong-operator/issues/2154 (which you 👍d 🙃 )

---

> Yes, you're right that observation [#2141 (comment)](https://github.com/Kong/kong-operator/pull/2141#issuecomment-3241338409) is 💯 unrelated, but I think it's good to have an issue to investigate it further. We shouldn't leave any leftovers

Created https://github.com/Kong/kong-operator/issues/2156 for tracking this.

charts/kong-operator/CHANGELOG.md
```diff
@@ -1,5 +1,26 @@
...
+
+Moreover, `env.enable_conversion_webhook` must be set to `false` in all
+releases except the first one that enables it, otherwise operator deployments may fail.
+The operator deployment that has been installed as first, will handle the conversions.
+
+### Changes
+
+- Add support for helm generated self signed certs for conversion webhook
+  [#2141](https://github.com/Kong/kong-operator/pull/2141)
+
```
pmalek, line 23: @programmer04 PTAL.

programmer04 (approved):

PR #2143, opened 2025-08-27
feat: Gateway API hybrid mode underlying APIs enforcement

### LINKED ISSUE #2108: Create Generic reconciler enforcement logic
## Problem Statement

Create interfaces that declare the needed functions and aim at implementing the following functionalities:

- given an amount of expected resources:
  -  fetch them
  - delete duplicates
  - create the missing ones
  - return the result of the enforcement operation (error, operation, etc.)

## Acceptance Criteria

- [ ] A generic enforcement logic that can be implemented for specific types is added to the codebase

mlavacca:

Implementation of the generic logic to enforce in the cluster objects previously translated by the Gateway API full hybrid converter.

**Which issue this PR fixes**

Fixes #2108 

**PR Readiness Checklist**:


controller/fullhybrid/converter/converter.go
```diff
@@ -4,36 +4,37 @@ import (
...
+	SetRootObject(obj t)
+	// LoadStore loads and initializes any required internal store or cache, using the provided context.
+	LoadStore(ctx context.Context) error
+	// Translate performs the conversion or translation logic for the root object, returning an error if the process fails.
+	Translate() error
+	// GetStore returns a slice of unstructured.Unstructured objects representing the current state of the store, using the provided context.
+	GetStore(ctx context.Context) []unstructured.Unstructured
+	// Reduce returns a slice of utils.ReduceFunc functions that can be applied to the given unstructured.Unstructured object for reduction.
+	Reduce(obj unstructured.Unstructured) []utils.ReduceFunc
+	// ListExistingObjects lists all existing unstructured.Unstructured objects relevant to the converter, using the provided context, and returns them along with any error encountered.
```
fgiudici: We use Objects both here and in [Get,Set]RootObject. Here we mean the "destination" object (e.g., KongService as per the dummy implementation) while with RootObject we meant the "source" object (e.g., the Service).
Instead of "relevant to the converter"  I would consider a more straightforward comment, something like:

```suggestion
	// ListExistingObjects lists all existing unstructured.Unstructured objects of the destination API kind, using the provided context, and returns them along with any error encountered.
```

controller/fullhybrid/converter/converter.go
```diff
@@ -4,36 +4,37 @@ import (
...
+	// GetRootObject returns the current root object of type t.
+	GetRootObject() t
+	// SetRootObject sets the root object to the provided value of type t.
+	SetRootObject(obj t)
+	// LoadStore loads and initializes any required internal store or cache, using the provided context.
+	LoadStore(ctx context.Context) error
+	// Translate performs the conversion or translation logic for the root object, returning an error if the process fails.
+	Translate() error
+	// GetStore returns a slice of unstructured.Unstructured objects representing the current state of the store, using the provided context.
+	GetStore(ctx context.Context) []unstructured.Unstructured
```
fgiudici: I found the LoadStore and GetStore names a bit misleading as they refer to two different kind of stores: the former "Loads" "source" data while the latter returns the "output" store, which contains "Translate(d)" data.
To be more clear, what bothers me is that once one calls LoadStore() and data is loaded the GetStore() returns nothing. One would need to call  Translate() first.
So, what about making the "Translate()" operation implicit (performed during the LoadStore()) or change the LoadStore() name to something like "InitStore()" or "LoadCache()" ?

controller/fullhybrid/reconciler_utils.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	resourcesToDelete := make([]unstructured.Unstructured, 0)
+	for _, v := range ownedResourceMap {
+		// mark for deletion all the resources with no hits.
+		if v.hits == 0 {
+			resourcesToDelete = append(resourcesToDelete, v.resources...)
+		}
+		// mark for deletion all the resources with duplicates
+		reduceDuplicates(v.resources, conv.Reduce(v.resources[0])...)
```
fgiudici:
```suggestion
		resourcesToDelete = append(resourcesToDelete, reduceDuplicates(v.resources, conv.Reduce(v.resources[0])...)...)
```

controller/fullhybrid/converter/converter.go
```diff
@@ -4,36 +4,37 @@ import (
...
+	// Reduce returns a slice of utils.ReduceFunc functions that can be applied to the given unstructured.Unstructured object for reduction.
```
fgiudici: I would add an hint that it is needed to drop the duplicates, something like:
```suggestion
	// Reduce returns a slice of utils.ReduceFunc functions that can be applied to the given unstructured.Unstructured object to get a list of duplicates to be removed.
```

fgiudici (commented):
Nice PR, great step forward! Interesting the use of the unstructured type.

controller/fullhybrid/reconciler_utils.go
mlavacca: Nice catch!

controller/fullhybrid/converter/converter.go
mlavacca: Yes, that makes total sense! I agree that having the `GetOutput` include the `Translate` is the best solution, but I am not sure we can do it straight away. First, I need to write the controller logic, and eventually, we can streamline the whole interface it if feasible. In the meantime, I changed the functions' names. LMKWYT.

fgiudici (approved):
Thanks for the changes, looks good here!

PR #2153, opened 2025-08-28
Add `KonnectCloudDataPlaneGroupConfig` admission policy

### LINKED ISSUE #1779: Konnect entities: migrate away from Cloud Gateways DataPlaneGroup `ConfigurationDataPlaneGroupAutoscaleStatic`
### Problem statement

When upgrading https://github.com/Kong/sdk-konnect-go/ to [v0.4.0](https://github.com/Kong/sdk-konnect-go/releases/tag/v0.4.0) `sdkkonnectcomp.ConfigurationDataPlaneGroupAutoscaleStatic` appears to have been deprecated:

```
controller/konnect/ops/ops_konnectdataplanegroupconfigurations.go:266:49: SA1019: sdkkonnectcomp.ConfigurationDataPlaneGroupAutoscaleStatic is deprecated: This will be removed in a future release, please migrate away from it as soon as possible. (staticcheck)
                        ConfigurationDataPlaneGroupAutoscaleStatic: &sdkkonnectcomp.ConfigurationDataPlaneGroupAutoscaleStatic{
                                                                     ^
```

This issue covers handling this deprecation.

fgiudici:

Emit a warning when the `dataplane_groups.autoscale.type` is set to the deprecated `static` value.
Currently, deprecation warnings at field level require a validating admission policy
(see https://github.com/kubernetes/kubernetes/issues/131817).

**Which issue this PR fixes**

Fixes #1779 

**PR Readiness Checklist**:


charts/kong-operator/templates/validation-policy-konnect.yaml
```diff
@@ -0,0 +1,36 @@
+{{- if (semverCompare ">= 1.5.0" (include "kong.effectiveVersion" .)) }}
```
pmalek: This is incorrect (leftover from porting the https://github.com/Kong/charts/tree/main/charts/gateway-operator).

This should be removed.

```suggestion
```

charts/kong-operator/templates/validation-policy-konnect.yaml
```diff
@@ -0,0 +1,36 @@
...
+---
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingAdmissionPolicyBinding
+metadata:
+  name: "biding-autoscale.dataplanegroups.konnectclouddataplanegroupconfig.konnect.konghq.com"
+spec:
+  policyName: "autoscale.dataplanegroups.konnectclouddataplanegroupconfig.konnect.konghq.com"
+  validationActions: [Warn,Audit]
+{{- end -}}
+{{- end -}}
```
pmalek:
```suggestion
```

pmalek (changes requested):
We will need an option in `values.yaml` to control this so that users can deploy more than 1 helm release and do not get a conflict.

See related issue: https://github.com/Kong/kong-operator/issues/2154

pmalek (commented):
I'm wondering if we should rethink the CRD validation tests for VAPs.

Currently we have https://github.com/Kong/kong-operator/blob/85e364f9bcd34560feb9e966f77b14cc439e2c1e/test/crdsvalidation/dataplane_validatingadmissionpolicy_test.go with the ports VAP being installed with `kustomize` from https://github.com/Kong/kong-operator/blob/f2e367d6766b8653c2ade9195b489f40cb947cfb/config/default/validating_policies/kustomization.yaml.

I think we should unify everything to use the chart. WDYT? 

Bringing this up because I thought it would be worth to add a CRD validation test against the added VAP.

fgiudici:
> I'm wondering if we should rethink the CRD validation tests for VAPs.
> 
> Currently we have https://github.com/Kong/kong-operator/blob/85e364f9bcd34560feb9e966f77b14cc439e2c1e/test/crdsvalidation/dataplane_validatingadmissionpolicy_test.go with the ports VAP being installed with `kustomize` from https://github.com/Kong/kong-operator/blob/f2e367d6766b8653c2ade9195b489f40cb947cfb/config/default/validating_policies/kustomization.yaml.
> 
> I think we should unify everything to use the chart. WDYT?
> 
> Bringing this up because I thought it would be worth to add a CRD validation test against the added VAP.

Now the VAP tests use the templated resources from the chart.
Our testcase suite does not allow to check/catch warnings, so for the Konnect VAP added some code directly in the test to manage that.
The right way to address it is to extend our testcase suite with an "ExpectWarningMessage" functionality and then leverage that one: something to be addressed in a separate PR on kong/kubernetes-config (https://github.com/Kong/kubernetes-configuration/pull/602).

pmalek:
> Now the VAP tests use the templated resources from the chart. Our testcase suite does not allow to check/catch warnings, so for the Konnect VAP added some code directly in the test to manage that. The right way to address it is to extend our testcase suite with an "ExpectWarningMessage" functionality and then leverage that one: something to be addressed in a separate PR on kong/kubernetes-config ([Kong/kubernetes-configuration#602](https://github.com/Kong/kubernetes-configuration/pull/602)).

This looks nice, I wasn't aware of that feature 👍 

Nonetheless, since the VAP is added in chart's templates, we'd need to use the warning handler in KO repo in a test suite that I believe doesn't exist yet (install chart and then check errors + **warnings** after applying custom resources), right? So what you're proposing in kcfg#602 I think we'd want to add in KO, otherwise we can't add test the VAP.

fgiudici:
> > Now the VAP tests use the templated resources from the chart. Our testcase suite does not allow to check/catch warnings, so for the Konnect VAP added some code directly in the test to manage that. The right way to address it is to extend our testcase suite with an "ExpectWarningMessage" functionality and then leverage that one: something to be addressed in a separate PR on kong/kubernetes-config ([Kong/kubernetes-configuration#602](https://github.com/Kong/kubernetes-configuration/pull/602)).
> 
> This looks nice, I wasn't aware of that feature 👍
> 
> Nonetheless, since the VAP is added in chart's templates, we'd need to use the warning handler in KO repo in a test suite that I believe doesn't exist yet (install chart and then check errors + **warnings** after applying custom resources), right? So what you're proposing in kcfg#602 I think we'd want to add in KO, otherwise we can't add test the VAP.

Yep, the right way is to have kcfg#602 in first and than use that in KO (we import the test suite for the chart from there: https://github.com/Kong/kong-operator/blob/main/test/crdsvalidation/dataplane_validatingadmissionpolicy_test.go#L13).
In this PR I added the logic to do the check right after the test suite is run but the test is flaky and keeps failing here: my guess is that the test suite ends too quick for the later check to capture the warning (it should be done inside the test suite).

So going to remove the test of the `KonnectCloudDataPlaneGroupConfig` admission policy as it is right now.
I would start merging this PR without the check on the warning message of the KonnectCloudDataPlaneGroupConfig and in the meanwhile test and get kcfg#602 in. At that point I would open a new PR to add the test for `KonnectCloudDataPlaneGroupConfig` that will be straightforward.
Otherwise it  could wait here till kcfg PR is in. WDYT?

pmalek:
> Yep, the right way is to have kcfg#602 in first and than use that in KO (we import the test suite for the chart from there: https://github.com/Kong/kong-operator/blob/main/test/crdsvalidation/dataplane_validatingadmissionpolicy_test.go#L13). In this PR I added the logic to do the check right after the test suite is run but the test is flaky and keeps failing here: my guess is that the test suite ends too quick for the later check to capture the warning (it should be done inside the test suite).
> 
> So going to remove the test of the `KonnectCloudDataPlaneGroupConfig` admission policy as it is right now. I would start merging this PR without the check on the warning message of the KonnectCloudDataPlaneGroupConfig and in the meanwhile test and get kcfg#602 in. At that point I would open a new PR to add the test for `KonnectCloudDataPlaneGroupConfig` that will be straightforward. Otherwise it could wait here till kcfg PR is in. WDYT?

👍 Let's merge the 602 first and then handle it here properly with the test.

fgiudici:
~Needs a rebase after https://github.com/Kong/kubernetes-configuration/pull/602 is merged and KO repo imports an~ ~updated kconf release to implement the test dependencies.~ _(done)_

test/helpers/apply/apply.go
```diff
@@ -30,8 +32,44 @@ import (
 
 var decUnstructured = yaml.NewDecodingSerializer(unstructured.UnstructuredJSONScheme)
 
-// Apply applies a kustomization to the cluster using the given rest.Config.
-func Apply(ctx context.Context, t *testing.T, cfg *rest.Config, dir string) {
+// Template applies templated resources to the cluster using the given rest.Config.
+func Template(t *testing.T, cfg *rest.Config, chartPath string, templates []string) {
```
pmalek: I _think_ the change in these package is OK, going from tool oriented to verb oriented but I'm concerned that down the line when we want to add new actions we'd require a new package vs in the old package structure, we'd only need a new func. WDYT?

I believe though that in this context we'd want to rename this function and `templateRenderer` below so that they clearly mention they should be use with helm. WDYT?

test/helpers/apply/apply.go
```diff
@@ -30,8 +32,44 @@ import (
...
+	require.NoError(t, err)
+	for _, r := range res {
+		t.Logf("Result: %s", r)
+	}
+}
+
+func templateRenderer(t *testing.T, chartPath string, templates []string, helmArgs ...string) string {
+	t.Helper()
+	t2 := &testing.T{}
+	t2.Log("\nT2\n")
```
pmalek: ?

```suggestion
```

fgiudici: 🤦🏼  nice catch, a leftover

test/helpers/apply/apply.go
fgiudici: > I _think_ the change in these package is OK, going from tool oriented to verb oriented but I'm concerned that down the line when we want to add new actions we'd require a new package vs in the old package structure, we'd only need a new func. WDYT?
> 

So, the idea here was to render the template in order to use any template resource available from the chart.
I was aiming at a new package but then I realized that the Kustomization package had already all what was needed (apply(), readYaml(), buildDynamicResourceClient()) to apply the resource, only the template rendering was missing. 
Moreover, the "old" kustomize.Apply() function (renamed "Kustomization()") is no more used but it could be handy to be kept around.
This is why instead of creating a new package (and copying quite some code from kustomize) or renaming the kustomize one (and lose the kustomize.Apply()) I came up with this name-neutral package. 

I would be also in favor to split that change in 3 packages: the 3 common functions with the apply() exported (which is the one used by Kustomization and the Template one), keeping the kustomize package and adding a new one for the template rendering.
What would you prefer? (EDIT: pushed  a commit on top splitting in the 3 packages)

> I believe though that in this context we'd want to rename this function and `templateRenderer` below so that they clearly mention they should be use with helm. WDYT?

Something like helmRenderer() ? Ok (EDIT: now the function is helm.ApplyTemplate())

pmalek: > I would be also in favor to split that change in 3 packages: the 3 common functions with the apply() exported (which is the one used by Kustomization and the Template one), keeping the kustomize package and adding a new one for the template rendering.
What would you prefer?

That could work 👍 Be wary of naming packages common or util though :) 

> Something like helmRenderer() ? Ok

This would idiomatically match an interface name in Go. I'd think more of something like `renderTemplate` (assuming the package is called `helm`)

test/crdsvalidation/dataplane_validatingadmissionpolicy_test.go
```diff
@@ -1,32 +1,104 @@
...
+	"github.com/kong/kong-operator/test/helpers/helm"
 )
 
 const (
-	// KustomizePathValidatingPolicies is the path to the Kustomize directory containing the validation policies.
-	KustomizePathValidatingPolicies = "config/default/validating_policies/"
+	// ChartPath is the path relative to the project to the kong-operator chart
+	ChartPath = "charts/kong-operator"
+
+	// ValidationPolicy_ contain validation policies path relative to the kong-operator chart
```
pmalek: Let's make this so that each `const` have its own go doc.

test/crdsvalidation/dataplane_validatingadmissionpolicy_test.go
```diff
@@ -1,32 +1,104 @@
...
+	ValidationPolicyDataplane = "templates/validation-policy-dataplane.yaml"
```
pmalek, line 33: This seems unused. I've checked out the repo on this branch and can't see any references to this

```suggestion
```

pmalek (commented):
Looks good, 2 minor nits, other than that we're ready to merge.

test/crdsvalidation/dataplane_validatingadmissionpolicy_test.go
fgiudici: Oh, right... I meant to use the const in the test but replicated the string.
Good catch! Now we use the const string in the test.

pmalek (approved):

PR #2172, opened 2025-09-02
Chore(ci): Add cleanup job for removing leftover Konnect control planes

randmonkey:

Add a cleanup workflow to remove the leftover control planes. 
Standard of cleanup:

- Created in KO's integration tests (has the label `operator-test-id`) 
- Created more than 1 hour before

Also deletes the roles related to the deleted CPs.

**Which issue this PR fixes**

Fixes #

Basically copied from the cleanup code and workflow file in KIC. Removed the cleanup of the `Role`s in Konnect.

**PR Readiness Checklist**:

### DIFF
--- a/.github/workflows/cleanup.yaml
+++ b/.github/workflows/cleanup.yaml
@@ -2,7 +2,7 @@ name: cleanup
 
 on:
   schedule:
-    - cron: '*/30 * * * *'
+    - cron: '8 */1 * * *' # Run once per hour
   workflow_dispatch: {}
 
 permissions:
--- a/.github/workflows/cleanup.yaml
+++ b/.github/workflows/cleanup.yaml
@@ -1,0 +1,26 @@
+name: cleanup
+
+on:
+  schedule:
+    - cron: '*/30 * * * *'
+  workflow_dispatch: {}
+
+permissions:
+  contents: read
+
+jobs:
+  konnect:
+    timeout-minutes: ${{ fromJSON(vars.GHA_DEFAULT_TIMEOUT) }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - name: setup golang
+        uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # v5.5.0
+        with:
+          go-version-file: go.mod
+      - name: cleanup orphaned test clusters
+        run: go run ./hack/cleanup konnect
+        env:
+          KONG_TEST_KONNECT_ACCESS_TOKEN: ${{ secrets.KONG_TEST_KONNECT_ACCESS_TOKEN }}
+          KONG_TEST_KONNECT_SERVER_URL: us.api.konghq.tech
--- a/.github/workflows/cleanup.yaml
+++ b/.github/workflows/cleanup.yaml
@@ -24,3 +24,5 @@ jobs:
         env:
           KONG_TEST_KONNECT_ACCESS_TOKEN: ${{ secrets.KONG_TEST_KONNECT_ACCESS_TOKEN }}
           KONG_TEST_KONNECT_SERVER_URL: us.api.konghq.tech
+          # Ref: https://github.com/Kong/sdk-konnect-go/issues/20
+          KONG_CUSTOM_DOMAIN: konghq.tech
--- a/hack/cleanup/konnect_control_planes.go
+++ b/hack/cleanup/konnect_control_planes.go
@@ -1,0 +1,206 @@
+package main
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"time"
+
+	sdkkonnectgo "github.com/Kong/sdk-konnect-go"
+	sdkkonnectcomp "github.com/Kong/sdk-konnect-go/models/components"
+	sdkkonnectops "github.com/Kong/sdk-konnect-go/models/operations"
+	"github.com/go-logr/logr"
+	"github.com/samber/lo"
+
+	"github.com/kong/kong-operator/controller/konnect/ops"
+	"github.com/kong/kong-operator/test"
+)
+
+const (
+	konnectControlPlanesLimit     = int64(100)
+	timeUntilControlPlaneOrphaned = time.Hour
+
+	kindKonnectGatewayControlPlane = "KonnectGatewayControlPlane"
+)
+
+// cleanupKonnectControlPlanes deletes orphaned control planes created by the tests and their roles.
+func cleanupKonnectControlPlanes(ctx context.Context, log logr.Logger) error {
+	// NOTE: The domain for global endpoints is overridden in cleanup.yaml workflow.
+	// See https://github.com/Kong/sdk-konnect-go/issues/20 for details
+	sdk := sdkkonnectgo.New(
+		sdkkonnectgo.WithSecurity(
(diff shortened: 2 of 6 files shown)


hack/cleanup/konnect_control_planes.go
```diff
@@ -0,0 +1,133 @@
...
+		body, err := io.ReadAll(response.RawResponse.Body)
+		if err != nil {
+			body = []byte(err.Error())
+		}
+		return nil, fmt.Errorf("failed to list control planes, status: %d, body: %s", response.GetStatusCode(), body)
+	}
+
+	var orphanedControlPlanes []string
+	for _, ControlPlane := range response.ListControlPlanesResponse.Data {
+		if ControlPlane.Labels[ops.KubernetesKindLabelKey] != kindKonnectGatewayControlPlane {
```
randmonkey: REVIEW: should we add more checks to check whether the CP is created in integration tests? Do we have some labels specifically used in the tests? (or should we add such labels?)

hack/cleanup/konnect_control_planes.go
```diff
@@ -0,0 +1,133 @@
...
+	kindKonnectGatewayControlPlane = "KonnectGatewayControlPlane"
+)
+
+// cleanupKonnectControlPlanes deletes orphaned control planes created by the tests and their roles.
+func cleanupKonnectControlPlanes(ctx context.Context, log logr.Logger) error {
+	serverURL, err := canonicalizedServerURL()
+	if err != nil {
+		return fmt.Errorf("invalid server URL %s: %w", test.KonnectServerURL(), err)
+	}
+
```
randmonkey, line 34: REVIEW: We removed the check and delete of the roles here. Do we still need them?

.github/workflows/cleanup.yaml
```diff
@@ -0,0 +1,26 @@
+name: cleanup
+
+on:
+  schedule:
+    - cron: '*/30 * * * *'
```
randmonkey: REVIEW: This is copied from KIC. Should we run the cleanup once half an hour?

hack/cleanup/konnect_control_planes.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+// findOrphanedControlPlanes finds control planes that were created by the tests and are older than timeUntilControlPlaneOrphaned.
+func findOrphanedControlPlanes(
+	ctx context.Context,
+	log logr.Logger,
+	c *sdkkonnectgo.ControlPlanes,
+) ([]string, error) {
+	response, err := c.ListControlPlanes(ctx, sdkkonnectops.ListControlPlanesRequest{
+		PageSize: lo.ToPtr(konnectControlPlanesLimit),
```
randmonkey, line 102: REVIEW: Could we add label filter here? Is there docs to tell how to use the options in the list request?

hack/cleanup/konnect_control_planes.go
pmalek: Let's add those 👍 sdk-konnect-go uses labels like : https://github.com/Kong/sdk-konnect-go/blob/0deeb38e6d822b55ee2e0d4e705f484a7f314948/test/integration/funcs.go#L17-L23.

Let's add something similar to ensure that these particular CPs are from KO's tests.

.github/workflows/cleanup.yaml
pmalek: Once an hour I believe should be plenty enough.

hack/cleanup/konnect_control_planes.go
pmalek: You can find the API docs here https://developer.konghq.com/api/konnect/control-planes/v2/#/operations/list-control-planes and use SDK's

https://github.com/Kong/sdk-konnect-go/blob/971395bf6f415948b69846a1de94b9a03afb9fe2/models/operations/listcontrolplanes.go#L17-L18

hack/cleanup/konnect_control_planes.go
pmalek: ~AFAIR this has been fixed in Konnect and roles are deleted on CP deletion, that was the reason for removing this.~

I can still see KIC's job handling role deletion https://github.com/Kong/kubernetes-ingress-controller/blame/c29a8c84afdb1d6a04b93789b6a0faf8f9cf24e3/hack/cleanup/konnect_control_planes.go#L51-L58 so I believe we need to do that as well here.

randmonkey: ~I tried locally and the same token succeeded to list and delete CPs but failed to get user info from `GetUsersMe` with `401` response. I already asked konnect team in slack and I think we could temporarily disable them and track it in a new issue.~
Added the deletion of roles back and set the `KONG_CUSTOM_DOMAIN` in the workflow.

hack/cleanup/konnect_control_planes.go
randmonkey: ~We already added label `konghq.com/test-id` in tests. I added the filter in the listing of cps.~ 
The label is in the k8s resource instead of on the Konnect CP. Konnect does not support label keys starting with `kong`. I'll add another one.

pmalek: Let's add something like `kong-operator:true` or `konghq.com/kong-operator:true` or `repo:github.com/kong/kong-operator` to indicate which tests this CP is coming from, WDYT?

hack/cleanup/konnect_control_planes.go
randmonkey: Added the label filter to filter CPs created in KO tests.

.github/workflows/cleanup.yaml
randmonkey: Done.

hack/cleanup/konnect_control_planes.go
randmonkey: Added a label `operator-test-id` in integration tests.

test/integration/konnect_entities_test.go
```diff
@@ -57,6 +57,7 @@ func TestKonnectEntities(t *testing.T) {
 
 	cp := deploy.KonnectGatewayControlPlane(t, GetCtx(), clientNamespaced, authCfg,
 		deploy.WithTestIDLabel(testID),
+		deploy.KonnectGatewayControlPlaneLabel(deploy.KonnectTestIDLabel, testID),
```
pmalek, line 60: I think that we're duplicating the information here, the same `testID` is set with `WithTestIDLabel` (label name there is `konghq.com/test-id`, and the values is `testID` as well).

How about we use different key value pair? For example `konghq.com/kong-operator` with value `true`?

randmonkey: This is the label added to the created Konnect control plane but not in the k8s CR's labels. Moreover, Konnect control planes does not allow label keys starting with `kong`.
Added more comments on the `KonnectGatewayControlPlaneLabel`.

pmalek: But `WithTestIDLabel` already adds a label `konghq.com/test-id` so that has to be allowed.

randmonkey: > But `WithTestIDLabel` already adds a label `konghq.com/test-id` so that has to be allowed.

No. The `WithTestIDLabel` only adds the label in k8s metadata but not in Konnect CP. The labels in k8s metadata are not added to the CP. Only the labels in `spec.createCotnrolPlaneRequest.labels` are added to the Konnect CP.
 
Also the `WithTestIDLabel` is applied on all the Konnect entities but not only CPs.

pmalek (approved):

PR #2227, opened 2025-09-08
chore: delete old controlplane managed resources

pmalek:

This PR introduces a clean up step in `ControlPlane` reconciliation where old managed resources are deleted (managed by KGO 1.6 and older).

**Which issue this PR fixes**

Fixes #

This will be removed as part of https://github.com/Kong/kong-operator/issues/2228.

Tentatively added to KO 2.2 milestone.

**PR Readiness Checklist**:

### DIFF
--- a/controller/controlplane/controller.go
+++ b/controller/controlplane/controller.go
@@ -10,10 +10,14 @@ import (
 
 	"github.com/go-logr/logr"
 	"github.com/samber/lo"
+	admregv1 "k8s.io/api/admissionregistration/v1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/rest"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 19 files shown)


controller/controlplane/controller.go
```diff
@@ -130,6 +134,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
+	if err := r.cleanupOldManagedResources(ctx, cp); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to cleanup old managed resources: %w", err)
+	}
+
```
alacuku, line 140: Should we run the cleanup only once?

pmalek: We could add an annotation after a successful cleanup but that would add complexity and the client calls are using cache anyway. Do you think that would be worth the hassle?

alacuku: Why not tracking the state in the reconciler? A simple bool should do the trick and avoid to run the cleanup at every reconcile.

pmalek: This would be a map of CP namespaced names to bools though right? 

And we'd run the cleanup first time the operator boots up anyway right?

alacuku: That's right, but if you feel that's overkill then feel free to ignore me.

pmalek: Added the annotation handling, it's not that much work/logic anyway.

PTAL : 2c3f4877d414f82e029e23a484a4ab7765a7f648

alacuku (dismissed):

controller/controlplane/controller.go
```diff
@@ -818,3 +826,106 @@ func controlPlaneStatusEqual(
...
+	var networkPolicy networkingv1.NetworkPolicy
+	err = r.DeleteAllOf(ctx, &networkPolicy, &client.DeleteAllOfOptions{
+		ListOptions: client.ListOptions{
+			Namespace: cp.Namespace,
+			LabelSelector: labels.SelectorFromSet(labels.Set{
+				consts.GatewayOperatorManagedByLabel: "gateway",
+			}),
+		},
+	})
```
pmalek, line 868: This could also delete the `NetworkPolicy` that is managed by the KO as it currently is being done in https://github.com/Kong/kong-operator/blob/409bb080eefd9c6a2701267f8836fabd81917d45/controller/gateway/controller_reconciler_utils.go#L296-L340 but 

- since we've added the annotation to only perform the clean up once this is safe
- the `NetworkPolicy` is being enforced actively anyway in the linked snippet so we're OK

PR #2280, opened 2025-09-12
fix(tests): adjust address of konghq.tech

programmer04:

This PR resolves the problem with CI

image

It's changed, and we do not have control over this address, but it's still better to have this assertion than not (even taking into account that it may break) to ensure that URLs are set as expected by the controller.

### DIFF
--- a/config/samples/dataplane-konnect.yaml
+++ b/config/samples/dataplane-konnect.yaml
@@ -44,13 +44,13 @@ spec:
             - name: KONG_CLUSTER_MTLS
               value: pki
             - name: KONG_CLUSTER_CONTROL_PLANE
-              value: <YOUR_CP_ID>.cp0.konghq.tech:443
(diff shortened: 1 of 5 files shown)


pmalek:
Is there a related conversion somewhere (slack or otherwise)?

programmer04:
Nope, I've just discovered it and I'm trying to fix it

config/samples/dataplane-konnect.yaml
pmalek, line 1: 👍

pmalek (approved):

pmalek:
This will also be needed in other branches that we still support, e.g. 1.6, it has failed in https://github.com/Kong/kong-operator/actions/runs/17672414362/job/50228323377?pr=2279#step:5:8264.

Are you back porting this to 1.6?

PR #2289, opened 2025-09-15
fix: `nil` dereference and setting status in `EnforceStatus`

programmer04:

The condition could dereference a `nil` pointer. It verifies only whether both are not `nil`, but still one of them can be `nil`, and the next part of the alternative is checked, which may lead to a panic. See the test

image

[src](https://github.com/Kong/kong-operator/actions/runs/17731865551/job/50384577499?pr=2287#step:7:11539)

Adding respective unit tests allowed discovering that the status is incorrectly set for uninitiated, which is fixed in this PR too.

### DIFF
--- a/controller/hybridgateway/route/httproute_test.go
+++ b/controller/hybridgateway/route/httproute_test.go
@@ -1,0 +1,423 @@
+package route
+
+import (
+	"context"
+	"testing"
+
+	"github.com/go-logr/logr"
+	"github.com/samber/lo"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	"github.com/kong/kong-operator/controller/pkg/op"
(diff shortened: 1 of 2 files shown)


pmalek (commented):
Can we add a UT for this? 

I can see that the test file is missing for this so I think it's a good  moment to add it, WDYT?

programmer04:
@pmalek PTAL, adding a unit test allowed discovery of another issue which is fixed too (description of the PR is updated accordingly)

controller/hybridgateway/route/httproute_test.go
```diff
@@ -0,0 +1,424 @@
...
+												Name: "test-service",
+											},
+										},
+									},
+								},
+							},
+						},
+					},
+				}
+				// Set up existing parent status with non-nil namespace to trigger the panic
```
pmalek: I'm not sure I follow: we're setting this to trigger panic but does the panic actually happen? Where should it be caught?

programmer04: Sorry, it's a leftover that used to trigger panic, but PR includes a fix. I'm removing

pmalek (approved):
👍 

cc: @fgiudici

PR #2323, opened 2025-09-22
chore(tests): proper conformance report for KO

### LINKED ISSUE #1872: Merge conformance test suites into a single one
## Problem Statement

Currently KO conformance tests with bootstrapping code are in `./test/conformance/...` and  `ingress-controller` (old KIC) are in `./ingress-controller/test/conformance/...`. Target `make test.conformance` runs them sequentially, but they should be merged into one

## Acceptance Criteria
- [ ] one setup and teardown code
- [ ] one place in the repo
- [ ] single report generated (unblocks #131)

programmer04:

This PR ensures that only conformance tests are run for KO and the maximum number of tests is enabled. Reports are generated and saved as an artifact with `.yaml` file name that matches [what is in the GWAPI repo](https://github.com/kubernetes-sigs/gateway-api/tree/main/conformance/reports/v1.2.1/kong-kubernetes-ingress-controller). 

Each disabled test for the expression router has a respective issue in the KO repo.

**Which issue this PR fixes**

Closes https://github.com/Kong/kong-operator/issues/1872


test/envvars.go
```diff
@@ -75,3 +75,30 @@ func IsTelepresenceDisabled() bool {
...
+	return keepTestCluster
+}
+
+// IsCI indicates whether or not the tests are running in a CI environment.
+func IsCI() bool {
+	// It's a common convention that e.g. GitHub, GitLab, and other CI providers
+	// set the CI environment variable.
+	envVar := strings.ToLower(os.Getenv("CI"))
+	isCI := envVar == "true" || envVar == "1"
+	fmt.Printf("INFO: running in CI: %t\n", isCI)
```
pmalek, line 95: Can we pass `*testing.T` here and use it for logging?

test/envvars.go
```diff
@@ -75,3 +75,30 @@ func IsTelepresenceDisabled() bool {
...
 	return ret
 }
+
+// KeepTestCluster indicates whether the caller wants the cluster created by the test suite
+// to persist after the test for inspection. This has a nil effect when an existing cluster
+// is provided, as cleanup is not performed for existing clusters.
+func KeepTestCluster() bool {
+	envVar := strings.ToLower(os.Getenv("KONG_TEST_CLUSTER_PERSIST"))
+	keepTestCluster := envVar == "true" || envVar == "1"
+	fmt.Printf("INFO: keeping test cluster after tests: %t\n", keepTestCluster)
```
pmalek, line 85: Ditto about logging with `*testing.T`.

test/envvars.go
```diff
@@ -75,3 +75,30 @@ func IsTelepresenceDisabled() bool {
 	}
 	return ret
 }
+
+// KeepTestCluster indicates whether the caller wants the cluster created by the test suite
+// to persist after the test for inspection. This has a nil effect when an existing cluster
+// is provided, as cleanup is not performed for existing clusters.
```
pmalek:
```suggestion
// to persist after the test for inspection. This has no effect when an existing cluster
// is provided, as cleanup is not performed for existing clusters.
```

test/envvars.go
programmer04: Both `KeepTestCluster` and `IsCI`  are used in `SkipCleanup`, which is used in `TestMain` https://github.com/Kong/kong-operator/blob/6386170ed47099ce0b5cf7652b67beefdc3b15bb/test/conformance/suite_test.go#L51

and unfortunately, there is no `*testing.T` to pass. Do you have any other ideas on how to address it? Or we're leaving it as is

pmalek: WDYT about accepting an interface with the Log function and making a tiny wrapper on testing.M which would print using fmt.Printf ?

programmer04: Only if you insist IMHO :D They are currently used only in TestMain, so it's coding for the future

pmalek: ;) Alright. Let's do it the proposed way "for now" :)

.github/workflows/tests.yaml
```diff
@@ -512,7 +512,7 @@ jobs:
       uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
       with:
         name: conformance-report-${{ matrix.router-flavor }}
-        path: standard-*-report.yaml
+        path: experimental-*-report.yaml
```
pmalek, line 515: Why is this changing? I don't see the experimental suite being explicitly set anywhere 🤔

programmer04: Hmmm, there is a stanza

```yml
gatewayAPIChannel: experimental
```

in a generated report and we used to have it like that for KIC, see https://github.com/kubernetes-sigs/gateway-api/tree/main/conformance/reports/v1.2.1/kong-kubernetes-ingress-controller

pmalek: Ok, this look like its deduced from the installed CRDs

https://github.com/kubernetes-sigs/gateway-api/blob/c26512f2207d7c6fd0c12bf6f98ff10bed50b05f/conformance/utils/suite/suite.go#L666-L695 

Let's leave it as proposed.

pmalek (approved):

PR #2199, opened 2025-09-04
chore(refactor): remove obsolete KGO's KIC's webhook config generator

programmer04:
**What this PR does / why we need it**:

In the KGO KIC's webhook used to be bootstrapped differently. Therefore, a dedicated script for generating the respective configuration was needed - `hack/generators/kic/webhook-config-generator`. It is unused in KO, hence this PR removes it entirely without affecting the setup.

Adjustments in generation scritps are done to satisfy `golangci-lint`

**Which issue this PR fixes**

closes https://github.com/Kong/kong-operator/issues/1733

PR #2354, opened 2025-09-25
chore: remove redundant go.mod & go.sum, fix odd number of args in logs

programmer04:

PR with dependency bump
- https://github.com/Kong/kong-operator/pull/2206

by mistake reintroduced `hack/generators/go.mod` and `hack/generators/go.sum`, which were removed in 

- https://github.com/Kong/kong-operator/pull/2199

It's not needed because there are no specific dependencies that are not used in KO directly.


tao12345666333:
I have re-run the CI

tao12345666333 (approved):

Issue #1658, opened 2025-05-21
Rearrange codebase


programmer04:
## Problem Statement

KGO used to have two versions, OSS and EE. In task

- https://github.com/Kong/gateway-operator/issues/1493

EE stuff has been reworked and added to the OSS repo. Some cleanup has been done, but there is still room for improvement. First of all, some packages can be moved to internal, because they don't need to be importable anymore. 

Furthermore KIC codebase became part of KO

- https://github.com/Kong/gateway-operator/issues/1567

So some code can be deduplicated.

Ideas for such improvements have been captured in 

- https://github.com/Kong/gateway-operator/issues/223
- https://github.com/Kong/gateway-operator/issues/100

That may serve as additional food for thought.

## Proposed Solution

What does not need to be public should be internal; code shouldn't be duplicated in the repo either. 

The scope may be too broad, so additional issues can be created.

## Acceptance Criteria
- [ ] Analyze what can be done
- [ ] Possibly create additional issues with a detailed description
- [ ] Propose/identify and tackle the most important improvements right away

sandromodarelli:
punting to KO 2.1.0 due to other priorities

programmer04:
It can be considered as a done (see all linked PRs), and created more specific spinoff issues

- https://github.com/Kong/kong-operator/issues/2780
- https://github.com/Kong/kong-operator/issues/2781

for further improvements.

pmalek:
Shouldn't we also make sure that there's only 1 Makefile file eventually? Or would completing these 2 issues imply that we can get rid of it?

programmer04:
Completing #2780 implies it, I'm adding explicitly to its ACs

PR #2358, opened 2025-09-29
chore: remove obsolete code needed for KIC as a  separate `Pod`

programmer04:

Get rid of all code related to the KIC distributed as a separate image (old KGO+KIC setup) and run in its own `Deployment`. Mostly leftovers (dead code) are removed in this PR, like test helpers, versions, dedicated webhook etc.

There was a test dedicated to Kong

- [`TestUpgradeKICWithExistingPlugins` in `ingress-controller/test/integration/isolated/upgrade_test.go`](https://github.com/Kong/kong-operator/pull/2358/files#diff-430ed83151cf8da7164ed345179a2937aef12fd929b2b209df7089bad1ab7f33)

which should be preserved, but run for an upgrade of the whole KO, so a dedicated issue has been created

- https://github.com/Kong/kong-operator/issues/2360

to implement it

**Which issue this PR fixes**

Part of

- https://github.com/Kong/kong-operator/issues/1658

The content of this PR is essentially yak shaving for issues related to reworking test suites to become KO native.


pkg/consts/controlplane.go
```diff
@@ -20,47 +18,3 @@ const (
 	// by the controlplane controller.
 	ControlPlaneManagedLabelValue = "controlplane"
```
fgiudici, line 19: uh... only one const left in this file 😆

fgiudici (approved):
LGTM

pkg/consts/controlplane.go
programmer04: Two `ControlPlanePrefix` and `ControlPlaneManagedLabelValue`. One of them is imported into multiple modules, and having them in a separate file is consistent, as it is for `DataPlane`. So I'm leaving it for now 🙂

Issue #2335, opened 2025-09-22
Hybrid gateways - HTTPRoute hostname and TLS support


fgiudici:
## Problem Statement

Add support to HTTPRoute hostname and TLS conversion on top of #2262 

## Proposed Solution

...

## Additional Information

...

## Acceptance Criteria

- [ ] Documentation is updated with this new feature / enhancement
- [ ] Tests are added for this feature / enhancement
- [ ] Helm chart is updated to work with this feature / enhancement
- [ ] As a user, I can ...

mlavacca:
@fgiudici Is this related to the hostname field of `HTTPRoute` or the Hostname intersection between `HTTPRoute` and `Gateway`?

fgiudici:
> [@fgiudici](https://github.com/fgiudici) Is this related to the hostname field of `HTTPRoute` or the Hostname intersection between `HTTPRoute` and `Gateway`?

It is related to the latter, i.e., the intersection of the HTTPRoute and Gateway Hostnames.
The hostname intersection part should be already covered by #2346, the TLS part is completely missing instead.

mlavacca:
Closing as not relevant anymore. TLS will be implemented by #2270.

PR #2359, opened 2025-09-29
fix(hybridgateway): accept all connections when no hostname is specified

fgiudici:

accept all connections when no hostname is specified in both the `Gatway` listener and the `HTTPRoute`.

**Which issue this PR fixes**

related to #2335 

**PR Readiness Checklist**:

(No need to update the CHANGELOG as the hostname support in hybridgateway has already  been added and not released yet)

### DIFF
--- a/controller/hybridgateway/converter/http_route.go
+++ b/controller/hybridgateway/converter/http_route.go
@@ -266,15 +266,23 @@ func (c *httpRouteConverter) addHostnames(ctx context.Context) error {
 					// This listener doesn't match the section reference, skip it
 					continue
 				}
+				if listener.Port != *pRef.Port {
+					// This listener doesn't match the port reference, skip it
+					continue
+				}
 			}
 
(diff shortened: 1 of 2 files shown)


controller/hybridgateway/converter/http_route.go
```diff
@@ -266,15 +266,23 @@ func (c *httpRouteConverter) addHostnames(ctx context.Context) error {
 					// This listener doesn't match the section reference, skip it
 					continue
 				}
+				if listener.Port != *pRef.Port {
```
programmer04: Are we sure that `pRef.Port` won't be `nil`? If not, we can do 


```suggestion
				if listener.Port != lo.FromPtr(pRef.Port) {
```

fgiudici: good catch! The port should be mandatory but using the `lo` package would be safer (moreover is better also for consistency as we use the `lo` package extensively).

programmer04 (approved):

PR #2424, opened 2025-10-10
feat(konnect): Fetch and update adopted entity (KongService) from Konnect

randmonkey:
**What this PR does / why we need it**:

Initial implementation of adopting a service from Konnect. This PR includes the general process of adoption and the adoption for services.

**Which issue this PR fixes**

Part of #2115 

**Special notes for your reviewer**:

**PR Readiness Checklist**:

Complete these before marking the PR as `ready to review`:

- [x] the `CHANGELOG.md` release notes have been updated to reflect significant changes

PR #2429, opened 2025-10-11
feat: implement the Match mode for adoption for Konnect cloud gateway…

### LINKED ISSUE #2116: Implement the "Match" mode adoption for Konnect cloud gateway entities
## Problem Statement

The Konnect cloud gateway resources are considered "immutable" with most of the fields cannot be changed. They also involves external resources (like AWS VPCs, transit gateways) in creation. The CRDs includes:
- `KonnectCloudGatewayNetwork`
- `KonnectCloudGatewayDataPlaneGroupConfiguration`
- `KonnectCloudGatewayTransitGateway`

We cannot update the entities like normal Kong gateway entities. The issue tells about the "match" mode adoption that does not change the existing entity.

## Proposed Solution

- Implement a "match" mode for those kind of resources. In this mode, the entity is successfully adopted only when the spec of the CR matches configuration of the existing entity in Konnect. When the entity is successfully adopted and the CR gets deleted in k8s, KO should also delete the entity in Konnect.

## Additional Information

...

## Acceptance Criteria

For the resources:
- `KonnectCloudGatewayNetwork`
- `KonnectCloudGatewayDataPlaneGroupConfiguration`
- `KonnectCloudGatewayTransitGateway`
- [ ] The controller can properly set the status and condition for the CRs adopting existing Konnect gateway resources after fetched them and verified the configuration matches
- [ ] The other resources referencing these resources adopting existing Konnect entities can be reconciled correctly
- [ ] The controller can delete the entity in Konnect correctly when the entity is adopted by the CR and the CR gets deleted

tao12345666333:
… entities

**Which issue this PR fixes**

Fixes https://github.com/Kong/kong-operator/issues/2116

~Due to the numerous changes in this modification, for convenience in review, I have split it into two PRs. We can choose to directly review this PR or first review https://github.com/Kong/kong-operator/pull/2455 , and then I will rebase this PR after it is merged.~

**PR Readiness Checklist**:


controller/konnect/ops/ops_adopt.go
```diff
@@ -0,0 +1,653 @@
...
+func AdoptMatch[
+	T constraints.SupportedKonnectEntityType,
+	TEnt constraints.EntityType[T],
+](
+	ctx context.Context,
+	sdk sdkops.SDKWrapper,
+	cl client.Client,
+	ent TEnt,
+) error {
+	adoptable, ok := any(ent).(interface {
```
randmonkey: I have defined an interface in #2424. You can reuse it here,

tao12345666333:
https://github.com/Kong/kong-operator/actions/runs/18467784217/job/52619461965

```
go: toolchain upgrade needed to resolve sigs.k8s.io/controller-runtime/tools/setup-envtest
go: sigs.k8s.io/controller-runtime/tools/setup-envtest@v0.0.0-20251013122658-1ad089b0ee87 requires go >= 1.25.0 (running go 1.24.6; GOTOOLCHAIN=local)
ERROR ~/work/kong-operator/kong-operator/ingress-controller/bin/plugins/setup-envtest/bin/install failed
Error: 
   0: Failed to install asdf:setup-envtest@0.21.0: 
         0: ~/work/kong-operator/kong-operator/ingress-controller/bin/plugins/setup-envtest/bin/install exited with non-zero status: exit code 1

      Location:
         src/plugins/script_manager.rs:184

      Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
      Run with RUST_BACKTRACE=full to include source snippets.

```

xref: https://github.com/Kong/kong-operator/pull/2149

ingress-controller/pkg/manager/config/consts.go
```diff
@@ -12,12 +12,12 @@ const (
 const (
 	// DefaultDataPlanesReadinessReconciliationInterval is the interval at which the manager will run DPs readiness reconciliation loop.
 	// It's the same as the default interval of a Kubernetes container's readiness probe.
-	DefaultDataPlanesReadinessReconciliationInterval = 10 * time.Second
+	DefaultDataPlanesReadinessReconciliationInterval = 20 * time.Second
```
randmonkey, line 15: Why did we increase the interval here? If we just want to fix CIs, I think only changing the test code should be enough. This changes the behavior of controllers, so I think it should not be brought inside the PR.

controller/konnect/reconciler_generic.go
```diff
@@ -528,13 +528,35 @@ func (r *KonnectEntityReconciler[T, TEnt]) adoptFromExistingEntity(
...
+		retErr         error
 	)
 	status := ent.GetKonnectStatus()
 	logger.Info("Adopting from existing entity",
 		"type", ent.GetTypeName(), "konnect_id", adoptOptions.Konnect.ID)
-	_, err := ops.Adopt(ctx, sdk, r.SyncPeriod, r.Client, r.MetricRecorder, ent, *adoptOptions)
-	if err != nil {
-		logger.Error(err, "failed to adopt entity", "type", ent.GetTypeName(), "konnect_id", adoptOptions.Konnect.ID)
+	// If adopting in match mode for Konnect Cloud Gateway resources, use the read-only match adopter.
+	if adoptOptions.Mode == "" || adoptOptions.Mode == commonv1alpha1.AdoptModeMatch {
```
randmonkey: I think we could just put the adopt process for these types in the `Adopt` function and keep the call of `ops.Adopt` unchanged here. This seems redundant.

controller/konnect/ops/ops_adopt.go
```diff
@@ -0,0 +1,669 @@
...
+	commonv1alpha1 "github.com/kong/kong-operator/api/common/v1alpha1"
+	kcfgkonnect "github.com/kong/kong-operator/api/konnect"
+	konnectv1alpha1 "github.com/kong/kong-operator/api/konnect/v1alpha1"
+	"github.com/kong/kong-operator/controller/konnect/constraints"
+	sdkops "github.com/kong/kong-operator/controller/konnect/ops/sdk"
+)
+
+// AdoptMatch performs adoption of a Konnect entity in "match" mode. When successful, the existing
+// entity in Konnect is managed by the operator without modifying it.
+func AdoptMatch[
```
randmonkey: I think we do not need an `AdoptMatch` function here because the `Adopt` already does the setting of status and dispatching to different types. You can modify the `Adopt` function for your requirement of implementation.

controller/konnect/ops/ops_adopt.go
```diff
@@ -0,0 +1,669 @@
...
+		}
+		return err
+	}
+
+	SetKonnectEntityAdoptedConditionTrue(ent)
+	SetKonnectEntityProgrammedConditionTrue(ent)
+	return nil
+}
+
+func adoptKonnectCloudGatewayNetworkMatch(
```
randmonkey: I think naming the functions just as `adoptKonnectCloudGatewayNetwork` and move the check of adopt mode is better.
Also there is a file for each supported entity type, I think the functions are better to be placed there to make the diff on the single file smaller.

controller/konnect/ops/ops_adopt.go
```diff
@@ -0,0 +1,669 @@
...
+		if network.GetKonnectID() == "" {
+			return "", fmt.Errorf("KonnectCloudGatewayNetwork %s does not have a Konnect ID", key)
+		}
+		return network.GetKonnectID(), nil
+	default:
+		return "", fmt.Errorf("unsupported network ref type %q", ref.Type)
+	}
+}
+
+func equalStringSets(a, b []string) bool {
```
randmonkey: There is `lo.ElementsMatch` for comparing slices so I think no need for this.

controller/konnect/ops/ops_adopt.go
```diff
@@ -0,0 +1,669 @@
...
+			return diff
+		}
+	default:
+		return fmt.Sprintf("unsupported transit gateway type %q", spec.Type)
+	}
+
+	return ""
+}
+
+func compareTransitGatewayDNSConfig(
```
randmonkey: I think there is no need  to marshal them to JSON and then compare as you already sorted the string slices inside the `TransitGatewayDNSConfig` here. Just comparing the slices using `lo.ElementsMatch` should be enough.

controller/konnect/reconciler_generic.go
tao12345666333: Updated.

controller/konnect/ops/ops_adopt.go
tao12345666333: Updated.

ingress-controller/pkg/manager/config/consts.go
tao12345666333: reverted.

controller/konnect/ops/ops_adopt.go
tao12345666333: OK

controller/konnect/ops/ops_konnectdataplanegroupconfigurations.go
```diff
@@ -295,3 +297,260 @@ func dataPlaneGroupsResponseToStatus(
...
+) (string, error) {
+	if manifest.GetControlPlaneID() != cfg.GetControlPlaneID() {
+		return fmt.Sprintf(
+			"controlPlaneRef mismatch spec=%q konnect=%q",
+			cfg.GetControlPlaneID(),
+			manifest.GetControlPlaneID(),
+		), nil
+	}
+
+	specAPIAccess := sdkkonnectcomp.APIAccessPrivatePlusPublic
```
randmonkey, line 356: I have a `equalWithDefault` helper function used to compare nullable fields with default values. Do you think it can be used here to simplify the code?

randmonkey (commented):
Mostly LGTM, only some minor comments on simplifying comparison methods.

controller/konnect/ops/ops_adopt_test.go
```diff
@@ -0,0 +1,501 @@
...
+		Once()
+
+	_, err := Adopt(ctx, *sdk, 0, cl, metricRecorder, network, *network.Spec.Adopt)
+	require.NoError(t, err)
+	assert.Equal(t, "net-1", network.GetKonnectID())
+	assert.Equal(t, string(sdkkonnectcomp.NetworkStateReady), network.Status.State)
+	assertProgrammedCondition(t, network.GetConditions(), metav1.ConditionTrue, string(konnectv1alpha1.KonnectEntityProgrammedReasonProgrammed))
+}
+
+func TestAdoptMatchNetworkMismatch(t *testing.T) {
```
randmonkey: Could we also move the tests for specific types to `ops_*_test.go` instead of `ops_adopt_test.go`?

controller/konnect/ops/ops_konnectdataplanegroupconfigurations.go
tao12345666333: Great! But I found there is a bug in the `equalWithDefault` function; it needs an else branch. I prefer that we keep the current implementation, fixing the function, then consider how to simplify.

controller/konnect/ops/ops_adopt_test.go
tao12345666333: OK

randmonkey (approved):

PR #2490, opened 2025-10-21
feat(konnect): Implement adoption for KongVault

### LINKED ISSUE #2446: Adopting Konnect entities: vaults
## Problem Statement

The sub issue for adopting vaults in Konnect by `KongVault`.

## Proposed Solution

...

## Additional Information

...

## Acceptance Criteria

- [ ] KO can adopt vaults from Konnect.

tao12345666333:

**Which issue this PR fixes**

Fixes https://github.com/Kong/kong-operator/issues/2446

**PR Readiness Checklist**:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -105,6 +105,8 @@
     [#2429](https://github.com/Kong/kong-operator/pull/2429)
   - Implement adoption support for `KongCertificate`, `KongCACertificate` and `KongSNI`
     [#2484](https://github.com/Kong/kong-operator/pull/2484)
+  - Implement adoption support for `KongVault`.
+    [#2490](https://github.com/Kong/kong-operator/pull/2490)
 - HybridGateway:
   - Added controller-runtime watches for Gateway and GatewayClass resources to the hybridgateway controller.
   - HTTPRoutes are now reconciled when related Gateway or GatewayClass resources change.
--- a/controller/konnect/ops/ops_kongvault.go
+++ b/controller/konnect/ops/ops_kongvault.go
@@ -3,12 +3,15 @@ package ops
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"reflect"
 
 	sdkkonnectcomp "github.com/Kong/sdk-konnect-go/models/components"
 	sdkkonnectops "github.com/Kong/sdk-konnect-go/models/operations"
 	"github.com/samber/lo"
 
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -103,6 +103,8 @@
     [#2424](https://github.com/Kong/kong-operator/pull/2424)
   - Implement the Match mode for adoption for Konnect cloud gateway entities
     [#2429](https://github.com/Kong/kong-operator/pull/2429)
+  - Implement adoption for KongValut.
```
randmonkey:
```suggestion
  - Implement adoption for `KongVault`.
```

controller/konnect/ops/ops_kongvault.go
```diff
@@ -95,6 +157,29 @@ func kongVaultToVaultInput(vault *configurationv1alpha1.KongVault) (sdkkonnectco
...
+	}
+
+	return konnectVault.Name == expected.Name &&
+		konnectVault.Prefix == expected.Prefix &&
+		equalWithDefault(konnectVault.Description, expected.Description, "") &&
+		vaultConfigMatch(konnectVault.Config, expected.Config)
+}
+
+func vaultConfigMatch(a map[string]any, b map[string]any) bool {
+	if len(a) == 0 && len(b) == 0 {
```
randmonkey, line 177:
```
Map values are deeply equal when all of the following are true: they are both nil or both non-nil, they have the same length, and either they are the same map object or their corresponding keys (matched using Go equality) map to deeply equal values.
```
https://pkg.go.dev/reflect#DeepEqual 
So is it required to check `nil` or length of each map before checking two maps equal using `DeepEqual`?

randmonkey (commented):
Generally LGTM, some minor comments.

controller/konnect/ops/ops_kongvault.go
tao12345666333: Yes!

In our code:
   •  `expected.Config` is always a non-nil empty map (`vaultConfig := map[string]any{}`)
   •  `konnectVault.Config` from the Konnect API could be nil when no config is present

Without the `len(a) == 0 && len(b) == 0` check:
   •  `reflect.DeepEqual(nil, map[string]any{})`  will returns false
   •  This would create a mismatch, even though both semantically represent "no configuration."

The explicit length check ensures that nil maps and empty maps are treated as equivalent, which is the correct semantic behavior for configuration comparison.

randmonkey (approved):

PR #2502, opened 2025-10-27
feat: implement comprehensive EndpointSlices-based target resolution for HybridGateway

### LINKED ISSUE #2264: Hybrid gateways - EndpointSlice conversion
## Problem Statement

In #2262 the HTTPRoute BackendRefs are converted to `KongTarget`s pointing to Kubernetes Services.
This issues explodes the Kubernetes Services in Endpoints and the `KongTarget`s should point to actual Endpoints.
The weight assigned to each resulting `KongTarget` endpoint should be computed taking into account the weight assigned to the `Service` in the BackendRefs section of the  `HTTPRoute` and the number of `Endpoint`s belonging to the `Service`.

Configure `EndpointSlice` in Konnect Hybrid `Gateway`s. To achieve this high-level result, the following set of tasks needs to be addressed:
- **Conversion**: Create a new `EndpointSliceConverter` struct that implements the `APIConverter` interface and converts one `EndpointSlice` into the following set of resources:
  - `KongUpstream`s
  - One `KongTarget` per raw endpoint
- **Controller**: Create a new instance of the `HybridGatewayReconciler` that reconciles `EndpointSlice`s and uses the converter defined above
- **Watches**: Define the set of `Watches` and `Owns` for the controller above, in order to be sure that all the needed APIs are properly used to trigger new reconciliation events
- **RBACs**: Ensure the Controller has the set of RBACs needed to perform the CRUD operations on the underlying APIs.
- **Status update**: Ensure the status of the `HTTPRoute`s is enriched with the proper set of conditions related to the `KongUpstream` and `KongTarget` CRs.

## Acceptance Criteria

- [ ] All the points described in the problem statement are properly implemented

### LINKED ISSUE #2266: Hybrid gateways - backend cross-namespace references
## Problem Statement

Allow Cross-namespace references between `HTTPRoute` and `service`s.  At the moment, the cross-namespace references are not supported in the controllers. We need to allow cross-namespace by watching and using `ReferenceGrant` to see if the reference has been explicitly allowed.

## Acceptance Criteria

- [ ] Cross-namespace references on `HTTPRoute` backends are allowed and fully supported.

alacuku:

HybridGateway controllers need a robust mechanism to resolve Kubernetes services into Kong targets that supports all service types and provides accurate traffic distribution.

##### Target Resolution (`target.go`)
- `TargetsForBackendRefs()`: Main entry point that processes HTTPRoute BackendRefs
- `filterValidBackendRefs()`: Validates services, ports, and ReferenceGrants
- `resolveServiceEndpoints()`: Handles different service types:
  - Regular services: Uses EndpointSlices API
  - Headless services: Direct endpoint resolution
  - ExternalName services: External DNS targets
  - FQDN mode: Service DNS names
- `recalculateWeightsAcrossBackendRefs()`: Applies weight algorithm across backends

##### Weight Distribution (`weight.go`)
- `CalculateEndpointWeights()`: Converts service weights to per-endpoint weights
- Uses GCD/LCM math to maintain proportional traffic distribution
- Handles edge cases (zero weights, missing endpoints)

##### Integration
- Updates HTTP route converter to use new target resolution
- Maintains existing resource generation patterns

### Service Type Support

| Service Type | Resolution Method | Target Format |
|--------------|-------------------|---------------|
| ClusterIP | EndpointSlices | `pod-ip:target-port` |
| Headless | EndpointSlices | `pod-ip:target-port` |
| ExternalName | DNS | `external-name:service-port` |
| FQDN Mode | Service DNS | `service.ns.svc.cluster.local:service-port` |

#### Testing

- Unit tests for all resolution paths and edge cases
- Weight calculation validation with mathematical examples
- Error handling for missing services, invalid ports, ReferenceGrant scenarios
- Mock client testing for Kubernetes API interactions

#### Files

- `controller/hybridgateway/target/target.go` - Target resolution logic
- `controller/hybridgateway/target/weight.go` - Weight calculation algorithm  
- `controller/hybridgateway/target/*_test.go` - Test coverage
- `controller/hybridgateway/converter/http_route.go` - Integration point

**Which issue this PR fixes**

This PR adds a target resolution system that uses EndpointSlices to discover service endpoints and generate Kong targets with proper weight distribution.

Fixes #2264 #2266

**PR Readiness Checklist**:

### DIFF
--- a/controller/hybridgateway/builder/kongtarget.go
+++ b/controller/hybridgateway/builder/kongtarget.go
@@ -52,25 +52,23 @@ func (b *KongTargetBuilder) WithLabels(route *gwtypes.HTTPRoute, parentRef *gwty
 	return b
 }
 
-// WithBackendRef sets the target specification based on the given HTTPRoute and backend reference.
-func (b *KongTargetBuilder) WithBackendRef(httpRoute *gwtypes.HTTPRoute, bRef *gwtypes.HTTPBackendRef) *KongTargetBuilder {
-	// Build the dns name of the service for the backendRef.
-	// TODO(alacuku): We need to handle the cluster domain properly for the cluster where we are running.
-	var namespace string
-	if bRef.Namespace == nil || *bRef.Namespace == "" {
-		namespace = httpRoute.Namespace
+// WithTarget sets the target address for the KongTarget.
+// It combines the host and port into a single target string in the format "host:port".
+// The host can be an IP address, hostname, or FQDN, and the port must be a valid port number.
+func (b *KongTargetBuilder) WithTarget(host string, port int) *KongTargetBuilder {
+	b.target.Spec.Target = net.JoinHostPort(host, strconv.Itoa(port))
+	return b
+}
+
+// WithWeight sets the weight for the KongTarget, which determines the proportion of traffic
+// this target will receive relative to other targets in the same upstream.
+// If weight is nil, it defaults to 100. Higher weights receive more traffic.
+func (b *KongTargetBuilder) WithWeight(weight *int32) *KongTargetBuilder {
+	// Weight is optional, default to 100 if not specified.
+	if weight != nil {
+		b.target.Spec.Weight = int(*weight)
 	} else {
-		namespace = string(*bRef.Namespace)
-	}
-
-	host := string(bRef.Name) + "." + namespace + ".svc.cluster.local"
-	port := strconv.Itoa(int(*bRef.Port))
-	target := net.JoinHostPort(host, port)
-	b.target.Spec.Target = target
-
-	// Weight is optional, default to 100 if not specified
-	if bRef.Weight != nil {
-		b.target.Spec.Weight = int(*bRef.Weight)
+		b.target.Spec.Weight = 100
 	}
 	return b
 }
--- a/controller/hybridgateway/builder/kongtarget_test.go
+++ b/controller/hybridgateway/builder/kongtarget_test.go
@@ -6,7 +6,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
+	"k8s.io/utils/ptr"
 
 	configurationv1alpha1 "github.com/kong/kong-operator/api/configuration/v1alpha1"
 	gwtypes "github.com/kong/kong-operator/internal/types"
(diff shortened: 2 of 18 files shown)


controller/hybridgateway/converter/http_route.go
```diff
@@ -216,23 +221,26 @@ func (c *httpRouteConverter) translate(ctx context.Context) error {
...
-				if err != nil {
-					// TODO: decide how to handle build errors in converter
-					// For now, skip this resource
-					continue
-				}
-				c.outputStore = append(c.outputStore, &target)
+			// Build the KongTarget resources using the new rule-based approach.
+			targets, err := target.TargetsForBackendRefs(
+				ctx,
+				logr.Discard(), // TODO: pass proper logger.
```
tao12345666333, line 227: Do you plan to complete this modification in this PR?

controller/hybridgateway/target/weight.go
```diff
@@ -0,0 +1,180 @@
...
+	for i := 1; i < len(activeDenominators); i++ {
+		overallLCM = lcm(overallLCM, activeDenominators[i])
+	}
+
+	// Step 3: Calculate the un-simplified, integer-based weights.
+	unsimplifiedWeights := make([]uint32, 0, len(fractions))
+	nonZeroWeights := make([]uint32, 0)
+	for _, f := range fractions {
+		// The weight is the fraction's value multiplied by the LCM.
+		weight := (f.num * overallLCM) / f.den
```
tao12345666333: I'm worried there might be an overflow here. Should we consider using uint64?

controller/hybridgateway/builder/kongtarget_test.go
```diff
@@ -58,129 +57,95 @@ func TestKongTargetBuilder_WithLabels(t *testing.T) {
...
+
+func TestKongTargetBuilder_WithWeight(t *testing.T) {
+	tests := []struct {
+		name           string
+		weight         *int32
+		expectedWeight int
+	}{
+		{
+			name:           "with weight 50",
+			weight:         func() *int32 { w := int32(50); return &w }(),
```
fgiudici: the `github.com/samber/lo` package can be handy here
```suggestion
			weight:         lo.ToPtr(50)
```

controller/hybridgateway/target/weight.go
alacuku: Added the logic to prevent integer overflow in weigh calculation algorithm.

controller/hybridgateway/builder/kongtarget_test.go
```diff
@@ -58,129 +57,95 @@ func TestKongTargetBuilder_WithLabels(t *testing.T) {
...
+			require.NoError(t, err)
+			assert.Equal(t, tt.expectedTarget, target.Spec.Target)
+		})
+	}
+}
+
+func TestKongTargetBuilder_WithWeight(t *testing.T) {
+	tests := []struct {
+		name           string
+		weight         *int32
```
fgiudici: if we take the pointer in the test loop it will be more easy...
```suggestion
		weight         int32
```

controller/hybridgateway/builder/kongtarget_test.go
```diff
@@ -58,129 +57,95 @@ func TestKongTargetBuilder_WithLabels(t *testing.T) {
...
+			name:           "with nil weight (defaults to 100)",
+			weight:         nil,
+			expectedWeight: 100,
+		},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			builder := NewKongTarget().WithBackendRef(tt.httpRoute, tt.backendRef)
+			builder := NewKongTarget().WithWeight(tt.weight)
```
fgiudici:
```suggestion
			builder := NewKongTarget().WithWeight(&tt.weight)
```

controller/hybridgateway/converter/http_route.go
alacuku: This is intentionally out of scope for this particular PR. I'll be handling that change in a follow-up PR dedicated to revising the hybrid gateway logging.

controller/hybridgateway/target/target.go
```diff
@@ -0,0 +1,411 @@
...
+
+	configurationv1alpha1 "github.com/kong/kong-operator/api/configuration/v1alpha1"
+	"github.com/kong/kong-operator/controller/hybridgateway/builder"
+	"github.com/kong/kong-operator/controller/hybridgateway/route"
+	"github.com/kong/kong-operator/controller/hybridgateway/utils"
+	gwtypes "github.com/kong/kong-operator/internal/types"
+)
+
+// ValidBackendRef represents a BackendRef that has passed all validation checks.
+type ValidBackendRef struct {
```
fgiudici: ValidBackendRef is an internal type, used for the endpoint validation pipeline.
We use it as an internal type indeed.
We may want to not export it.
```suggestion
type validBackendRef struct {
```

controller/hybridgateway/converter/http_route.go
```diff
@@ -34,15 +35,19 @@ type httpRouteConverter struct {
 	outputStore           []client.Object
 	expectedGVKs          []schema.GroupVersionKind
 	referenceGrantEnabled bool
+	fqdnMode              bool
+	clusterDomain         string
 }
 
 // NewHTTPRouteConverter returns a new instance of httpRouteConverter.
-func newHTTPRouteConverter(httpRoute *gwtypes.HTTPRoute, cl client.Client, referenceGrantEnabled bool) APIConverter[gwtypes.HTTPRoute] {
+func newHTTPRouteConverter(httpRoute *gwtypes.HTTPRoute, cl client.Client, referenceGrantEnabled bool, fqdnMode bool, clusterDomain string) APIConverter[gwtypes.HTTPRoute] {
```
fgiudici, line 43: I would consider having the fqdnMode/clusterDomain evaluated per HTTPRoute.
Maybe marked with an annotation? 🤔  Having it set once for all HTTPRoutes may be a limit in some conditions.

Anyway, not something for this PR, all good here! Food for thoughts to consider when we will enable selecting the fqdnMode.

fgiudici (dismissed):
Looks good here!
Would just consider unexporting the ValidBackend struct, but is a minor one.

controller/hybridgateway/builder/kongtarget.go
```diff
@@ -52,25 +52,23 @@ func (b *KongTargetBuilder) WithLabels(route *gwtypes.HTTPRoute, parentRef *gwty
...
-		b.target.Spec.Weight = int(*bRef.Weight)
+// WithWeight sets the weight for the KongTarget, which determines the proportion of traffic
+// this target will receive relative to other targets in the same upstream.
+// If weight is nil, it defaults to 100. Higher weights receive more traffic.
+func (b *KongTargetBuilder) WithWeight(weight *int32) *KongTargetBuilder {
+	// Weight is optional, default to 100 if not specified.
+	if weight != nil {
+		b.target.Spec.Weight = int(*weight)
+	} else {
+		b.target.Spec.Weight = 100
```
mlavacca, line 71: Gateway API spec has a default value, which means weight can't be nil here. However, I agree we need to make code bulletproof and check nilness. In that case, what do you think about sticking to API specifications and using 1 as well?

https://github.com/mlavacca/gateway-api/blob/main/apis/v1/shared_types.go#L313-L324

controller/hybridgateway/converter/http_route.go
```diff
@@ -216,23 +221,26 @@ func (c *httpRouteConverter) translate(ctx context.Context) error {
...
+				&pRef,
+				upstreamName,
+				c.referenceGrantEnabled,
+				c.fqdnMode,
+				c.clusterDomain,
+			)
+			if err != nil {
+				// TODO: decide how to handle target creation errors in converter
+				// For now, skip this rule
+				continue
```
mlavacca, line 240: Can we add a log line here?

controller/hybridgateway/route/status.go
```diff
@@ -1017,3 +1009,46 @@ func IsHTTPReferenceGranted(grantSpec gwtypes.ReferenceGrantSpec, backendRef gwt
...
+
+	// Check if any ReferenceGrant permits this reference
+	for _, grant := range grantList.Items {
+		if IsHTTPReferenceGranted(grant.Spec, *bRef, routeNamespace) {
+			return true, true, nil
+		}
+	}
+
+	// ReferenceGrants exist but none permit this reference
+	return false, true, nil
```
mlavacca, line 1053: How is `false, false, nil` different from `false, true, nil`? Do we really need to distinguish between "grant does not exist" and "grant does not allow"?

controller/hybridgateway/target/target.go
```diff
@@ -0,0 +1,411 @@
...
+}
+
+// extractReadyEndpointAddresses extracts all ready endpoint addresses that match the service port.
+func extractReadyEndpointAddresses(endpointSlices *discoveryv1.EndpointSliceList, svcPort *corev1.ServicePort) []string {
+	var addresses []string
+
+	for _, endpointSlice := range endpointSlices.Items {
+		for _, p := range endpointSlice.Ports {
+			// Skip ports that don't match the service port.
+			if p.Port == nil || *p.Port < 0 || *p.Protocol != svcPort.Protocol || *p.Name != svcPort.Name {
```
mlavacca, line 121: why `Port<0`? Shouldn't we check `p.port == svcPort.Port`?

mlavacca:
Huge PR, great job!

controller/hybridgateway/builder/kongtarget.go
alacuku: As you noted the weigh can't be nil, but i decided to set the default value of the kongTarget: https://github.com/Kong/kong-operator/blob/22e30ec3834c8bc02aa9bec4797f753925207ffc/api/configuration/v1alpha1/kongtarget_types.go#L71C2-L72C28

controller/hybridgateway/route/status.go
alacuku: We need to distinguish for the condition message in the `ResolvedRefs` condition.

controller/hybridgateway/target/target.go
alacuku: You can call it defensive programming, but you are right, it should never happen. The question about `p.port == svcPort.Port`: we do not really check that, since we should check that `p.port == svc.TargetPort` but checking `p.Name != svcPort.Name` is enough.

controller/hybridgateway/converter/http_route.go
alacuku: We should handle the error properly, and there is already a TODO about that.

controller/hybridgateway/builder/kongtarget.go
mlavacca: I see. So, we need to map the interval 0-1000000 (Gateway API) to the interval 0-65536 (KongTarget). Are we already doing that in some place I missed?

What I am specifically thinking is: on Gateway API, two backends (with single endpoints, not to overcomplicate things) with weights 100 and 300: we can't just set the target weights to 100 and 300, but we should remap them on the new interval (which is narrower than the Gateway API one). I've seen all the logic you wrote around rebalancing weights. Is this case taken into account there?

controller/hybridgateway/target/target.go
mlavacca: What if the endpointslice is manually forged? I think I can create an EPSlices, bind it to a service, but misconfigure it such that a port name matches a different port. Probably it never gets ready, but I think it is still possible. WDYT?

controller/hybridgateway/builder/kongtarget.go
alacuku: Here is the issue that tracks this use case: https://github.com/Kong/kong-operator/issues/2515

controller/hybridgateway/target/target.go
alacuku: This case suits the headless services without selectors: https://github.com/Kong/kong-operator/issues/2516

controller/hybridgateway/converter/http_route.go
alacuku: I opened a tracking issue for this: https://github.com/Kong/kong-operator/issues/2517

mlavacca (approved):

Issue #2496, opened 2025-10-27
Flaky integration test case: `TestGatewayEssentials`


randmonkey:
Recently `TestGatewayEssentials` test case in KO integration test cases is flaky and it fails several times in merge queue CIs before a PR gets merged. 
The issue is created to track resolving the flakiness.

randmonkey:
Closed this as it is seems not flaky now.

PR #2504, opened 2025-10-29
test(integration): Use watch to wait gateway and listeners set to not programmed in `TestGatewayEssentials`

randmonkey:

Use `client.Watch` to watch events of gateways and catch the transient state of gateway and its listeners marked as not programmed.

**Which issue this PR fixes**

Part of #2496

Note: In this PR I moved `WatchFor` to `helpers` package to run a more generic watch because the `setupWatch` in `envtest` cannot run on types other than our own CRD. The other types does not have our own methods so we have to use the generic method to run watches for them.

**PR Readiness Checklist**:

### DIFF
--- a/test/helpers/watch.go
+++ b/test/helpers/watch.go
@@ -28,6 +28,8 @@ func WatchFor[
 ) T {
 	t.Helper()
 
+	require.Greater(t, timeout, time.Duration(0), "Must provide a duration greater than 0 as the timeout")
+
 	ctx, cancel := context.WithTimeout(ctx, timeout)
 	defer cancel()
 
--- a/test/helpers/watch.go
+++ b/test/helpers/watch.go
@@ -1,0 +1,63 @@
+package helpers
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/kr/pretty"
+	"github.com/stretchr/testify/require"
+	apiwatch "k8s.io/apimachinery/pkg/watch"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// WatchFor watches for an event of type eventType using the provided watch.Interface.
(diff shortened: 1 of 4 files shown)


test/helpers/watch.go
```diff
@@ -0,0 +1,63 @@
...
+	ctx context.Context,
+	w apiwatch.Interface,
+	eventType apiwatch.EventType,
+	timeout time.Duration,
+	predicate func(T) bool,
+	failMsg string,
+) T {
+	t.Helper()
+
+	ctx, cancel := context.WithTimeout(ctx, timeout)
```
tao12345666333, line 33: Should we check if the timeout is always greater than 0?

test/integration/gateway_test.go
```diff
@@ -48,6 +50,14 @@ func TestGatewayEssentials(t *testing.T) {
 	require.NoError(t, err)
 	cleaner.Add(gatewayClass)
 
+	t.Log("setting up watch for Gateways")
+	cl, err := client.NewWithWatch(GetEnv().Cluster().Config(), client.Options{
+		Scheme: scheme.Get(),
+	})
+	require.NoError(t, err, "failed to setup a client for watching gateways")
+	wGateway, err := cl.Watch(GetCtx(), &gatewayv1.GatewayList{}, client.InNamespace(namespace.Name))
```
tao12345666333, line 58: Should we add a cleanup logic?

test/helpers/watch.go
randmonkey: Added the check.

test/integration/gateway_test.go
randmonkey: Added it to cleanup.

tao12345666333 (approved):

alacuku (approved):

Issue #2520, opened 2025-10-30
HybridGateway: enforce underlying resources CRUD ops ordering


mlavacca:
## Problem Statement

At the moment, the creation of the underlying resources in the HybridGateway controller is neither checked nor taken into consideration. This may lead to some problems:

- if a service gets deleted before the routes attached, the delete operation fails
- if a route gets deleted before the attached plugins, the delete operation fails
- if a HTTPRoute defines rules with plugins attached, the targets should be created only once all the other building blocks are created.
- ...

The goal of the issue is to figure out the proper order of creation/update/delete of resources and ensures such order is applied in the code

## Acceptance Criteria

- [ ] The underlying resources are created/updated/deleted according to a specific order that needs to be figured out as part of this issue.

mlavacca:
Closing as the issue is already split and captured by other issues

Issue #2562, opened 2025-11-05
Hybrid Gateway - Set KongService Protocol


mlavacca:
## Problem Statement

In Hybrid Gateways, when we create a KongService, the Protocol field is not set. We need to set it according to the route type (HTTP, gRPC, etc.)

## Acceptance Criteria

- [ ] In the context of hybrid Gateway configurations, the KongService Protocol is always set.

PR #2566, opened 2025-11-06
feat(hybridgateway): update naming for autogenerated resources

### LINKED ISSUE #2542: Hybrid Gateway - update naming of the created Kong CRDs
## Problem Statement
Currently the created Kong CRDs names are derived from the HTTPRoute resource name and an hash based on relevant part of the HTTPRoute spec.

This result in creating different KongService / Upstream resources mapping the same backend Services (BackendRefs) if they are referenced in different HTTPRoute resources.

## Proposed Solution

The HTTPRoute name should be removed from the generated names for KongService and Upstreams.
The naming should instead depend upon the namespace of the backend Services and their namespace (if namespace is not defined, i.e., the namespace is the same of the HTTPRoute resource, the namspace of the HTTPRoute should be considered to avoid duplicating KongServices/Upstreams when references are cross-namespaces).

## Acceptance Criteria

fgiudici:
Currently Kong CRDs names are derived from the HTTPRoute resource name and an hash based on the relevant part of the HTTPRoute spec.
This result in creating different KongService / Upstream resources mapping the same backend Services (BackendRefs) if they are referenced in different HTTPRoute resources.
HTTPRoute name from the generated names of KongService and KongUpstreams should be dropped.

While there, move all the name generation logic to the namegen package.

**Which issue this PR fixes**

Fixes #2542 

This PR introduces an issue with the ownership of the created resources: till now the Owner was the source HTTPRoute, but at this point different HTTPRoutes may share the same KongService/KongUpstream/KongPlugin/KongTargets.
This issue is going to be addressed while working on #2520.

**PR Readiness Checklist**:

### DIFF
--- a/controller/hybridgateway/converter/http_route.go
+++ b/controller/hybridgateway/converter/http_route.go
@@ -312,13 +312,10 @@ func (c *httpRouteConverter) translate(ctx context.Context, logger logr.Logger) 
 	log.Debug(logger, "Found supported parent references",
 		"parentRefCount", len(supportedParentRefs))
 
-	httpRouteName := c.route.Namespace + "-" + c.route.Name
-
 	for _, pRefData := range supportedParentRefs {
 		pRef := pRefData.parentRef
 		cp := pRefData.cpRef
 		hostnames := pRefData.hostnames
-		cpRefName := "cp" + utils.Hash32(cp)
 
 		log.Debug(logger, "Processing parent reference",
 			"parentRef", pRef,
@@ -331,8 +328,9 @@ func (c *httpRouteConverter) translate(ctx context.Context, logger logr.Logger) 
 				"backendRefCount", len(rule.BackendRefs),
 				"matchCount", len(rule.Matches),
 				"filterCount", len(rule.Filters))
+
 			// Build the KongUpstream resource.
-			upstreamName := namegen.NewName(httpRouteName, cpRefName, utils.Hash32(rule.BackendRefs)).String()
+			upstreamName := namegen.NewKongUpstreamName(cp, rule)
 			log.Trace(logger, "Building KongUpstream resource",
 				"upstream", upstreamName,
 				"controlPlane", cp.KonnectNamespacedRef)
@@ -385,7 +383,7 @@ func (c *httpRouteConverter) translate(ctx context.Context, logger logr.Logger) 
 			}
 
 			// Build the KongService resource.
-			serviceName := namegen.NewName(httpRouteName, cpRefName, utils.Hash32(rule.BackendRefs)).String()
+			serviceName := namegen.NewKongServiceName(cp, rule)
 			log.Trace(logger, "Building KongService resource",
 				"service", serviceName,
 				"upstream", upstreamName)
@@ -411,7 +409,7 @@ func (c *httpRouteConverter) translate(ctx context.Context, logger logr.Logger) 
 				"service", serviceName)
 
 			// Build the kong route resource.
-			routeName := namegen.NewName(httpRouteName, cpRefName, utils.Hash32(rule.Matches)).String()
+			routeName := namegen.NewKongRouteName(c.route, cp, rule)
 			log.Trace(logger, "Building KongRoute resource",
 				"kongRoute", routeName,
 				"service", serviceName,
@@ -450,8 +448,7 @@ func (c *httpRouteConverter) translate(ctx context.Context, logger logr.Logger) 
 				"filterCount", len(rule.Filters))
 
 			for _, filter := range rule.Filters {
-				filterHash := utils.Hash32(filter)
-				pluginName := namegen.NewName(httpRouteName, cpRefName, filterHash).String()
+				pluginName := namegen.NewKongPluginName(filter)
 
 				log.Trace(logger, "Building KongPlugin resource",
 					"plugin", pluginName,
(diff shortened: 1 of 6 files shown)


controller/hybridgateway/namegen/name.go
```diff
@@ -1,19 +1,28 @@
...
-	parentRefID string
-	sectionID   string
+	httpRouteID    string
+	controlPlaneID string
+	sectionID      string
 }
 
+const defaultHTTPRoutePrefix = "http"
+const defaultCPPrefix = "cp"
+const defaultResPrefix = "res"
```
programmer04:
```suggestion
const (
  defaultHTTPRoutePrefix = "http"
  defaultCPPrefix = "cp"
  defaultResPrefix = "res"
)
```

nit

controller/hybridgateway/namegen/name.go
```diff
@@ -64,10 +70,54 @@ func (h *Name) String() string {
 }
 
 // NewName creates a new Name instance with the given components.
-func NewName(httpRouteID, parentRefID, sectionID string) *Name {
+func NewName(httpRouteID, controlPlaneID, sectionID string) *Name {
```
programmer04: As far as I see, this public `NewName` is nowhere used outside this package. Do you think it may be worth making it `newName`?

fgiudici: yep, makes sense... it was public once as we used it directly. Now there is no more reason to keep it public.
Good catch!

mlavacca (dismissed):
I'm not 100% sold on the code structure around how we generate the names and the `New<resource>Name` functions, as I think we could semplify that part of code and make it more readable, but it would touch sections untouched by this PR. Let's merge it as it is and discuss it further.

fgiudici:
> I'm not 100% sold on the code structure around how we generate the names and the `New<resource>Name` functions, as I think we could semplify that part of code and make it more readable, but it would touch sections untouched by this PR. Let's merge it as it is and discuss it further.

Well, the main idea was to isolate the code to generate the name in the namegen package: if we decide to change it at least hopefully would not affect the translate function anymore.
If what you refer to is how the names are generated in the namegen package... well, honestly that is a bit convoluted today, mainly because the package initially was used in another way. Fully agree that should be simplified. 
Pushed on top a simplified version.

controller/hybridgateway/converter/http_route.go
```diff
@@ -450,8 +448,7 @@ func (c *httpRouteConverter) translate(ctx context.Context, logger logr.Logger)
 				"filterCount", len(rule.Filters))
 
 			for _, filter := range rule.Filters {
-				filterHash := utils.Hash32(filter)
-				pluginName := namegen.NewName(httpRouteName, cpRefName, filterHash).String()
+				pluginName := namegen.NewKongPluginName(filter)
```
tao12345666333, line 451: Since this modification does not include information related to route or cp, I would like to know if this is what we expected? Also, will debugging become more complicated after this change?

If possible, I suggest adding some comments here for clarification.

fgiudici: Yes, this is expected: the KongPlugin just carries its configuration only and is used to be linked to a KongRoute (or other entities) via a KongPluginBinding. There is no need to keep KongPlugins with exactly the same configuration: the same KongPlugin can be referenced by multiple Kong entities.
Since the KongPlugin configuration is derived directly from the HTTPRoute filter, the filter itself is used to identify the KongPlugin.
I understand the confusion as resource naming is not straightforward.
Just added a doc about the naming in the PR, is that of help?

fgiudici:
(branch rebased)

controller/hybridgateway/namegen/name.go
```diff
@@ -3,71 +3,76 @@ package namegen
...
+	namegenPrefix   = "ngn"
+	maxLen          = 253
+)
 
-	parts := []string{}
-	if h.httpRouteID != "" {
-		parts = append(parts, h.httpRouteID)
+// newName generates a name by concatenating the given components if the length is within the limit of
+// Kubernetes resource names, otherwise it returns a hashed version of the concatenated elements.
+func newName(elements ...string) string {
```
mlavacca, line 22: Great to see that! 👍🏻

controller/hybridgateway/namegen/name.go
```diff
@@ -3,71 +3,76 @@ package namegen
...
+}
 
-	httpRouteID := h.httpRouteID
-	parentRefID := h.parentRefID
-	sectionID := h.sectionID
+// NewKongServiceName generates a KongService name based on the ControlPlaneRef and HTTPRouteRule passed as arguments.
+func NewKongServiceName(cp *commonv1alpha1.ControlPlaneRef, rule gatewayv1.HTTPRouteRule) string {
+	return newName(
+		defaultCPPrefix+utils.Hash32(cp),
+		utils.Hash32(rule.BackendRefs),
```
mlavacca, line 44: I remember that in the previous implementation, we had the protocol prefix. I think we should keep it, as when in the future we'll add `GRPCRoute` support, we'll need to set the protocol field accordingly (see https://github.com/Kong/kong-operator/issues/2562)

fgiudici: Good point! Added 'http' prefix to the generated KongServices... and KongRoutes too!
The other resources should not need the prefix.

mlavacca (approved):
🚀

controller/hybridgateway/converter/http_route.go
fgiudici: Added detailed doc explaining the naming.

Issue #2622, opened 2025-11-17
When reconciling Hybrid Gateway unexpected code branch with Gateway set to `nil` is reached


pmalek:
### Problem statement

Inspecting tests logs I've encountered the following log entry:

```
2025-11-17T11:16:32Z	ERROR	Reconciler error	{"controller": "gateway", "controllerGroup": "gateway.networking.k8s.io", "controllerKind": "Gateway", "Gateway": {"name":"38f4b206-834e-4de0-9ecc-a45deabb826b","namespace":"efc03950-15bd-45ec-87d7-77737fd0dacb"}, "namespace": "efc03950-15bd-45ec-87d7-77737fd0dacb", "name": "38f4b206-834e-4de0-9ecc-a45deabb826b", "reconcileID": "eebc2162-3217-492d-a562-d8e97670c33a", "error": "unexpected error, KonnectGatewayControlPlane is nil. Returning to avoid panic"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler
	/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.21.0/pkg/internal/controller/controller.go:353
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem
	/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.21.0/pkg/internal/controller/controller.go:300
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func2.1
	/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.21.0/pkg/internal/controller/controller.go:202
```

https://github.com/Kong/kong-operator/actions/runs/19427385460/job/55578012517?pr=2589#step:7:20749

It would seem that this code branch should be reached.

### Acceptance criteria

- [ ] Log `unexpected error, KonnectGatewayControlPlane is nil. Returning to avoid panic` is not emitted.

Issue #1586, opened 2025-05-08
Konnect entities: handle `KonnectAPIAuthConfiguration` deletion


pmalek:
## Problem Statement

Currently, when working with Konnect entities, the resource holding the API authentication information - `KonnectAPIAuthConfiguration` - does not have any mechanisms in place which would prevent its deletion, before the deletion of resources that utilize the Konnect API.

Since objects mapping Konnect entities (like `KonnectGatewayControlPlane`) do have finalizers themselves, the effect is that `KonnectAPIAuthConfiguration` gets deleted but the Konnect entity resource remains with the following status condition:

```
    conditions:
    - lastTransitionTime: "2025-05-08T17:07:58Z"
      message: Referenced KonnectAPIAuthConfiguration / not found
      observedGeneration: 2
      reason: RefNotFound
      status: "False"
      type: APIAuthResolvedRef
```

## Proposed Solution

Add a `gateway.konghq.com/api-auth-in-use` finalizer to `KonnectAPIAuthConfiguration`s that are being used to prevent its deletion before all entities that use it get deleted first.

## Acceptance Criteria
- [ ] As a user I can delete a namespace which contains `KonnectAPIAuthConfiguration` as well Konnect resources using that auth configuration and can expect that the mapped Konnect entities will get deleted. After that happens all the resources and the `KonnectAPIAuthConfiguration` will get deleted, unblocking namespace deletion.

amontalban:
I just hit this problem during the upgrading tests from 1.5 to 1.6 and was going to report it today 😄.

As a workaround I did the following, in case it helps someone in the meantime:

```
 kubectl -n NAMESPACE get secret konnect-extension-konnect-config-j79pv -o=json | jq '.metadata.finalizers = null' | kubectl apply -f -
 kubectl -n NAMESPACE get konnectextensions.konnect.konghq.com konnect-config -o=json | jq '.metadata.finalizers = null' | kubectl apply -f -
```

Replace the secret and extension accordingly.

pmalek:
Still an issue, removing `stale` label and adding to 2.1 milestone tentatively.

PR #2625, opened 2025-11-17
fix(konnect): improve KonnectExtension reconciliation and cleanup error handling

tao12345666333:

**Which issue this PR fixes**

Related to https://github.com/Kong/kong-operator/issues/2622

In this PR, I tried using Copilot to analyze the failure logs and let AI make the changes, but I found that it still couldn't accurately complete all the modifications. 🤣

**PR Readiness Checklist**:

~- [ ] the `CHANGELOG.md` release notes have been updated to reflect significant changes~


controller/konnect/konnectextension_controller.go
```diff
@@ -645,3 +689,55 @@ func (r *KonnectExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		RequeueAfter: r.SyncPeriod,
 	}, nil
 }
+
+func (r *KonnectExtensionReconciler) skipKonnectCleanup(ctx context.Context, logger logr.Logger, ext *konnectv1alpha2.KonnectExtension) (ctrl.Result, error) {
```
pmalek: Can we add godoc for this method explaining what it does (and why)?

controller/konnect/konnectextension_controller_test.go
```diff
@@ -0,0 +1,78 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            "konnect-cert",
+			Namespace:       "default",
+			Finalizers:      []string{KonnectCleanupFinalizer, consts.KonnectExtensionSecretInUseFinalizer},
+			ResourceVersion: "1",
+		},
+	}
+
+	client := fakeclient.NewClientBuilder().
+		WithScheme(testScheme).
```
pmalek:
```suggestion
		WithScheme(scheme.Get()).
```

controller/konnect/konnectextension_controller_test.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/kong/kong-operator/modules/manager/scheme"
+	"github.com/kong/kong-operator/pkg/consts"
+)
+
+func TestSkipKonnectCleanupRemovesFinalizers(t *testing.T) {
+	testScheme := scheme.Get()
+	utilruntime.Must(corev1.AddToScheme(testScheme))
+	utilruntime.Must(operatorv1beta1.AddToScheme(testScheme))
+	utilruntime.Must(konnectv1alpha2.AddToScheme(testScheme))
+
```
pmalek:
```suggestion
```

controller/konnect/konnectextension_controller.go
```diff
@@ -186,6 +187,9 @@ func (r *KonnectExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
 			if k8serrors.IsConflict(err) {
 				return ctrl.Result{Requeue: true}, nil
 			}
+			if k8serrors.IsNotFound(err) {
+				return ctrl.Result{}, nil
+			}
```
pmalek: This seems to repeat quite a lot. I believe it would be a good idea to extract this into a function which would for example return a bool indicating whether to return from the caller. WDYT?

mlavacca: I don't think we want to silently ignore `NotFound` errors. These checks are after Update attempt, so `NotFound` means that something weird is happening, and proper error handling is required in my opinion.

controller/konnect/konnectextension_controller.go
mlavacca: This function, in case there is no possibility to clean up the Konnect environment, leaves leftovers in Konnect and instead cleans up the finalizer from the `KonnectExtension`. I strongly think we should go in the other direction: in case konnect communication is broken, fix it first, then all the resources will be properly collected. This will be eased by https://github.com/Kong/kong-operator/issues/1586 and https://github.com/Kong/kong-operator/issues/1586

controller/konnect/konnectextension_controller.go
pmalek: This might happen for example when extension did not have the finalizer set yet and it got deleted between being queued for reconciliation and the `.Update()` call.

If the extension (or in general an object) is deleted, then there's no other action that we can do for it (apart from handling other resources that depend on this) IMHO.

WDYT?

mlavacca:
This PR is touching many places that seem unrelated to the issue in the object. I've opened https://github.com/Kong/kong-operator/pull/2712 to address that specific issue. Let's discuss this one separately, as it aims at fixing other potential issues.

tao12345666333:
OK, let's review #2712 first. I will reply to these review comments

controller/konnect/konnectextension_controller.go
tao12345666333: Sure.

test/helpers/setup.go
```diff
@@ -31,7 +31,9 @@ func SetupTestEnv(t *testing.T, ctx context.Context, env environments.Environmen
 		t.Helper()
 
 		t.Log("performing test cleanup")
-		ctx, cancel := context.WithTimeout(context.Background(), time.Minute*5)
+		// Use a longer timeout for cleanup to avoid flakiness when namespaces
+		// take longer to terminate due to finalizers on CRDs (e.g., Konnect resources).
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
```
pmalek: I don't think this will help here, 5 minutes should already be enough to clean up the tests, konnect resources and finalizers.

IMHO if something is taking more, then the issue somewhere else.

controller/konnect/reconciler_generic.go
```diff
@@ -167,6 +167,9 @@ func (r *KonnectEntityReconciler[T, TEnt]) Reconcile(
 					if k8serrors.IsConflict(err) {
 						return ctrl.Result{Requeue: true}, nil
 					}
+					if k8serrors.IsNotFound(err) {
+						return ctrl.Result{}, nil
+					}
```
pmalek: As commented elsewhere: I believe we can refactor all these into a single function, WDYT?

IMO returning no error here when the updated entity is not found (during finalizer removal) is OK: there's nothing more to be done now, the CR is gone.

controller/konnect/konnectextension_controller_utils.go
```diff
@@ -262,22 +265,26 @@ func enforceKonnectExtensionStatus(
...
-		ClusterType:    konnectClusterTypeToCRDClusterType(cp.Status.ClusterType),
-		AuthRef:        &apiAuthRef,
-	}
 
-	if cp.Status.Endpoints != nil {
-		expectedKonnectStatus.Endpoints = konnectv1alpha2.KonnectEndpoints{
-			ControlPlaneEndpoint: cp.Status.Endpoints.ControlPlaneEndpoint,
-			TelemetryEndpoint:    cp.Status.Endpoints.TelemetryEndpoint,
+	// Only update KonnectExtension.Status.Konnect when Konnect endpoints are available.
+	// The Status schema requires non-empty endpoints; updating with empty values triggers validation errors.
```
pmalek: I _think_ that making these fields [required](https://github.com/Kong/kong-operator/blob/14594485dd0045852d1579cb141e5684f8d6dfa9/api/konnect/v1alpha2/konnect_extension_types.go#L243-L259) was a mistake: now when for example the CP gets missing/deleted/? the status cannot be updated to reflect this fact, it will keep the stale information.

This is a discussion for another time though. Just leaving a comment here to make sure we have an anchor for the discussion.

controller/konnect/konnectextension_controller.go
```diff
@@ -627,6 +648,9 @@ func (r *KonnectExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		&ext,
 		readyCondition,
 	); err != nil || updated || !res.IsZero() {
+		if err != nil && k8serrors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
```
pmalek:
```suggestion
		if k8serrors.IsNotFound(err) {
			return ctrl.Result{}, nil
		}
```

It's safe to use `nil` with `IsNotFound()`

```
func TestX(t *testing.T) {
	var err error
	require.False(t, k8serrors.IsNotFound(err)) // Passes
}
```

controller/konnect/konnectextension_controller.go
```diff
@@ -596,22 +608,31 @@ func (r *KonnectExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		&ext,
 		certProvisionedCond,
 	); err != nil || updated || !res.IsZero() {
+		if err != nil && k8serrors.IsNotFound(err) {
```
pmalek:
```suggestion
		if k8serrors.IsNotFound(err) {
```

controller/konnect/konnectextension_controller.go
```diff
@@ -381,6 +386,9 @@ func (r *KonnectExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
 			readyCondition,
 			certProvisionedCond,
 		); err != nil || updated || !res.IsZero() {
+			if err != nil && k8serrors.IsNotFound(err) {
```
pmalek:
```suggestion
			if k8serrors.IsNotFound(err) {
```

Same elsewhere

test/helpers/setup.go
tao12345666333: OK, I will revert this change. https://github.com/Kong/kong-operator/pull/2625/commits/c717e2b8d7ee23bdb9488fc256b2dbc643131177

controller/konnect/reconciler_generic.go
tao12345666333: Thanks, updated https://github.com/Kong/kong-operator/pull/2625/commits/8756177d943ebd2cf70bff34dfca93149ad6f77d

controller/konnect/konnectextension_controller.go
tao12345666333: Sure, updated. https://github.com/Kong/kong-operator/pull/2625/commits/8756177d943ebd2cf70bff34dfca93149ad6f77d

tao12345666333: Yes. When `Update()` returns `NotFound`, the object is already gone, so there's nothing more to do.

controller/konnect/konnectextension_controller.go
tao12345666333: @mlavacca thanks for the comment.

Just to clarify your concern, are you suggesting that we should use finalizers on `KonnectAPIAuthConfiguration` to prevent it from being deleted before dependent resources are cleaned up?

Note: #2726 has already added protection for `KonnectAPIAuthConfiguration` deletion. This `skipKonnectCleanup` is meant as a fallback for edge cases (e.g., KonnectExtension never successfully connected), WDYT?

controller/dataplane/controller.go
```diff
@@ -219,6 +222,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	deployment, res, err := deploymentBuilder.BuildAndDeploy(ctx, dataplane, r.EnforceConfig, r.ValidateDataPlaneImage)
 	if err != nil {
+		// If Konnect cert Secret isn't ready yet, requeue instead of logging an error stack.
+		if errors.Is(err, certificates.ErrKonnectSecretMissing) {
+			log.Debug(logger, "Konnect certificate Secret not ready yet; requeuing", "dataplane", dpNn)
```
pmalek: The logger that we get [from the context](https://github.com/Kong/kong-operator/blob/5cf7bf832e1cfc2535b1c754c4453438511e5851/controller/dataplane/controller.go#L71) and pass around should already contain dataplane's NN

```suggestion
			log.Debug(logger, "Konnect certificate Secret not ready yet; requeuing")
```

controller/konnect/konnectextension_controller.go
```diff
@@ -817,22 +825,31 @@ func (r *KonnectExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
-		log.Debug(logger, "updating KonnectExtension status")
-		err := r.Client.Status().Update(ctx, &ext)
-		if k8serrors.IsConflict(err) {
-			// in case the err is of type conflict, don't return it and instead trigger
-			// another reconciliation.
-			// This is just to prevent spamming of conflict errors.
-			return ctrl.Result{RequeueAfter: ctrlconsts.RequeueWithoutBackoff}, nil
+	// Only update status when cp is available (not during cleanup with missing ControlPlane).
+	// During deletion, if the ControlPlane has been removed, we don't need to update the status.
+	if cp != nil {
```
pmalek: `cp != nil` is the happy path. What I'd suggest is to invert the check here to make it `if cp == nil ...` and update status there and return.

Then we continue with reconciliation if `cp` is not `nil`. WDYT?

controller/konnect/konnectextension_controller.go
```diff
@@ -817,22 +825,31 @@ func (r *KonnectExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+		authRef := konnectv1alpha2.KonnectAPIAuthConfigurationRef{
+			Name: apiAuth.Name,
+		}
+		if enforceKonnectExtensionStatus(*cp, authRef, *certificateSecret, &ext) {
+			log.Debug(logger, "updating KonnectExtension status")
+			// Ensure we don't write an incomplete Konnect status that fails CRD validation.
+			if ext.Status.Konnect != nil {
+				if ext.Status.Konnect.Endpoints.ControlPlaneEndpoint == "" || ext.Status.Konnect.Endpoints.TelemetryEndpoint == "" {
```
pmalek: @mlavacca This is a Q to you I guess: I recall a discussion about making these fields required (ref: https://github.com/Kong/kong-operator/commit/40f078703a2f08ed58f5f237210af5edc19a7ba4).

This unfortunately precludes updates like this which would unset the fields. I think we should revisit this decision and relax the requirement here to make these fields optional.

Leaving stale status in the object feels wrong IMHO.

Related comment in this PR: https://github.com/Kong/kong-operator/pull/2625/files#r2570880136

controller/konnect/konnectextension_controller.go
```diff
@@ -912,3 +932,59 @@ func enforceSecretInUseFinalizer(
...
+
+	res, certificateSecret, err := r.getCertificateSecret(ctx, *ext, true)
+	if client.IgnoreNotFound(err) != nil {
+		return ctrl.Result{}, err
+	}
+	if res != op.Noop {
+		return ctrl.Result{}, nil
+	}
+
+	if err == nil {
```
pmalek: Inverse the logic here and make the happy path least indented?

controller/konnect/reconciler_utils.go
```diff
@@ -0,0 +1,27 @@
...
+//   - For not found errors: returns (ctrl.Result{}, nil, true)
+//   - For other errors: returns (ctrl.Result{}, err, true)
+//   - For nil error: returns (ctrl.Result{}, nil, false)
+func handleUpdateError(err error) (ctrl.Result, error, bool) {
+	if err == nil {
+		return ctrl.Result{}, nil, false
+	}
+	switch {
+	case k8serrors.IsConflict(err):
+		return ctrl.Result{Requeue: true}, nil, true
```
pmalek: [`Requeue`](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.22.4/pkg/reconcile#Result.Requeue) is deprecated and we should use `RequeueAfter` instead.

I'd suggest we use https://github.com/Kong/kong-operator/blob/0d9479acad3af9ca336edc81926fb9a196586ad3/controller/consts/requeue.go#L5-L8 for this.

pmalek (commented):
@tao12345666333 Do you mind adding a description to this PR elaborating why we need these changes?

controller/konnect/konnectextension_controller.go
pmalek: The reason I'm bringing this up is because this comes up in test failures a lot:

```
2025-12-08T11:17:55Z	ERROR	Reconciler error	{"controller": "konnectextension", "controllerGroup": "konnect.konghq.com", "controllerKind": "KonnectExtension", "KonnectExtension": {"name":"konnect-extension-wgd4g","namespace":"0fd9cfe0-4375-4ab1-a035-dcd290345526"}, "namespace": "0fd9cfe0-4375-4ab1-a035-dcd290345526", "name": "konnect-extension-wgd4g", "reconcileID": "cb922a8b-1208-42bd-85ad-4abb37d74aff", "error": "KonnectExtension.konnect.konghq.com \"konnect-extension-wgd4g\" is invalid: [status.konnect.endpoints.controlPlane: Invalid value: \"\": konnect.endpoints.controlPlane in body should be at least 1 chars long, status.konnect.endpoints.telemetry: Invalid value: \"\": konnect.endpoints.telemetry in body should be at least 1 chars long]"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler
	/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.22.4/pkg/internal/controller/controller.go:474
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem
	/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.22.4/pkg/internal/controller/controller.go:421
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func1.1
	/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.22.4/pkg/internal/controller/controller.go:296
```

controller/konnect/konnectextension_controller.go
tao12345666333: If we use `if err != nil` here, we can't return directly.

It will looks like this:

```
if err != nil {
  // we can't return directly, need to handle finalizers
}
// secret exists, handle finalizers 
```

WDYT?

pmalek: That's a valid point. Let's leave this as is.

controller/konnect/konnectextension_controller.go
```diff
@@ -817,22 +825,31 @@ func (r *KonnectExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+		}
+		if enforceKonnectExtensionStatus(*cp, authRef, *certificateSecret, &ext) {
+			log.Debug(logger, "updating KonnectExtension status")
+			// Ensure we don't write an incomplete Konnect status that fails CRD validation.
+			if ext.Status.Konnect != nil {
+				if ext.Status.Konnect.Endpoints.ControlPlaneEndpoint == "" || ext.Status.Konnect.Endpoints.TelemetryEndpoint == "" {
+					// Drop Konnect status until endpoints are available.
+					ext.Status.Konnect = nil
+				}
+			}
```
pmalek: I believe this should be done inside of `enforceKonnectExtensionStatus` to perform this one task in one place. WDYT?

mlavacca (commented):
Hi @tao12345666333, as I said in another comment, I am afraid that this PR is potentially very disruptive if we don't carefully proceed evaluating every change independently. The code is pretty complex here, and the chain of unforeseen consequences can be pretty high.

Though my proposal: instead of having a generic "KonnectExtension improvement" scope, what do you think if we scope it down and address a specific problem? I am referring to what @pmalek brought up about the Endpints being required. The reasoning behind that design decision is that in case the outer struct `enforceKonnectExtensionStatus` is set, all the inner fields must be set. What should be fixed, however, is that in case the CP is not found or doesn't exist, we should clean up that outer struct altogether, rather than focusing on the inner fields. If you think we can scope it down to fix that exact problem, I am more than happy to push this PR to the finish line, I'm sorry it's taking so long.

Let me know what you think.

controller/konnect/konnectextension_controller.go
tao12345666333: Good point. I agree with you.

tao12345666333:
@mlavacca Sure. We can scope it down and address the specific problem.

Here's my plan:

  This PR will focus only on `enforceKonnectExtensionStatus`

  I'll split the other changes into separate PRs:
  - skipKonnectCleanup logic for handling deletion when Konnect credentials are unavailable
  - handleUpdateError helper and related refactoring
  - IsNotFound error handling improvements
  - Other cleanup/improvements

  I'll update this PR shortly with only the scoped-down changes. Let me know if this approach works for you.

mlavacca:
> @mlavacca Sure. We can scope it down and address the specific problem.
> 
> Here's my plan:
> 
> This PR will focus only on `enforceKonnectExtensionStatus`
> 
> I'll split the other changes into separate PRs:
> 
>     * skipKonnectCleanup logic for handling deletion when Konnect credentials are unavailable
> 
>     * handleUpdateError helper and related refactoring
> 
>     * IsNotFound error handling improvements
> 
>     * Other cleanup/improvements
> 
> 
> I'll update this PR shortly with only the scoped-down changes. Let me know if this approach works for you.

That would be great, thank you! 

I have a proposal, though: what do you think if we create separate issues for the other points you mentioned, and we prioritize them properly?

tao12345666333:
Sure! I will do that.

pmalek (approved):

PR #2643, opened 2025-11-18
chore(refactor): move all scripts, generators to `hack` directory

programmer04:

Move all generators to `hack/generators` remove redundant files, and ensure that all generators work (generated diff changes order of import, probably related to how it's done by default formatting after generation).

Furthemore, for `cleanup` script port gke too, now it's not a part of CI run, but in the future (when e2e tests will run again) it may be useful as it's in KIC. Basically, it's continuation of work initally done in https://github.com/Kong/kong-operator/pull/2172

**Which issue this PR fixes**

Part of
- #1658


hack/cleanup/gke_clusters.go
```diff
@@ -12,11 +12,12 @@ import (
 	"github.com/go-logr/logr"
 	"google.golang.org/api/option"
 
-	"github.com/kong/kong-operator/ingress-controller/test/e2e"
+	e2e "github.com/kong/kong-operator/ingress-controller/test/e2e"
```
randmonkey: Why should we assign an alias of the import here?

randmonkey (commented):
LGTM, only a small question.

hack/cleanup/gke_clusters.go
programmer04: My bad, I fixed it. Thx

tao12345666333 (approved):

PR #2656, opened 2025-11-19
feat(hybridgateway): rework generated resources lifecycle

### LINKED ISSUE #2683: HybridGateway: rework generated resources ownership
## Problem Statement

Currently the generated resources are created adding the source HTTPRoute as Owner reference.
This enables automatic clean-up when the HTTPRoute is deleted.
With the change introduced with #2542 we now support sharing the same resources among different HTTPRoutes to avoid duplication.
Anyway, now each generated Resource may have multiple HTTPRoute "owners". The Owner reference is no more a good way to manage the resources lifecycle.

## Proposed Solution
Use a custom annotation on each generated resource to track the source HTTPRoutes.
The annotation should carry all the HTTPRoutes that own the generated resource as a CSV value.
This will require checking actual resource state during the translation phase in the HybridGateway controller, which is a pre-requisite to manage CRUD operations ordering.

fgiudici:

This PR reworks the life-cycle of the Kong resources generated by the Hybrid Gateway controller.
With the introduction of #2566 the Kong resources generated from `HTTPRoute`s can be shared by different `HTTPRoute`s.
This requires to drop the single Owner reference in the generated Kong resources in favor of a multi-ownership strategy.

This PR drops the Owner references from the generated resources and adds a tracking annotation where all the `HTTPRoute`s sharing the resource will add themselves.
A generated resource is removed only when no more `HTTPRoute`s appear in the tracking annotation.

The tracking annotation is:
`gateway-operator.konghq.com/hybrid-routes`

**Which issue this PR fixes**

Fixes #2683 

The current code has some duplication that can be avoided with better factorization.
This is left for #2684 and the plan is to address it once all the functionality is there (i.e., after #2685).

**PR Readiness Checklist**:


controller/hybridgateway/controller.go
```diff
@@ -129,6 +131,22 @@ func (r *HybridGatewayReconciler[t, tPtr]) Reconcile(ctx context.Context, req ct
 	gvk := obj.GetObjectKind().GroupVersionKind()
 	log.Debug(logger, "Reconciling object", "Group", gvk.Group, "Kind", gvk.Kind)
 
+	// Handle deletion and finalizer cleanup
+	if obj.GetDeletionTimestamp() != nil {
+		return r.handleDeletion(ctx, logger, obj, rootObj)
```
mlavacca, line 137: Can we handle deletion gracefully? Something similar to https://github.com/Kong/kong-operator/blob/aa12e0421b12a21a4436b2c163fcf37ddf29bd7e/controller/gateway/controller_cleanup.go#L50-L57

controller/hybridgateway/controller.go
```diff
@@ -129,6 +131,22 @@ func (r *HybridGatewayReconciler[t, tPtr]) Reconcile(ctx context.Context, req ct
...
+	}
+
+	// Ensure finalizer is present
+	if !controllerutil.ContainsFinalizer(obj, routeconst.RouteFinalizer) {
+		log.Debug(logger, "Adding finalizer")
+		controllerutil.AddFinalizer(obj, routeconst.RouteFinalizer)
+		if err := r.Update(ctx, obj); err != nil {
```
mlavacca: We should use `Patch` here, instead of Update, to minimize the risks of conflicts

controller/hybridgateway/controller.go
```diff
@@ -234,3 +252,64 @@ func (r *HybridGatewayReconciler[t, tPtr]) Reconcile(ctx context.Context, req ct
...
+// and removing the finalizer. This ensures that all Kong resources generated from the Route
+// are properly cleaned up before the Route is deleted from the cluster.
+func (r *HybridGatewayReconciler[t, tPtr]) handleDeletion(ctx context.Context, logger logr.Logger, obj tPtr, rootObj t) (ctrl.Result, error) {
+	log.Debug(logger, "Handling Route deletion")
+
+	// Only process deletion if our finalizer is present
+	if !controllerutil.ContainsFinalizer(obj, routeconst.RouteFinalizer) {
+		log.Debug(logger, "Finalizer not present, skipping cleanup")
+		return ctrl.Result{}, nil
+	}
```
mlavacca: I can see why doing this filtering here, but I do not think it's harmful to run the cleanup logic even for Routes without a finalizer. I can imagine weird scenarios where the resource is in deletion and the finalizer (maybe during the grace period) is manually removed. This would skip the cleanup logic, leaving leftovers. It's definitely a corner case, but I don't see potential issues in running the logic below regardless of the finalizer's existence. WDYT?

controller/hybridgateway/controller.go
```diff
@@ -234,3 +252,64 @@ func (r *HybridGatewayReconciler[t, tPtr]) Reconcile(ctx context.Context, req ct
...
+			obj,
+			corev1.EventTypeNormal,
+			routeconst.EventReasonOrphanCleanupSucceeded,
+			"Route deletion cleanup completed successfully",
+		)
+	}
+
+	// Remove finalizer
+	controllerutil.RemoveFinalizer(obj, routeconst.RouteFinalizer)
+	if err := r.Update(ctx, obj); err != nil {
```
mlavacca: Aditto about Update

controller/hybridgateway/reconciler_utils.go
```diff
@@ -272,11 +272,28 @@ func cleanOrphanedResources[t converter.RootObject, tPtr converter.RootObjectPtr
...
+				if !am.RemoveRouteFromAnnotation(&item, rootObjPtr) {
+					log.Trace(logger, "Route annotation not found, skipping resource", "kind", item.GetKind(), "obj", client.ObjectKeyFromObject(&item))
+					continue
+				}
+
+				// Not in desired output, Route resource found and removed from the hybrid-routes annotation.
+				// If other Routes are still present in the annotation, update the resource.
+				if hybridAnnotations := am.GetRoutes(&item); len(hybridAnnotations) > 0 {
+					log.Debug(logger, "Updating hybrid-routes annotation", "kind", item.GetKind(), "obj", client.ObjectKeyFromObject(&item))
+					if err := cl.Update(ctx, &item); err != nil && !errors.IsNotFound(err) {
```
mlavacca: Patch vs. Update

controller/hybridgateway/service/service.go
```diff
@@ -0,0 +1,101 @@
...
+		return &service, nil
+	}
+
+	// KongService exists, update annotations to include current HTTPRoute
+	log.Debug(logger, "KongService found")
+	annotationManager := metadata.NewAnnotationManager(logger)
+	service.Annotations[consts.GatewayOperatorHybridRoutesAnnotation] = existingService.Annotations[consts.GatewayOperatorHybridRoutesAnnotation]
+	annotationManager.AppendRouteToAnnotation(&service, httpRoute)
+
+	// TODO: we should check that the existingService.Spec matches what we expect
```
mlavacca, line 97: Can we create an issue for this and link it here?

controller/hybridgateway/controller.go
fgiudici: After discussion we decided to not deal with the grace period for HTTPRoute deletion.

controller/hybridgateway/controller.go
fgiudici: Makes sense, I don't see anything bad that could happen if we run the cleanup logic in such a corner case.
So, no need to even manage this case 👍🏼

fgiudici: done

controller/hybridgateway/service/service.go
fgiudici: 😒 done

mlavacca (approved):

PR #2662, opened 2025-11-20
fix(ingress-controller): Add translation of healthchecks.threshold in upstreams

randmonkey:

Migration of https://github.com/Kong/kubernetes-ingress-controller/pull/7784 to fix the missing of translating `healthchecks.threshold`.

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -152,6 +152,9 @@
   [#2427](https://github.com/Kong/kong-operator/pull/2427)
 - Hybrid Gateway: attach KongService generation to BackendRefs and fix filter/plugin conversion.
   [#2456](https://github.com/Kong/kong-operator/pull/2456)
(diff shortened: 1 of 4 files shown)


pmalek (approved):
👍 

Since [Kong's documentation explicitly states](https://developer.konghq.com/gateway/entities/upstream/#schema-upstream-healthchecks-threshold) that threshold can be in the range: `>= 0` x `<= 100` I'd even suggest that we can:

- verify if and report and error if it's outside of this range ( this can be done separately, as users would get configuration apply failed event anyway for this)
- add CRD validation rule to enforce the range of threshold in https://github.com/Kong/kong-operator/blob/d6e8e3abf0b4a3af0cc9b01d0405196c807506e8/api/configuration/v1beta1/kongupstreampolicy_types.go#L177-L179

These can be done separately but I believe it would be worth doing.

Issue #2176, opened 2025-09-02
Enable validation webhook for test suites


programmer04:
## Problem Statement

Spinoff of

- https://github.com/Kong/kong-operator/issues/1367

Read the description for more context.

Validation webhook is enabled by default for the controller, but in test suites, e.g., `integration`, it's disabled. 

Enhance the machinery of running those suites to include all the stuff required for the webhook to run, to be able to cover it with some integration tests too.

> It probably makes sense to do it when KO will have well-defined test suites 

## Acceptance Criteria

- [x] identify test suites that should be run with webhook enabled
- [x] allow running them with webhook enabled

programmer04:
It's done, two more specific issues for porting some leftover tests have been created
- https://github.com/Kong/kong-operator/issues/2895
- https://github.com/Kong/kong-operator/issues/2896

PR #2675, opened 2025-11-20
chore(tests): re-enable validating webhook tests

programmer04:

**Which issue this PR fixes**

Fixes #

- #2176


programmer04:
This PR contains some messy code that allows running the integration test suite with the webhook enabled. Connectivity from the cluster to the local is established with the telepresence replace command. There are two problems with this approach. The minor one - mounting secrets referred to by K8s Deployment locally requires a lot of gymnastics on macOS [like installing a kernel module for having sshfs](https://github.com/Kong/kong-operator/pull/2866). Thus workaround is used.

The critical problem that makes this approach unsuitable is stability. The Deployment that represents the operator in the cluster and intercepts gets evicted a couple of times during the run of the test suite, which leads to a lack of connection and failures. The only information is 

```yml
- lastProbeTime: null
  lastTransitionTime: "2025-12-07T17:26:26Z"
  message: 'Eviction API: evicting'
  reason: EvictionByEvictionAPI status: "True"
  type: DisruptionTarget
```

Spending quite a lot of time with this problem hasn't brought any resolution. Furthermore, there is a similar issue in the telerpesence repo

- https://github.com/telepresenceio/telepresence/issues/3998

without any additional info.

To sum up, this way can't be used now, maybe in the future.

Issue #733, opened 2024-10-10
Quick subsequent reconciliations cause errors caused by using cached client


pmalek:
### Problem statement

`controller-runtime` (use by the operator) uses a cached client.

This causes issues when quick, successive reconciliations of a resource happen and the subsequent iterations do not get an up to date view on an updated/patched object. For example:

- calls an API (either external or kubeapi-server's) to create a resource
- the request was successfull, object's status is patched with ID, controller returns from reconciliation loop (update/patch will cause a requeue)
- another reconciliation happens
- controller (using controller-runtime's client) gets the object and sees that there is no ID in status, attempts to call create again
- this can result in
  - an error if the call specificies unique createria (like ID, or name which has to be unique)
  - or duplicated resources

Up until now, this has been taken care of by calling one of the [`Reduce*()` helper functions](https://github.com/Kong/gateway-operator/blob/f567172b20074fd176ef048562fb4dec6f3caf3b/pkg/utils/kubernetes/reduce/reduce.go) which are aware of the objects' schema and their priorities so that surplus objects can be removed without consequences.

With integration with external APIs (e.g. Konnect API in #370) this does not work as the resource is created in external API and more importantly the create API calls will result in `HTTP 409 Conflict`s.

Exemplar logs

```
request:
POST /v2/control-planes HTTP/1.1
Host: eu.api.konghq.tech
Content-Length: 290
Accept: application/json
Authorization: Bearer kpat_XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
Content-Type: application/json
Accept-Encoding: gzip

{"name":"test1","labels":{"app":"test1","k8s-generation":"1","k8s-group":"konnect.konghq.com","k8s-kind":"KonnectGatewayControlPlane","k8s-name":"test1","k8s-namespace":"default","k8s-uid":"0c2b1653-2839-4a4e-9c42-15ec8bcacc01","k8s-version":"v1alpha1","key1":"test1","session":"eng-demo"}}
response:
HTTP/2.0 201 Created
Connection: close
Access-Control-Allow-Credentials: true
Access-Control-Expose-Headers: Date,x-datadog-trace-id,Konnect-Acting-As,Konnect-Feature-Set
[...]

PR #2696, opened 2025-11-24
fix(konnect): use uncached API reader for secret lookups

asabitov:
## Problem

When using `KonnectAPIAuthConfiguration` with `type: secretRef` to reference secrets managed by External Secrets Operator (ESO), the operator fails with `Secret "konnect-token" not found` errors even though:
- The secret exists in the cluster
- The secret has the correct label (`konghq.com/credential: konnect`)
- RBAC permissions are correct
- Test pods can successfully read the secret

This issue affects both:
- `KonnectAPIAuthConfiguration` resources directly
- `KonnectGatewayControlPlane` and other Konnect entities that reference `KonnectAPIAuthConfiguration`

## Root Cause

The reconcilers use the cached client (`client.Client`) for secret lookups. The cache can become stale when secrets are created/updated by ESO, causing the operator to miss recently created secrets.

## Solution

This PR changes the secret lookup to use the uncached API reader (`mgr.GetAPIReader()`) which directly queries the API server, bypassing the cache and ensuring we always get the latest secret data.

**Note:** We tested the requeue approach suggested during review, but it didn't resolve the cache staleness issue. The controller's cache doesn't properly sync with dynamically created secrets from ESO, even after multiple requeues. Using an uncached client is the only reliable way to access ESO-managed secrets in this scenario.

### `KonnectAPIAuthConfigurationReconciler`
- Added `apiReader client.Reader` field to `KonnectAPIAuthConfigurationReconciler`
- Initialize `apiReader` with `mgr.GetAPIReader()` in `SetupWithManager()`
- Use `apiReader` instead of cached `client` for secret lookups in `getTokenFromKonnectAPIAuthConfiguration()`
- Updated function signature to accept `client.Reader` instead of `client.Client`

### `KonnectEntityReconciler` (Generic Reconciler)
- Added `apiReader client.Reader` field to `KonnectEntityReconciler` struct
- Initialize `apiReader` with `mgr.GetAPIReader()` in `SetupWithManager()`
- Use `apiReader` instead of cached `Client` for secret lookups in `getTokenFromKonnectAPIAuthConfiguration()`
- This ensures `KonnectGatewayControlPlane` and other Konnect entities can reliably access ESO-managed secrets

## Testing

✅ Tested in production EKS cluster  
✅ Verified `KonnectAPIAuthConfiguration` status changes from `Invalid` to `Valid` after fix  
✅ Verified `KonnectGatewayControlPlane` can successfully use ESO-managed secrets  
✅ Confirmed secret is successfully read and token validated  
✅ Tested requeue approach - confirmed it doesn't resolve cache staleness with ESO secrets

## Related Issues

This fixes the issue where ESO-managed secrets cannot be used with `KonnectAPIAuthConfiguration` and Konnect entities that depend on it.

### DIFF
--- a/controller/konnect/reconciler_generic.go
+++ b/controller/konnect/reconciler_generic.go
@@ -44,6 +44,7 @@ type KonnectEntityReconciler[T constraints.SupportedKonnectEntityType, TEnt cons
 	sdkFactory              sdkops.SDKFactory
 	CacheSyncTimeout        time.Duration
 	Client                  client.Client
+	apiReader               client.Reader
 	LoggingMode             logging.Mode
 	MaxConcurrentReconciles uint
 	SyncPeriod              time.Duration
@@ -111,6 +112,7 @@ func NewKonnectEntityReconciler[
 
 // SetupWithManager sets up the controller with the given manager.
 func (r *KonnectEntityReconciler[T, TEnt]) SetupWithManager(ctx context.Context, mgr ctrl.Manager) error {
+	r.apiReader = mgr.GetAPIReader()
 	var (
 		e              T
 		ent            = TEnt(&e)
@@ -356,7 +358,7 @@ func (r *KonnectEntityReconciler[T, TEnt]) Reconcile(
 		return res, retErr
 	}
 
-	token, err := getTokenFromKonnectAPIAuthConfiguration(ctx, r.Client, &apiAuth)
+	token, err := getTokenFromKonnectAPIAuthConfiguration(ctx, r.apiReader, &apiAuth)
 	if err != nil {
 		if res, errStatus := patch.StatusWithCondition(
 			ctx, r.Client, &apiAuth,
--- a/controller/konnect/reconciler_konnectapiauth.go
+++ b/controller/konnect/reconciler_konnectapiauth.go
@@ -30,6 +30,7 @@ import (
 type KonnectAPIAuthConfigurationReconciler struct {
 	sdkFactory  sdkops.SDKFactory
 	client      client.Client
+	apiReader   client.Reader
 	loggingMode logging.Mode
 }
 
@@ -59,6 +60,7 @@ func NewKonnectAPIAuthConfigurationReconciler(
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *KonnectAPIAuthConfigurationReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager) error {
+	r.apiReader = mgr.GetAPIReader()
 	secretLabelPredicate, err := predicate.LabelSelectorPredicate(
 		metav1.LabelSelector{
 			MatchLabels: map[string]string{
@@ -116,7 +118,7 @@ func (r *KonnectAPIAuthConfigurationReconciler) Reconcile(
 		return ctrl.Result{}, nil
 	}
 
-	token, err := getTokenFromKonnectAPIAuthConfiguration(ctx, r.client, &apiAuth)
+	token, err := getTokenFromKonnectAPIAuthConfiguration(ctx, r.apiReader, &apiAuth)
 	if err != nil {
 		if res, errStatus := patch.StatusWithCondition(
 			ctx, r.client, &apiAuth,
@@ -238,7 +240,7 @@ func (r *KonnectAPIAuthConfigurationReconciler) Reconcile(
 
 // getTokenFromKonnectAPIAuthConfiguration returns the token from the secret reference or the token field.
 func getTokenFromKonnectAPIAuthConfiguration(
-	ctx context.Context, cl client.Client, apiAuth *konnectv1alpha1.KonnectAPIAuthConfiguration,
+	ctx context.Context, cl client.Reader, apiAuth *konnectv1alpha1.KonnectAPIAuthConfiguration,
 ) (string, error) {
 	switch apiAuth.Spec.Type {
 	case konnectv1alpha1.KonnectAPIAuthTypeToken:


pmalek:
Thanks for submitting the PR. Much appreciated. Before we move forward with more details review of this I just wanted to throw it out there: possibly this is just an error reporting issue. Rather than reporting an error in https://github.com/Kong/kong-operator/blob/9ed7701d9a013def8e28ea437388ce55e2cbac20/controller/konnect/reconciler_konnectapiauth.go#L258-L260 _I think_ we should requeue, this way the controller will **eventually** get to the consistent state where it's able to observe the actual state (new `Secret`).

I'd think twice before change usage of cached to uncached client simply to prevent additional load on the API server (there's a significant cohort of users who monitor that metric quite carefully).

asabitov:
> Thanks for submitting the PR. Much appreciated. Before we move forward with more details review of this I just wanted to throw it out there: possibly this is just an error reporting issue. Rather than reporting an error in
> 
> https://github.com/Kong/kong-operator/blob/9ed7701d9a013def8e28ea437388ce55e2cbac20/controller/konnect/reconciler_konnectapiauth.go#L258-L260
> 
> _I think_ we should requeue, this way the controller will **eventually** get to the consistent state where it's able to observe the actual state (new `Secret`).
> I'd think twice before change usage of cached to uncached client simply to prevent additional load on the API server (there's a significant cohort of users who monitor that metric quite carefully).

@pmalek 

Thank you for the feedback! I tried the requeue approach as you suggested, but unfortunately it didn't resolve the underlying issue.

### What I Tried

I implemented the requeue mechanism with a configurable delay as suggested. However, the problem persisted:

1. **The cache remains stale**: Even after multiple requeues, the cached client still couldn't see the ESO-managed secret
2. **Continuous requeues**: The controller would requeue indefinitely without ever seeing the secret, even though:
   - The secret exists in the cluster (verified via `kubectl get secret`)
   - RBAC permissions are correctly configured
   - The secret is accessible from other pods/controllers

### Root Cause

The issue isn't about timing or eventual consistency - it's that the controller's cache doesn't properly sync with secrets created by External Secrets Operator (ESO). The ESO creates secrets dynamically, and the controller's cache doesn't pick them up even after waiting.

### Why Uncached Client Works

Using `mgr.GetAPIReader()` (uncached client) directly queries the API server, bypassing the cache entirely. This allows the operator to see ESO-managed secrets immediately after they're created, which is exactly what we need for this use case.

### Trade-offs

I understand the concern about API server load. However:
- Secret lookups only happen during reconciliation, which is already a relatively infrequent operation
- The secret is only read once per reconciliation cycle
- For ESO-managed secrets, this is the only reliable way to access them

I'm open to alternative approaches if you have suggestions, but the requeue mechanism didn't solve the cache staleness issue in our testing.

programmer04:
x-post for visibility https://github.com/Kong/kong-operator/issues/733 issue in the same area

pmalek:
I'm quite puzzled why the requeueing wouldn't eventually make the operator see the `Secret`.

In any case, this patch could make some users unhappy due to the API server load this could potentially incur: even if relatively minimal in some environments, change like the [one in `KonnectEntityReconciler`](https://github.com/Kong/kong-operator/pull/2696/changes#diff-5b6aadc0fe9b4d0ba8cfd31756eef41ffdcd65687cd077d3b8d427294dcc81b5L359) would cause _all_ Konnect Custom Resources (which are reconciled periodically to enforce the state in Konnect) to perform an API call to API server.

While I don't have a solution for this at this moment, there is work being done in controller-runtime (in https://github.com/kubernetes-sigs/controller-runtime/issues/3320) which will help with this and rather than try to solve this now I'd suggest we wait until that produces something useful for us.

An alternative could be using APIReader but with an opt-in gate but as mentioned, I'd prefer to use an upstream feature rather than going for something like this.

asabitov:
> I'm quite puzzled why the requeueing wouldn't eventually make the operator see the `Secret`.
> 
> In any case, this patch could make some users unhappy due to the API server load this could potentially incur: even if relatively minimal in some environments, change like the [one in `KonnectEntityReconciler`](https://github.com/Kong/kong-operator/pull/2696/changes#diff-5b6aadc0fe9b4d0ba8cfd31756eef41ffdcd65687cd077d3b8d427294dcc81b5L359) would cause _all_ Konnect Custom Resources (which are reconciled periodically to enforce the state in Konnect) to perform an API call to API server.
> 
> While I don't have a solution for this at this moment, there is work being done in controller-runtime (in [kubernetes-sigs/controller-runtime#3320](https://github.com/kubernetes-sigs/controller-runtime/issues/3320)) which will help with this and rather than try to solve this now I'd suggest we wait until that produces something useful for us.
> 
> An alternative could be using APIReader but with an opt-in gate but as mentioned, I'd prefer to use an upstream feature rather than going for something like this.

Thanks for the update, Patryk! Waiting for the upstream controller-runtime fix makes sense.

Just to add more data: we've reproduced this in a completely different setup (Pulumi, fresh EKS cluster, Kong operator v2.0.5, ESO v0.20.2). Same behaviour, -  the secret exists and is accessible via kubectl, but the cached client never sees it, even after pod restarts.

Looking forward to seeing this resolved upstream. Happy to help test when a fix is available!

pmalek:
Some work has already started upstream: https://github.com/kubernetes-sigs/controller-runtime/issues/3320.

In that spirit I'll close this PR as we're more interested in a general solution provided by upstream rather than trying to fix in several places only.

LMK your thoughts @asabitov. Thanks again for looking into this issue and providing your details findings 🙇

PR #2718, opened 2025-11-25
feat(hybridgateway): generate resources in the parentRef namespace

fgiudici:

Till now the resources were created in the source HTTPRoute namespace. This patch will instead use the parentRef namespace in order to share generated resources among HTTPRoutes living in different namespaces but referencing the same gateway (parentRef).

### DIFF
--- a/controller/hybridgateway/metadata/namespace.go
+++ b/controller/hybridgateway/metadata/namespace.go
@@ -1,16 +1,24 @@
 package metadata
 
 import (
-	gwtypes "github.com/kong/kong-operator/internal/types"
+	"github.com/samber/lo"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
+
+	gwtypes "github.com/kong/kong-operator/internal/types"
 )
 
 // NamespaceFromParentRef returns the namespace for a given Route ParentReference.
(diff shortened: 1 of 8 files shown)


controller/hybridgateway/metadata/namespace.go
```diff
@@ -0,0 +1,16 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// NamespaceFromParentRef returns the namespace for a given Route ParentReference.
+// If the ParentReference specifies a namespace, that namespace is returned.
+// Otherwise, the namespace of the provided resource is returned.
+func NamespaceFromParentRef(obj metav1.Object, pRef *gwtypes.ParentReference) string {
+	if pRef.Namespace != nil && *pRef.Namespace != "" {
+		return string(*pRef.Namespace)
+	}
```
programmer04:
```suggestion
	if ns := lo.FromPtr(pRef.Namespace); ns != "" {
		return string(ns)
	}
```
WDYT?

Furthermore, it is rather invalid to pass `pRef = nil` to this function. Are you checking it on the caller side?

fgiudici: Good catch! While the call stack heading here in current code ensures that pRef will not be nil, the check in this function of the metadata package is mandatory IMHO! 👍🏼

programmer04 (approved):

PR #2725, opened 2025-11-26
feat(testing): Bootstrap hybrid gateway e2e tests with chainsaw

alacuku:
**What this PR does / why we need it**:

This PR introduces the `chainsaw` testing framework for end-to-end (e2e) tests of the hybrid gateway controller. It serves as a pilot for future e2e tests using this framework. A basic `HTTPRoute` test scenario is included to validate the initial setup. This test covers the full lifecycle of the `HTTPRoute` resource, including creation, updates, and deletion, while also verifying data plane connectivity.

**Which issue this PR fixes**

Fixes #2522

**Special notes for your reviewer**:

The CI integration for these `chainsaw` tests is not yet in place. For now, these tests will be run manually. Subsequent PRs will focus on automating these tests within our CI pipelines.

**PR Readiness Checklist**:

Complete these before marking the PR as `ready to review`:

- [ ] the `CHANGELOG.md` release notes have been updated to reflect significant changes

PR #2757, opened 2025-11-27
chore: single library for all CRDs

programmer04:

Continuation of 

- https://github.com/Kong/kong-operator/pull/2740

that ensures the single source of truth for all CRDs paths, furthermore GWAPI CRDs are taken from the module directly to simplify tracking and avoid unnecessary network calls. 

**Which issue this PR fixes**

Part of 
- https://github.com/Kong/kong-operator/issues/1658

Code in files that contain `e2e` in the path is basically dead. All of them will be removed or migrated to this approach
- https://github.com/Kong/kong-operator/pull/2725 

so this PR only ensures it compiles even by the cost of simple comment out.

For envtest the two suites will be merged in the future, so it'll be a great opportunity for a dedicated helper.

Validating policies and installing them from Chart (`test/crdsvalidation/dataplane_validatingadmissionpolicy_test.go`) is more complicated, and it's not in the scope of this PR, so it's mostly left as it was. A dedicated issue will be created.


test/e2e/environment.go
```diff
@@ -216,8 +216,8 @@ func CreateEnvironment(t *testing.T, ctx context.Context, opts ...TestEnvOption)
 	require.NoError(t, err)
 
 	if opt.InstallViaKustomize {
-		t.Logf("deploying Gateway APIs CRDs from %s", testutils.GatewayExperimentalCRDsKustomizeURL)
-		require.NoError(t, clusters.KustomizeDeployForCluster(ctx, env.Cluster(), testutils.GatewayExperimentalCRDsKustomizeURL))
+		// t.Logf("deploying Gateway APIs CRDs from %s", testutils.GatewayExperimentalCRDsKustomizeURL)
+		// require.NoError(t, clusters.KustomizeDeployForCluster(ctx, env.Cluster(), testutils.GatewayExperimentalCRDsKustomizeURL))
```
pmalek: So this is not necessary for e2e tests?

```suggestion
```

ingress-controller/test/e2e/features_test.go
```diff
@@ -394,15 +393,15 @@ func TestDeployAllInOneDBLESSNoLoadBalancer(t *testing.T) {
...
-	require.NoError(t, clusters.KustomizeDeployForCluster(ctx, env.Cluster(), consts.GatewayExperimentalCRDsKustomizeURL))
+	// t.Logf("deploying Gateway APIs CRDs from %s", consts.GatewayExperimentalCRDsKustomizeURL)
+	// require.NoError(t, clusters.KustomizeDeployForCluster(ctx, env.Cluster(), consts.GatewayExperimentalCRDsKustomizeURL))
 
 	deployment := getManifestDeployments(dblessPath).GetController(ctx, t, env)
 	t.Log("updating controller deployment to enable Gateway feature gate")
 	for i, container := range deployment.Spec.Template.Spec.Containers {
 		if container.Name == controllerContainerName {
 			deployment.Spec.Template.Spec.Containers[i].Env = append(deployment.Spec.Template.Spec.Containers[i].Env,
-				corev1.EnvVar{Name: "CONTROLLER_FEATURE_GATES", Value: consts.DefaultFeatureGates})
```
pmalek, line 405: Why don't we leave this const as is by replace all occurrences with hardcoded value?

ingress-controller/test/e2e/features_test.go
```diff
@@ -394,15 +393,15 @@ func TestDeployAllInOneDBLESSNoLoadBalancer(t *testing.T) {
 	verifyIngressWithEchoBackends(ctx, t, env, numberOfEchoBackends)
 
 	// ensure that Gateways with no addresses come online and start ingesting routes
-	t.Logf("deploying Gateway APIs CRDs from %s", consts.GatewayExperimentalCRDsKustomizeURL)
-	require.NoError(t, clusters.KustomizeDeployForCluster(ctx, env.Cluster(), consts.GatewayExperimentalCRDsKustomizeURL))
+	// t.Logf("deploying Gateway APIs CRDs from %s", consts.GatewayExperimentalCRDsKustomizeURL)
+	// require.NoError(t, clusters.KustomizeDeployForCluster(ctx, env.Cluster(), consts.GatewayExperimentalCRDsKustomizeURL))
```
pmalek: ?

ingress-controller/test/e2e/features_test.go
```diff
@@ -274,8 +273,8 @@ func TestDeployAllInOneDBLESSGateway(t *testing.T) {
 		return true
 	}, time.Minute, 3*time.Second)
 
-	t.Logf("deploying Gateway APIs CRDs in standard channel from %s", consts.GatewayStandardCRDsKustomizeURL)
-	require.NoError(t, clusters.KustomizeDeployForCluster(ctx, env.Cluster(), consts.GatewayStandardCRDsKustomizeURL))
+	// t.Logf("deploying Gateway APIs CRDs in standard channel from %s", consts.GatewayExperimentalCRDsKustomizeURL)
+	// require.NoError(t, clusters.KustomizeDeployForCluster(ctx, env.Cluster(), consts.GatewayExperimentalCRDsKustomizeURL))
```
pmalek: ?

ingress-controller/test/e2e/features_test.go
```diff
@@ -346,8 +345,8 @@ func TestDeployAllInOneDBLESSGateway(t *testing.T) {
 	gw, err = gc.GatewayV1().Gateways(corev1.NamespaceDefault).Get(ctx, gw.Name, metav1.GetOptions{})
 	require.NoError(t, err)
 
-	t.Logf("deploying Gateway APIs CRDs in experimental channel from %s", consts.GatewayExperimentalCRDsKustomizeURL)
-	require.NoError(t, clusters.KustomizeDeployForCluster(ctx, env.Cluster(), consts.GatewayExperimentalCRDsKustomizeURL))
+	// t.Logf("deploying Gateway APIs CRDs in experimental channel from %s", consts.GatewayExperimentalCRDsKustomizeURL)
+	// require.NoError(t, clusters.KustomizeDeployForCluster(ctx, env.Cluster(), consts.GatewayExperimentalCRDsKustomizeURL))
```
pmalek: ?

test/conformance/conformance_test.go
```diff
@@ -107,7 +107,7 @@ func TestGatewayConformance(t *testing.T) {
 	timeoutConfig.HTTPRouteMustHaveCondition = looserTimeout
 
 	opts := conformance.DefaultOptions(t)
-	opts.BaseManifests = testutils.GatewayRawRepoURL + "/conformance/base/manifests.yaml"
+	opts.BaseManifests = kcfg.GatewayAPIRawURL() + "/conformance/base/manifests.yaml"
```
pmalek: Can't we also rely on the manifest from the Go module or would that be too big of a hack?

pmalek (changes requested):

ingress-controller/test/e2e/features_test.go
programmer04: I think by the end of the day we'll get rid of it entirely, but you're right, for now it should be const

test/e2e/environment.go
programmer04: Yeah, I can just remove it to keep it clean and clear

test/conformance/conformance_test.go
programmer04: done, but the implementation in upstream is trickier than it is supposed to be IMHO (without looking into src impossible to figure out different behavior for URL and path), I've even started questioning my sanity ;p

test/e2e/helm_install_upgrade_test.go
```diff
@@ -44,8 +43,9 @@ func TestHelmUpgrade(t *testing.T) {
 	const waitTime = 3 * time.Minute
 
 	var (
-		ctx      = t.Context()
-		chartDir = filepath.Join(testutils.ProjectRootPath(), "charts/kong-operator")
+		ctx = t.Context()
+		// chartDir = filepath.Join(testutils.ProjectRootPath(), "charts/kong-operator")
+		chartDir = ""
```
pmalek: ?

pmalek (approved):

PR #2847, opened 2025-12-10
fix: generating NetworkPolicy for Gateway with validation enabled

### LINKED ISSUE #161: Follow up research on broken queued up reconciliation after changing `NetworkPolicy` ObjectMeta
### Problem statement

Kong/gateway-operator-archive#993 attempted to fix a flaky test `TestGatewayDataPlaneNetworkPolicy` described in Kong/gateway-operator#184.

This issue aims to track the effort of performing research on what was a particular root cause of this.

Whether it was an `Update` of ObjectMeta that didn't cause an update of the underlying object and hence the spec of related `NetworkPolicy` didn't get an `Update` in a subsequent reonciliation request? Or maybe it was something else.

programmer04:

PR re-enables the check for proper regeneration of `NetworkPolicy` in case of changes in `GatewayConfiguration`. It contains the fix that makes this regeneration more stable (get rid of flakiness) - change of order.

The test used to check the regeneration of `NetworkPolicy` after adjustment of `KONG_PORT_MAPS` in `GatewayConfiguration`, but now it can't be tested with `config/default/validating_policies` applied. Update of `DataPlane` results in 

```
Patch Failed dataplanes.gateway-operator.konghq.com "2e6ca3c7-d34c-4ccb-9f18-c136938d8ffd-4mvlq" is forbidden: ValidatingAdmissionPolicy 'ports.dataplane.gateway-operator.konghq.com' with binding 'binding-ports.dataplane.gateway-operator.konghq.com' denied request: Each port from spec.network.services.ingress.ports has to have an accompanying port in KONG_PORT_MAPS env
```

which is a symptom of a bigger issue described in detail in a dedicated issue

- https://github.com/Kong/kong-operator/issues/2854

So, for now, that part has actually been removed for now.

**Which issue this PR fixes**

Fixes #161

The test was run in the loop with `GOMAXPROCS=1` set, because flake often occurred. Now it's gone. For DataPlane `ObservedGeneration: 3` misses in case of flake for triggering regeneration of network policy, bringing it close just to after provisioning of `DataPlane` ensures that `NetworkPolicy` is regenerated for updated `DataPlane`. It's also true for the case that is removed from the test (it was run without a validating policy).

### DIFF
--- a/controller/gateway/controller.go
+++ b/controller/gateway/controller.go
@@ -313,6 +313,20 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	// condition to the Gateway status if the dataplane is ready. If not ready
 	// the status DataPlaneReady=False will be set instead.
 	dataplane, provisionErr := r.provisionDataPlane(ctx, logger, &gateway, gatewayConfig, konnectExtension)
+	if provisionErr == nil && k8sutils.RunningOnKubernetes() {
+		// DataPlane NetworkPolicies
+		// Only create network policies if KO is running inside k8s.
+		// If the code is run outside of k8s (like in envtest or integration test), do not create network policies.
+		log.Trace(logger, "ensuring DataPlane's NetworkPolicy exists")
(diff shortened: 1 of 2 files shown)


pmalek (approved):
🤞

PR #2852, opened 2025-12-11
feat: `KonnectAPIAuthConfiguration` cross-namespace reference

mlavacca:
**What this PR does / why we need it**:

Implementation of the cross-namespace reference between `KonnectGatewayControlPlane` and `KonnectAPIAuthConfiguration` by leveraging `KongReferenceGrant`.

**Which issue this PR fixes**

Fixes #2404 

**Special notes for your reviewer**:

This PR addresses the underlying (`KonnectGatewayControlPlane`->KonnectAPIAuthConfiguration) cross-namespace reference. The upper-level (`GatewayConfiguration`->`KonnectAPIAuthConfiguration`) check is not performed at the moment and will be tackled separately by https://github.com/Kong/kong-operator/issues/2483.

**PR Readiness Checklist**:

Complete these before marking the PR as `ready to review`:

- [x] the `CHANGELOG.md` release notes have been updated to reflect significant changes

PR #2855, opened 2025-12-11
feat: add `KongReferenceGrant` CRD for x-ns Secret ref

### LINKED ISSUE #2853: Introduce `KongReferenceGrant`
## Problem Statement

We need an API to natively support x-namespace references in Konnect entities. For this purpose, we can create a 1:1 mapping or the upstream `ReferenceGrant` API tailored to that use case.

## Acceptance Criteria

- [ ] `KongReferenceGrant` CRD is available to the user

mlavacca:

Implements KongReferenceGrant API to enable secure cross-namespace references in Kong configuration resources. This allows resources like KongCertificate to reference Secrets in different namespaces while maintaining security through explicit trust relationships.

**Which issue this PR fixes**

Fixes #2853 

**PR Readiness Checklist**:


alacuku (approved):

api/configuration/v1alpha1/KongReferenceGrant_types.go
```diff
@@ -0,0 +1,134 @@
...
+	// Group is the group of the referent.
+	//
+	// +required
+	// +kubebuilder:validation:Enum=configuration.konghq.com
+	Group Group `json:"group"`
+
+	// Kind is the kind of the referent.
+	//
+	// +required
+	// +kubebuilder:validation:Enum=KongCertificate
```
pmalek, line 103: Constraining this on enum now is good. I'm concerned about scaling this to have different combination of from+to. For example we'll want to allow references:
- from `KongCertificate` to `KonnectGatewayControlPlane`
- but also from `KongCertificate` to `Secret`

This is definitely doable in CEL but I'm just pointing it out that with time these rules will become complex.

api/configuration/v1alpha1/KongReferenceGrant_types.go
```diff
@@ -0,0 +1,134 @@
...
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:scope=Namespaced
+// +kubebuilder:resource:categories=kong
+// +kubebuilder:storageversion
+// +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
```
pmalek, line 27: Do you think adding additional print columns for `from` and `to` fields would make sense? Would it be too verbose?

Without those I feel that printing out the grant with `kubectl` might feel a bit naked, not providing any valuable information.

fgiudici (approved):
Looks fine here

api/configuration/v1alpha1/KongReferenceGrant_types.go
mlavacca: In #2852 I'm going to add stricter rules as we are adding CP-Auth as a possibility. Something along the lines of 

```go
// +kubebuilder:validation:XValidation:rule="self.group == 'configuration.konghq.com' && self.kind == 'KongCertificate' || self.group == 'konnect.konghq.com' && self.kind == 'KonnectGatewayControlPlane'",message="pair group/kind is not allowed"
```

Other than this, do you have any ideas on how we can do things differently?

api/configuration/v1alpha1/KongReferenceGrant_types.go
mlavacca: I've been thinking about it and I am not sure. We have:
- From:
  - group
  - kind
  - namespace
- To:
  - group
  - kind
  - name (optional)
 
How would you format this information?

api/configuration/v1alpha1/KongReferenceGrant_types.go
pmalek: This is exactly what I think is possible and what I'm afraid will be hard to understand when we keep adding combinations there.

Sadly, I don't have a solution for this other than expanding the rule.

I'll resolve this comment as I believe we're on the same page on this.

api/configuration/v1alpha1/KongReferenceGrant_types.go
pmalek: I'd be tempted to say that adding all of those might actually be a good approach?

So adding printcolumns:

- `To-Group`
- `To-Kind`
- `To-Name`
- `From-Group`
- `From-Kind`
- `From-Namespace`

WDYT?

Sadly, combined printcolumns is not supported yet: https://github.com/kubernetes/kubernetes/issues/127246 https://github.com/kubernetes/enhancements/issues/4595

mlavacca: what do you think if we go with

- `To-GK` (e.g., `KonnectAPIAuthConfiguration.konnect.konghq.com`)
- `To-Name`
- `From-GK` (e.g., `KonnectGatewayControlPlane.konnect.konghq.com`)
- `From-Namespace`

To save a couple of columns? I'm worried we can get too verbose

pmalek: > To save a couple of columns? I'm worried we can get too verbose

I wouldn't be too concerned about this. A lot of in-tree resource have several printcolumns and they look good. There's also the `-o wide` supported by `kubectl` but I don't know how to designate a column to be only displayed when that's enabled.

---

> what do you think if we go with
> 
> * `To-GK` (e.g., `KonnectAPIAuthConfiguration.konnect.konghq.com`)
> * `To-Name`
> * `From-GK` (e.g., `KonnectGatewayControlPlane.konnect.konghq.com`)
> * `From-Namespace`

This would be possible if KEP I linked was already implemented. Do you know how to achieve this now? If yes then I'm all in for it.

mlavacca: Actually, printing nice columns is not possible, as `from` and `to` are slices. There is no single value to print.

pmalek: 😢 

I thought that since `Service` has support for arrays in printcolumn (its ports) then that could work as well.

But apparently there has to be some special sauce in kubectl (or elsewhere) that makes it happen because there are no special annotations in the source: https://github.com/kubernetes/api/blob/v0.34.2/core/v1/types.go#L6152-L6160

---

I've tried adding a custom type:

```
type ReferenceGrantFromList []ReferenceGrantFrom
```

but when I set the `printcolumn` type as `string` I get:

```
kg kongreferencegrant -A
NAMESPACE   NAME                   AGE   FROM
default     allow-cert-to-secret   8s    [{"group":"configuration.konghq.com","kind":"KongCertificate","namespace":"cert-namespace"}]
```

which is not easily readable.

I guess we're OK proceeding without any additional print columns in that case.

pmalek (approved):

Issue #2684, opened 2025-11-21
HybridGateway: get rid of duplicated code in the translation phase


fgiudici:
## Problem Statement

Current  code for the translation from HTTPRoutes to Kong resources brings a package to translate each Kong resource. The code in the packages is quite similar.

## Proposed Solution

Factorize the actual code and reuse the core logic by means of Interfaces or Generics.

fgiudici:
completed

PR #2857, opened 2025-12-12
feat(hybridgateway): order creation/deletion of Kong resources

### LINKED ISSUE #2685: HybridGateway: add dependencies during creation and deletion of Kong resources
## Problem Statement

The translation from HTTPRoute to Kong resources in the HybridGateway controller happens all at once, generating a list of Kong resources derived from the HTTPRoute which are created all together.
The deletion happens in a similar way: once an HTTPRoute is deleted, all the resources are marked for deletion at the same time.
Creating/deleting Kong resources without accounting the dependencies among them can cause different issues, from triggering temporary incomplete configurations on the data plane (e.g., enabling a Route without a plugin) to causing transient errors on the Konnect side (e.g., deleting a KongPlugin while it is still referenced by a KongPluginBinding).

## Proposed Solution
Actively manage the resource dependencies, creating/deleting the resources only when the dependencies are satisfied, performing the operations in multiple reconcile steps.

fgiudici:
This PR adds logic to order the creation and deletion of Kong resources translated from HTTPRoutes.
The full creation (and deletion) of the set of derived resources is now performed in multiple steps that take into account dependencies among the generated resources to avoid errors caused by generating a resource before the resources it depends on.

While there the translation code that checks for already existing resources has been shared among the different resource kinds.

**Which issue this PR fixes**

Fixes #2685 , #2684 

**PR Readiness Checklist**:


controller/hybridgateway/converter/converter.go
```diff
@@ -16,13 +16,16 @@ import (
...
-	// Translate performs the conversion or translation logic for the root object, returning the number of Kong resources created and an error if the process fails.
-	Translate(ctx context.Context, logger logr.Logger) (int, error)
+	// Translate performs the conversion or translation logic for the root object, returning a boolean indicating if a requeue is needed and an error if the process fails.
+	Translate(ctx context.Context, logger logr.Logger) (bool, error)
 	// GetRootObject returns the current root object of type t.
 	GetRootObject() t
 	// GetOutputStore returns a slice of unstructured.Unstructured objects representing the current state of the store, using the provided context.
 	GetOutputStore(ctx context.Context, logger logr.Logger) ([]unstructured.Unstructured, error)
+	// GetOutputStoreLen returns the number of objects in the output store.
+	GetOutputStoreLen(ctx context.Context, logger logr.Logger) int
```
alacuku: Do we really need to add this new function to the interface? We have already a function that returns the OutputStore. If we are going to keep it, then let's remove unused arguments and the logger.

controller/hybridgateway/controller.go
```diff
@@ -224,8 +226,14 @@ func (r *HybridGatewayReconciler[t, tPtr]) Reconcile(ctx context.Context, req ct
 		return ctrl.Result{Requeue: true}, nil
 	}
 
+	// If translation is not finished and needs requeue, do it now.
+	if needsRequeue {
```
alacuku: The `needsRequeue` value is returned by the translate function, but it is checked after we have enforced the state by calling `enforceState`. At this point, I would make the `needsRequeue` variable and internal state of the converter implementation and check it in the `enforceState` function which already returns a values that is used re-queue the reconciled object.

controller/hybridgateway/reconciler_utils.go
```diff
@@ -207,25 +208,33 @@ func enforceStatus[t converter.RootObject](ctx context.Context, logger logr.Logg
...
 // Returns:
-//   - bool: true if any orphaned resources were deleted from the cluster
+//   - orphansDeleted: true if any orphaned resources were deleted in this iteration
+//   - requeueNeeded: true if a requeue is needed to continue the multi-step deletion process
 //   - error: Any error that occurred during the cleanup process
 //
 // The function uses ownership labels to identify resources managed by the root object
 // and only deletes resources that are no longer present in the converter's desired output.
-func cleanOrphanedResources[t converter.RootObject, tPtr converter.RootObjectPtr[t]](ctx context.Context, cl client.Client, logger logr.Logger, conv converter.APIConverter[t]) (bool, error) {
+func cleanOrphanedResources[t converter.RootObject, tPtr converter.RootObjectPtr[t]](
```
alacuku, line 238: This function is not more generic since it has some specific handling for the httpRoute or at least we are introducing ordered cleanup. In the future we are going to introduce new converters that do not need this specific logic. I would make the cleanup logic part of the `converter interface`.

controller/hybridgateway/converter/http_route.go
```diff
@@ -91,11 +94,11 @@ func (c *httpRouteConverter) GetRootObject() gwtypes.HTTPRoute {
 // Returns:
 //   - int: Number of Kong resources created during translation
 //   - error: Aggregated translation errors or nil if successful
```
alacuku, line 96: Let's update the comment to reflect the changes.

controller/hybridgateway/translator/translator.go
```diff
@@ -0,0 +1,77 @@
...
+//   - err: Any error encountered during the process.
+func VerifyAndUpdate[T client.Object](
+	ctx context.Context,
+	logger logr.Logger,
+	cl client.Client,
+	obj T,
+	route client.Object,
+	exclusiveRoute bool,
+) (exists bool, err error) {
+	existingObj := obj.DeepCopyObject().(T)
```
alacuku, line 40: Why are we deep copying the original object?

controller/hybridgateway/translator/translator.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	if exclusiveRoute {
+		if len(routes) > 1 || strings.TrimSpace(routes[0]) != metadata.ObjectToNameString(route) {
+			err = fmt.Errorf("existing %s object %s/%s is associated with multiple routes %s",
+				obj.GetObjectKind().GroupVersionKind().Kind, existingObj.GetNamespace(), existingObj.GetName(), routesCSV)
+			log.Error(logger, err, "Tracking annotation exclusive source Route check failed")
+			return true, err
+		}
+	}
```
alacuku, line 76: Commenting what this code does and why we need it would be helpful. Like which entities we expect to be associated with more than one route and which ones are exclusive.

controller/hybridgateway/reconciler_utils.go
```diff
@@ -307,14 +327,34 @@ func cleanOrphanedResources[t converter.RootObject, tPtr converter.RootObjectPtr
 		if orphansForGVK > 0 {
 			log.Debug(logger, "Deleted orphaned resources for GVK", "gvk", gvk.String(), "orphansDeleted", orphansForGVK)
 		} else {
-			log.Debug(logger, "No orphaned resources found for GVK", "gvk", gvk.String())
+			if orphansBeingDeleted > 0 {
+				log.Debug(logger, "Resources are being deleted for GVK, will requeue", "gvk", gvk.String(), "orphansBeingDeleted", orphansBeingDeleted)
+			} else {
+				log.Debug(logger, "No orphaned resources found for GVK", "gvk", gvk.String())
+			}
+		}
```
alacuku: Do we really need this check? We are checking again the same thing in line 340. Can we drop this?

controller/hybridgateway/reconciler_utils.go
```diff
@@ -307,14 +327,34 @@ func cleanOrphanedResources[t converter.RootObject, tPtr converter.RootObjectPtr
...
+		// This ensures we wait for the current GVK's resources to be fully deleted before
+		// moving to the next GVK, maintaining the proper deletion order for security.
+		if orphansBeingDeleted > 0 {
+			log.Debug(logger, "Requeueing to wait for resource deletion to complete", "gvk", gvk.String())
+			return totalOrphansDeleted > 0, true, nil
+		}
+
+		// If we deleted any orphans for this GVK, also requeue to verify they are fully deleted
+		// before proceeding to the next GVK type.
+		if orphansForGVK > 0 {
```
alacuku, line 339: We are making the same check as in line 327, but logging a different message. Can we merge these two checks?

alacuku (commented):
Left some comments.

controller/hybridgateway/converter/converter.go
fgiudici: didn't looked that bad adding it honestly: I slightly preferred it than having Translate returning the len of the obj in the OutputStore.
Regarding the arguments, since all the other functions in the interface (with the notable exception of the GetRootObject()) have exactly the same, I would prefer keeping the same here too.
I do not have strong opinions on keeping this function, but would have on the signature: so let's drop it and have Translate back returning also the number of generated Kong resources.

controller/hybridgateway/controller.go
fgiudici: Are you suggesting to extend the Converter interface with some sort of "GetNeedsRequeue()" function? Otherwise how would you get access to an implementation specific state? 

Re-looking at the code seems we may get rid of the needsRequeue as the enforceState will *always* update or create something if the Translate produced something anew. Not on the super-safe side as we implicitly rely on the flow of the code, but cannot image any case in which this would not work.

controller/hybridgateway/reconciler_utils.go
fgiudici: The only thing that has been introduced is sequential clean-up of the resources by their kind. The order comes from the conv.GetExpectedGVKs().
This is generic.
I would argue that *any* converter that has multiple resource kinds translated from the same ***Route needs some ordering during deletion.

The only thing I can think of is few resource kinds that may not need to wait for their deletion before moving on. The easiest way to manage it that comes to my mind is to extend the GetExpectedGVKs() to carry extra information regarding if the complete deletion of the current resource kind is needed before moving on or not. But still , the logic would be in this same function and not that different. Do you mean something like that? Otherwise, can you elaborate a bit more of what you have in mind?

controller/hybridgateway/converter/http_route.go
fgiudici: Done

controller/hybridgateway/translator/translator.go
fgiudici: Looked prettier than reflection:
`existingObj := reflect.New(reflect.TypeOf(obj).Elem()).Interface().(T)`

controller/hybridgateway/translator/translator.go
fgiudici: If the generated resource should be associated with one or multiple routes it is not decided here: exclusiveRoute is a parameter of the function.
I can add a link here to the docs about naming (https://github.com/Kong/kong-operator/blob/main/docs/internal/hybridgateway/autogen-resource-naming.md), that may help.

The exclusiveRoute is something we expect for KongRoutes or KongBindings.
To be honest I could not envision any real scenario in which the exclusiveRoute branch could fail: it would require same HTTPRoute with two rules that have exactly the same matches (which makes no sense).
I could add some comment about this scenario.

controller/hybridgateway/reconciler_utils.go
fgiudici: Uh, right, this part should be improved.

controller/hybridgateway/reconciler_utils.go
fgiudici: Yep, makes sense.

controller/hybridgateway/converter/converter.go
fgiudici: (Dropped)

controller/hybridgateway/translator/translator.go
fgiudici: Added some more comments hope it is more clear now.

controller/hybridgateway/reconciler_utils.go
fgiudici: Reworked

controller/hybridgateway/reconciler_utils.go
fgiudici: So, changed the logic and simplified it a bit:. how does it look now?

controller/hybridgateway/controller.go
fgiudici: After offline discussion we decided to move the logic to ensure ordering during resource creation in the enforceState() function.
The translate() function now always translates all the object, so needsRequeue was removed.

fgiudici:
Moved the logic to order resources creation in the enforceState() function.

fgiudici:
(updated CHANGELOG and rebased)

alacuku (approved):

PR #2825, opened 2025-12-08
fix: data race in admission webhook

programmer04:
**What this PR does / why we need it**:

Fixes datarace during concurrent access and modification of the map used for registering/unregistering validators

```log
==================
WARNING: DATA RACE
Read at 0x00c002d183f0 by goroutine 7536:
  github.com/kong/kong-operator/ingress-controller/internal/admission.(*RequestHandler).dispatchValidationNoMatcher()
      /Users/jakub.warczarek@konghq.com/Documents/work/kong-operator/ingress-controller/internal/admission/handler.go:63 +0x138
  github.com/kong/kong-operator/ingress-controller/internal/admission.RequestHandler.handleSecret()
      /Users/jakub.warczarek@konghq.com/Documents/work/kong-operator/ingress-controller/internal/admission/handler.go:404 +0x188
  github.com/kong/kong-operator/ingress-controller/internal/admission.RequestHandler.handleValidation()
      /Users/jakub.warczarek@konghq.com/Documents/work/kong-operator/ingress-controller/internal/admission/handler.go:251 +0x728
  github.com/kong/kong-operator/ingress-controller/internal/admission.RequestHandler.ServeHTTP()
      /Users/jakub.warczarek@konghq.com/Documents/work/kong-operator/ingress-controller/internal/admission/handler.go:162 +0x2a4
  github.com/kong/kong-operator/ingress-controller/internal/admission.(*RequestHandler).ServeHTTP()
      <autogenerated>:1 +0x60
  net/http.serverHandler.ServeHTTP()
      /opt/homebrew/Cellar/go/1.25.5/libexec/src/net/http/server.go:3340 +0x270
  net/http.(*conn).serve()
      /opt/homebrew/Cellar/go/1.25.5/libexec/src/net/http/server.go:2109 +0x9b0
  net/http.(*Server).Serve.gowrap3()
      /opt/homebrew/Cellar/go/1.25.5/libexec/src/net/http/server.go:3493 +0x48

Previous write at 0x00c002d183f0 by goroutine 2818:
  runtime.mapaccess2_faststr()
      /opt/homebrew/Cellar/go/1.25.5/libexec/src/internal/runtime/maps/runtime_faststr_swiss.go:162 +0x29c
  github.com/kong/kong-operator/ingress-controller/internal/admission.(*RequestHandler).RegisterValidator()
      /Users/jakub.warczarek@konghq.com/Documents/work/kong-operator/ingress-controller/internal/admission/handler.go:136 +0x318
[...]

PR #2865, opened 2025-12-12
chore(tests): enable validating webhook in integration tests

programmer04:

Running the whole integration test suite with the webhook enabled allows catching issues like that
- https://github.com/Kong/kong-operator/pull/2825 (it's been resolved)
- https://github.com/Kong/kong-operator/issues/2867 (discovered during work on this PR, hence test is skipped)

Thus, this PR enables it and cleans up other suites from webhook-related stuff. Envtests are not feasible anymore to run webhook tests, because the architecture after the merge of KGO and KIC is too complicated to bootstrap the environment properly and be sure about correctness. The old integration test suite of KIC is not suitable anymore.

PR uses a hack to obtain connectivity from inside the cluster to locally run the operator, as it used to be used in KIC. Unfortunately telepresence approach doesn't work reliably, see https://github.com/Kong/kong-operator/pull/2675#issuecomment-3651753585.

What is more, it contains a couple of opportunistic cleanups and reorganizations.

Check
- https://github.com/Kong/kong-operator/pull/2866

to see a ported test from envtests to integration run with this machinery. 

**Which issue this PR fixes**

Part of https://github.com/Kong/kong-operator/pull/2176 (porting all skipped tests is required for completion)


ingress-controller/pkg/validation/consts/consts.go
```diff
@@ -0,0 +1,15 @@
+package consts
+
+// NOTE: These paths have to match paths used in Helm Chart.
+// E.g. in
```
pmalek: in ... ?

pmalek (commented):
Minor nit, otherwise 👍

pmalek (dismissed):

pmalek (approved):

PR #2882, opened 2025-12-15
CI: run chainsaw in CI

### LINKED ISSUE #2761: Run chainsaw e2e tests in CI
## Problem Statement

https://github.com/Kong/kong-operator/pull/2725 introduced a new e2e test suite based on chainsaw. We need to introduce automation to run the tests for every PR.

## Acceptance Criteria

- [ ] The Chainsaw e2e tests are run in CI

tao12345666333:

**Which issue this PR fixes**

Fixes https://github.com/Kong/kong-operator/issues/2761

**PR Readiness Checklist**:

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -830,6 +830,7 @@ jobs:
           --create-namespace \
           --set image.repository=${{ env.IMG }} \
           --set image.tag=${{ env.TAG }} \
+          --set env.enable_controller_konnect=true \
           --wait
 
     - name: run chainsaw e2e tests
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -811,7 +811,6 @@ jobs:
       uses: helm/kind-action@92086f6be054225fa813e0a4b13787fc9088faab # v1.13.0
       with:
         cluster_name: ${{ env.CLUSTER_NAME }}
-        version: v0.30.0
         node_image: kindest/node:${{ needs.matrix_k8s_node_versions.outputs.latest }}
 
     - uses: jdx/mise-action@146a28175021df8ca24f8ee1828cc2a60f980bd5 # v3.5.1
@@ -821,13 +820,17 @@ jobs:
     - name: Load image to kind cluster
       run: kind load docker-image ${{ env.IMG }}:${{ env.TAG }} --name ${{ env.CLUSTER_NAME }}
 
-    - name: Install CRDs
-      run: make install
+    - name: Install cert-manager
+      run: make install.helm.cert-manager
 
-    - name: Deploy operator
-      env:
-        VERSION: ${{ env.TAG }}
-      run: make deploy
+    - name: Deploy operator with Helm
+      run: |
+        helm install kong-operator charts/kong-operator \
+          --namespace kong-system \
+          --create-namespace \
+          --set image.repository=${{ env.IMG }} \
+          --set image.tag=${{ env.TAG }} \
+          --wait
 
     - name: run chainsaw e2e tests
       run: make test.e2e.chainsaw
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -807,21 +807,17 @@ jobs:
     - name: build docker image
       run: make docker.build
 
-    - name: Create k8s KinD Cluster
-      uses: helm/kind-action@92086f6be054225fa813e0a4b13787fc9088faab # v1.13.0
-      with:
-        cluster_name: ${{ env.CLUSTER_NAME }}
-        node_image: kindest/node:${{ needs.matrix_k8s_node_versions.outputs.latest }}
-
     - uses: jdx/mise-action@146a28175021df8ca24f8ee1828cc2a60f980bd5 # v3.5.1
       with:
         install: false
 
+    - name: Create k8s KinD Cluster with ktf
+      run: |
+        go install github.com/kong/kubernetes-testing-framework/cmd/ktf@latest
+        ktf environments create --name ${{ env.CLUSTER_NAME }} --addon metallb --addon cert-manager
+
     - name: Load image to kind cluster
       run: kind load docker-image ${{ env.IMG }}:${{ env.TAG }} --name ${{ env.CLUSTER_NAME }}
-
-    - name: Install cert-manager
-      run: make install.helm.cert-manager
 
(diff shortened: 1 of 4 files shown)


tao12345666333:
The chainsaw test job: https://github.com/Kong/kong-operator/actions/runs/20244643743/job/58121450108?pr=2882

.github/workflows/tests.yaml
```diff
@@ -698,6 +698,61 @@ jobs:
...
+        go-version-file: go.mod
+
+    - name: build docker image
+      run: make docker.build
+
+    - name: Create k8s KinD Cluster
+      uses: helm/kind-action@92086f6be054225fa813e0a4b13787fc9088faab # v1.13.0
+      with:
+        cluster_name: ${{ env.CLUSTER_NAME }}
+        version: v0.30.0
```
pmalek: Do we need this here?

0.30.0 is already the default in 1.13.0: https://github.com/helm/kind-action/blob/v1.13.0/action.yml#L11

.github/workflows/tests.yaml
```diff
@@ -698,6 +698,61 @@ jobs:
...
+    - name: Load image to kind cluster
+      run: kind load docker-image ${{ env.IMG }}:${{ env.TAG }} --name ${{ env.CLUSTER_NAME }}
+
+    - name: Install CRDs
+      run: make install
+
+    - name: Deploy operator
+      env:
+        VERSION: ${{ env.TAG }}
+      run: make deploy
```
pmalek: Can we use helm for this? I think that since `helm` is the only officially supported way of installing the operator we should use it here as well. WDYT?

.github/workflows/tests.yaml
tao12345666333: Thank you, we can delete it.

tao12345666333: Thank you, we can delete it.

.github/workflows/tests.yaml
tao12345666333: Make sense.

One additional question, as you said, is Helm the only official supported way to install KO, should we change `make deploy` to use helm command?

pmalek: I'd say let's leave this topic aside in this PR.

Personally I'd be okay with leaving this as is for the long term.

.github/workflows/tests.yaml
```diff
@@ -706,6 +706,61 @@ jobs:
...
+    - name: checkout repository
+      uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+
+    - name: setup golang
+      uses: actions/setup-go@4dc6199c7b1a012772edbd06daecab0f50c9053c # v6.1.0
+      with:
+        go-version-file: go.mod
+
+    - name: build docker image
+      run: make docker.build
```
pmalek, line 808: We're already building the docker image for e2e tests here: https://github.com/Kong/kong-operator/blob/b2632163c600504a75c3abd72ecb8cf5d4f1ab12/.github/workflows/tests.yaml#L730-L734

We should be able to build it once and use it across jobs. WDYT?

tao12345666333: Great! I can cache it maybe.

tao12345666333: Since these are two separate jobs, sharing images through artifacts doesn't bring any significant benefits and would only complicate the workflow.  I created an issue for tracking https://github.com/Kong/kong-operator/issues/2948

pmalek: > Since these are two separate jobs, sharing images through artifacts doesn't bring any significant benefits and would only complicate the workflow

Actually, it **would** bring benefits: a single build job. I do agree that this would complicate things a bit but in the long run I believe it's worth it. Let's handle it separate as suggested.

.github/workflows/tests.yaml
```diff
@@ -706,6 +706,61 @@ jobs:
...
+      run: |
+        go install github.com/kong/kubernetes-testing-framework/cmd/ktf@latest
+        ktf environments create --name ${{ env.CLUSTER_NAME }} --addon metallb --addon cert-manager
+
+    - name: Load image to kind cluster
+      run: kind load docker-image ${{ env.IMG }}:${{ env.TAG }} --name ${{ env.CLUSTER_NAME }}
+
+    - name: Deploy operator with Helm
+      run: |
+        helm install kong-operator charts/kong-operator \
```
pmalek, line 824:
```suggestion
        helm install kong-operator ./charts/kong-operator \
```

Don't we want to use the in tree chart?

tao12345666333: Good catch!

tao12345666333: By the way, using the in-tree chart explicitly makes things clearer.

Before this change, we were in fact already using the in-tree chart. :smile:

test/e2e/chainsaw/hybridgateway/basic-httproute/chainsaw-test.yaml
```diff
@@ -40,6 +40,72 @@ spec:
...
+              echo ""
+              echo "=== Pod details and logs ==="
+              for pod in $(kubectl get pods -n kong -o jsonpath='{.items[*].metadata.name}'); do
+                echo "--- Pod: $pod ---"
+                kubectl describe pod $pod -n kong | tail -50
+                echo ""
+                echo "--- Logs for $pod ---"
+                kubectl logs $pod -n kong --all-containers --tail=50 2>/dev/null || echo "No logs available"
+                echo ""
+              done
```
mlavacca, line 99: Can we delete this catch block or it's something we want to keep?

mlavacca (commented):
Apart from an inline comment, lgtm!

test/e2e/chainsaw/hybridgateway/basic-httproute/chainsaw-test.yaml
tao12345666333: @mlavacca I think we can keep this catch block, in case the CI run fails, it'll give us a clear stack trace for quick debugging.

tao12345666333: removed the inline comments https://github.com/Kong/kong-operator/pull/2882/commits/29a732b84bd4469f9f0e3a08ae2bee0d5fcaee69

mlavacca (approved):

Issue #2851, opened 2025-12-11
Cross namespace references for `KongService` and `KongUpstream` → `KonnectGatewayControlPlane`


mlavacca:
## Problem Statement

We need to enable x-namespace references for `KongService`  and `KongUpstream` resources. For scenarios where the `KonnectGatewayControlPlane` is in Namespace A, we need to allow `KongService` and `KongUpstream` in namespace B to reference it.

## Proposed Solution

Leverage the `KongReferenceGrant` CRD to safely allow x-namespace reference between
- `KongService` and `KonnectGatewayControlPlane`, and
- `KongUpstream` and `KonnectGatewayControlPlane`

## Additional Information

...

## Acceptance Criteria

- [ ] as a user, I can x-ns reference a `KonnectGatewayControlPlane` from `KongService`s and `KongUpstream`s in different namespaces. 
- [ ] Tests are added
- [ ] User facing documentation which explains how to use `KongReferenceGrant` with `KongService` and `KonnectGatewayControlPlane` is added

jerry153fish:
hi team, any plan to add support for the KongUpstream?

pmalek:
> hi team, any plan to add support for the KongUpstream?

Hi @jerry153fish 👋 

Are you referring to the cross namespace reference between `KongUpstream` and `KonnectGatewayControlPlane` as described in this issue? If so then yes, that's the plan.

PR #2892, opened 2025-12-16
feat: add support for cross-namespace references for KongServices to KonnectGatewayControlPlanes

pmalek:

**Which issue this PR fixes**

Part of #2851

**PR Readiness Checklist**:


api/configuration/v1alpha1/conditions.go
```diff
@@ -0,0 +1,14 @@
+package v1alpha1
+
+const (
+	// KongReferenceGrantConditionType is the condition type used to indicate
+	// whether a KongReferenceGrant is valid for cross-namespace references.
+	KongReferenceGrantConditionType = "KongReferenceGrantValid"
```
mlavacca: Can we re-use the `ResolvedRefs` condition type and set `KongReferenceGrantInvalid` as its reason to be `false`?

mlavacca (commented):
Mostly lgtm. Just a comment inline

api/configuration/v1alpha1/conditions.go
pmalek: Sure we can, do you believe there's merit to using just 1 condition type for this rather than adding a dedicated one?

mlavacca: It's mainly to map how  this thing works upstream: missing referenceGrant is mapped as a reason on the `ResolvedRefs` condition.

controller/konnect/watch_kongservice.go
```diff
@@ -100,3 +108,57 @@ func enqueueKongServiceForKongRoute() func(ctx context.Context, obj client.Objec
 		}
 	}
 }
+
+// enqueueKongServiceForKongReferenceGrant returns a function that enqueues
+// reconcile.Requests for KongServices that are allowed by the KongReferenceGrant.
+// This is useful for situations where a KongReferenceGrant is created/updated/deleted
+// and we need to reconcile the KongServices that are affected by it.
+func enqueueKongServiceForKongReferenceGrant(cl client.Client) func(ctx context.Context, obj client.Object) []reconcile.Request {
```
mlavacca, line 116: Can we make this function generic so that it can be shared among all the controllers that need it?

pmalek: Agreed. Although I'd suggest to adapt it when we're actually going to need it. Subsequent PRs adding other cross namespace references will do it. WDYT?

mlavacca: sounds good to me 👍🏻

mlavacca (approved):

PR #2893, opened 2025-12-16
chore(tests): port and re-enable `TestAdmissionWebhook_KongConsumers` and deduplicate code

programmer04:

Porting `TestAdmissionWebhook_KongConsumers` to integration the test that used to be skipped with

```
t.Skip("skipping until https://github.com/Kong/kong-operator/issues/2176 is resolved")
```

Furthermore, since it is a subsequent PR in the series, the right moment to provide some refactoring to deduplicate code. 

**Which issue this PR fixes**

Part of https://github.com/Kong/kong-operator/issues/2176


pmalek (dismissed):
I believe that in the long run we could isolate those tests into a separate test dir and run them in a separate GH job to prevent flakes and long CI times.

I'm happy to hear your thoughts on this.

programmer04:
Yes, we can do this (all validation one be a part of a separate suite). But I think the whole integration test suite still should be run with the webhook enabled

pmalek (approved):

PR #3018, opened 2026-01-07
HybridGateway: backendRef pointer aliasing breaks KongTarget generation for multiple BackendRefs

tao12345666333:

Because the loop variable is reused, all stored pointers can end up referencing the last BackendRef, causing incorrect downstream behavior (e.g. KongTarget names/weights derived from the wrong backend, multiple backends effectively treated as one)

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:

### DIFF
--- a/controller/hybridgateway/target/target.go
+++ b/controller/hybridgateway/target/target.go
@@ -273,7 +273,8 @@ func filterValidBackendRefs(
 ) ([]validBackendRef, error) {
 	var validBackendRefs []validBackendRef
 
-	for _, bRef := range backendRefs {
+	for i := range backendRefs {
+		bRef := backendRefs[i]
 		// Check if the backendRef is supported.
 		if !route.IsBackendRefSupported(bRef.Group, bRef.Kind) {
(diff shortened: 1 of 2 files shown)


controller/hybridgateway/target/target.go
```diff
@@ -334,7 +335,10 @@ func filterValidBackendRefs(
 
 		// If we reach here, the BackendRef is valid and has endpoints.
 		validBackendRefs = append(validBackendRefs, validBackendRef{
-			backendRef:     &bRef,
+			// IMPORTANT: take the address of the slice element, not the loop variable.
+			// The loop variable is reused across iterations which would make all pointers
+			// alias the same BackendRef.
+			backendRef:     &backendRefs[i],
```
programmer04, line 341: It's no longer true. It used to be like that for Go < 1.22, see [the official guide](https://go.dev/wiki/CommonMistakes#using-reference-to-loop-iterator-variable). Furthermore, without any changes in `controller/hybridgateway/target/target.go`, the testcase added in `controller/hybridgateway/target/target_test.go` passes. So it's not the culprit

programmer04 (changes requested):

PR #3025, opened 2026-01-08
ci(chainsaw): test conversion webhook for `ControlPlane`

programmer04:

Add chainsaw e2e scenario for `ControlPlane` conversion from `v1beta1` to `v2beta1` to cover conversion webhook with e2e test

**Which issue this PR fixes**

Part of #1986

### DIFF
--- a/test/e2e/chainsaw/conversion_webhook/controlplane/00-assert-controlplane.yml
+++ b/test/e2e/chainsaw/conversion_webhook/controlplane/00-assert-controlplane.yml
@@ -1,0 +1,101 @@
+apiVersion: gateway-operator.konghq.com/v1beta1
+kind: DataPlane
+metadata:
+  name: ($dataplane_name)
+  namespace: ($kong_namespace)
+status:
+  (conditions[?type == 'Ready']):
+    - status: 'True'
+      reason: Ready
+---
+
+apiVersion: gateway-operator.konghq.com/v2beta1
+kind: ControlPlane
+metadata:
+  name: ($controlplane_name)
+  namespace: ($kong_namespace)
+spec:
+  configDump:
+    dumpSensitive: disabled
+    state: disabled
+  dataplane:
+    ref:
+      name: ($dataplane_name)
+    type: ref
+  dataplaneSync:
+    reverseSync: disabled
+  ingressClass: kong
+  konnect:
+    consumersSync: enabled
+  translation:
+    combinedServicesFromDifferentHTTPRoutes: enabled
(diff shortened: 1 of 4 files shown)


test/e2e/chainsaw/conversion_webhook/controlplane/00-controlplane.yml
```diff
@@ -0,0 +1,61 @@
...
+apiVersion: gateway-operator.konghq.com/v1alpha1
+kind: WatchNamespaceGrant
+metadata:
+  name: watchnamespacegrant
+  namespace: ($kong_namespace)
+spec:
+  from:
+  - group: gateway-operator.konghq.com
+    kind: ControlPlane
+    namespace: ($kong_namespace)
```
pmalek: Do we actually need all of the watch namespace related config given that the CP and DP are in the same ns?

test/e2e/chainsaw/conversion_webhook/controlplane/00-assert-controlplane.yml
```diff
@@ -0,0 +1,101 @@
...
+  (featureGates[?name == 'GatewayAlpha']):
+    - state: enabled
+  (featureGates[?name == 'KongServiceFacade']):
+    - state: disabled
+  (featureGates[?name == 'SanitizeKonnectConfigDumps']):
+    - state: enabled
+  (featureGates[?name == 'FallbackConfiguration']):
+    - state: disabled
+  (featureGates[?name == 'RewriteURIs']):
+    - state: enabled
```
pmalek, line 75: I'm wondering if this is not going to be a problem in the long run, when we'd potentially change the defaults.

WDYT about changing the asserts to check whether those exist rather than asserting the state value?

test/e2e/chainsaw/conversion_webhook/controlplane/00-controlplane.yml
programmer04: For the `spec.watchNamespaces.type` `list` yes. Without specifying anything, it defaults to `all`, also you can set `own`, and it's not needed too. So I removed the explicit specification for the sake of clarity

test/e2e/chainsaw/conversion_webhook/controlplane/00-assert-controlplane.yml
programmer04: Good point, I left the assertion for a specific value only for those speciefied explicitly by env var

pmalek (approved):

PR #3039, opened 2026-01-12
feat(hybridgateway): add support to PathPrefixMatch URL rewrite

### LINKED ISSUE #2902: Hybrid gateway - extend supported HTTPRoute filters - URLRewrite - replacePrefixMatch
## Current Behavior
The `URLRewriteFilter` `replacePrefixMatch` path configuration is not currently honored.

## Expected Behavior
Add support to the  `URLRewriteFilter` `replacePrefixMatch` path configuration.

## Notes
This should be similar and/or worked on together with #2466 .

fgiudici:

This PR adds support to the `PathPrefixMatch` path rewrite functionality for the Gateway API `URLRewrite` filter.

**Which issue this PR fixes**

Fixes #2902

The `PathPrefixMatch` path matching performed via Kong entities requires a `KongPlugin` as usual but also to set up a `capture group` in the `KongRoute` used to translate the matching path of the  Gateway API `HTTPRoute` `Match`.
That's the way to capture the remaining part of the URL after the matching path is removed, which is then to be appended by the Kong plugin to the translated prefix path.

**PR Readiness Checklist**:

### DIFF
--- a/controller/hybridgateway/kongroute/route.go
+++ b/controller/hybridgateway/kongroute/route.go
@@ -104,7 +104,6 @@ func needsCaptureGroup(rule gwtypes.HTTPRouteRule) bool {
 			filter.RequestRedirect.Path.ReplacePrefixMatch != nil:
 			return true
 		}
-
 	}
 	return false
 }
--- a/controller/hybridgateway/kongroute/route.go
+++ b/controller/hybridgateway/kongroute/route.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/go-logr/logr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
 
 	commonv1alpha1 "github.com/kong/kong-operator/api/common/v1alpha1"
 	configurationv1alpha1 "github.com/kong/kong-operator/api/configuration/v1alpha1"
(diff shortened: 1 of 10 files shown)


controller/hybridgateway/kongroute/route.go
```diff
@@ -81,3 +86,25 @@ func RouteForRule(
...
+			filter.URLRewrite.Path != nil &&
+			filter.URLRewrite.Path.ReplacePrefixMatch != nil:
+			return true
+		case filter.Type == gatewayv1.HTTPRouteFilterRequestRedirect &&
+			filter.RequestRedirect != nil &&
+			filter.RequestRedirect.Path != nil &&
+			filter.RequestRedirect.Path.ReplacePrefixMatch != nil:
+			return true
+		}
+
```
tao12345666333:
```suggestion
```

controller/hybridgateway/plugin/converter.go
```diff
@@ -287,7 +302,9 @@ func translateURLRewrite(filter gwtypes.HTTPRouteFilter) (transformerData, error
 		case gatewayv1.FullPathHTTPPathModifier:
 			pluginConf.Replace.Uri = translateURLRewritePathFullPath(ur.Path.ReplaceFullPath)
 		case gatewayv1.PrefixMatchHTTPPathModifier:
-			fallthrough
+			pluginConf.Replace.Uri = translateURLRewritePathPrefixMatch(
+				normalizePath(ur.Path.ReplacePrefixMatch),
+				normalizePath(&path))
 		default:
 			return pluginConf, fmt.Errorf("unsupported URLRewrite path modifier type: %s", ur.Path.Type)
```
tao12345666333, line 309: After this PR,  do we still need this default case?

fgiudici: Good catch, the default case cannot happen with the current API specification.
Anyway the Gateway API standard states that new values may be added in the future and implementation should not crash if that is the case (https://gateway-api.sigs.k8s.io/reference/spec/#httppathmodifier).
Moreover the path modifier type is a string, so yes, I think to be future proof we should keep it.

tao12345666333: make sense

mlavacca (approved):
lgtm!

alacuku (approved):

Issue #2156, opened 2025-09-01
Move operator's CA Secret management outside of the operator


pmalek:
### Problem statement

Currently in the operator [`caManager`](https://github.com/Kong/kong-operator/blob/cde9d5c8545d8ab08bee314aee08aacf64f0c6d4/modules/manager/run.go#L398-L439) is responsible for creating the CA certificate Secret for the operator.

This Secret is not removed when the operator is deleted.

As KO 2.0 is moving towards helm self signed certs and cert-manager integration, it might be a good idea to use these mechanism to manage that Secret.

### Acceptance criteria

- [ ] As a user I can use helm self signed certs or cert-manager integration to create and manage operator's CA certificate
- [ ] No Secret leftover is left behind when operator is uninstalled.

pmalek:
@programmer04 This is done, right? Or is there anything else to do here?

programmer04:
@pmalek Cert manager integration is left. I'm going to reorganize all issue related this inititative

Issue #1666, opened 2025-05-22
Handle expiration of certificates in `ControlPlane` controller


programmer04:
## Problem Statement

PR #1376 introduced KIC in KGO [during review ](https://github.com/Kong/gateway-operator/pull/1376#discussion_r2102402203) lack fo handling expiration of certificates was spotted. It has to be resolved, certificate should be renewed automatically. Otherwise it will fail after a certain time.

## Proposed Solution

Expiration of a certificate is handled gracefully

## Acceptance Criteria
- [ ] Everything works when certificates expire (it will be replaced)

programmer04:
x-post 

- https://github.com/Kong/kong-operator/issues/2156
- https://github.com/Kong/kong-operator/issues/198

programmer04:
closed by https://github.com/Kong/kong-operator/pull/3745

Issue #198, opened 2022-08-19
CA secret reconciliation


mlavacca:
### Problem Statement

At controller startup time the operator tries to retrieve the ca secret (name and namespace are passed as args or defaulted to `kong-system/kong-operator-ca`). If such a secret does not exist, it is created. The existence of the CA is never enforced. Thus, if the secret containing the CA is deleted the operator will not be able to issue mTLS certificates anymore (unless it restarts).

### Proposed Solution

If the CA secret is not provided by the user but is created by the operator, it needs to be enforced in the cluster. This can be achieved through a controller that watches all the secrets with a specific label and whenever such a secret triggers an event, the desired state (i.e., having a valid CA secret) is enforced. 

### Additional information

This issue may considered one step in resolving Kong/gateway-operator#199, in which a secret controller has been mentioned as well.

### Acceptance Criteria

- [ ] If the CA secret provided by the operator is deleted, it gets automatically recreated by the operator.
- [ ] certificates that were created using a deleted CA get re-issued

shaneutt:
> CA certificate recreation would effectively invalidate all the PKI that depends on being signed by that CA. Therefore, this feature likely needs rotation of all the certificates dependent on the deleted CA.

I've updated the acceptance criteria to incorporate this.

pmalek:
Related to #199. When this gets implemented, the AC from #199 

> After the CA certificate changes, the operator issues new certificates and updates all mTLS certificate Secrets.

would need to be implemented.

PR #3084, opened 2026-01-16
feat: CA bootstrap moved to operator chart

programmer04:

Creation of the CA Secret used to be managed by the operator on its own. It led to an orphaned Secret after helm chart uninstallation and doesn't allow the use of the CA provided by a user.

This PR moves CA Secret management to Helm Chart (it creates a self-signed Secret and manages its lifecycle) in a similar way to webhooks. Also, a CA can be provided by a user. Based on the CA parameters of all created certificates, e.g., RSA vs ECDSA, etc. Hence, flags `--cluster-ca-key-type` and `--cluster-ca-key-size` are removed.

To provide backward compatibility and seamless upgrade in case of an existing CA (Secret with a particular name in a particular namespace), it will be reused instead of recreated. 

Check 

- `charts/kong-operator/CHANGELOG.md`
- `CHANGELOG.md` 

for more specific info.

**Which issue this PR fixes**

Part of 

- #2156

It doesn't include the introduction of cert manager. For more specific stuff like autorenew, expiration, etc. there are dedicated issues #1666 and #198.

Basically, the level of automation etc. is pretty much the same as it used to be. 

During the work, it's been spotted that for CA generated like that 

```
openssl genrsa -out /tmp/ko-makefile-ca.key 4096 2>/dev/null && openssl req -x509 -new -key /tmp/ko-makefile-ca.key -days 3650 -out 
```

Code related to certificates doesn't work properly, and it has been fixed now. See `PKCS1` vs `PKCS8` for the private key.

It seems that code related to dealing with certificates could be deduplicated and streamlined, but it's not in the scope of this PR.

**PR Readiness Checklist**:


controller/pkg/secrets/cert.go
```diff
@@ -382,6 +385,34 @@ func generateTLSDataSecret(
 	return op.Created, generatedSecret, nil
 }
 
+// DetectCertType inspects a PEM-encoded certificate and returns its KeyConfig.
+func DetectCertType(certPEM []byte) (KeyConfig, error) {
+	block, _ := pem.Decode(certPEM)
```
tao12345666333: This part reminded me of https://github.com/Kong/kong-operator/pull/2671

Currently, the implementation does not support multiple PEM crets(chain). I think we should add a comment to clarify this limitation and handle the possible failure gracefully. WDYT?

controller/pkg/secrets/cert.go
```diff
@@ -435,29 +466,36 @@ func ensureContainerImageUpdated(container *corev1.Container, imageVersionStr st
...
+	case "PRIVATE KEY":
+		privAny, err := x509.ParsePKCS8PrivateKey(pemBlock.Bytes)
+		if err != nil {
+			return nil, signatureAlgorithm, err
+		}
+		rsaKey, ok := privAny.(*rsa.PrivateKey)
+		if !ok {
+			return nil, signatureAlgorithm, fmt.Errorf("unsupported private key type in PKCS#8 wrapping: %T", privAny)
+		}
+		return rsaKey, x509.SHA256WithRSA, nil
```
tao12345666333: Since we accept `*ecdsa.PublicKey` 
https://github.com/Kong/kong-operator/pull/3084/changes#diff-465e5135e663acc85bb2b22849039cb833ae715282354383ac12ee3d638b67feR406

I think we should accept `*ecdsa.PrivateKey`, WDYT?

controller/pkg/secrets/cert.go
programmer04: Thanks for pointing out, I added code to handle it gracefully

modules/cli/cli.go
```diff
@@ -53,8 +51,7 @@ func New(m metadata.Info) *CLI {
 	flagSet.StringVar(&cfg.ControllerName, "controller-name", "", "Custom controller name, required only in multi-tenant setups.")
 	flagSet.StringVar(&cfg.ClusterCASecretName, "cluster-ca-secret", "kong-operator-ca", "Specifies the Secret name that contains the cluster CA certificate.")
 	flagSet.StringVar(&deferCfg.ClusterCASecretNamespace, "cluster-ca-secret-namespace", "", "Specifies the namespace of the Secret that contains the cluster CA certificate.")
-	flagSet.Var(&cfg.ClusterCAKeyType, "cluster-ca-key-type", "Type of the key used for the cluster CA certificate (possible values: ecdsa, rsa). Default: ecdsa.")
-	flagSet.IntVar(&cfg.ClusterCAKeySize, "cluster-ca-key-size", mgrconfig.DefaultClusterCAKeySize, "Size (in bits) of the key used for the cluster CA certificate. Only used for RSA keys.")
```
pmalek, line 57: We don't want to remove these flags in a minor release.

I'd suggest to move them somewhere below in this file (for visual separation), add a `WARNING: ` prefix to description and log a warning/error when they are set.

CHANGELOG.md
```diff
@@ -75,6 +75,11 @@
 - Removed the `KonnectID` type of control plane reference in CRDs for Konnect
   entities as it is not supported.
   [#2966](https://github.com/Kong/kong-operator/pull/2966)
+- Move management of bootstrapping CA certificate to Helm Chart, remove flags
+  `--cluster-ca-key-type` and  `--cluster-ca-key-size` now those values are
+  inferred automatically based on the CA certificate Secret. Read more in
+  Helm Chart release notes.
```
pmalek: I'd mention here which CA cert specifically we're talking about here (one used for signing CP-DP communication certs)

charts/kong-operator/values.yaml
```diff
@@ -144,3 +144,20 @@ global:
...
+    # Type of the Secret must be "kubernetes.io/tls" and labels must contain
+    # "konghq.com/secret: internal", fields "tls.crt" and "tls.key" must be present.
+    # Based on the type of this certificate (ECDSA or RSA, along with its key size),
+    # all other certificates created will follow the same configuration.
+    secretName: ""
+    # Namespace of the Secret containing the CA certificate and key.
+    # If not set, defaults to the operator's installation namespace.
+    # I can be set only when certificateAuthority.secretName
+    # is set to a custom value, otherwise returns error.
+    secretNamespace: ""
```
pmalek: Nit: how about making this a nested tree:

```
secret:
  name: ""
  namespace: ""
```

?

charts/kong-operator/CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## Unreleased
+
+### Changed
+
+- Bootstrapping CA certificate is done automatically by Helm.
```
pmalek: Also mention which CA certificate are we talking about here.

test/helpers/operator_resources.go
```diff
@@ -0,0 +1,72 @@
...
+	cert, key := certificate.MustGenerateCertPEMFormat(
+		certificate.WithCommonName("Kong Operator CA"),
+		certificate.WithCATrue(),
+	)
+	caSecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "kong-operator-ca",
+			Namespace: namespace,
+			Labels: map[string]string{
+				"konghq.com/secret": "true",
```
tao12345666333: This one does not match the other part. `konghq.com/secret: internal`

https://github.com/Kong/kong-operator/pull/3084/changes#diff-036652936a17bce4a9fee13123d8f2df9bc000519f3fd8a183602ae011eba1d5R155

charts/kong-operator/templates/kong-operator-ca.yaml
```diff
@@ -0,0 +1,16 @@
...
+{{ $caSecretNamespace := (include "kong.caSecretNamespace" .) }}
+{{ if not (lookup "v1" "Secret" $caSecretNamespace $caSecretName) }}
+{{ $ca := genCA "Kong Operator CA" 3650 }}
+apiVersion: v1
+kind: Secret
+metadata:
+    name: {{ $caSecretName }}
+    namespace: {{ $caSecretNamespace }}
+    labels:
+        konghq.com/secret: internal
```
tao12345666333, line 11: I think we should add an annotation `helm.sh/resource-policy: keep` for the Secret resource, WDYT?

https://helm.sh/docs/howto/charts_tips_and_tricks/#tell-helm-not-to-uninstall-a-resource

programmer04: Good point, I mentioned it in values.yaml too

controller/pkg/secrets/cert.go
```diff
@@ -382,6 +386,49 @@ func generateTLSDataSecret(
 	return op.Created, generatedSecret, nil
 }
 
+// DetectCertType inspects a PEM-encoded certificate and returns its KeyConfig.
+func DetectCertType(certPEM []byte) (KeyConfig, error) {
```
tao12345666333, line 390: We can add some test cases for this function. But it's not a blocker. I'm fine if you want to add it in the follow-up PR

tao12345666333 (approved):

controller/pkg/secrets/cert.go
programmer04: I'll do it in the futre pr

PR #3102, opened 2026-01-20
feat(telemetry): Add telemetry provider for gateway count

### LINKED ISSUE #2836: Hybrid Gateway telemetry
## Problem Statement

Since we are building a brand new Gateway API-related feature, we should enable telemetry on both the feature enablement and the amount of resources created.

## Proposed Solution

Improve our telemetry with data about:

- Konnect controllers enabled
- Amount of hybrid Gateways created
- Amount of HTTPRoutes attached to Hybrid Gateways

## Acceptance Criteria

- [x] Our telemetry is improved with data coming from the Hybrid Gateways feature

randmonkey:

Adds a gateway count provider providing:
- Count of reconciled `Gateway`s
- Count of programmed `Gateway`s
- Count of attached routes to programmed `Gateway`s
- The 3 metrics above for Konnect hybrid gateways

**Which issue this PR fixes**

Fixes #2836

`controller_konnect_enabled` is already present for whether Konnect controller is enabled.

**PR Readiness Checklist**:


internal/telemetry/provider_gatewaycount.go
```diff
@@ -0,0 +1,179 @@
...
+		report[ProgrammedHybridGatewayCountKey] = programmedHybridGatewayCount
+		report[HybridGatewayAttachedRouteKey] = hybridGatewayAttachedRouteCount
+	}
+	return report, nil
+}
+
+// getGatewayConfigForGatewayClass gets the GatewayConfiguration specified for the given GatewayClass.
+// If the spec.parametersRef is empty, it returns a default GatewayConfiguration.
+// If the kind in spec.parametersRef does not match, or namespace/name is not specified, it returns error.
+// TODO: DRY -- extract the getGatewayConfigForParametersRef in gateway reconciler elsewhere and reuse it.
```
pmalek: Can we do it in this PR? If that's a lot of work let's create an issue and add a link here

internal/telemetry/provider_gatewaycount.go
```diff
@@ -0,0 +1,179 @@
...
+			Continue: continueToken,
+		})
+
+		if err != nil {
+			return nil, err
+		}
+		// Check each Gateway and increase the count if the Gateway satisfies the conditions of counting.
+		for _, gw := range gatewayList.Items {
+			// Check if the gateway is reconciled and continue checking only when it is.
+			gwc, err := gatewayclass.Get(ctx, p.cl, string(gw.Spec.GatewayClassName))
```
pmalek, line 117: Why getting `Gateway`'s `GatewayClass` is understood as the `Gateway` being reconciled?

I believe what's missing here is checking the controller name on that `GatewayClass` and whether it matches the controller name set in https://github.com/Kong/kong-operator/blob/57307f0d56c4c5e109baab66696d3b957c1ab5a5/pkg/vars/gateway.go#L36-L48.

Otherwise we'll also match `Gateway`s from other controllers.

internal/telemetry/provider_gatewaycount.go
```diff
@@ -0,0 +1,179 @@
...
+	HybridGatewayCountKey = "konnect_hybrid_gateways_count"
+	// ProgrammedHybridGatewayCountKey is the key for count of programmed Konnect hybrid gateways.
+	ProgrammedHybridGatewayCountKey = "konnect_hybrid_gateways_programmed_count"
+	// HybridGatewayAttachedRouteKey is the total count of attached routes to the programmed hybrid gateways.
+	HybridGatewayAttachedRouteKey = "konnect_hybrid_gateways_attached_routes_count"
+
+	defaultPageSize = 1000
+)
+
+// gatewayCountProvider is the provider to provide telemetry data of count of reconciled and programmed Gateways.
```
pmalek: I believe we may need to make this more verbose to include:

- an exemplar report contents (in JSON)
- an explanation how hybrid and non hybrid metrics compare to each other (as I understand this hybrid metrics will always be less than or equal to the non hybrid metrics, correct)?

internal/telemetry/provider_gatewaycount.go
```diff
@@ -0,0 +1,179 @@
...
+)
+
+const (
+	// GatewayCountProviderName is the name of the gateway count provider.
+	GatewayCountProviderName = "gateways"
+	// GatewayCountProviderKind is the kind of the gateway count provider.
+	GatewayCountProviderKind = telemetryprovider.Kind("gateway_count")
+
+	// ReconciledGatewayCountKey is the key for count of gateways reconciled by the controller.
+	ReconciledGatewayCountKey = "k8s_gateways_reconciled_count"
```
pmalek, line 26: I believe we may also add a metric for all gateways (regardless if they match kong's controller name) to understand if there are other providers being used.

internal/telemetry/provider_gatewaycount.go
randmonkey: The `gatewayclass.Get` is the same fucntion used in the reconciler to judge whether the `Gateway` is reconciled.
It returns error when the `GatewayClass`'s controller name does not match. So when the `gatewayclass.Get` returns no error, the `Gateway` can be understood to be reconciled by the controller.
Added comments here.

internal/telemetry/provider_gatewaycount.go
randmonkey: Done.

internal/telemetry/provider_gatewaycount.go
randmonkey: Added comments.

internal/telemetry/provider_gatewaycount.go
randmonkey: Added `k8s_gateways_count` metric.

internal/telemetry/provider_gatewaycount.go
```diff
@@ -0,0 +1,193 @@
...
+	defaultPageSize = 1000
+)
+
+// gatewayCountProvider is the provider to provide telemetry data of count of reconciled and programmed Gateways,
+// and sum of attached routes to the reconciled gateways.
+// The data for the Konnect hybrid gateways are collected in the reports separately.
+// Examplar reports could be:
+//
+//	{
+//	    "k8s_gateway_count":4,
```
pmalek:
```suggestion
//	    "k8s_gateways_count":4,
```

pmalek (approved):

PR #3114, opened 2026-01-22
fix konnect entities duplicate

### LINKED ISSUE #2587: Kong entities duplicated in Konnect when KongService/KongRoute missing konnect name
## Current Behavior

When a `KongService` or `KongRoute` resource is created without the entity name specified in the `spec` (that is, the field that would map to Konnect entity's name/identifier), the generic reconciler in this operator ends up creating two Konnect entities instead of one.

## Expected Behavior

Only one Konnect entity should be created per Kubernetes resource.

## Steps To Reproduce

1. Create a `KongService` (or `KongRoute`) CR without setting `spec.konnect.name` (or the appropriate konnect name field). Example:

```yaml
kind: KongService
metadata:
    name: example-service
    namespace: default
spec:
    controlPlaneRef:
        type: konnectNamespacedRef
        konnectNamespacedRef:
            name: test
    host: example-backend.default.svc.cluster.local
    port: 80
  # note: no konnect.name set here
```

2. Apply the resource to a cluster where the kong-operator is running and connected to Konnect.

3. Observe Konnect dashboard or query Konnect API for created entities.

## Operator Version

main

tao12345666333:

**Which issue this PR fixes**

Fixes https://github.com/Kong/kong-operator/issues/2587

**PR Readiness Checklist**:

### DIFF
--- a/controller/konnect/ops/ops_kongroute.go
+++ b/controller/konnect/ops/ops_kongroute.go
@@ -23,6 +23,18 @@ func createRoute(
 		return CantPerformOperationWithoutControlPlaneIDError{Entity: route, Op: CreateOp}
 	}
 
+	if route.Spec.Name == nil || *route.Spec.Name == "" {
+		existingID, err := getKongRouteForUID(ctx, sdk, route)
+		if err == nil {
+			route.SetKonnectID(existingID)
+			return nil
+		}
+		var notFound EntityWithMatchingUIDNotFoundError
+		if !errors.As(err, &notFound) {
+			return err
+		}
+	}
+
 	resp, err := sdk.CreateRoute(ctx, route.Status.Konnect.ControlPlaneID, kongRouteToSDKRouteInput(route))
 
 	if errWrap := wrapErrIfKonnectOpFailed(err, CreateOp, route); errWrap != nil {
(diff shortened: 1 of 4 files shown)


controller/konnect/ops/ops_kongroute.go
```diff
@@ -23,6 +23,18 @@ func createRoute(
 		return CantPerformOperationWithoutControlPlaneIDError{Entity: route, Op: CreateOp}
 	}
 
+	if route.Spec.Name == nil || *route.Spec.Name == "" {
+		existingID, err := getKongRouteForUID(ctx, sdk, route)
```
pmalek: Why do we want to list the existing (in Konnect) entities and use those?

As far as I can see Konnect API for e.g. service creation [API doc](https://developer.konghq.com/api/konnect/control-planes-config/v2/#/operations/create-service) does allow creating a service without a name.

BTW: I'm not sure I understand using the existing route ID here to fill in the Konnect API request. Can you elaborate?

tao12345666333: @pmalek Yes, the Konnect API allows creating a service without a name.

But when we use this API to create a service, it will create a new resource every time. It's not idempotent.

By querying for existing entities with matching UID tag, we can detect if one was already created and recover its ID, avoiding the duplicate.

Maybe we can try to use [Upsert](https://developer.konghq.com/api/konnect/control-planes-config/v2/#/operations/upsert-service) API, WDYT?

pmalek: If that would work I believe it would be slightly better (prevents the additional list operation).

tao12345666333: OK, let me try to use the upsert

pmalek (approved):

PR #3121, opened 2026-01-23
chore(hybridgateway): surface backend context in names

### LINKED ISSUE #2814: Hybrid gateways - allow Konnect users to understand which workloads are associated to the generated resource
## Problem Statement
[The actual naming of the generated resources](https://github.com/Kong/kong-operator/blob/v2.1.0-alpha.0/docs/internal/hybridgateway/autogen-resource-naming.md) fulfills engineering needs (avoids duplication, hashes the relevant parts of the source XXXRoute and is unique) but is not human friendly.
Example of a generated KongService name:
`http.cpdabefe21.22834143`

The KongService is used everywhere in the Konnect platform: Service Catalog, Analytics, Metering or Developer Portal.
Moreover, current naming of generated resources could make reading analytics very hard.

...

## Proposed Solution
Instead of (or besides of) hashing the relevant part from which a resource is generated, use the names from the source resources.
For example:
- instead of having the hash of the control plane in the resource name add the name of the backendRef gateway (or the name of the control plane).
- for the KongService and/or the KongUpstream consider if the Backends part of the name can include the referenced service names.
-  ... 

...

## Additional Information
Request coming from feedback during Alpha testing.

sandromodarelli:

Improves hybridgateway-generated KongService/KongUpstream/KongRoute names by embedding human-readable backend and route context while preserving uniqueness with stable hash suffixes. This matters for UX because Konnect now shows meaningful names instead of opaque hashes, making it much easier to map routes to the services they target. Names are capped to 1 backends with a N marker for backendRef count, sorted for stability, and truncated safely so we always retain the unique hash suffix within the 253-char limit.
I considered that Kubernetes resource names' length is 253 and that in Konnect routes and services' name have 512 max length while upstreams' name length is capped to 256.

Example (from the provided HTTPRoute):
```
kind: HTTPRoute
metadata:
  name: echo-route
  namespace: kong
spec:
  parentRefs:
    - group: gateway.networking.k8s.io
      kind: Gateway
      name: hybrid-gateway
  rules:
    - matches:
        - path:
            type: PathPrefix
            value: /echo
      backendRefs:
        - name: echo-svc
          port: 1027
        - name: echo-svc1
          port: 1027
        - name: echo-svc2
          port: 1027
```
New names (hash values are illustrative):

- KongService: `http.kong-echo-svc-1027.3.cp<hash>.<backendRefsHash>`
- KongUpstream: `http.kong-echo-svc-1027.3.cp<hash>.<backendRefsHash>`
- KongRoute: `http.kong-echo-route.cp<hash>.<matchesHash>`

Konnect will look like this:

image

**Which issue this PR fixes**

Fixes #2814 

- Backend display names are sorted and capped at 1 with a N suffix showing count of backendRefs to keep names readable and stable.
- When the full name would exceed 253 chars, the readable prefix is truncated so the hash suffix always remains.

**PR Readiness Checklist**:


controller/hybridgateway/namegen/name.go
```diff
@@ -39,31 +44,73 @@ func newName(elements ...string) string {
 	return namegenPrefix + utils.Hash64(elements)
 }
 
+func newNameWithHashSuffix(readableElements []string, hashElements []string) string {
+	allElements := append(append([]string{}, readableElements...), hashElements...)
```
pmalek, line 44:
```suggestion
	allElements := append(readableElements, hashElements...)
```

controller/hybridgateway/namegen/name.go
```diff
@@ -14,6 +16,9 @@ const (
 	// httpProcolPrefix is the prefix used for HTTP-related resources.
 	httpProcolPrefix = "http"
 
+	// backendRefDisplayLimit is the maximum number of backend references to render.
+	backendRefDisplayLimit = 2
```
pmalek: I think I'd personally go with 1 here.

This will greatly simplify the logic and prevent ambiguities. For instance: what will the name look like if there are 2 backendRefs with names close to the allowed object name (253 characters)?

controller/hybridgateway/namegen/name_test.go
```diff
@@ -184,16 +211,68 @@ func TestNewKongServiceName(t *testing.T) {
...
+	}
+
+	result := NewKongServiceName(route, cp, rule)
+	parts := strings.Split(result, ".")
+	require.GreaterOrEqual(t, len(parts), 6)
+	assert.Equal(t, httpProcolPrefix, parts[0])
+	assert.Equal(t, "default-svc-a-8080", parts[1])
+	assert.Equal(t, "default-svc-b-8080", parts[2])
+	assert.Equal(t, "more-1", parts[3])
+	assert.True(t, strings.HasPrefix(parts[len(parts)-2], defaultCPPrefix))
```
pmalek: Why aren't these tests deterministic, i.e. asserting a concrete name being returned from a concrete set of inputs?

(BTW: I've noticed the same in other placed e.g. https://github.com/Kong/kong-operator/blob/4729dd882cb84e5977e11661415bf280dc396ec9/controller/hybridgateway/namegen/name_test.go#L134-L198. I believe this should be changed to expected a concrete output and that expectation to change if the implementation changes).

controller/hybridgateway/namegen/name_test.go
```diff
@@ -184,16 +211,68 @@ func TestNewKongServiceName(t *testing.T) {
...
-			assert.GreaterOrEqual(t, len(parts), 3, fmt.Sprintf("should have at least 3 parts: %q, cp hash, and backend refs hash", httpProcolPrefix))
+			assert.GreaterOrEqual(t, len(parts), 3, fmt.Sprintf("should have at least 3 parts: %q, backend info, cp hash, and backend refs hash", httpProcolPrefix))
 			assert.Equal(t, httpProcolPrefix, parts[0], fmt.Sprintf("first part should be %q", httpProcolPrefix))
-			assert.True(t, strings.HasPrefix(parts[1], defaultCPPrefix), fmt.Sprintf("second part should start with %q", defaultCPPrefix))
+			assert.True(t, strings.HasPrefix(parts[len(parts)-2], defaultCPPrefix), fmt.Sprintf("cp hash should start with %q", defaultCPPrefix))
 		})
 	}
 }
 
+func TestNewKongServiceName_BackendDisplayLimit(t *testing.T) {
```
pmalek, line 224: Related to my comment above: could we add a test case with 2 backendrefs with very long names, e.g. 250 chars long?

controller/hybridgateway/namegen/name.go
sandromodarelli: > I think I'd personally go with 1 here.
> 
> This will greatly simplify the logic and prevent ambiguities. For instance: what will the name look like if there are 2 backendRefs with names close to the allowed object name (253 characters)?

about the value for `backendRefDisplayLimit` i don't have a strong opinion.

about service name's length capped to 63 chars because of its nature (DNS related). with `backendRefDisplayLimit = 2` we should be safe, but for a even better consistency, there's a total string cap to 253 handled by `newNameWithHashSuffix` that trims the readable part of the string to have always displayed the cp and backend hashes.

a backendRef like:
```
backendRefs:       
  - name: echo-svc-rmebsulbxilfoswmjaeghltbtjxjihuavtkczqksnkgpnjzbqjvyi
    port: 1027  
  - name: echo-svc1-rmebsulbxilfoswmjaeghltbtjxjihuavtkczqksnkgpnjzbqjvyi
    port: 1027
  - name: echo-svc2-rmebsulbxilfoswmjaeghltbtjxjihuavtkczqksnkgpnjzbqjvyi
    port: 1027  
```

will generate a KongService named:
```
http.kong-echo-svc-rmebsulbxilfoswmjaeghltbtjxjihuavtkczqksnkgpnjzbqjvyi-1027.kong-echo-svc1-rmebsulbxilfoswmjaeghltbtjxjihuavtkczqksnkgpnjzbqjvyi-1027.more-1.cp8c3a12e.b121685d
```

controller/hybridgateway/namegen/name_test.go
sandromodarelli: tests refactored in a deterministic way

controller/hybridgateway/namegen/name.go
sandromodarelli: i won't change that because your suggestion can extend the same backing array and overwrite elements in `readableElements`, which is then reused later in the function.

as a proof:
```
package main

import "fmt"

func main() {
	readable := make([]string, 2, 4)
	readable[0] = "http"
	readable[1] = "svc-a"
	hash := []string{"cp123", "deadbeef"}

	all := append(readable, hash...)
	///all := append(append([]string{}, readable...), hash...)

	fmt.Println("all:", all)
	fmt.Println("readable (len):", readable)
	fmt.Println("readable (full cap):", readable[:cap(readable)])
}
```

controller/hybridgateway/namegen/name_test.go
sandromodarelli: edge cases tests added

controller/hybridgateway/namegen/name.go
pmalek: Very good point!

Nonetheless, `readableElements` is only used in:

```
readablePart := strings.Join(readableElements, ".")
```

which reads it (and does not change it) and allocates `readablePart` which is then returned. It would seem to me then that my suggestion is safe. Are there any concerns here that I have not addressed?

sandromodarelli: if we have to implement some other manipulation in the future, the risk is that a ghost element could lead to headaches. out of specific semantic, do you have any other reason for changing it?

controller/hybridgateway/namegen/name.go
sandromodarelli: changed to `backendRefDisplayLimit = 1` and tests adjusted accordingly

controller/hybridgateway/namegen/name.go
pmalek: I understand your concern. I believe that tests _should_ catch these cases but I also see that there might be some cases where we wouldn't catch it and hit reusing the backing array of `readableElements`.

Let's leave this as is then.

controller/hybridgateway/namegen/name.go
```diff
@@ -39,31 +41,73 @@ func newName(elements ...string) string {
 	return namegenPrefix + utils.Hash64(elements)
 }
 
+func newNameWithHashSuffix(readableElements []string, hashElements []string) string {
+	allElements := append(append([]string{}, readableElements...), hashElements...)
+	name := strings.Join(allElements, ".")
+	if len(name) <= maxLen {
+		return name
+	}
```
pmalek: Nit: can we revert the logic here to keep the "happy path least indented" (ref: https://maelvls.dev/go-happy-line-of-sight/) ?

controller/hybridgateway/namegen/name.go
```diff
@@ -39,31 +41,73 @@ func newName(elements ...string) string {
...
+
+	hashPart := strings.Join(hashElements, ".")
+	if hashPart == "" {
```
pmalek: This can only happen if `len(hashElements)` is 0. Can we check it this way before calling `strings.Join`?

controller/hybridgateway/namegen/name.go
```diff
@@ -39,31 +41,73 @@ func newName(elements ...string) string {
...
+		return namegenPrefix + utils.Hash64(allElements)
+	}
+	if len(hashPart) > maxLen {
+		return namegenPrefix + utils.Hash64(allElements)
+	}
+
+	readablePart := strings.Join(readableElements, ".")
```
pmalek: This can move right before usage below.

controller/hybridgateway/namegen/name.go
```diff
@@ -39,31 +41,73 @@ func newName(elements ...string) string {
...
+	remaining := maxLen - len(hashPart) - 1
+	if remaining <= 0 {
+		return hashPart
+	}
+	if len(readablePart) > remaining {
+		readablePart = strings.TrimRight(readablePart[:remaining], ".-")
+	}
+	if readablePart == "" {
+		return hashPart
+	}
```
pmalek: This can only happen if `len(readableElements)` is 0. Can we test it that way?

controller/hybridgateway/namegen/name.go
sandromodarelli: i didn't get your observation because the code seems already in the "happy path" to me. can you elaborate a bit more? 🙏

pmalek: Git diff might not show it well but the basic idea is to keep the happy path (`return name`) in the left most indented line.

The "special case" - when we don't fit - is handled in the `if` statement.

```diff
diff --git a/controller/hybridgateway/namegen/name.go b/controller/hybridgateway/namegen/name.go
index dddb1ce56..4107526a1 100644
--- a/controller/hybridgateway/namegen/name.go
+++ b/controller/hybridgateway/namegen/name.go
@@ -44,36 +44,34 @@ func newName(elements ...string) string {
 func newNameWithHashSuffix(readableElements []string, hashElements []string) string {
 	allElements := append(append([]string{}, readableElements...), hashElements...)
 	name := strings.Join(allElements, ".")
-	if len(name) <= maxLen {
-		return name
+	// If the name is too long, truncate it.
+	if len(name) > maxLen {
+		// No hash elements: fall back to a deterministic hash of everything.
+		if len(hashElements) == 0 {
+			return namegenPrefix + utils.Hash64(allElements)
+		}
+
+		hashPart := strings.Join(hashElements, ".")
+		// If the hash part alone is too long, also fall back to hashing everything.
+		if len(hashPart) > maxLen {
+			return namegenPrefix + utils.Hash64(allElements)
+		}
+
+		remaining := maxLen - len(hashPart) - 1 // space for readable + "." + hashPart
+		// Not enough room for "<readable>." prefix or no readable elements: return just the hash part.
+		if remaining <= 0 || len(readableElements) == 0 {
+			return hashPart
+		}
+
+		readablePart := strings.Join(readableElements, ".")
+		if len(readablePart) > remaining {
+			readablePart = strings.TrimRight(readablePart[:remaining], ".-")
+		}
+		return readablePart + "." + hashPart
 	}
 
-	// No hash elements: fall back to a deterministic hash of everything.
-	if len(hashElements) == 0 {
-		return namegenPrefix + utils.Hash64(allElements)
-	}
+	return name
 
-	hashPart := strings.Join(hashElements, ".")
-	// If the hash part alone is too long, also fall back to hashing everything.
-	if len(hashPart) > maxLen {
-		return namegenPrefix + utils.Hash64(allElements)
-	}
-
-	remaining := maxLen - len(hashPart) - 1 // space for readable + "." + hashPart
-	// Not enough room for "<readable>." prefix or no readable elements: return just the hash part.
-	if remaining <= 0 || len(readableElements) == 0 {
-		return hashPart
-	}
-
-	readablePart := strings.Join(readableElements, ".")
-	if len(readablePart) > remaining {
-		readablePart = strings.TrimRight(readablePart[:remaining], ".-")
-	}
-	if len(readablePart) == 0 {
-		return hashPart
-	}
-
-	return readablePart + "." + hashPart
 }
 
 // NewKongUpstreamName generates a KongUpstream name based on the ControlPlaneRef and HTTPRouteRule passed as arguments.
```

controller/hybridgateway/namegen/name.go
```diff
@@ -39,31 +41,79 @@ func newName(elements ...string) string {
...
+			return hashPart
+		}
+
+		readablePart := strings.Join(readableElements, ".")
+		if len(readablePart) > remaining {
+			readablePart = strings.TrimRight(readablePart[:remaining], ".-")
+		}
+		if len(readablePart) == 0 {
+			return hashPart
+		}
```
pmalek: We don't need this because we already check `len(readableElements) == 0` in line 62 and `len(readablePart) == 0` can only be satisfied when `len(readableElements) == 0` is hit.

controller/hybridgateway/namegen/name.go
```diff
@@ -100,3 +150,55 @@ func NewKongCertificateName(gatewayName, listenerPort string) string {
...
+		names = append(names, displayName)
+	}
+
+	if len(names) == 0 {
+		return nil
+	}
+
+	sort.Strings(names)
+	count := fmt.Sprintf("%d", len(refs))
+	return []string{names[0], count}
```
pmalek: We can slightly optimize this by not using a slice but just using a variable and comparing to it:


```suggestion
	var name string
	for _, ref := range refs {
		displayName := backendRefDisplayName(routeNamespace, ref)
		if displayName == "" {
			continue
		}
		if displayName < name || name == "" {
			name = displayName
		}
	}

	if len(name) == 0 {
		return nil
	}

	count := fmt.Sprintf("%d", len(refs))
	return []string{name, count}
```

Benchmark:

```
func BenchmarkBackendRefDisplayNames(b *testing.B) {
	port := func(value gatewayv1.PortNumber) *gatewayv1.PortNumber { return &value }
	ns := func(value string) *gatewayv1.Namespace { n := gatewayv1.Namespace(value); return &n }
	backendRef := func(name string, namespace *gatewayv1.Namespace, portNumber *gatewayv1.PortNumber) gatewayv1.HTTPBackendRef {
		return gatewayv1.HTTPBackendRef{
			BackendRef: gatewayv1.BackendRef{
				BackendObjectReference: gatewayv1.BackendObjectReference{
					Name:      gatewayv1.ObjectName(name),
					Namespace: namespace,
					Port:      portNumber,
				},
			},
		}
	}

	benchmarks := []struct {
		name    string
		routeNS string
		refs    []gatewayv1.HTTPBackendRef
	}{
		{
			name:    "single backend",
			routeNS: "default",
			refs: []gatewayv1.HTTPBackendRef{
				backendRef("service-a", nil, port(8080)),
			},
		},
		{
			name:    "three backends",
			routeNS: "default",
			refs: []gatewayv1.HTTPBackendRef{
				backendRef("service-b", nil, port(8080)),
				backendRef("service-a", nil, port(8080)),
				backendRef("service-c", nil, port(9090)),
			},
		},
		{
			name:    "ten backends with namespaces",
			routeNS: "default",
			refs: []gatewayv1.HTTPBackendRef{
				backendRef("svc-1", ns("ns-alpha"), port(8080)),
				backendRef("svc-2", ns("ns-beta"), port(8081)),
				backendRef("svc-3", ns("ns-gamma"), port(8082)),
				backendRef("svc-4", ns("ns-delta"), port(8083)),
				backendRef("svc-5", ns("ns-epsilon"), port(8084)),
				backendRef("svc-6", ns("ns-zeta"), port(8085)),
				backendRef("svc-7", ns("ns-eta"), port(8086)),
				backendRef("svc-8", ns("ns-theta"), port(8087)),
				backendRef("svc-9", ns("ns-iota"), port(8088)),
				backendRef("svc-10", ns("ns-kappa"), port(8089)),
			},
		},
		{
			name:    "long service names",
			routeNS: "default",
			refs: []gatewayv1.HTTPBackendRef{
				backendRef(strings.Repeat("a", 63), nil, port(8080)),
				backendRef(strings.Repeat("b", 63), nil, port(8080)),
				backendRef(strings.Repeat("c", 63), nil, port(8080)),
			},
		},
	}

	b.ResetTimer()
	for _, bm := range benchmarks {
		b.Run(bm.name, func(b *testing.B) {
			b.ReportAllocs()
			for b.Loop() {
				_ = backendRefDisplayNames(bm.routeNS, bm.refs)
			}
		})
	}
}
```

results current:

```
goos: darwin
goarch: arm64
pkg: github.com/kong/kong-operator/controller/hybridgateway/namegen
cpu: Apple M4 Max
BenchmarkBackendRefDisplayNames/single_backend-14         	 9757056	       110.3 ns/op	      64 B/op	       4 allocs/op
BenchmarkBackendRefDisplayNames/single_backend-14         	10916268	       111.2 ns/op	      64 B/op	       4 allocs/op
BenchmarkBackendRefDisplayNames/three_backends-14         	 4323079	       278.2 ns/op	     176 B/op	      11 allocs/op
BenchmarkBackendRefDisplayNames/three_backends-14         	 4385283	       274.9 ns/op	     176 B/op	      11 allocs/op
BenchmarkBackendRefDisplayNames/ten_backends_with_namespaces-14         	 1435428	       833.5 ns/op	     516 B/op	      33 allocs/op
BenchmarkBackendRefDisplayNames/ten_backends_with_namespaces-14         	 1468354	       815.0 ns/op	     516 B/op	      33 allocs/op
BenchmarkBackendRefDisplayNames/long_service_names-14                   	 4130460	       288.4 ns/op	     344 B/op	      11 allocs/op
BenchmarkBackendRefDisplayNames/long_service_names-14                   	 4141528	       291.3 ns/op	     344 B/op	      11 allocs/op
PASS
ok  	github.com/kong/kong-operator/controller/hybridgateway/namegen	9.784s
```

results changed:

```
goos: darwin
goarch: arm64
pkg: github.com/kong/kong-operator/controller/hybridgateway/namegen
cpu: Apple M4 Max
BenchmarkBackendRefDisplayNames/single_backend-14         	10304224	       108.7 ns/op	      64 B/op	       4 allocs/op
BenchmarkBackendRefDisplayNames/single_backend-14         	11578266	       107.7 ns/op	      64 B/op	       4 allocs/op
BenchmarkBackendRefDisplayNames/three_backends-14         	 4571641	       264.7 ns/op	     128 B/op	      10 allocs/op
BenchmarkBackendRefDisplayNames/three_backends-14         	 4494463	       264.2 ns/op	     128 B/op	      10 allocs/op
BenchmarkBackendRefDisplayNames/ten_backends_with_namespaces-14         	 1601481	       752.0 ns/op	     356 B/op	      32 allocs/op
BenchmarkBackendRefDisplayNames/ten_backends_with_namespaces-14         	 1591310	       756.5 ns/op	     356 B/op	      32 allocs/op
BenchmarkBackendRefDisplayNames/long_service_names-14                   	 4215264	       284.8 ns/op	     296 B/op	      10 allocs/op
BenchmarkBackendRefDisplayNames/long_service_names-14                   	 4356696	       276.2 ns/op	     296 B/op	      10 allocs/op
PASS
ok  	github.com/kong/kong-operator/controller/hybridgateway/namegen	9.902s
```

diff:

```
benchstat -alpha 0.5 slice single_name
goos: darwin
goarch: arm64
pkg: github.com/kong/kong-operator/controller/hybridgateway/namegen
cpu: Apple M4 Max
                                                       │    slice     │            single_name             │
                                                       │    sec/op    │    sec/op     vs base              │
BackendRefDisplayNames/single_backend-14                 110.8n ± ∞ ¹   108.2n ± ∞ ¹  -2.30% (p=0.333 n=2)
BackendRefDisplayNames/three_backends-14                 276.6n ± ∞ ¹   264.4n ± ∞ ¹  -4.38% (p=0.333 n=2)
BackendRefDisplayNames/ten_backends_with_namespaces-14   824.2n ± ∞ ¹   754.2n ± ∞ ¹  -8.49% (p=0.333 n=2)
BackendRefDisplayNames/long_service_names-14             289.9n ± ∞ ¹   280.5n ± ∞ ¹  -3.23% (p=0.333 n=2)
geomean                                                  292.5n         278.9n        -4.63%
¹ need >= 6 samples for confidence interval at level 0.95

                                                       │    slice    │             single_name              │
                                                       │    B/op     │    B/op      vs base                 │
BackendRefDisplayNames/single_backend-14                 64.00 ± ∞ ¹   64.00 ± ∞ ¹        ~ (p=1.000 n=2) ²
BackendRefDisplayNames/three_backends-14                 176.0 ± ∞ ¹   128.0 ± ∞ ¹  -27.27% (p=0.333 n=2)
BackendRefDisplayNames/ten_backends_with_namespaces-14   516.0 ± ∞ ¹   356.0 ± ∞ ¹  -31.01% (p=0.333 n=2)
BackendRefDisplayNames/long_service_names-14             344.0 ± ∞ ¹   296.0 ± ∞ ¹  -13.95% (p=0.333 n=2)
geomean                                                  211.5         171.4        -18.94%
¹ need >= 6 samples for confidence interval at level 0.95
² all samples are equal

                                                       │    slice    │             single_name             │
                                                       │  allocs/op  │  allocs/op   vs base                │
BackendRefDisplayNames/single_backend-14                 4.000 ± ∞ ¹   4.000 ± ∞ ¹       ~ (p=1.000 n=2) ²
BackendRefDisplayNames/three_backends-14                 11.00 ± ∞ ¹   10.00 ± ∞ ¹  -9.09% (p=0.333 n=2)
BackendRefDisplayNames/ten_backends_with_namespaces-14   33.00 ± ∞ ¹   32.00 ± ∞ ¹  -3.03% (p=0.333 n=2)
BackendRefDisplayNames/long_service_names-14             11.00 ± ∞ ¹   10.00 ± ∞ ¹  -9.09% (p=0.333 n=2)
geomean                                                  11.24         10.64        -5.38%
```

pmalek (approved):

PR #3180, opened 2026-01-30
feat(hybridgateway): generate KongPlugin copies for HTTPRoute filters

mlavacca:

Add logic to automatically create copies of KongPlugins when they are referenced by HTTPRoute extensionRef filters. When an HTTPRoute uses a KongPlugin via Gateway API filters, the controller now generates a dedicated copy of that plugin with proper ownership and annotations linking it back to the HTTPRoute.

**Which issue this PR fixes**

Fixes #

Manifests

```yaml
kind: Gateway
metadata:
  name: jw-gw
  namespace: default
spec:
  gatewayClassName: kong-v2beta1
  listeners:
  - name: http
    protocol: HTTP
    port: 80
    allowedRoutes:
      namespaces:
        from: All
---
kind: HTTPRoute
metadata:
  name: httproute-echo
  namespace: route-ns
spec:
  parentRefs:
  - name: jw-gw
    namespace: default
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /echo
    filters:
    - type: ExtensionRef
      extensionRef:
        kind: KongPlugin
        name: rate-limit-5-min
    backendRefs:
    - name: echo
      kind: Service
      port: 80
---
kind: KongPlugin
metadata:
 name: rate-limit-5-min
 namespace: route-ns
config:
 minute: 5
 policy: local
plugin: rate-limiting
```

**PR Readiness Checklist**:

### DIFF
--- a/controller/hybridgateway/plugin/plugin.go
+++ b/controller/hybridgateway/plugin/plugin.go
@@ -71,7 +71,6 @@ func PluginsForFilter(
 				log.Debug(logger, "Referenced KongPlugin not found")
 				return plugins, nil
 			}
-			log.Error(logger, err, "Failed to retrieve referenced KongPlugin")
 			return nil, fmt.Errorf("failed to retrieve referenced KongPlugin: %w", err)
 		}
 		pluginName := plugin.Name
@@ -85,7 +84,6 @@ func PluginsForFilter(
 			WithAnnotations(httpRoute, pRef).
 			Build()
 		if err != nil {
-			log.Error(logger, err, "Failed to build KongPlugin resource")
 			return nil, fmt.Errorf("failed to build KongPlugin %s: %w", pluginName, err)
 		}
 		plugins = append(plugins, pluginCopy)
@@ -111,7 +109,6 @@ func PluginsForFilter(
 			WithAnnotations(httpRoute, pRef).
 			Build()
 		if err != nil {
-			log.Error(logger, err, "Failed to build KongPlugin resource")
 			return nil, fmt.Errorf("failed to build KongPlugin %s: %w", pluginName, err)
 		}
 
--- a/controller/hybridgateway/plugin/plugin.go
+++ b/controller/hybridgateway/plugin/plugin.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 
 	"github.com/go-logr/logr"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
@@ -48,7 +49,6 @@ import (
 //
 // Returns:
 //   - kongPlugins: The translated plugin(s).
-//   - selfManaged: True if sourced from ExtensionRef.
 //   - err: Any error encountered.
 func PluginsForFilter(
 	ctx context.Context,
@@ -58,7 +58,7 @@ func PluginsForFilter(
 	rule gwtypes.HTTPRouteRule,
 	filter gwtypes.HTTPRouteFilter,
 	pRef *gwtypes.ParentReference,
-) ([]configurationv1.KongPlugin, bool, error) {
+) ([]configurationv1.KongPlugin, error) {
 	logger = logger.WithValues("filter-type", filter.Type)
 	plugins := []configurationv1.KongPlugin{}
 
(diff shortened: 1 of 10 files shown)


pmalek:
Can we add the problematic manifest to the PR description to see the actual problem?

Just for thought: copying the `KongPlugin` resources around will mean that we need to maintain these copies to be up to date with the original which complicates things. Wouldn't it be possible to generate the `KongReferenceGrant` instead to allow the reference?

mlavacca:
> Can we add the problematic manifest to the PR description to see the actual problem?
> 
> Just for thought: copying the `KongPlugin` resources around will mean that we need to maintain these copies to be up to date with the original which complicates things. Wouldn't it be possible to generate the `KongReferenceGrant` instead to allow the reference?

The plugin enforcement logic is already in place as we create plugins out of Gateway API filters. I agree this solution is not optimal and the creation of a `KongReferenceGrant` would be best here, but we lack all machinery to do that, while we already have everything in place to deal with Plugin duplication. We've been talking with @justindavies and he's fine with it, considering the time constraint. In case we want to iterate and implement the `KongReferenceGrant` solution, we can do it in a subsequent release.

controller/hybridgateway/plugin/plugin.go
```diff
@@ -58,31 +58,47 @@ func PluginsForFilter(
...
 	if filter.Type == gatewayv1.HTTPRouteFilterExtensionRef {
 		log.Debug(logger, "Filter is an ExtensionRef, retrieving referenced KongPlugin")
 		plugin, err := getReferencedKongPlugin(ctx, cl, httpRoute.Namespace, filter)
-		pluginName := plugin.Name
 		if err != nil {
+			if k8serrors.IsNotFound(err) {
+				log.Debug(logger, "Referenced KongPlugin not found")
+				return plugins, nil
+			}
 			log.Error(logger, err, "Failed to retrieve referenced KongPlugin")
```
pmalek: Nit: we can address this separately but here we log an error and we also return it to the caller (where it's also logged e.g. in `controller/hybridgateway/converter/http_route.go:448`)

This duplicates the error information being  logged and creates noise. I think a better approach here could be to drop the log here and just return the error.

controller/hybridgateway/plugin/plugin.go
```diff
@@ -58,31 +58,47 @@ func PluginsForFilter(
...
+		pluginCopy, err := builder.NewKongPlugin().
+			WithName(namegen.NewKongPluginName(filter, httpRoute.Namespace, plugin.PluginName)).
+			WithNamespace(metadata.NamespaceFromParentRef(httpRoute, pRef)).
+			WithLabels(httpRoute, pRef).
+			WithPluginName(plugin.PluginName).
+			WithPluginConfig(plugin.Config.Raw).
+			WithAnnotations(httpRoute, pRef).
+			Build()
+		if err != nil {
+			log.Error(logger, err, "Failed to build KongPlugin resource")
```
pmalek: Ditto here about not logging but just returning.


```suggestion
```

controller/hybridgateway/plugin/plugin.go
```diff
@@ -96,16 +112,16 @@ func PluginsForFilter(
 			Build()
 		if err != nil {
 			log.Error(logger, err, "Failed to build KongPlugin resource")
```
pmalek: Ditto: 
```suggestion
```

internal/utils/index/httproute.go
```diff
@@ -89,3 +99,27 @@ func GatewaysOnHTTPRoute(o client.Object) []string {
...
+	for _, rule := range httpRoute.Spec.Rules {
+		for _, filter := range rule.Filters {
+			if filter.Type != gatewayv1.HTTPRouteFilterExtensionRef || filter.ExtensionRef == nil {
+				continue
+			}
+			if filter.ExtensionRef.Group != gatewayv1.Group(configurationv1.GroupVersion.Group) || filter.ExtensionRef.Kind != "KongPlugin" {
+				continue
+			}
+			ns := httpRoute.Namespace
+			plugins = append(plugins, ns+"/"+string(filter.ExtensionRef.Name))
```
pmalek: This is the same for each iteration so we might as well use `httpRoute.Namespace` below or move the `ns` declaration out of the loop.

I personally suggest the latter.


```suggestion
			plugins = append(plugins, httpRoute.Namespace+"/"+string(filter.ExtensionRef.Name))
```

controller/hybridgateway/watch/mapfuncs_httproute.go
```diff
@@ -256,3 +256,43 @@ func MapHTTPRouteForKongResource[T kongResource](cl client.Client) handler.MapFu
...
+		for i, httpRoute := range httpRoutes.Items {
+			indexRequests[i] = reconcile.Request{
+				NamespacedName: client.ObjectKey{
+					Namespace: httpRoute.Namespace,
+					Name:      httpRoute.Name,
+				},
+			}
+		}
+
+		requests := MapHTTPRouteForKongResource[*configurationv1.KongPlugin](cl)(ctx, obj)
```
pmalek, line 295:
```suggestion
		// Add requests for HTTPRoutes referencing the plugin via annotations.
		requests := MapHTTPRouteForKongResource[*configurationv1.KongPlugin](cl)(ctx, obj)
```

pmalek (dismissed):
Overall, looks good to me.

Some minor nits but otherwise 👍

pmalek:
> > Can we add the problematic manifest to the PR description to see the actual problem?
> > Just for thought: copying the `KongPlugin` resources around will mean that we need to maintain these copies to be up to date with the original which complicates things. Wouldn't it be possible to generate the `KongReferenceGrant` instead to allow the reference?
> 
> The plugin enforcement logic is already in place as we create plugins out of Gateway API filters. I agree this solution is not optimal and the creation of a `KongReferenceGrant` would be best here, but we lack all machinery to do that, while we already have everything in place to deal with Plugin duplication. We've been talking with @justindavies and he's fine with it, considering the time constraint. In case we want to iterate and implement the `KongReferenceGrant` solution, we can do it in a subsequent release.

👍 Let's make sure that we capture this with an issue. Regardless if we are going to plan the work or make a decision to not pursue this I believe it will be valuable for project planning. WDYT?

mlavacca:
> > > Can we add the problematic manifest to the PR description to see the actual problem?
> > > Just for thought: copying the `KongPlugin` resources around will mean that we need to maintain these copies to be up to date with the original which complicates things. Wouldn't it be possible to generate the `KongReferenceGrant` instead to allow the reference?
> > 
> > 
> > The plugin enforcement logic is already in place as we create plugins out of Gateway API filters. I agree this solution is not optimal and the creation of a `KongReferenceGrant` would be best here, but we lack all machinery to do that, while we already have everything in place to deal with Plugin duplication. We've been talking with @justindavies and he's fine with it, considering the time constraint. In case we want to iterate and implement the `KongReferenceGrant` solution, we can do it in a subsequent release.
> 
> 👍 Let's make sure that we capture this with an issue. Regardless if we are going to plan the work or make a decision to not pursue this I believe it will be valuable for project planning. WDYT?

here it is: https://github.com/Kong/kong-operator/issues/3212

alacuku (approved):
/lgtm

pmalek (approved):

Issue #3187, opened 2026-01-30
Plugin deletion behaviour on self managed ControlPlane


justindavies:
Ok, here's the result of testing with self managed for annotations and ExtensioRef:

# Create a dedicated namespace for plugins
kubectl create namespace plugins
# Deploy the plugin
kubectl apply -f - <<EOF
apiVersion: configuration.konghq.com/v1
kind: KongPlugin
metadata:
  name: rate-limit-cross-repro
  namespace: plugins
plugin: rate-limiting
config:
  minute: 5
  policy: local
EOF
# Authorize the team1 namespace to use this plugin
kubectl apply -f - <<EOF
apiVersion: gateway.networking.k8s.io/v1beta1
kind: ReferenceGrant
metadata:
  name: allow-plugins-repro
  namespace: plugins
spec:
  from:
    - group: gateway.networking.k8s.io
      kind: HTTPRoute
      namespace: team1
  to:
    - group: configuration.konghq.com
      kind: KongPlugin
EOF

kubectl annotate httproute echo -n team1 \
  konghq.com/plugins=plugins:rate-limit-cross-repro --overwrite

Iterate a curl to test the rate limit...

kubectl delete kongplugin rate-limit-cross-repro -n plugins

Check kubectl get httproute echo -n team1 -o yaml

Status does not change on the route - I do see this in the events:

Warning  KongConfigurationTranslationFailed  httproute/echo-cross-deletion  referenced KongPlugin or KongClusterPlugin "plugins:rate-limit-cross-deletion" does not exist

  - conditions:
    - lastTransitionTime: "2026-01-30T22:43:36Z"
      reason: Accepted
      status: "True"
      type: Accepted
    - lastTransitionTime: "2026-01-30T22:43:36Z"
      reason: ResolvedRefs
      status: "True"      # <--- Should be False if a filter ref is missing?

ExtensionRef Scenario

kubectl apply -f - <<EOF
apiVersion: configuration.konghq.com/v1
kind: KongPlugin
metadata:
  name: rate-limit-extref-repro
  namespace: team1
plugin: rate-limiting
config:
  minute: 5
  policy: local
EOF

kubectl apply -f - <<EOF
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: echo-extref-repro
  namespace: team1
spec:
  parentRefs: [{name: kong, namespace: kong}]
  rules:
[...]

Issue #1399, opened 2025-03-25
Simplify management of Kong Plugins


programmer04:
## Problem Statement

Currently, there are two dedicated CRDs 
- [KongPlugin](https://docs.konghq.com/kubernetes-ingress-controller/latest/reference/custom-resources/#kongplugin)
- [KongClusterPlugin](https://docs.konghq.com/kubernetes-ingress-controller/latest/reference/custom-resources/#kongclusterplugin)

How Kong Plugins can be configured
- Plugins can be associated with the `Ingress`, `Service`, `HTTPRoute`, `KongConsumer`, or `KongConsumerGroup` object in Kubernetes using [konghq.com/plugins annotation](https://docs.konghq.com/kubernetes-ingress-controller/latest/reference/annotations/#konghqcomplugins).
- For `KongClusterPlugin`s global label is used, which is not well documented, you can find the description in [this example](https://docs.konghq.com/kubernetes-ingress-controller/latest/get-started/proxy-caching/#create-a-proxy-cache-kongclusterplugin) and [PR KIC #112](https://github.com/Kong/kubernetes-ingress-controller/pull/112).
- For [KongPlugin cross-namespace reference with a ReferenceGrant is allowed too](https://docs.konghq.com/kubernetes-ingress-controller/3.2.x/plugins/cross-namespace/), which basically can replace KongClusterPlugins entirely. See [KIC issue #5355](https://github.com/Kong/kubernetes-ingress-controller/issues/5355) for more context.

Everything aforementioned happens in annotations and labels.

## Proposed Solution

Leave only KongPlugin CRD that works with ReferenceGrant and [enhance KongPluginBinding](https://github.com/Kong/kubernetes-ingress-controller/issues/6139#issuecomment-2610404987) to handle all cases that used to be handled with annotations explicitly with CRDs.

## Additional Information

Problems with the current approach that will be resolved
- https://github.com/Kong/kubernetes-ingress-controller/issues/6139
- https://github.com/Kong/gateway-operator/issues/660

## Acceptance Criteria
- [ ] only `KongPlugin` that works with `ReferenceGrant` and enhanced `KongPluginBinding`

czeslavo:
@programmer04 will update the issue description with links to the design doc with more details about the proposed solution.

sandromodarelli:
punting to KO 2.1.0 due to other priorities, maybe can land in future milestones

sandromodarelli:
[...]

PR #3206, opened 2026-02-04
fix: update httproute resolved refs for plugins

tao12345666333:

This PR fixes the `ResolvedRefs` status condition on `HTTPRoute` (and other route types) in self-managed ControlPlane mode. Currently, when a `KongPlugin` referenced by an `HTTPRoute` is deleted, the route's `ResolvedRefs` condition remains True, which is incorrect and misleading.

This is a bug fix for existing behavior; it does not introduce annotations as a new plugin reference mechanism. Both extensionRef and annotation-based plugin references are already supported in self-managed mode; this PR simply ensures that the `ResolvedRefs` status accurately reflects whether those references are valid.

**Which issue this PR fixes**

Part of #3187

- This PR is scoped to self-managed mode only (`ingress-controller/` path). It does not affect hybrid mode.
- Per the current KO 2.1 strategy: self-managed supports both annotations and extensionRef for plugin references. This PR only fixes the status reporting for these existing mechanisms.
- The long-term plan to deprecate annotations (#1399) is not affected — this fix ensures correct behavior while annotations remain supported.

**PR Readiness Checklist**:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -591,7 +591,7 @@ test.unit.pretty:
 	@$(MAKE) _test.unit GOTESTSUM_FORMAT=pkgname GOTESTFLAGS="$(GOTESTFLAGS)" UNIT_TEST_PATHS="$(UNIT_TEST_PATHS)"
 
 ENVTEST_TEST_PATHS := ./test/envtest/...
-ENVTEST_TIMEOUT ?= 10m
+ENVTEST_TIMEOUT ?= 20m
 PKG_LIST=./controller/...,./internal/...,./pkg/...,./modules/...
 TEST_DIR ?= $(PROJECT_DIR)
 
--- a/Makefile
+++ b/Makefile
@@ -591,7 +591,7 @@ test.unit.pretty:
 	@$(MAKE) _test.unit GOTESTSUM_FORMAT=pkgname GOTESTFLAGS="$(GOTESTFLAGS)" UNIT_TEST_PATHS="$(UNIT_TEST_PATHS)"
 
 ENVTEST_TEST_PATHS := ./test/envtest/...
-ENVTEST_TIMEOUT ?= 20m
+ENVTEST_TIMEOUT ?= 15m
 PKG_LIST=./controller/...,./internal/...,./pkg/...,./modules/...
 TEST_DIR ?= $(PROJECT_DIR)
 
--- a/ingress-controller/internal/controllers/gateway/grpcroute_controller.go
+++ b/ingress-controller/internal/controllers/gateway/grpcroute_controller.go
@@ -38,6 +38,8 @@ import (
 type GRPCRouteReconciler struct {
 	client.Client
 
+	APIReader client.Reader
+
 	Log             logr.Logger
 	Scheme          *runtime.Scheme
 	DataplaneClient controllers.DataPlane
@@ -55,6 +57,10 @@ type GRPCRouteReconciler struct {
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *GRPCRouteReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	if r.APIReader == nil {
+		r.APIReader = mgr.GetAPIReader()
+	}
+
 	blder := ctrl.NewControllerManagedBy(mgr).
 		// set the controller name
 		Named("grpcroute-controller").
@@ -106,7 +112,7 @@ func (r *GRPCRouteReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return blder.
 		For(&gatewayapi.GRPCRoute{},
 			builder.WithPredicates(
-				IsRouteAttachedToReconciledGatewayPredicate[*gatewayapi.GRPCRoute](r.Client, mgr.GetLogger(), r.GatewayNN),
+				IsRouteAttachedToReconciledGatewayPredicate[*gatewayapi.GRPCRoute](mgr.GetAPIReader(), mgr.GetLogger(), r.GatewayNN),
 			),
 		).
 		Complete(r)
@@ -303,7 +309,11 @@ func (r *GRPCRouteReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	// we need to pull the Gateway parent objects for the grpcroute to verify
 	// routing behavior and ensure compatibility with Gateway configurations.
 	debug(log, grpcroute, "Retrieving GatewayClass and Gateway for route")
-	gateways, err := getSupportedGatewayForRoute(ctx, log, r.Client, grpcroute, r.GatewayNN)
+	reader := r.APIReader
+	if reader == nil {
+		reader = r.Client
+	}
+	gateways, err := getSupportedGatewayForRoute(ctx, log, reader, grpcroute, r.GatewayNN)
 	if err != nil {
 		if errors.Is(err, ErrNoSupportedGateway) {
 			// if there's no supported Gateway then this route could have been previously
--- a/ingress-controller/internal/controllers/gateway/grpcroute_controller.go
+++ b/ingress-controller/internal/controllers/gateway/grpcroute_controller.go
@@ -38,8 +38,6 @@ import (
 type GRPCRouteReconciler struct {
 	client.Client
 
-	APIReader client.Reader
-
 	Log             logr.Logger
 	Scheme          *runtime.Scheme
 	DataplaneClient controllers.DataPlane
@@ -57,10 +55,6 @@ type GRPCRouteReconciler struct {
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *GRPCRouteReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	if r.APIReader == nil {
-		r.APIReader = mgr.GetAPIReader()
-	}
-
 	blder := ctrl.NewControllerManagedBy(mgr).
 		// set the controller name
 		Named("grpcroute-controller").
@@ -112,7 +106,7 @@ func (r *GRPCRouteReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return blder.
 		For(&gatewayapi.GRPCRoute{},
 			builder.WithPredicates(
-				IsRouteAttachedToReconciledGatewayPredicate[*gatewayapi.GRPCRoute](mgr.GetAPIReader(), mgr.GetLogger(), r.GatewayNN),
+				IsRouteAttachedToReconciledGatewayPredicate[*gatewayapi.GRPCRoute](r.Client, mgr.GetLogger(), r.GatewayNN),
 			),
 		).
 		Complete(r)
@@ -309,11 +303,7 @@ func (r *GRPCRouteReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	// we need to pull the Gateway parent objects for the grpcroute to verify
 	// routing behavior and ensure compatibility with Gateway configurations.
 	debug(log, grpcroute, "Retrieving GatewayClass and Gateway for route")
-	reader := r.APIReader
-	if reader == nil {
-		reader = r.Client
-	}
-	gateways, err := getSupportedGatewayForRoute(ctx, log, reader, grpcroute, r.GatewayNN)
+	gateways, err := getSupportedGatewayForRoute(ctx, log, r.Client, grpcroute, r.GatewayNN)
 	if err != nil {
 		if errors.Is(err, ErrNoSupportedGateway) {
 			// if there's no supported Gateway then this route could have been previously
--- a/ingress-controller/internal/controllers/gateway/httproute_controller.go
+++ b/ingress-controller/internal/controllers/gateway/httproute_controller.go
@@ -46,6 +46,8 @@ import (
 type HTTPRouteReconciler struct {
 	client.Client
 
+	APIReader client.Reader
+
 	Log              logr.Logger
 	Scheme           *runtime.Scheme
 	DataplaneClient  controllers.DataPlane
@@ -65,6 +67,10 @@ type HTTPRouteReconciler struct {
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *HTTPRouteReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	if r.APIReader == nil {
+		r.APIReader = mgr.GetAPIReader()
+	}
+
 	// We're verifying whether ReferenceGrant CRD is installed at setup of the HTTPRouteReconciler
 	// to decide whether we should run additional ReferenceGrant watch and handle ReferenceGrants
 	// when reconciling HTTPRoutes.
@@ -142,7 +148,7 @@ func (r *HTTPRouteReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return blder.
 		For(&gatewayapi.HTTPRoute{},
 			builder.WithPredicates(
-				IsRouteAttachedToReconciledGatewayPredicate[*gatewayapi.HTTPRoute](r.Client, mgr.GetLogger(), r.GatewayNN),
+				IsRouteAttachedToReconciledGatewayPredicate[*gatewayapi.HTTPRoute](mgr.GetAPIReader(), mgr.GetLogger(), r.GatewayNN),
 			),
 		).
 		Complete(r)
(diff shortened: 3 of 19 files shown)


controller/konnect/reconciler_controlplaneref.go
```diff
@@ -169,7 +169,7 @@ func ensureKongReferenceGrant[
 			consts.ConditionType(configurationv1alpha1.KongReferenceGrantConditionTypeResolvedRefs),
 			metav1.ConditionFalse,
 			configurationv1alpha1.KongReferenceGrantReasonRefNotPermitted,
-			fmt.Sprintf("KongReferenceGrants do not allow access to KonnectGatewayControlPlane %s", cpRef),
+			fmt.Sprintf("KongReferenceGrants do not allow access to KonnectGatewayControlPlane %s", (&cpRef).String()),
```
pmalek:
```suggestion
			fmt.Sprintf("KongReferenceGrants do not allow access to KonnectGatewayControlPlane %s", cpRef.String()),
```

We don't really need the reference/pointer: when a method is defined on a pointer receiver it's also available for values

controller/konnect/reconciler_controlplaneref.go
```diff
@@ -181,7 +181,7 @@ func ensureKongReferenceGrant[
 		consts.ConditionType(configurationv1alpha1.KongReferenceGrantConditionTypeResolvedRefs),
 		metav1.ConditionTrue,
 		configurationv1alpha1.KongReferenceGrantReasonResolvedRefs,
-		fmt.Sprintf("KongReferenceGrants allow access to KonnectGatewayControlPlane %s", cpRef),
+		fmt.Sprintf("KongReferenceGrants allow access to KonnectGatewayControlPlane %s", (&cpRef).String()),
```
pmalek:
```suggestion
		fmt.Sprintf("KongReferenceGrants allow access to KonnectGatewayControlPlane %s", cpRef.String()),
```

Makefile
```diff
@@ -591,7 +591,7 @@ test.unit.pretty:
 	@$(MAKE) _test.unit GOTESTSUM_FORMAT=pkgname GOTESTFLAGS="$(GOTESTFLAGS)" UNIT_TEST_PATHS="$(UNIT_TEST_PATHS)"
 
 ENVTEST_TEST_PATHS := ./test/envtest/...
-ENVTEST_TIMEOUT ?= 10m
+ENVTEST_TIMEOUT ?= 20m
```
```suggestion
ENVTEST_TIMEOUT ?= 15m
```

If we need to bump this let's be conservative about it.

test/integration/suite.go
```diff
@@ -165,6 +174,25 @@ func Suite(m *testing.M) {
 	}
 }
 
+func ensureKubeconfig() (string, func()) {
```
pmalek: Why do we need this `func`? It seems to create an empty kubeconfig when it's not specified. I'm having hard time trying to understand the reason 🤔

ingress-controller/internal/controllers/gateway/grpcroute_controller.go
```diff
@@ -38,6 +38,8 @@ import (
 type GRPCRouteReconciler struct {
 	client.Client
 
+	APIReader client.Reader
```
pmalek: Why did we decide to go with reader (which is uncached as far as I know) instead of using `client.Client`?

ingress-controller/internal/controllers/gateway/httproute_controller.go
```diff
@@ -132,12 +148,64 @@ func (r *HTTPRouteReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-				IsRouteAttachedToReconciledGatewayPredicate[*gatewayapi.HTTPRoute](r.Client, mgr.GetLogger(), r.GatewayNN),
+				IsRouteAttachedToReconciledGatewayPredicate[*gatewayapi.HTTPRoute](mgr.GetAPIReader(), mgr.GetLogger(), r.GatewayNN),
 			),
 		).
 		Complete(r)
 }
 
+const httpRoutePluginRefIndexKey = "httproute-pluginref"
+
+func setupHTTPRouteIndices(mgr ctrl.Manager) error {
```
pmalek: Can we split index related functions to a separate file with readability?

ingress-controller/internal/controllers/gateway/httproute_controller.go
```diff
@@ -132,12 +148,64 @@ func (r *HTTPRouteReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+	if len(refs) == 0 {
+		return []string{}
+	}
+
+	keys := make([]string, 0, len(refs))
+	for key := range refs {
+		keys = append(keys, key)
+	}
+	return keys
```
pmalek: This is essentially:

```suggestion
	return slices.Collect(maps.Keys(keys))
```

Or if we want to trade verbosity for performance:

```suggestion
	return slices.AppendSeq(make([]string, 0, len(keys)), maps.Keys(keys))
```

ingress-controller/internal/controllers/gateway/httproute_controller.go
```diff
@@ -669,10 +765,142 @@ func (r *HTTPRouteReconciler) getHTTPRouteRuleReason(ctx context.Context, httpRo
...
+					nil
+			}
+
+			if referenceGrants == nil {
+				grants, err := r.listReferenceGrants(ctx)
+				if err != nil {
+					return "", "", err
+				}
+				referenceGrants = grants
+			}
```
pmalek: This is essentially done once, whenever we have at least 1 pluginRef. Can we move this outside of the loop to make this clearer?

ingress-controller/internal/controllers/gateway/httproute_controller.go
```diff
@@ -669,10 +765,142 @@ func (r *HTTPRouteReconciler) getHTTPRouteRuleReason(ctx context.Context, httpRo
...
+	}
+
+	return gatewayapi.RouteReasonResolvedRefs, "", nil
+}
+
+func (r *HTTPRouteReconciler) listReferenceGrants(ctx context.Context) ([]*gatewayapi.ReferenceGrant, error) {
+	referenceGrantList := &gatewayapi.ReferenceGrantList{}
+	if err := r.List(ctx, referenceGrantList); err != nil {
+		return nil, err
+	}
```
pmalek, line 834: Can we limit the list operation here just to the relevant namespace?

This would mean that we need to do it more than once but OTOH in clusters with a lot of namespaces and a lot of `ReferenceGrant`s this would potentially save a lot of compute. WDYT?

ingress-controller/internal/controllers/gateway/httproute_controller.go
```diff
@@ -190,6 +258,29 @@ func referenceGrantHasHTTPRouteFrom(obj client.Object) bool {
 	return false
 }
 
+// listHTTPRoutesForKongPlugin returns the list of HTTPRoutes that reference the given KongPlugin.
+func (r *HTTPRouteReconciler) listHTTPRoutesForKongPlugin(ctx context.Context, obj client.Object) []reconcile.Request {
+	plugin, ok := obj.(*configurationv1.KongPlugin)
+	if !ok {
+		r.Log.Error(fmt.Errorf("invalid type"), "Found invalid type in event handlers", "expected", "KongPlugin", "found", reflect.TypeOf(obj))
```
pmalek: Can we extract the error here to a named sentinel error?

I can see it could be used in other watch helpers like this (to be addressed separately). WDYT?

ingress-controller/internal/controllers/gateway/grpcroute_controller.go
tao12345666333: Yes, it is uncached.

TBH, it's just copied/pasted the existing code ..


https://github.com/Kong/kong-operator/blob/ba227565e5c943bc97c3d04c55e99a702be756a6/ingress-controller/internal/controllers/gateway/tcproute_controller.go#L41

ingress-controller/internal/controllers/gateway/httproute_controller.go
tao12345666333: Yes, thank you! I will change it.

ingress-controller/internal/controllers/gateway/httproute_controller.go
tao12345666333: good point! let me update

ingress-controller/internal/controllers/gateway/grpcroute_controller.go
pmalek: This https://github.com/Kong/kong-operator/blame/ba227565e5c943bc97c3d04c55e99a702be756a6/ingress-controller/internal/controllers/gateway/tcproute_controller.go#L41 is not on `main`.

https://github.com/Kong/kong-operator/blame/db41f6db6dbf7e81639f05bc025a28e0322920a0/ingress-controller/internal/controllers/gateway/tcproute_controller.go#L38-L49 is on `main` and does not use the APIReader.

ingress-controller/internal/controllers/gateway/httproute_controller.go
tao12345666333: Sure https://github.com/Kong/kong-operator/pull/3206/changes/fb6228b0c7279553d57c7b6390d59adf48daca0d#diff-f8b3c5216c236ba08bae5e8f1041e67ef177c7cd42f238815075bf43ee0b6ea5

ingress-controller/internal/controllers/gateway/grpcroute_controller.go
tao12345666333: Thank you! Reverted. https://github.com/Kong/kong-operator/pull/3206/commits/25f7c364d2d6d5526c89082da1c5578914e0e770

test/integration/suite.go
tao12345666333: removed.

ingress-controller/internal/controllers/utils/ingress_predicates.go
```diff
@@ -77,6 +77,14 @@ func IsIngressClassEmpty(obj client.Object) bool {
...
-	_, err := restMapper.KindsFor(gvr)
-	return err == nil
+	if _, err := restMapper.KindsFor(gvr); err == nil {
+		return true
+	} else if meta.IsNoMatchError(err) {
+		if resettable, ok := restMapper.(interface{ Reset() }); ok {
+			resettable.Reset()
+			_, err = restMapper.KindsFor(gvr)
+			return err == nil
+		}
```
pmalek, line 90: Can we comment here why we want to `Reset()`? In what circumstances restMapper is going to have this method?

ingress-controller/test/kongintegration/expression_router_test.go
```diff
@@ -22,8 +22,6 @@ import (
 )
 
 func TestExpressionsRouterMatchers_GenerateValidExpressions(t *testing.T) {
-	t.Parallel()
-
```
pmalek, line 26: Can we extract this change to a separate PR? (similar with other removals of `t.Parallel()` in kongintegration tests)

ingress-controller/internal/controllers/gateway/route_predicates.go
```diff
@@ -115,9 +115,25 @@ func IsRouteAttachedToReconciledGateway[routeT gatewayapi.RouteT](
...
+	cl client.Reader, log logr.Logger, gatewayNN controllers.OptionalNamespacedName, e event.UpdateEvent,
 ) bool {
 	oldObj, newObj := e.ObjectOld, e.ObjectNew
-	return IsRouteAttachedToReconciledGateway[routeT](cl, log, gatewayNN, oldObj) ||
-		IsRouteAttachedToReconciledGateway[routeT](cl, log, gatewayNN, newObj)
+	if IsRouteAttachedToReconciledGateway[routeT](cl, log, gatewayNN, oldObj) ||
+		IsRouteAttachedToReconciledGateway[routeT](cl, log, gatewayNN, newObj) {
+		return true
+	}
+	return routeHasKongParentStatus[routeT](oldObj) || routeHasKongParentStatus[routeT](newObj)
```
pmalek, line 129: I don't think we should rely on status fields. Status should be mostly informational.

ingress-controller/internal/controllers/gateway/route_predicates.go
```diff
@@ -95,14 +95,14 @@ func IsRouteAttachedToReconciledGateway[routeT gatewayapi.RouteT](
 			err := cl.Get(context.Background(), k8stypes.NamespacedName{Namespace: namespace, Name: string(parentRef.Name)}, &gateway)
 			if err != nil {
 				log.Error(err, "Failed to get Gateway in HTTPRoute watch")
-				return false
+				return true
```
pmalek, line 100: Why are we changing this (and below) to return true?

test/envtest/httproute_controller_test.go
```diff
@@ -185,16 +186,24 @@ func TestHTTPRouteReconcilerProperlyReactsToReferenceGrant(t *testing.T) {
...
-				Status: "False",
-				Reason: "RefNotPermitted",
-			},
-		),
+		func() bool {
+			if _, err := reconciler.Reconcile(ctx, reconcileRequest); err != nil {
+				t.Logf("reconcile failed: %v", err)
+				return false
+			}
+			return containsInitial()
```
pmalek: I'm not sure I understand why we'd need to sprinkle the `Reconcile()` calls across this test.

The reconciler itself is started in line 71 and the whole point of that is that it runs in the background and reacts to API events.

I suggest that we remove this and if the test does not pass we'll need to change watch criteria or something else to ensure that it does pass.

pmalek (changes requested):

ingress-controller/internal/controllers/gateway/route_predicates.go
tao12345666333: In controller-runtime predicates, returning false drops that watch event and there's no retry for it. 

If the predicate can't determine attachment because a Gateway/GatewayClass lookup fails (cache staleness, API errors, or even the object being deleted), returning false risks skipping reconciliation until some unrelated event happens to re-enqueue the route. 

Returning true fails open: it may cause extra reconciles, but reconcile is where we already re-check attachment/ownership and can no-op or clean up status/dataplane state. 

I prefer extra reconciliation over silently skipping potentially relevant routes. WDYT?

tao12345666333: The `routeHasKongParentStatus` check serves as a fallback for cases where attachment changes due to external object modifications (e.g, GatewayClass controller name changes) that cause both old and new parentRef-based checks to return false. By checking if we previously set a parent status (identified by our ControllerName), we ensure we still reconcile the route to clean up stale status/dataplane state.

ingress-controller/internal/controllers/gateway/route_predicates.go
tao12345666333: https://github.com/Kong/kong-operator/pull/3206#discussion_r2804916604

ingress-controller/internal/controllers/gateway/route_predicates.go
pmalek: I see your point and I agree.

I'm only concerned about the the unnecessary cost of these reconcile requests but as you mentioned, this is the cost of consistency. Otherwise we'd be waiting for controller-runtime sync or other events to trigger.

Good point! Let's leave this as is.

ingress-controller/test/kongintegration/expression_router_test.go
tao12345666333: OK, I will revert these changes in this PR.

ingress-controller/internal/controllers/utils/ingress_predicates.go
tao12345666333: Sure

test/envtest/httproute_controller_test.go
tao12345666333: OK

tao12345666333: removed https://github.com/Kong/kong-operator/pull/3206/commits/4f2403ce40bb714037e47e81efdd8aa6a8e83559

ingress-controller/test/kongintegration/expression_router_test.go
tao12345666333: https://github.com/Kong/kong-operator/pull/3206/changes/13816c9f188313db5db39056a70a5e710a8cb17b

ingress-controller/internal/controllers/utils/ingress_predicates.go
tao12345666333: https://github.com/Kong/kong-operator/pull/3206/changes/13816c9f188313db5db39056a70a5e710a8cb17b#diff-0d258b33ad9da5bb4e186a34a855a5fa89d7c1a96fd7eeb14226f797d9629805R83-R85

pmalek (approved):

PR #3248, opened 2026-02-09
chore(lint): enable golangci-lint: `embeddedstructfieldcheck`

programmer04:

One from the series, to keep it reviewable. 

Enable and apply fixes for [embeddedstructfieldcheck](https://golangci-lint.run/docs/linters/configuration/#embeddedstructfieldcheck). It enforces the consistent order of embedded struct fields and forbids embedding mutexes in such a way. It makes codebase more consistent

**Which issue this PR fixes**

Part of 

- https://github.com/Kong/kong-operator/issues/1847


pmalek (dismissed):

pmalek (approved):

api/konnect/v1alpha1/konnect_cloudgateway_dataplane_configuration_types.go
```diff
@@ -220,7 +220,7 @@ type KonnectCloudGatewayDataPlaneGroupConfigurationStatus struct {
 	// +patchMergeKey=type
 	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,1,rep,name=conditions"`
 
-	konnectv1alpha2.KonnectEntityStatusWithControlPlaneRef `json:",inline"`
+	konnectv1alpha2.KonnectEntityStatusWithControlPlaneRef `json:",inline"` //nolint:embeddedstructfieldcheck
```
pmalek, line 223: Why do we need to keep these `nolint` directives here? Otherwise we'd need to move this field to be the first in this type?

programmer04: because it conflicts with the rule from ` make lint.api` to keep conditions at the top, and it makes sense, I think, to always have conditions in docs on the top WDYT? Unfortunately, there isn't a more elegant way to make this rule conditional...

pmalek: Ok, fair enough. I wonder if it would make sense to submit a change to https://github.com/kubernetes-sigs/kube-api-linter for the conditions rule so that those are allowed to be preceded by embedded fields.

PR #3277, opened 2026-02-11
chore: increase telepresence timeout

pmalek:

Set telepresence helm timeout to 60s (from default 30s: https://telepresence.io/docs/reference/config#timeouts:~:text=15%20seconds-,helm,30%20seconds,-Local%20Overrides%E2%80%8B)

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:

### DIFF
--- a/scripts/telepresence-manager.sh
+++ b/scripts/telepresence-manager.sh
@@ -5,7 +5,7 @@
 set -e
 set -o pipefail
 
-readonly CONFIG="./.config/telepresence/config.yml"
+readonly CONFIG="./.config_telepresence.yaml"
 
 # Function to log messages with different levels.
 log() {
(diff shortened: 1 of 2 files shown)


scripts/telepresence-manager.sh
```diff
@@ -5,6 +5,8 @@
 set -e
 set -o pipefail
 
+readonly CONFIG="./.config/telepresence/config.yml"
```
programmer04: Maybe instead of introducing such a complicated path, just put in the root of the repo `.telepresence_config.yaml` and it will be consistent with other configs

pmalek: Moved to `.config_telepresence.yaml` 960bd793739e9841e679b1bcc214e1b5a35adcde.

This way we can have a share prefix for config files. WDYT?

programmer04 (approved):

PR #3278, opened 2026-02-11
chore(lint): enable golangci-lint: `testifylint`

programmer04:

One from the series, to keep it reviewable. 

Enable and apply fixes for [testifylint](https://golangci-lint.run/docs/linters/configuration/#testifylint), which are resonable

Rule [require-error](https://github.com/Antonboom/testifylint?tab=readme-ov-file#require-error) has been disabled, because it's not crucial and leads to hundreds of changes, so it's not worth doing.

Rule [error-is-as](https://github.com/Antonboom/testifylint?tab=readme-ov-file#error-is-as) will be enabled in a separate PR because it finds legit issues with the codebase that have to be addressed.

**Which issue this PR fixes**

Part of 

- https://github.com/Kong/kong-operator/issues/1847

### DIFF
--- a/test/envtest/run.go
+++ b/test/envtest/run.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/samber/lo"
 	"github.com/samber/mo"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/zap/zaptest/observer"
 	k8stypes "k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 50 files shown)


test/envtest/run.go
tao12345666333, line 252: I think we can keep the `require.NoError` here, or change to:

```
mgr, err := manager.NewManager(ctx, mgrID, logger, cfg)
if !assert.NoError(t, err) {
    return nil
}
```

Because if `require.NoError` failed, it will stop immediately, but the `assert` will continue the test.

programmer04: It was changed due to

```
test/envtest/telemetry_test.go:53:3: go-require: RunManager contains assertions that must only be used in the goroutine running the test function (testifylint)
                RunManager(ctx, t, envcfg, AdminAPIOptFns(),
```

But you're right, hence in bd46ef58aa18066b9ee228de909bf6b13bd251c2 I refactored this, PTAL

pmalek (dismissed):

pmalek (approved):

PR #3342, opened 2026-02-17
chore(lint): resolve all TODOs from `.golangci.yaml`

### LINKED ISSUE #1847: Adjust golangci-lint configuration

programmer04:

Removes all TODOs from `.golangci.yaml`. 

Diff for 
```
# Uncomment or remove - TODO: https://github.com/Kong/kong-operator/issues/1847
        # - pattern: ^.*Dataplane[^G].*$
        #   msg: Please use camel case 'DataPlane' instead of 'Dataplane'
```

is enormous and in sdk `Dataplane` is used, so interfaces etc. have to match.

For section

```
# Remove/adjust whole section - TODO: https://github.com/Kong/kong-operator/issues/1847
```

too many false positives.

Closes #1847

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -85,9 +85,6 @@ linters:
       default-signifies-exhaustive: true
     forbidigo:
       forbid:
-        # Uncomment or remove - TODO: https://github.com/Kong/kong-operator/issues/1847
-        # - pattern: ^.*Dataplane[^G].*$
-        #   msg: Please use camel case 'DataPlane' instead of 'Dataplane'
         - pattern: ^.*Controlplane.*$
           msg: Please use camel case 'ControlPlane' instead of 'Controlplane'
         - pattern: ^.*operatorv1beta1.ControlPlane.*$
@@ -208,7 +205,6 @@ linters:
           - gosec
         path: .*_test\.go
         text: Use of weak random number generator
-      # Remove/adjust whole section - TODO: https://github.com/Kong/kong-operator/issues/1847
       - linters:
           - gosec
         path: .*_test\.go


.golangci.yaml
```diff
@@ -85,9 +85,6 @@ linters:
       default-signifies-exhaustive: true
     forbidigo:
       forbid:
-        # Uncomment or remove - TODO: https://github.com/Kong/kong-operator/issues/1847
-        # - pattern: ^.*Dataplane[^G].*$
-        #   msg: Please use camel case 'DataPlane' instead of 'Dataplane'
```
pmalek, line 90: Why not uncomment this and fix the underlying issues?

programmer04: As in the description, the diff is enormous, and in sdk Dataplane is used, so interfaces etc., have to match, so exclusion will be needed. I considered this as not worth it. But if you insist, I can fight with it. So?

pmalek: Ah, I searched for exact matches for `Dataplane` and have not seen that it's also used as a substring.

Fair enough, let's leave this as is.

pmalek (approved):

PR #3357, opened 2026-02-18
fix: name in Konnect matches K8s name for `KonnectGatewayControlPlane`

programmer04:

To prevent

```
Message:               {"status":409,"title":"Conflict","instance":"","detail":"Key (org_id, name)=(8a6e97b1-b481-4fc2-8399-a59077076af6, gateway-conformance-infra-same-namespace) already exists."}
```

for `KonnectGatewayControlPlane` name specified in Spec section Create Control Plane Request in the field Name has to be unique across Konnect org.

In a real-world scenario, such colision may happen when a user has to separate K8s with the same configuration (e.g., `dev-1` and `dev-2`) in a single Konnect org. The fix proposed in this PR resolves the problem.

**Which issue this PR fixes**

Fixes #

To match the name in Konnect with the name in the K8s cluster, relying on server-side name generation is not an option, the code responsible for it has been incorporated in KO.

**PR Readiness Checklist**:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,10 @@
 - Fix `ResolvedRefs` status condition on `HTTPRoute` not being updated when a
   referenced `KongPlugin` is deleted in self-managed ControlPlane mode.
   [#3206](https://github.com/Kong/kong-operator/pull/3206)
+- Name of Konnect Gateway Control Plane resource created in Konnect matches
+  the name of the corresponding `KonnectGatewayControlPlane` resource in Kubernetes
+  also in for name with random suffix. It prevents collisions in Konnect.
+  [#3357](https://github.com/Kong/kong-operator/pull/3357)
 
 ## [v2.1.0]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,7 +66,7 @@
   [#3206](https://github.com/Kong/kong-operator/pull/3206)
 - Name of Konnect Gateway Control Plane resource created in Konnect matches
   the name of the corresponding `KonnectGatewayControlPlane` resource in Kubernetes
-  also in for name with random suffix. It prevents collisions in Konnect.
+  (the same random suffix is added). It prevents collisions in Konnect.
   [#3357](https://github.com/Kong/kong-operator/pull/3357)
 
 ## [v2.1.0]
--- a/pkg/utils/kubernetes/metadata.go
+++ b/pkg/utils/kubernetes/metadata.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilrand "k8s.io/apimachinery/pkg/util/rand"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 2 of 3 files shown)


mlavacca (commented):
With this change, we end up having the `KonnectGatewayControlPlane` Kubernetes resource name equal to the Konnect resource name. Is that correct?

CHANGELOG.md
```diff
@@ -64,6 +64,10 @@
 - Fix `ResolvedRefs` status condition on `HTTPRoute` not being updated when a
   referenced `KongPlugin` is deleted in self-managed ControlPlane mode.
   [#3206](https://github.com/Kong/kong-operator/pull/3206)
+- Name of Konnect Gateway Control Plane resource created in Konnect matches
+  the name of the corresponding `KonnectGatewayControlPlane` resource in Kubernetes
+  also in for name with random suffix. It prevents collisions in Konnect.
```
pmalek: > also in for name with random suffix

This does not sound correct. Can we rephrase it? What are we trying to say here?

pkg/utils/kubernetes/metadata.go
```diff
@@ -94,3 +95,22 @@ func TrimGenerateName(name string) string {
...
+	base = TrimGenerateName(base)
+
+	const (
+		maxNameLength          = 63
+		randomLength           = 5
+		maxGeneratedNameLength = maxNameLength - randomLength
+	)
+	if len(base) > maxGeneratedNameLength {
+		base = base[:maxGeneratedNameLength]
+	}
```
pmalek: `base` is provided with `-` and trimmer if too long.

This way we can loose the `-` and simply attach the random string to the base.

WDYT about changing the approach and moving appending the `-` inside the `GenerateName` to prevent this?

programmer04:
> With this change, we end up having the `KonnectGatewayControlPlane` Kubernetes resource name equal to the Konnect resource name. Is that correct?

Exactly @mlavacca

pmalek:
IIUC for already existing resources this won't change anything, right?

In that case I believe it's a fix that we should backport to 2.1.x

pkg/utils/kubernetes/metadata.go
programmer04: basically I left it to be consistent with copied function, putting `-` properly is ensured by `TrimGenerateName` called at the top. I can remove it, PTAL

programmer04:
> IIUC for already existing resources this won't change anything, right?
> 
> In that case I believe it's a fix that we should backport to 2.1.x

As I spoke with @mlavacca yes

mlavacca:
> IIUC for already existing resources this won't change anything, right?
> 
> In that case I believe it's a fix that we should backport to 2.1.x

Yes, we should backport it to 2.1.x as this is backward compatible. Already existing deployments will be left untouched

mlavacca (approved):
Looks good!

pmalek:
https://github.com/Kong/kong-operator/pull/3361 has been merged and fixes the failing kongintegration tests.

pmalek (approved):

pkg/utils/kubernetes/metadata.go
pmalek: 👍

pmalek:
2.1 backport in https://github.com/Kong/kong-operator/pull/3368

PR #3412, opened 2026-02-23
feat(gateway): handle labels and annotation from Gateway `spec.infrastructure`

### LINKED ISSUE #1791: Support labelling and annotating with `Gateway` `spec.infrastructure`
## Problem Statement
Support for spec.infrastructure to label / annotate infrastrcuture created as part of the Gateway.  This is useful for customers who need to placing billing or organisational metadata on Kong resources.  This is also useful for example to annotate that a Gateway should be part of Kong mesh.

## Acceptance Criteria
- [ ] Pass infrastructure annotations and labels to the Gateway dataplane

sandromodarelli:

Propagate Gateway.spec.infrastructure labels and annotations to DataPlane resources

When a Gateway has spec.infrastructure.labels or spec.infrastructure.annotations set, the operator now propagates those values to the DataPlane's ingress Service metadata and Deployment pod template metadata.

Merge priority: GatewayConfiguration settings are applied first; Gateway.spec.infrastructure values are merged on top (last-write-wins), so infrastructure-level metadata always takes effect.

**Which issue this PR fixes**

Fixes #1791 

**PR Readiness Checklist**:


api/gateway-operator/v1beta1/dataplane_types.go
```diff
@@ -302,6 +302,15 @@ type ServiceOptions struct {
...
 	Annotations map[string]string `json:"annotations,omitempty" protobuf:"bytes,12,rep,name=annotations"`
 
+	// Labels are an unstructured key value map that may be used to organize and
+	// categorize resources and that are propagated to the DataPlane's ingress Service
+	// by the operator.
+	//
+	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/
+	//
+	// +optional
+	Labels map[string]string `json:"labels,omitempty" protobuf:"bytes,13,rep,name=labels"`
```
pmalek: This is already an issue with other API that were introduced to specify objects labels and/or annotations but I believe we should start thinking about validating the names and values with CEL rules.

For example the labels are constrained as explained in https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set.

While this could be done as a separate PR I'd propose the following:
- either do this in this PR
- or schedule it in the same version to prevent releasing this and then breaking the API by constraining it in subsequent release.

pmalek (commented):
Looks good overall. I believe though, that there 3 things missing from this PR which would greatly improve it if added:

- a changelog entry
- a sample manifest showing how to use this
- CRD validation tests using this feature.

api/gateway-operator/v1beta1/dataplane_types.go
sandromodarelli: added CEL validation rules for label keys and values

sandromodarelli: i tried to implement a full validation with regex/string ops, but it turned out that the rule and the generated schema are exceeding the CRD validation cost budget. i set only validations for length

config/samples/gateway-with-infrastructure-labels_v2.yaml
```diff
@@ -0,0 +1,57 @@
...
+  name: kong
+  namespace: default
+spec:
+  gatewayClassName: kong
+  infrastructure:
+    labels:
+      environment: production
+      team: platform
+    annotations:
+      service.beta.kubernetes.io/aws-load-balancer-type: nlb
```
pmalek: I believe is not encouraging users to misuse this: `Service` specific annotations can be applied through dedicated options:

```
spec:
  dataPlaneOptions:
    network:
      services:
        ingress:
          annotations:
            foo: bar
```

If the proposed field is used then that annotation is applied to all managed resources - not only `Service`s - which IMHO is not correct (even though is allowed and will work).

My proposal then for this sample manifest is to use a generic annotation like above (`environment: production`) and a dedicated LB annotation with the above mentioned dedicated field.

api/gateway-operator/v1beta1/dataplane_types.go
pmalek: Thanks for addressing this.

I wanted to suggest using a separate type for this, like:

```
// LabelName is the name of a label that can be applied to a resource.
type LabelName string

// LabelValue is the value of a label that can be applied to a resource.
//
// +kubebuilder:validation:MaxLength=63
// +kubebuilder:validation:Pattern=`^[a-zA-Z0-9]([-a-zA-Z0-9]*[a-zA-Z0-9])?$`
type LabelValue string
```

That does create other conflicts in the code and requires to adapt code in various places to this change, e.g. this would make the types in https://github.com/Kong/kong-operator/blob/cf2ad33dd128bce10f6022c8b72a88196501a8e9/controller/gateway/controller_reconciler_utils.go#L269 incompatible but perhaps it could be worth it.

We can also try to reuse the validations used by Gateway API: https://github.com/kubernetes-sigs/gateway-api/blob/v1.4.1/apis/v1/gateway_types.go#L1006-L1022.

Is that what caused exceeding the budget?

sandromodarelli: added your suggested implementation and some tests to cover malformed labels

config/samples/gateway-with-infrastructure-labels_v2.yaml
sandromodarelli: done

config/samples/gateway-with-infrastructure-labels_v2.yaml
```diff
@@ -0,0 +1,62 @@
...
+  name: kong
+  namespace: default
+spec:
+  gatewayClassName: kong
+  infrastructure:
+    labels:
+      environment: production
+      team: platform
+    annotations:
+      enviroment: production
```
tao12345666333:
```suggestion
      environment: production
```

controller/dataplane/controller_utils.go
```diff
@@ -86,6 +86,32 @@ func extractDataPlaneIngressServiceAnnotations(dataplane *operatorv1beta1.DataPl
 	return anns
 }
 
+func addLabelsForDataPlaneIngressService(obj client.Object, dataplane operatorv1beta1.DataPlane) {
```
pmalek: This `func` should only work on `corev1.Service` objects right?

It relies on dp's `.Spec.Network.Services.Ingress.Labels` so allowing other types - through the use of `client.Object` - does not really make sense here right?

pmalek (commented):
1 nit and small proposal:

Since this feature is getting implemented, can we enable [the `GatewayInfrastructure` test](https://github.com/kubernetes-sigs/gateway-api/blob/b311b1c0d2a1a81fed447f70ea9db9869d3a8c12/conformance/tests/gateway-infrastructure.go) in conformance suite now?

controller/dataplane/controller_utils.go
sandromodarelli: i used the same approach of the existing `func addAnnotationsForDataPlaneIngressService` right above. https://github.com/Kong/kong-operator/blob/bc1c07f1b535faed525dfd2d81fceacff7e3564b/controller/dataplane/controller_utils.go#L61

i would say that your suggestion seems valid to me and i should change it in both functions, WDYT?

pmalek: Agreed 👍 

Missed `addAnnotationsForDataPlaneIngressService` since it's not changed in this PR

sandromodarelli: done. added also unit tests for `addAnnotationsForDataPlaneIngressService` func

pmalek (approved):
Approving as discussed on slack.

Let's make sure to unmark it fixing https://github.com/Kong/kong-operator/issues/1791 since that should also cover enabling the [the GatewayInfrastructure test](https://github.com/kubernetes-sigs/gateway-api/blob/b311b1c0d2a1a81fed447f70ea9db9869d3a8c12/conformance/tests/gateway-infrastructure.go) in Gateway API conformance suite.

sandromodarelli:
> 1 nit and small proposal:
> 
> Since this feature is getting implemented, can we enable [the `GatewayInfrastructure` test](https://github.com/kubernetes-sigs/gateway-api/blob/b311b1c0d2a1a81fed447f70ea9db9869d3a8c12/conformance/tests/gateway-infrastructure.go) in conformance suite now?

i checked that conformance test and it's not passing because KO is not compliant with https://gateway-api.sigs.k8s.io/geps/gep-1762/ where all underlying resources of a gateway must have a label `gateway.networking.k8s.io/gateway-name: GATEWAY_NAME` and the conformance test uses that as a label selector to check the propagation of `spec.infrastructure` labels and annotations into underlying resources.
my proposal is to merge this PR and have a dedicated issue to have KO compliant with GEP-1762 and enable the `GatewayInfrastructure` conformance test. 

tracking issue https://github.com/Kong/kong-operator/issues/3523

PR #3417, opened 2026-02-24
chore: generate `ControlPlane`'s `ValidatingWebhookConfiguration` based on kubebuilder code markers

### LINKED ISSUE #2186: IMPROVE: Generate `ControlPlane`'s `ValidatingWebhookConfiguration` based on kubebuilder code markers
### Problem statement

KIC has historically used kubebuilder code markers for generating `ValidatingWebhookConfiguration`.

For example:

https://github.com/Kong/kubernetes-ingress-controller/blob/59fadd5e714b70c7a96f06ae61168ed499f69ca2/internal/admission/handler.go#L177

To achieve KO 2.0 release, validating webhook has been generated once, and added to the repository in https://github.com/Kong/kong-operator/pull/2084.

This issue tracks the effort to make the `ValidatingWebhookConfiguration` manifest generated using controller-gen, just like it has been done in KIC, e.g.: https://github.com/Kong/kubernetes-ingress-controller/blob/d8eb9a0bb3f673ea1d52008567368a3fdb6b44da/Makefile#L280-L284

### Acceptance criteria

- [ ] `ValidatingWebhookConfiguration` for `ControlPlane`s is generated from source code.

### LINKED ISSUE #2544: Failure to start KO may cause validating webhooks to block all updates of services
## Current Behavior

In current charts for installing KO, if validating webhooks are successfully created but KO did not start successfully, ALL updates to `Service`s in the cluster will fail with the following message:
```
Internal error occurred: failed calling webhook "[services.validations.kong.konghq.com](http://services.validations.kong.konghq.com/)": failed to call webhook: Post "https://kong-operator-kong-operator-webhook.kong-system.svc:5443/?timeout=10s
": no endpoints available for service "kong-operator-kong-operator-webhook" (retried 5 times).
```

## Expected Behavior

When KO cannot start successfully, update to other resources should be operational in the cluster.

## Steps To Reproduce

Run the following command to install KO:
```
 helm upgrade --install kong-operator kong/kong-operator -n kong-system   --create-namespace     --set env.ENABLE_CONTROLLER_KONNECT=true --set global.webhooks.options.certManager.enabled=true --set image.tag=2.0.5 
```
and block KO to start (for example, change the image tag to some non-exist image)
Then update any service in the cluster.

## Operator Version
2.0.5 (charts 1.0.2)
## Acceptance criteria

- [ ] Migrate admission webhook validations from https://github.com/Kong/kong-operator/blob/6ac999759d0e46d9f4ba6b551353c1fd8e955da4/ingress-controller/internal/admission/handler.go#L577-L592 to `ValidatingAdmissionPolicy` in kong-operator chart
- [ ] Make that `ValidatingAdmissionPolicy` enabled by default but allow users to disable it.

~This is an issue should be resolved in charts. Currently `failurePolicy` are hard coded to `Fail` in the templates for all `ValidationWebhook`s.~
 ~We could allow setting the `failurePolicy` of the validating webhooks to resolve the issue.~

programmer04:

Generate `config/default/validating_webhook/manifests.yaml` based on kubebuilder makers with a dedicated `manifests.validating-webhook` target for cert-manager etc. introduce a patch.

This also moves the validation of service annotations to `validatingadmissionpolicy` to prevent block of updating services when KO crashes.

**Which issue this PR fixes**

Closes #2186 
Closes #2544 

The format of the webhook definition in `config/default/validating_webhook/manifests.yaml` in the object

```
apiVersion: admissionregistration.k8s.io/v1
kind: ValidatingWebhookConfiguration
metadata:
  name: controlplane-configuration-validations
```

changed a little bit (you can see manual adjustment in 5430e95a68aa247639171f0048d847746dc73976 to be able to compare it easily with a generated one). There is a pattern of separating concerns. Furthermore, it's the only way supported by kubebuilder markers. From a functional pov, it's the same. It's been tested manually for upgrade, etc., and it works smoothly.


charts/kong-operator/templates/validation-policy-service-annotations.yaml
programmer04, line 1: The respective file is missing in `config/default/validating_policies` as we treat this dir as a source of truth. Kustomize is run for it, and some adjustment made with the script `hack/generators/validating-policy/main.go` and as a result, we have `charts/kong-operator/templates/validation-policy-dataplane.yaml`

randmonkey: So looks like the generating scripts should also be updated to supprt multiple files. It looks weird if the content is under the file `charts/kong-operator/templates/validation-policy-dataplane.yaml`.
 Which solution do you think better to improve the scripts in `hack`?
- A 1-1 file mapping from the `config/default/validating_policies` to the template files in `charts`
- Create a folder for each CRD that has a `validatingadmissionpolicy`attached and create a template for each CRD
- Rename the template file name in charts from `validation-policy-dataplane.yaml` to `validation-policy.yaml` and still merge all of the files into one (Not recommended because it cannot control enable/disable of each validation admission policy)

programmer04: 🤔 it depends on the level of granularity we want for enabling/disabling validation. In case the whole is enabled or disabled, renaming to `validation-policy.yaml` and having everything from `config/default/validating_policies` squashed to one file in the Chart, it's not such a bad idea. WDYT?

randmonkey: From the current file and values in charts, I think we should keep the fine granuale of enabling validations. So I would like to choose (2) to keep the scope not changed.

randmonkey: @programmer04 Chose (2) and updated the generating script to support multiple directories/files in charts.

programmer04 (commented):
It looks good, I think the only missing pieceis to cover it with a test in `test/crdsvalidation`, but you can handle it in a separate PR

randmonkey (approved):

alacuku (approved):
/lgtm

Issue #1614, opened 2025-05-15
Add tests to ensure DPs configuration does not get affected by Konnect interruption of service


mlavacca:
## Problem Statement

We need to improve our test suite to ensure that in case the Konnect platform goes down, no dataplane configuration is affected.

## Proposed Solution

- create a `DataPlane`
- create a set of service-routes-EE plugins to properly configure the dataplane with some EE-only features
- test the configuration and check it works as expected
- start to drop all the traffic from the CP to Konnect (we could use calico `GlobalNetworkPolicies`)
- test the configuration and check it works as expected
- rollout restart the CP deployment
- wait for the new CP pod to get ready
- Test the configuration and check it works as expected 

## Additional Information

These test improvements should be addressed once https://github.com/Kong/kubernetes-ingress-controller/issues/7421 and https://github.com/Kong/kubernetes-ingress-controller/issues/7422 get completed

## Acceptance Criteria
- [ ] a new test checks the in case Konnect goes down, no dataplane configuration gets affected

sandromodarelli:
punting to KO 2.1.0 to handle priorities

sandromodarelli:
unblocked since https://github.com/Kong/kubernetes-ingress-controller/issues/7421 and https://github.com/Kong/kubernetes-ingress-controller/issues/7422 have been done

sandromodarelli:
punting due to priorities

PR #3463, opened 2026-03-02
fix(hybridgateway): dataplane deletion in case of konnect connectivity issues

sandromodarelli:

This PR fixes a critical bug where transient Konnect API connectivity issues were causing the Gateway controller to permanently delete all DataPlane pods.

Root cause: The provisionKonnectExtension function in the Gateway controller treated all cases where `KonnectExtension.ControlPlaneRefValid=False` as permanent failures and deleted every DataPlane associated with the Gateway. This triggered a Kubernetes garbage collection cascade: DataPlane → Deployment → ReplicaSet → Pods destroyed.

However, `ControlPlaneRefValid=False` can be caused by transient conditions (when the Konnect API returns 5xx errors, timeouts, or connection resets), the KonnectGatewayControlPlane loses its `Programmed=True` condition, which propagates to the extension and then incorrectly triggers DataPlane deletion. Since the DataPlane controller handles missing extensions gracefully (by setting KonnectExtensionApplied=False while continuing to serve traffic), there is no reason to ever delete DataPlanes.

Changes:
 - New `ControlPlaneRefReasonNotProgrammed` condition reason: differentiates transient failures ("CP exists but not yet programmed in Konnect") from permanent failures ("CP not found / invalid ref"). Previously both used the same Invalid reason, making it impossible for downstream controllers to distinguish them.
- Updated reconciler_controlplaneref.go and konnectextension_controller_utils.go: The "CP not programmed yet" path now sets reason=NotProgrammed instead of reason=Invalid.
- Fixed `provisionKonnectExtension` in the Gateway controller: replaced the unconditional DataPlane-delete-then-extension-delete logic with a two-case check:
  - Transient failure (extension's CP ref matches the current KonnectGatewayControlPlane): just log and return, the KonnectExtension controller will re-validate on its own schedule.
  - Stale ref (extension's CP ref points to a different/old KGCP, e.g. after manual deletion+recreation): delete only the KonnectExtension for recreation (the `spec.konnect.controlPlane.ref` field is immutable per CEL validation, so recreation is the correct path).
  - DataPlanes are never deleted in either case.

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,12 @@
   [#3206](https://github.com/Kong/kong-operator/pull/3206)
 - Fix handling removal of annotations for DataPlane's Services
   [#3402](https://github.com/Kong/kong-operator/pull/3402)
+- Fix Gateway controller deleting all DataPlanes when KonnectExtension's
+  `ControlPlaneRefValid` condition is temporarily False due to transient Konnect
+  API failures. DataPlanes now continue serving traffic during Konnect
+  connectivity issues. Added `NotProgrammed` condition reason to differentiate
+  transient failures from permanent reference errors.
+  [#3463](https://github.com/Kong/kong-operator/pull/3463)
 
 ## [v2.1.1]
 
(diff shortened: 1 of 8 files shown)


programmer04:
Does it basically cover the issue https://github.com/Kong/kong-operator/issues/1614? 🤔

mlavacca:
> Does it basically cover the issue #1614? 🤔

#1614 is about adding a new integration/e2e test to simulate loss of connectivity. This PR addresses a specific behavior when connectivity is lost. #1614 still makes sense after this PR.

mlavacca (approved):

pmalek (approved):

PR #3470, opened 2026-03-02
chore: migrate to envtest `DataPlane` reconciler tests

### LINKED ISSUE #172: IMPROVE: Introduce `envtest` based tests for reconciler tests
### Problem statement

In several reconciler tests (e.g. `DataPlaneReconciler` tests: https://github.com/Kong/gateway-operator/blob/b6f20ea8180ccb5aaba7d708e96bc3c10315eb3b/controllers/dataplane_controller_test.go#L366-L370) we call their `.Reconcile()` method to trigger the reconciliation.

The problem with this approach is that with more steps in the `Reconcile()`'s logic, we need to call it more times to achieve a particular effect that is in specific order inside the function's logic: which is an implementation detail which no one should rely on.

### Proposed solution

This issue track the effort to rewrite those tests to something similar as its done for KIC in https://github.com/Kong/kubernetes-ingress-controller/tree/main/test/envtest

- https://github.com/Kong/kubernetes-ingress-controller/tree/main/test/envtest is placed in an open source repo to be used by all projects (KIC, KGO and more)
- `envtest` based tests are added in KGO

### Acceptance

- [ ] There is a framework/suite envtest setup that allows developers to add new envtests for their functionality.
- [ ] Existing reconciler unit tests that require repeatedly calling `Reconcile()` in the test body are replaced with envtests.
- [ ] The integration test split introduced in https://github.com/Kong/gateway-operator/pull/1503 is consolidated into a single (webhook on) test, with envtests or unit tests covering any webhook off behavior.

programmer04:

**Which issue this PR fixes**

Fixes #172


alacuku (approved):
/lgtm

tao12345666333 (approved):

PR #3491, opened 2026-03-04
chore(deps): bump sigs.k8s.io/gateway-api from 1.4.1 to 1.5.0 and make required adjustments

programmer04:

- bump of GWAPI to 1.5
- adjustment of conformance tests to 2 separate modules, as the GWAPI team made such a change
- usage of `gatewayv1` `ReferenceGrant` (and related resources) instead of `gatewayv1beta1` aliaset to `gwtypes`
- other minor fixes, linting, and regeneration of Charts, etc.


controller/specialized/aigateway_controller_watch.go
```diff
@@ -85,7 +85,7 @@ func (r *AIGatewayReconciler) listAIGatewaysForGatewayClass(ctx context.Context,
 // listAIGatewaysForReferenceGrants lists AIGateways whose group, kind and namespace appeared in `spec.from` of ReferenceGrants.
 // The listed AIGateways in are allowed to reference the resources in the `spec.to` of the ReferenceGrant.
 func (r *AIGatewayReconciler) listAIGatewaysForReferenceGrants(ctx context.Context, obj client.Object) []reconcile.Request {
-	referenceGrant, ok := obj.(*gatewayv1beta1.ReferenceGrant)
+	referenceGrant, ok := obj.(*gatewayv1.ReferenceGrant)
```
pmalek: Since we already have all the aliases defined in `internal/types/gatewaytypes.go` we could use this opportunity to move the explicit usage of v1/v1beta1 etc to those, WDYT?

programmer04: Good suggestion, it's done now

pmalek (approved):

PR #3493, opened 2026-03-05
configure dataplane options for TLS listeners

### LINKED ISSUE #64: Enable TLS listeners on `Gateway`s
## Problem Statement

KGO 1.2 allows HTTP and HTTPS listeners only. KIC already supports `TLSRoute`s, therefore we need to enable users to create `Gateway`s with `TLS` listeners with `Passthrough` TLS mode enabled.

## Proposed Solution

- In case a Gateway with TLS listener is created, it is accepted (only in case the TLS mode is `Passthrough`)
- Enable the `GatewayAlpha` FG in the `ControlPlane`
- The `DataPlane` gets the STREAM_LISTEN env var set accordingly
- The `DataPlane` Ingress service gets configured to expose the TLS port of the `DataPlane` 

## Acceptance Criteria
- [ ] As a user, I can create a `Gateway` with a TLS listener and use `TLSRoute`s to configure the `DataPlane`

randmonkey:

Configure `DeploymentOptions` and ingress service ports in `DataPlane` for `TLS` listeners.

It configures the `KONG_STREAM_LISTEN` to start L4 proxy ports on Kong DP and `KONG_PORT_MAPS` env in deployment options, and open a port on the ingress service for each TLS listeneer.

**Which issue this PR fixes**

Fixes #64 

**PR Readiness Checklist**:

### DIFF
--- a/controller/gateway/controller_reconciler_utils.go
+++ b/controller/gateway/controller_reconciler_utils.go
@@ -588,11 +588,12 @@ func generateDataPlaneNetworkPolicy(
 	podLabels map[string]string,
 ) (*networkingv1.NetworkPolicy, error) {
 	var (
-		protocolTCP     = corev1.ProtocolTCP
-		adminAPISSLPort = intstr.FromInt(consts.DataPlaneAdminAPIPort)
-		proxyPort       = intstr.FromInt(consts.DataPlaneProxyPort)
-		proxySSLPort    = intstr.FromInt(consts.DataPlaneProxySSLPort)
-		metricsPort     = intstr.FromInt(consts.DataPlaneMetricsPort)
+		protocolTCP       = corev1.ProtocolTCP
+		adminAPISSLPorts  = []intstr.IntOrString{intstr.FromInt(consts.DataPlaneAdminAPIPort)}
+		proxyPorts        = []intstr.IntOrString{intstr.FromInt(consts.DataPlaneProxyPort)}
+		proxySSLPorts     = []intstr.IntOrString{intstr.FromInt(consts.DataPlaneProxySSLPort)}
+		metricsPorts      = []intstr.IntOrString{intstr.FromInt(consts.DataPlaneMetricsPort)}
+		streamListenPorts = []intstr.IntOrString{}
 		// The label keys to match Kong operator pod.
 		// To not create new NetworkPolicy on upgrade of , we just keep the keys marking the application
 		// and remove the keys related to versions such as `version`,`pod-template-hash`,`helm.sh/chart`.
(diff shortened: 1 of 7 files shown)


controller/gateway/controller_reconciler_utils.go
```diff
@@ -1160,10 +1168,121 @@ func (g *gatewayConditionsAndListenersAwareT) setListenersStatus(status metav1.C
...
+		case gatewayv1.TLSProtocolType:
+			portNumber := int(l.Port)
+			// TODO: support multiple listeners using the same port:
+			// https://github.com/Kong/kong-operator/issues/3511
+			tlsPorts = append(tlsPorts, portNumber)
+			// Assign another port if the listener's port is already allocated on Kong DP.
+			// REVIEW: Also re-assign a port if known ports  (<1024) are used? Or always assign a random port?
+			if _, occupied := kongPortOccupied[portNumber]; occupied {
+				for {
+					assignedPortNumber := rand.Intn(1024) + 16384 //nolint:gosec // No Need to use the crypto safe algorithm to assign an available port, I think.
```
tao12345666333: I think we should use a deterministic method to generate the port, rather than generating a random port like this. Could generating a random port lead to some uncertain reconciliations?

controller/gateway/controller_reconciler_utils_test.go
```diff
@@ -580,7 +715,13 @@ func TestSetDataPlaneIngressServicePorts(t *testing.T) {
 					Protocol: gatewayv1.HTTPSProtocolType,
 					Port:     gatewayv1.PortNumber(443),
 				},
+				{
+					Name:     "tls",
+					Protocol: gatewayv1.HTTPSProtocolType,
```
tao12345666333:
```suggestion
					Protocol: gatewayv1.TLSProtocolType,
```

controller/gateway/controller_reconciler_utils.go
```diff
@@ -1160,10 +1168,121 @@ func (g *gatewayConditionsAndListenersAwareT) setListenersStatus(status metav1.C
...
+	if err != nil {
+		return err
+	}
+
+	return setDataPlaneIngressServicePorts(opts, listeners, listenersOpts, listenerPortToKongListenPort)
+}
+
+// setDataPlaneListenPorts configures deploymentOptions to set listen ports of the generated DataPlane.
+// It returns the map from the listener's port to the port.
+func setDataPlaneDeploymentListenPorts(
```
tao12345666333: The godoc says `setDataPlaneListenPorts` but the function is `setDataPlaneDeploymentListenPorts`

controller/gateway/controller_reconciler_utils.go
randmonkey: Updated to use a diterminisitc method to assign ports.

controller/gateway/controller_reconciler_utils_test.go
randmonkey: Done

tao12345666333 (commented):
I thought of a related point: Currently, we also need to open the corresponding protocol ports in our NetworkPolicy, right?

https://github.com/Kong/kong-operator/blob/8b8995b2001a7de15afb0f171a8f170236e3a4d6/controller/gateway/controller_reconciler_utils.go#L664-L669

controller/gateway/controller_reconciler_utils.go
randmonkey: Done.

randmonkey:
> I thought of a related point: Currently, we also need to open the corresponding protocol ports in our NetworkPolicy, right?
> 
> https://github.com/Kong/kong-operator/blob/8b8995b2001a7de15afb0f171a8f170236e3a4d6/controller/gateway/controller_reconciler_utils.go#L664-L669

Updated the function to generate `NetworkPolicy` from envs.
Also, the kong listen envs can have multiple SSL/non-SSL ports (See the example in [docs-proxy-listen](https://developer.konghq.com/gateway/configuration/#proxy-listen) ) so I also changed the `parseKongListenEnv`.
See: https://github.com/Kong/kong-operator/pull/3493/commits/1316b734b6b59c8a5e4f6d1f0f4552598f2df934

controller/gateway/controller_reconciler_utils.go
```diff
@@ -1160,10 +1195,137 @@ func (g *gatewayConditionsAndListenersAwareT) setListenersStatus(status metav1.C
...
+					if _, occupied := kongPortOccupied[assignedPortNumber]; !occupied {
+						listenerPortToKongListenPort[portNumber] = assignedPortNumber
+						kongPortOccupied[assignedPortNumber] = struct{}{}
+						break
+					}
+				}
+				// Although it should not happen where no ports can be assigned for the listener,
+				// we attach an error if the case really happens.
+				if assignedPortNumber >= assignedPortMax {
+					errs = errors.Join(errs, fmt.Errorf("listener %d's port %d already occupied and no available ports can be assinged", i, portNumber))
```
tao12345666333:
```suggestion
					errs = errors.Join(errs, fmt.Errorf("listener %d's port %d already occupied and no available ports can be assigned", i, portNumber))
```

controller/gateway/controller_reconciler_utils.go
```diff
@@ -653,25 +671,44 @@ func generateDataPlaneNetworkPolicy(
...
+	}
+
+	if len(streamListenPorts) > 0 {
+		allowStreamIngress := networkingv1.NetworkPolicyIngressRule{
+			Ports: lo.Map(streamListenPorts, func(port intstr.IntOrString, _ int) networkingv1.NetworkPolicyPort {
+				return networkingv1.NetworkPolicyPort{Protocol: &protocolTCP, Port: &port}
+			}),
+		}
+		ingressRules = append(ingressRules, allowStreamIngress)
 	}
```
tao12345666333, line 715: I think you can use the same approach here, checking for non-empty lists, when handling other NetworkPolicy rules (for example, adminAPISSLPorts), because in NetworkPolicy, passing an empty list means all traffic is allowed.

randmonkey: > I think you can use the same approach here, checking for non-empty lists, when handling other NetworkPolicy rules (for example, adminAPISSLPorts), because in NetworkPolicy, passing an empty list means all traffic is allowed.

If I add a rule with an empty port list, it means that the rule allows ALL ingress traffics. Thus the whole `NetworkPolicy` allows ALL ingress traffics. However, the other ports are guaranteed to be non-empty so it is not possible to add an allow-all policy.

If we want to change the existing code for making the code less error prone, we can add it in another PR since it is not so releavant.
See: https://github.com/kubernetes/api/blob/91061ea648b73ce8944e1c1ff2c1e99d6bacdb54/networking/v1/types.go#L112. I'll add comments here.

controller/gateway/controller_reconciler_utils.go
randmonkey: Done.

controller/gateway/controller_reconciler_utils.go
randmonkey: Added comments.

tao12345666333 (approved):

PR #3524, opened 2026-03-06
fix: fix clearing Routes' statuses

pmalek:

When the operator deploys a dedicated KIC instance per Gateway, each instance is scoped to a specific Gateway via the `GatewayNN` field. All these instances share the same controllerName (e.g. `konghq.com/kic-gateway-controller`).

The problem arises when a route is enqueued by a watch that isn't filtered by GatewayNN — for example, a `ReferenceGrant` create/update triggers `listHTTPRoutesForReferenceGrant`, which enqueues all HTTPRoutes in the matching namespace regardless of which gateway they belong to. A KIC instance that receives such an enqueue for a route it doesn't own will call getSupportedGatewayForRoute, which returns [ErrNoSupportedGateway](https://github.com/Kong/kong-operator/blob/5dce811c807ddb7a7da11ea65ee4e028ee2f094d/ingress-controller/internal/controllers/gateway/route_utils.go#L39) (because the route's parentRef doesn't match this instance's GatewayNN). It then calls `ensureGatewayReferenceStatusRemoved`, which previously removed **all** parent statuses matching controllerName — including statuses written by the correct KIC instance that actually manages that route.

Once the route's parent status is wiped, the correct KIC instance reconciles again but now finds no prior accepted status, and the route's ResolvedRefs condition is not properly populated. This prevents the cross-namespace backend reference (permitted by the `ReferenceGrant`) from being resolved, causing the `HTTPRouteReferenceGrant` conformance test to fail: requests never reach web-backend.

This PR fixes the issue by scoping both `ensureGatewayReferenceStatusRemoved` and `routeHasKongParentStatus` to the specific Gateway when GatewayNN is set. They now only consider/remove parent status entries whose parentRef matches the gateway this controller instance manages, leaving statuses owned by other KIC instances untouched.

**Which issue this PR fixes**

Fixes #

This should improve the stability of tests in CI.

**PR Readiness Checklist**:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,12 @@
   [#3490](https://github.com/Kong/kong-operator/pull/3490)
 - Fix reducing `Secret`s with in use finalizers.
   [#3506](https://github.com/Kong/kong-operator/pull/3506)
+- Fix `ensureGatewayReferenceStatusRemoved` and `routeHasKongParentStatus` not
+  scoping to the specific Gateway when `GatewayNN` is set.
+  This could cause one ingress-controller instance to erroneously remove route
+  parent statuses set by another instance managing a different Gateway,
+  breaking cross-namespace HTTPRoute backend references via `ReferenceGrant`.
+  [#3524](https://github.com/Kong/kong-operator/pull/3524)
 
 ## [v2.1.2]
 
--- a/ingress-controller/internal/controllers/gateway/grpcroute_controller.go
+++ b/ingress-controller/internal/controllers/gateway/grpcroute_controller.go
@@ -309,7 +309,7 @@ func (r *GRPCRouteReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 			// if there's no supported Gateway then this route could have been previously
 			// supported by this controller. As such we ensure that no supported Gateway
 			// references exist in the object status any longer.
-			if _, err := ensureGatewayReferenceStatusRemoved(ctx, r.Client, log, grpcroute); err != nil {
+			if _, err := ensureGatewayReferenceStatusRemoved(ctx, r.Client, log, grpcroute, r.GatewayNN); err != nil {
 				// some failure happened so we need to retry to avoid orphaned statuses
 				return ctrl.Result{}, err
 			}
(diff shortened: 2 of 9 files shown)


pmalek:
Flaky failures in hybrid conformance suite seem unrelated:

```
    http.go:285: 2026-03-06T15:30:26.398599012Z: Response expectation failed for request: {URL: {Scheme:http Opaque: User: Host:172.18.128.1 Path:/add Fragment: RawQuery: RawPath: RawFragment: ForceQuery:false OmitHost:false}, Host: , Protocol: HTTP, Method: GET, Headers: map[Some-Other-Header:[val] X-Echo-Set-Header:[]], UnfollowRedirect: false, ServerName: , ServerCertificate: <truncated>, ClientCertificate: <truncated>, ClientCertificateKey: <truncated>}  not ready yet: expected status code to be one of [200], got 404. CRes: &{404 103 HTTP/1.1 map[Content-Length:[103] Content-Type:[application/json; charset=utf-8] Date:[Fri, 06 Mar 2026 15:30:26 GMT] Server:[kong/3.13.0.1-enterprise-edition] X-Kong-Request-Id:[a0e8723d3cd19d7692257b3b2247e905] X-Kong-Response-Latency:[0]] <nil> []} (after 29.087191068s)
    http.go:266: 2026-03-06T15:30:27.310252369Z: timeout while waiting after 31 attempts, 0/3 successes
    http.go:266: 2026-03-06T15:30:27.31027538Z: timeout while waiting after 30 attempts, 0/3 successes
--- FAIL: TestGatewayConformance/HTTPRouteRequestHeaderModifier/3_request_to_'/add'_with_headers_should_go_to_infra-backend-v1 (30.00s)
--- FAIL: TestGatewayConformance/HTTPRouteRequestHeaderModifier/1_request_to_'/set'_with_headers_should_go_to_infra-backend-v1 (30.00s)
    http.go:266: 2026-03-06T15:30:27.310420014Z: timeout while waiting after 32 attempts, 0/3 successes
--- FAIL: TestGatewayConformance/HTTPRouteRequestHeaderModifier/2_request_to_'/add'_with_headers_should_go_to_infra-backend-v1 (30.00s)
    apply.go:283: 2026-03-06T15:30:27.310641945Z: Deleting request-header-modifier HTTPRoute
--- FAIL: TestGatewayConformance/HTTPRouteRequestHeaderModifier (0.32s)
```

programmer04 (approved):

PR #3526, opened 2026-03-06
fix(konnect): handle Programmed=False condition in case of APIAuthConfiguration missing reference

### LINKED ISSUE #3330: KonnectGatewayControlPlane missing Programmed=False condition when APIAuthResolvedRef is False
## Current Behavior

When a KonnectGatewayControlPlane cannot resolve its KonnectAPIAuthConfiguration reference (e.g., because the required KongReferenceGrant for a cross-namespace reference is missing), the APIAuthResolvedRef condition is correctly set to False with reason RefNotPermitted, but the Programmed condition is never updated to False.

Only APIAuthResolvedRef=False is set. The Programmed condition remains unchanged (either absent or stale from a previous reconciliation).

## Expected Behavior

The KonnectGatewayControlPlane should have both APIAuthResolvedRef=False and Programmed=False when the auth reference cannot be resolved.

## Steps To Reproduce

1. Create a namespace ns-auth for the KonnectAPIAuthConfiguration.
2. Create a Secret with the Konnect token in ns-auth.
3. Create a KonnectAPIAuthConfiguration in ns-auth referencing that secret.
4. Create a GatewayConfiguration in namespace ns-gateway with authRef pointing to the KonnectAPIAuthConfiguration in ns-auth.
5. Create a GatewayClass referencing the GatewayConfiguration.
6. Create a Gateway using that GatewayClass.
7. Do not create a KongReferenceGrant allowing the cross-namespace reference.
8. Observe the KonnectGatewayControlPlane status conditions:
    * APIAuthResolvedRef is False with reason RefNotPermitted (correct).
    * Programmed condition is missing or stale (incorrect).

sandromodarelli:

When a `KonnectGatewayControlPlane` cannot resolve its `KonnectAPIAuthConfiguration` reference (e.g. the auth config doesn't exist, or a cross-namespace reference is missing a `KongReferenceGrant`), the `APIAuthResolvedRef` condition was correctly set to `False` but the `Programmed` condition was never updated. This left `Programmed` absent or stale, making it impossible to determine the overall health of the resource from conditions alone.

Fix: pass a `Programmed=False` depending condition to both `handleAPIAuthStatusCondition()` call sites in reconciler_generic.go, so both conditions are patched atomically whenever auth ref resolution fails. This mirrors the pattern already used by the KonnectExtension controller.

**Which issue this PR fixes**

Fixes #3330

The fix is a one-liner per call site. Two envtests cover the two failure paths:

- `RefNotFound`: auth config does not exist (table-driven test in `konnect_entities_gatewaycontrolplane_test.go`)
- `RefNotPermitted`: cross-namespace ref without a `KongReferenceGrant` (standalone `TestKonnectGatewayControlPlane_CrossNamespaceRefNotPermitted`)

The standalone test uses mgr.GetClient() instead of the table suite's client.NewNamespacedClient because the cross-namespace grant check requires listing KongReferenceGrants in a different namespace.

**PR Readiness Checklist**:


pmalek (dismissed):
Looks good.

Linter complains about not using `new`:

```
--- a/test/envtest/konnect_entities_gatewaycontrolplane_test.go
+++ b/test/envtest/konnect_entities_gatewaycontrolplane_test.go
@@ -1006,7 +1006,7 @@ func TestKonnectGatewayControlPlane_CrossNamespaceRefNotPermitted(t *testing.T)
 			cp.Name = "cp-xns-no-grant"
 			cp.Namespace = cpNs.Name
 			cp.Spec.CreateControlPlaneRequest.Name = "cp-xns-no-grant"
-			cp.Spec.KonnectConfiguration.APIAuthConfigurationRef.Namespace = lo.ToPtr(authNs.Name)
+			cp.Spec.KonnectConfiguration.APIAuthConfigurationRef.Namespace = new(authNs.Name)
```

pmalek (approved):

PR #3532, opened 2026-03-08
hybridgateway: ordered state apply with Programmed=True dependency gating

tao12345666333:

Gate creation of dependent Kong resources until their prerequisites report Programmed=True, and split enforceState results to distinguish “applied changes” vs “waiting for deps”. This reduces transient Konnect errors and flapping in Hybrid Gateway reconciliations.

**Which issue this PR fixes**

split from https://github.com/Kong/kong-operator/pull/3522

**PR Readiness Checklist**:


controller/hybridgateway/reconciler_utils.go
```diff
@@ -184,11 +235,20 @@ func enforceState[t converter.RootObject](ctx context.Context, cl client.Client,
...
-	waitingForSkipped := objectsSkipped > 0
-	return stateChanged && waitingForSkipped, nil
+	// Report whether we applied anything or are waiting for prerequisites.
+	applied = (objectsCreated + objectsUpdated) > 0
+	waiting = objectsSkipped > 0
+	return applied, waiting, nil
+}
+
+// hasProgrammedTrue checks whether the Programmed condition is set to True.
+func hasProgrammedTrue(conds []metav1.Condition) bool {
```
randmonkey: I think we can use `k8sutils.HasConditionTrue` for the Konnet CRs because we have added `GetConditions()` for them.

controller/hybridgateway/controller.go
```diff
@@ -221,6 +222,12 @@ func (r *HybridGatewayReconciler[t, tPtr]) Reconcile(ctx context.Context, req ct
 		return ctrl.Result{Requeue: true}, nil
 	}
 
+	// If we are intentionally waiting for prerequisites (e.g., Programmed deps),
+	// schedule a short requeue as a safety net in case watch events are delayed.
+	if waiting {
+		return ctrl.Result{RequeueAfter: time.Second}, nil
```
randmonkey: Could we define the `time.Second` as a variable/constant?

controller/hybridgateway/reconciler_utils.go
```diff
@@ -82,13 +84,62 @@ func enforceState[t converter.RootObject](ctx context.Context, cl client.Client,
...
+					continue
+				}
+				if !hasProgrammedTrue(svc.Status.Conditions) {
+					log.Debug(logger, "Service not Programmed yet for route, waiting", "service", svcName)
+					objectsSkipped++
+					stopAtKind = "KongService"
+					continue
+				}
+			}
+		}
```
randmonkey, line 177: `KongPluginBinding` also has dependencies to other entities. Could we consider that?

controller/hybridgateway/reconciler_utils.go
```diff
@@ -57,18 +59,18 @@ func translate[t converter.RootObject](conv converter.APIConverter[t], ctx conte
 //
 // The function uses server-side apply with the "gateway-operator" field manager to ensure
 // proper ownership and conflict resolution when multiple controllers manage the same resources.
-func enforceState[t converter.RootObject](ctx context.Context, cl client.Client, logger logr.Logger, conv converter.APIConverter[t]) (bool, error) {
+func enforceState[t converter.RootObject](ctx context.Context, cl client.Client, logger logr.Logger, conv converter.APIConverter[t]) (applied bool, waiting bool, err error) {
```
randmonkey, line 63: I think there should be some unit tests for it. The logic of this function is quite complicated now.

controller/hybridgateway/reconciler_utils.go
```diff
@@ -82,13 +84,62 @@ func enforceState[t converter.RootObject](ctx context.Context, cl client.Client,
 	// In order to ensure proper ordering of resource creation/update, track the kind of the last created resource in the
 	// loop and skip further processing if we move to a different desired kind which may dependend on the just generated
 	// resource.
+	//
+	// Intentional: when dependency gating sets a specific kind in stopAtKind (e.g., waiting for
+	// KongService/KongUpstream to become Programmed), we conservatively skip all other kinds until
+	// that prerequisite kind is enforced. This avoids out-of-order application that could cause Konnect
+	// rejects (e.g., creating Routes/Targets before Services/Upstreams are ready).
```
randmonkey, line 92: So we should stop the creation/update of `KongTarget`s and all the resources after `KongTarget`s in the object slice when there is at least one `KongUpstream` in the desired objects does not exist or is not programmed, right?

tao12345666333: Yes. Current logic sets `stopAtKind="KongUpstream"` when the referenced upstream is absent or not `Programmed`. Since outputStore order is Upstream → Service → Route  → Plugin/Binding → Target, hitting a gated Target means we skip all remaining items in this reconcile pass, which is intended to avoid out-of-order programming. I’ll add unit tests to make this behavior explicit.

controller/hybridgateway/reconciler_utils.go
tao12345666333: Added https://github.com/Kong/kong-operator/pull/3532/changes/a0c51859012fb4ff8c272bc1be4df509992fcc70

controller/hybridgateway/controller.go
tao12345666333: Added https://github.com/Kong/kong-operator/pull/3532/changes/a0c51859012fb4ff8c272bc1be4df509992fcc70

controller/hybridgateway/reconciler_utils.go
tao12345666333: updated https://github.com/Kong/kong-operator/pull/3532/changes/a0c51859012fb4ff8c272bc1be4df509992fcc70

controller/hybridgateway/reconciler_utils.go
tao12345666333: Added https://github.com/Kong/kong-operator/pull/3532/changes/a0c51859012fb4ff8c272bc1be4df509992fcc70

controller/hybridgateway/controller.go
```diff
@@ -164,8 +168,38 @@ func (r *HybridGatewayReconciler[t, tPtr]) Reconcile(ctx context.Context, req ct
 		fmt.Sprintf("Successfully translated to %d Kong resources", resourceCount),
 	)
 
-	// Phase 2: State Enforcement.
-	stateChanged, err := enforceState(ctx, r.Client, logger, conv)
+	// Phase 1.5: Status Update (Accepted/ResolvedRefs/Programmed)
```
randmonkey: Why are we calling it `Phase 1.5`?

controller/hybridgateway/controller.go
```diff
@@ -164,8 +168,38 @@ func (r *HybridGatewayReconciler[t, tPtr]) Reconcile(ctx context.Context, req ct
...
+	// If the converter supports status updates for the root object (e.g., HTTPRoute),
+	// build and persist status conditions before enforcing state. This ensures that
+	// conformance tests expecting immediate status (e.g., NoMatchingParent) observe
+	// the conditions even when no resources are applied.
+	if su, ok := any(conv).(interface {
+		UpdateRootObjectStatus(context.Context, logr.Logger) (bool, bool, error)
```
randmonkey, line 209: After #3559 is done I think we should remove the type check.

controller/hybridgateway/controller.go
```diff
@@ -164,8 +168,38 @@ func (r *HybridGatewayReconciler[t, tPtr]) Reconcile(ctx context.Context, req ct
...
+			// Record status update failure event and exit.
+			r.eventRecorder.Event(
+				obj,
+				corev1.EventTypeWarning,
+				eventconst.EventReasonStatusUpdateFailed,
+				fmt.Sprintf("Status update failed: %v", err),
+			)
+			return ctrl.Result{}, err
+		}
+		if updated {
```
randmonkey, line 222: When status is updated, another round of reconcliation is triggered. But it looks like the current reconcliation continues when the status is updated. Would there be race conditions and should we return here?

controller/hybridgateway/controller.go
tao12345666333: Yes.  I will do a rebase again.

controller/hybridgateway/controller.go
tao12345666333: fixed Phase 3

controller/hybridgateway/controller.go
tao12345666333: done

controller/hybridgateway/controller.go
tao12345666333: I don't think so, `Status().Update()` only modifies the status sub-resource and does not affect the data used by enforceState's `GetOutputStore()` (which was already computed during the translation phase).

Added some notes here: https://github.com/Kong/kong-operator/pull/3532/changes#diff-7ea94c019df5b482065f6b5f18b7d42475aa7fda1b75f15852244acb8529e8c3R223-R227

randmonkey (approved):

randmonkey (approved):

PR #3553, opened 2026-03-10
chore: source of truth for chart's `validation-policy-konnect.yaml` in `config/`

programmer04:

The directory `config/default/validating_policies` is the single source of truth for all validation policies. Those in `charts/kong-operator/templates` are generated based on them. The konnect releated validation hasn't been done this way (it was put directly into the chart), hence this PR fixes it. 

**Which issue this PR fixes**

Follow-up of #3417


pmalek (approved):
🚀

PR #3575, opened 2026-03-12
ci: build e2e image once and reuse across jobs

### LINKED ISSUE #2948: Optimize CI: Share docker image between e2e test jobs
## Problem Statement

Currently, both e2e-tests and e2e-tests-chainsaw jobs in .github/workflows/tests.yaml build the docker image independently. We could potentially optimize CI resource usage by building the image once and sharing it across jobs.

## Proposed Solution

                                                                                          
  1. Create a dedicated build-image job that builds and uploads the image as an artifact    
  2. Both e2e test jobs download the artifact instead of building

## Additional Information

https://github.com/Kong/kong-operator/pull/2882#discussion_r2634392040

tao12345666333:

*  I randomly checked the build image time for two jobs, and it usually takes about 2 minutes and 36 seconds. ref: https://github.com/Kong/kong-operator/actions/runs/23004855651/job/66798832472
* after this change: 
    * download image and load it into docker daemon costs 3s, ref: https://github.com/Kong/kong-operator/actions/runs/23007295269/job/66808471296?pr=3575#step:5:17 
    * upload image costs 4s, ref: https://github.com/Kong/kong-operator/actions/runs/23007295269/job/66807870229?pr=3575#step:8:17

Thanks to the optimization in this PR, the GitHub Actions runner runs 2 minutes less per CI. However, the overall CI duration improvement may not be obvious, because previously the two jobs were running in parallel, but now they have to run one after the other.

**Which issue this PR fixes**

Fixes #2948 

**PR Readiness Checklist**:


.github/workflows/tests.yaml
```diff
@@ -398,7 +432,7 @@ jobs:
 
   e2e-tests:
     runs-on: ubuntu-latest
-    needs: [check-docs-only]
+    needs: [check-docs-only, build-e2e-image]
```
pmalek: Can we make it consistent and use the following syntax:


```suggestion
    needs:
      - check-docs-only
      - build-e2e-image
```

pmalek (dismissed):
Minor nit, overall this looks good

.github/workflows/tests.yaml
```diff
@@ -42,6 +42,40 @@ env:
...
+      - name: checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: setup golang
+        uses: actions/setup-go@4b73464bb391d4059bd26b0524d20df3927bd417 # v6.3.0
+        with:
+          go-version-file: go.mod
+
+      - name: build docker image
+        run: make docker.build
```
pmalek, line 67: Would it make sense to reuse the docker layers cache here?

https://github.com/Kong/kong-operator/blob/628d366fbb451bc3051787e2a524390e914e45e2/.github/workflows/__build-workflow.yaml#L167-L173

tao12345666333: Good point! I think it makes sense as a follow-up, but I would keep it out of this PR. 

The goal here is to remove the duplicate `make docker.build` work across the E2E jobs by building the image once and reusing it as an artifact.

After rebasing on the current `main`, the change naturally fits into the reusable E2E workflows. Reusing Docker layer cache here would likely mean switching this path to the same buildx/cache setup as the image build workflow, which feels like a larger CI refactor than the scope of this PR.

WDYT? @pmalek

pmalek: I'm fine with us following up on this later on 👍

pmalek (approved):

PR #3587, opened 2026-03-13
fix(ingress-controller): Add protocols in translated rotues

randmonkey:

Set `protocols` to `http,https` in translated routes when expression router is enabled.

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -132,6 +132,9 @@
   [#3576](https://github.com/Kong/kong-operator/pull/3576)
 - Do not try to list `Gateway`s for namespaces that are not being watched by controller
   [#3625](https://github.com/Kong/kong-operator/pull/3625)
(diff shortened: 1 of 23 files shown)


pmalek (approved):
LGTM, this got me thinking that we're not testing against different versions of Kong.

We have a [matrix of k8s versions we test against](https://github.com/Kong/kong-operator/blob/62ba86e4eae125a8b7da1624e8226081f3e1154d/.github/supported_k8s_node_versions.yaml) but we don't have this for Kong.

We could consider doing this at the very least for `kongintegration` test by defining an array of versions in https://github.com/Kong/kong-operator/blob/92d5c3100616b873765303e05bcc7d53cd6fd52c/test/test_dependencies.yaml#L17-L21. This would require some work in the golden files and test suite of course but I believe it would pay itself of.

WDYT?

Issue #2203, opened 2025-09-04
E2E CLEANUP: Re-enable Helm upgrade tests after releasing a version which includes #2201


pmalek:
### Problem statement

https://github.com/Kong/kong-operator/pull/2201 changed validating webhook's paths both in the operator and in helm chart.

To test it e2e in helm upgrade tests and using a single version of the chart we need to have 2 versions (to upgrade between) that support this new path.

### Acceptance

- [ ] Helm upgrade tests are reenabled after an image is published with #2201 and used in these tests

sandromodarelli:
punting to KO 2.2 with its parent issue

PR #3606, opened 2026-03-16
chore(tests): `TestHelmUpgrade` from the latest release to the current one

programmer04:

This PR modifies and re-enables `TestHelmUpgrade` to be really useful. Since currently Chart is bumped for every release of KO, it means that this is the path to test and upgrade (the whole Chart, instead of changing the KO image version).

It tests now upgrade from the currently publicly available chart to the local one with an image built from the source.

Assertion allows to discover any restarts of `DataPlane`s and the steps required during upgrade, so they will need some attention from time to time. See the modification of `CHANGELOG.md` to learn what it discovered for the upcoming KO 2.2.0 release. Probably reviewing this test should be added to the release checklist.

**Which issue this PR fixes**

Part of 

- #2203

In subsequent PR, enhance the test to not only assert on-prem Gateway, but also the hybrid one.


charts/kong-operator/UPGRADE.md
```diff
@@ -56,7 +56,7 @@ kustomize build github.com/Kong/kong-operator/config/crd/gateway-operator?ref=v2
 Upgrading [Gateway API][gwapi] to v1.2.0 requires running:
 
 ```sh
-kustomize build github.com/kubernetes-sigs/gateway-api/config/crd\?ref=v1.2.0 | kubectl apply -f -
+kustomize build github.com/kubernetes-sigs/gateway-api/config/crd\?ref=v1.5.1 | kubectl apply -f -
```
pmalek, line 59: Why update this? This is just an example isn't it? And if so the note on line 56 would need to be updated.

test/e2e/helm_install_upgrade_test.go
```diff
@@ -18,434 +20,288 @@ import (
...
-	const waitTime = 3 * time.Minute
+	ctx := t.Context()
 
-	var (
-		ctx      = t.Context()
-		chartDir = kcfg.ChartPath()
+	// This is the latest Chart available publicly (used by actual users) that we can upgrade from.
+	const (
+		lastReleasedChart        = "oci://docker.io/kong/kong-operator-chart"
+		lastReleasedChartVersion = "1.2.1" // renovate: datasource=docker depName=kong/kong-operator-chart versioning=docker
```
pmalek, line 42: I believe renovate config will have to be updated to take into account this file

pmalek (commented):
From this proposal I understand we'd ditch the idea of testing upgrade paths between different versions of operator apart from the one from last chart version to current right?

I'm concerned we'd loos some testing surface this way by testing only this way.

charts/kong-operator/UPGRADE.md
programmer04: I'm updating it because I'm referencing this file explicitly in `CHANGELOG.md:88` to make it as smooth for a user as possible. Line 56 will be updated too

test/e2e/helm_install_upgrade_test.go
programmer04: done

programmer04:
> From this proposal I understand we'd ditch the idea of testing upgrade paths between different versions of operator apart from the one from last chart version to current right?
> 
> I'm concerned we'd loos some testing surface this way by testing only this way.

Yes, and not only the operator now we're testing, but a whole Chart (which allows discovering, e.g., requirements for manual update of CRDs between versions that may be a friction for a user, and we haven't documented that need before this PR). This is, I think, the only recommended and supported by us way of upgrading since we're releasing new Chart for each KO release.

But your point is valid too. So, since those tests regarding upgrading were basically dead for a year (neither run nor maintained). Let's have this one, and I'd raise a topic on the next team-k8s sync about what has been introduced in this PR. And ask the team what we would ideally like to cover in this flavor of tests, and I'll create issues accordingly. WDYT?

test/e2e/environment.go
```diff
@@ -52,7 +53,6 @@ const (
 
 var (
 	existingCluster    = os.Getenv("KONG_TEST_CLUSTER")
-	imageOverride      = os.Getenv("KONG_TEST_KONG_OPERATOR_IMAGE_OVERRIDE")
```
pmalek, line 55: This is still specified in https://github.com/Kong/kong-operator/blob/5040435086436679d9cb483f52f0f9dd7ca1ca18/.github/workflows/__release-workflow.yaml#L247

pmalek:
> > From this proposal I understand we'd ditch the idea of testing upgrade paths between different versions of operator apart from the one from last chart version to current right?
> > I'm concerned we'd loos some testing surface this way by testing only this way.
> 
> Yes, and not only the operator now we're testing, but a whole Chart (which allows discovering, e.g., requirements for manual update of CRDs between versions that may be a friction for a user, and we haven't documented that need before this PR). This is, I think, the only recommended and supported by us way of upgrading since we're releasing new Chart for each KO release.
> 
> But your point is valid too. So, since those tests regarding upgrading were basically dead for a year (neither run nor maintained). Let's have this one, and I'd raise a topic on the next team-k8s sync about what has been introduced in this PR. And ask the team what we would ideally like to cover in this flavor of tests, and I'll create issues accordingly. WDYT?

Agreed, let's discuss the future of these tests. Ideally we'd make a release AI to update these.

pmalek (approved):

PR #3610, opened 2026-03-17
fix: fix operator restart sending empty config to `DataPlane` pods.

pmalek:

Fix operator restart (after or during no Konnect connectivity) causing `ControlPlane` to send an empty config to `DataPlane` pods, causing them to return 404 responses for several seconds until the K8s informer cache fully populates.

This PR addresses several distinct issues discovered during investigation:

### 1. Empty config push on restart (main fix)

**Root cause:** On the first synchronizer tick after operator restart, `Update()` calls
`TryFetchingValidConfigFromGateways` which successfully fetches the existing valid config
from the gateway (services, routes, upstreams), then proceeds unconditionally to
`BuildKongConfig()` which produces an empty KongState because the K8s informer cache
hasn't synced HTTPRoute objects yet. This empty config is pushed via `POST /config`,
wiping all routes and causing 404s for ~3 seconds until the next tick pushes the real
config.

**Fix:**

- Added `KongState.IsEmpty()` method that checks whether the state has any meaningful
  routing entities (services, upstreams, certificates, plugins, consumers, etc.).
  Licenses are deliberately excluded — a state with only a license is still "empty"
  from a routing perspective.
- Added a guard in `KongClient.Update()` between `BuildKongConfig()` and
  `sendOutToGatewayClients()`: if the new config is empty but a previously fetched
  valid config exists with real entities, the push is skipped. This is checked only once to allow sending empty configs afterwards.

### 2. Double base64 decode bug in license storage

**Root cause:** `Store()` uses `StringData` (Kubernetes auto-base64-encodes for storage),
but `Load()` reads from `secret.Data` (Kubernetes client auto-decodes) and then tries
`base64.StdEncoding.DecodeString()` again, causing `"illegal base64 data at input byte 0"`.
The existing test masked this by manually pre-encoding `Data` values.

**Fix:** Removed the redundant `base64.StdEncoding.DecodeString()` calls in `Load()`.
Fixed the test to use raw (non-encoded) values in `Data`, matching real Kubernetes
behavior. Added comprehensive `storage_internal_test.go` covering edge cases.

### 3. Hardcoded Konnect config in `ControlPlane`'s config

**Root cause:** `kicInKonnectDefaults()` in `controlplane.go` hardcoded its own polling
periods instead of using the shared defaults:

| Setting | Was (hardcoded) | Now (from defaults) |
|---|---|---|
| `UploadConfigPeriod` | 10s | 30s (`DefaultKonnectConfigUploadPeriod`) |
| `InitialLicensePollingPeriod` | 10s | 1m (`license.DefaultInitialPollingPeriod`) |
| `LicensePollingPeriod` | 10s | 12h (`license.DefaultPollingPeriod`) |
| `RefreshNodePeriod` | *(was missing)* | 60s (`DefaultKonnectNodeRefreshPeriod`) |

**Fix:** Extracted `DefaultKonnectConfig()` constructor in the config package and use it
in both `NewConfig()` and `kicInKonnectDefaults()`, ensuring a single source of truth
for all default values. Moved `DefaultRefreshNodePeriod` from the `konnect` package to
`config.DefaultKonnectNodeRefreshPeriod` for consistency.

### 4. Adding missing `Secret` label selector `SecretLicenseStore`

`SecretLicenseStore` now accepts a `secretLabelSelector` parameter, allowing the
operator to set custom labels on the license `Secret`.
The `Secret` label selector has been set to non empty value in KO 2.0 but `SecretLicenseStore` has not accounted for that change. This resulted in it not seeing the license `Secret`s.

This prevented it from creating or getting the license `Secret` from the cluster (when license storage was enabled).

License storage `Store()` handles `AlreadyExists` on create by deleting and
recreating the Secret (to enforce correct label selectors on pre-existing Secrets).

### 5. License storage was not enabled by default for `ControlPlane`s

Default config for `ControlPlane` [did include license storage set to `true`](https://github.com/Kong/kong-operator/blob/32fde06a1e38e90f7ba13d82b4fa72178da3cb9c/ingress-controller/pkg/manager/config/config.go#L147) but options for this config (based on `ControlPlane` spec) constructed with `constructControlPlaneManagerConfigOptions` where completely overriding it in https://github.com/Kong/kong-operator/blob/5dce811c807ddb7a7da11ea65ee4e028ee2f094d/controller/controlplane/manager_options.go#L565-L567 from options set by the extension processor (which did not include the license storage option enabled): https://github.com/Kong/kong-operator/blob/5dce811c807ddb7a7da11ea65ee4e028ee2f094d/controller/pkg/extensions/konnect/controlplane.go#L84-L112

Extracting the defaults `DefaultKonnectConfig` (as mentioned above) fixes that issue.

**Which issue this PR fixes**

Fixes #

FTI-7364.

**PR Readiness Checklist**:

### DIFF
--- a/ingress-controller/internal/dataplane/kong_client.go
+++ b/ingress-controller/internal/dataplane/kong_client.go
@@ -523,23 +523,24 @@ func (c *KongClient) Update(ctx context.Context) error {
 	parsingResult := c.kongConfigBuilder.BuildKongConfig()
 	translationDuration := time.Since(translationStart)
 
-	if failuresCount := len(parsingResult.TranslationFailures); failuresCount > 0 {
-		c.metricsRecorder.RecordTranslationFailure(translationDuration)
-		c.metricsRecorder.RecordTranslationBrokenResources(failuresCount)
-		c.recordResourceFailureEvents(parsingResult.TranslationFailures, KongConfigurationTranslationFailedEventReason)
-		c.logger.V(logging.DebugLevel).Info("Translation failures occurred when building data-plane configuration", "count", failuresCount)
+	kongState := parsingResult.KongState
+	if newKongState, skip := c.shouldSkipInitialEmptyConfigPush(parsingResult.KongState); skip {
+		kongState = newKongState
 	} else {
-		c.metricsRecorder.RecordTranslationSuccess(translationDuration)
-		c.metricsRecorder.RecordTranslationBrokenResources(0)
-		c.logger.V(logging.DebugLevel).Info("Successfully built data-plane configuration", "duration", translationDuration.String())
-	}
-
-	if c.shouldSkipInitialEmptyConfigPush(parsingResult.KongState) {
-		return nil
+		if failuresCount := len(parsingResult.TranslationFailures); failuresCount > 0 {
+			c.metricsRecorder.RecordTranslationFailure(translationDuration)
+			c.metricsRecorder.RecordTranslationBrokenResources(failuresCount)
+			c.recordResourceFailureEvents(parsingResult.TranslationFailures, KongConfigurationTranslationFailedEventReason)
+			c.logger.V(logging.DebugLevel).Info("Translation failures occurred when building data-plane configuration", "count", failuresCount)
+		} else {
+			c.metricsRecorder.RecordTranslationSuccess(translationDuration)
+			c.metricsRecorder.RecordTranslationBrokenResources(0)
+			c.logger.V(logging.DebugLevel).Info("Successfully built data-plane configuration", "duration", translationDuration.String())
+		}
 	}
 
 	const isFallback = false
-	shas, gatewaysSyncErr := c.sendOutToGatewayClients(ctx, parsingResult.KongState, c.kongConfig, isFallback)
+	shas, gatewaysSyncErr := c.sendOutToGatewayClients(ctx, kongState, c.kongConfig, isFallback)
 
 	// Taking into account the results of syncing configuration with Gateways and potential translation
 	// failures, calculate the config status and update it.
@@ -563,7 +564,7 @@ func (c *KongClient) Update(ctx context.Context) error {
 	}
 
 	// Send configuration to Konnect only when successfully applied configuration to Kong Gateways run in cluster.
-	c.maybeUpdateKonnectKongState(parsingResult.KongState, isFallback)
+	c.maybeUpdateKonnectKongState(kongState, isFallback)
 	// Gateways were successfully synced with the current configuration, so we can update the last valid cache snapshot.
 	c.maybePreserveTheLastValidConfigCache(cacheSnapshot)
 
@@ -591,24 +592,33 @@ func (c *KongClient) maybePreserveTheLastValidConfigCache(lastValidCache store.C
 	}
 }
 
-func (c *KongClient) shouldSkipInitialEmptyConfigPush(s *kongstate.KongState) bool {
+// shouldSkipInitialEmptyConfigPush returns true if the client should skip pushing
+// an initial empty configuration to the gateways, which can happen during startup
+// when the informer cache has not yet been fully populated and the client has
+// not yet successfully pushed any configuration.
+// In this case, if there is a last valid configuration available,
+// we can skip pushing an empty configuration and instead use the last valid
+// configuration as the one to be pushed, which can help avoid unnecessary downtime
+// due to pushing an empty configuration.
+// Last valid configuration is returned when function returns true.
+func (c *KongClient) shouldSkipInitialEmptyConfigPush(s *kongstate.KongState) (*kongstate.KongState, bool) {
 	if c.hasSuccessfullyPushedConfig {
-		return false
+		return nil, false
 	}
 
 	if !s.IsEmpty() {
-		return false
+		return nil, false
 	}
 
 	lastValidConfig, ok := c.kongConfigFetcher.LastValidConfig()
 	if !ok || lastValidConfig == nil || lastValidConfig.IsEmpty() {
-		return false
+		return nil, false
 	}
 
 	c.logger.V(logging.DebugLevel).Info(
 		"Skipping empty config push before first successful update because a last valid config is already available",
 	)
-	return true
+	return lastValidConfig, true
 }
 
 // maybeTryRecoveringFromGatewaysSyncError tries to recover from a configuration rejection if the error is of the expected
--- a/ingress-controller/internal/dataplane/kong_client.go
+++ b/ingress-controller/internal/dataplane/kong_client.go
@@ -198,6 +198,14 @@ type KongClient struct {
 	// konnectKongStateUpdater is used to update the current state seen by Konnect that will be picked asynchronously
 	// by the Konnect config synchronization loop.
 	konnectKongStateUpdater KonnectKongStateUpdater
+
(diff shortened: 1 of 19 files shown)


pmalek:
Just checked that adding `mgr.GetCache().WaitForCacheSync(ctx)` call in `Synchronizer` doesn't solve this issue:

```diff
diff --git a/ingress-controller/internal/dataplane/synchronizer.go b/ingress-controller/internal/dataplane/synchronizer.go
index b78637a48..eb41f080a 100644
--- a/ingress-controller/internal/dataplane/synchronizer.go
+++ b/ingress-controller/internal/dataplane/synchronizer.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
 
 	dpconf "github.com/kong/kong-operator/v2/ingress-controller/internal/dataplane/config"
 )
@@ -39,6 +40,7 @@ const (
 // the data-plane at regular intervals.
 type Synchronizer struct {
 	logger logr.Logger
+	mgr    manager.Manager
 
 	// dataplane client to send updates to the Kong Admin API
 	dataplaneClient Client
@@ -74,9 +76,10 @@ func WithInitCacheSyncDuration(period time.Duration) SynchronizerOption {
 // stagger time for data-plane updates to occur. Note that this starts some
 // background goroutines and the caller is resonsible for marking the provided
 // [context.Context] as "Done()" to shut down the background routines.
-func NewSynchronizer(logger logr.Logger, client Client, opts ...SynchronizerOption) (*Synchronizer, error) {
+func NewSynchronizer(logger logr.Logger, mgr manager.Manager, client Client, opts ...SynchronizerOption) (*Synchronizer, error) {
 	synchronizer := &Synchronizer{
 		logger:          logger,
+		mgr:             mgr,
 		stagger:         DefaultSyncInterval,
 		initWaitPeriod:  DefaultCacheSyncWaitDuration,
 		dataplaneClient: client,
@@ -101,12 +104,22 @@ func NewSynchronizer(logger logr.Logger, client Client, opts ...SynchronizerOpti
 //
 // To stop the server, the provided context must be Done().
 func (p *Synchronizer) Start(ctx context.Context) error {
-	select {
-	// This is a temporary mitigation to allow some time for controllers to
-	// populate their dataplaneClient cache.
-	case <-time.After(p.initWaitPeriod):
-	case <-ctx.Done():
-		return fmt.Errorf("Synchronizer Start() interrupted: %w", ctx.Err())
+	now := time.Now()
+sync:
+	for {
+		if p.mgr.GetCache().WaitForCacheSync(ctx) {
+			p.logger.Info("Initial cache sync completed", "duration", time.Since(now))
+			break sync
+		}
+
+		select {
+		// This is a temporary mitigation to allow some time for controllers to
+		// populate their dataplaneClient cache.
+		case <-time.After(p.initWaitPeriod):
+			p.logger.Info("Initial cache sync not complete yet")
+		case <-ctx.Done():
+			return fmt.Errorf("Synchronizer Start() interrupted: %w", ctx.Err())
+		}
 	}
 
 	p.lock.Lock()
diff --git a/ingress-controller/internal/manager/setup.go b/ingress-controller/internal/manager/setup.go
index 47ad5ca1e..68ebb8f10 100644
--- a/ingress-controller/internal/manager/setup.go
+++ b/ingress-controller/internal/manager/setup.go
@@ -189,6 +189,7 @@ func setupDataplaneSynchronizer(
 
 	dataplaneSynchronizer, err := dataplane.NewSynchronizer(
 		logger.WithName("dataplane-synchronizer"),
+		mgr,
 		dataplaneClient,
 		dataplane.WithStagger(proxySyncInterval),
 		dataplane.WithInitCacheSyncDuration(initCacheSyncWait),
```

ingress-controller/internal/konnect/license/storage.go
```diff
@@ -38,59 +38,69 @@ type Storer interface {
...
+		namespace:           namespace,
+		controlPlaneID:      controlPlaneID,
+		secretLabelSelector: secretLabelSelector,
 	}
 }
 
 // Store stores license to the secret `konnect-license-<cpid>`.
 func (s *SecretLicenseStore) Store(ctx context.Context, l license.KonnectLicense) error {
-	secret := &corev1.Secret{}
+	var secret corev1.Secret
```
randmonkey, line 67: Since we are always using the pointer, why should we define a non-pointer variable here?

ingress-controller/internal/dataplane/kong_client.go
```diff
@@ -579,6 +592,35 @@ func (c *KongClient) maybePreserveTheLastValidConfigCache(lastValidCache store.C
...
+	if c.hasSuccessfullyPushedConfig {
+		return nil, false
+	}
+
+	if !s.IsEmpty() {
+		return nil, false
+	}
+
+	lastValidConfig, ok := c.kongConfigFetcher.LastValidConfig()
+	if !ok || lastValidConfig == nil || lastValidConfig.IsEmpty() {
```
randmonkey, line 617: Is there a case that the `lastValidConfig` is really empty? In that case we should be OK to push an empty initial config right?

ingress-controller/internal/dataplane/kong_client.go
```diff
@@ -515,19 +523,24 @@ func (c *KongClient) Update(ctx context.Context) error {
 	parsingResult := c.kongConfigBuilder.BuildKongConfig()
 	translationDuration := time.Since(translationStart)
 
-	if failuresCount := len(parsingResult.TranslationFailures); failuresCount > 0 {
-		c.metricsRecorder.RecordTranslationFailure(translationDuration)
-		c.metricsRecorder.RecordTranslationBrokenResources(failuresCount)
-		c.recordResourceFailureEvents(parsingResult.TranslationFailures, KongConfigurationTranslationFailedEventReason)
-		c.logger.V(logging.DebugLevel).Info("Translation failures occurred when building data-plane configuration", "count", failuresCount)
+	kongState := parsingResult.KongState
+	if newKongState, skip := c.shouldSkipInitialEmptyConfigPush(parsingResult.KongState); skip {
```
randmonkey, line 527: The `if/else` is not so readable here. How about adding some comments like this:
```
if lastValidState, skip := c.shouldSkipInitialEmptyConfigPush(parsingResult.KongState); skip {
    // Use last valid state if we should skip the initial empty config.
    kongState = lastValidState
} else {
   // Otherwise, the translated Kong state is used so we should record the statistics of the translation.
}
```

randmonkey (commented):
Mostly LGTM, some minor comments.

ingress-controller/internal/dataplane/kong_client.go
pmalek: Added in b355d053b2fa590b983a60350ca932fd7a364134

ingress-controller/internal/dataplane/kong_client.go
pmalek: If last valid config is empty here, we're already checked if `s` is empty.

This means that both are empty and we're safe to proceed and send an empty config just like you're proposing, right?

ingress-controller/internal/konnect/license/storage.go
pmalek: Help prevent allocations. The variable is not returned - only used in this function - so it's quite possible this will not escape to heap and hence save an allocation.

randmonkey (approved):

PR #3634, opened 2026-03-18
chore: unit test for `listManagedGatewaysInNamespace`

programmer04:

Follow up for

- #3625

Since the test is added post factum, you can check that indeed test case `namespace not in watch list returns empty` fails when [this change is commented out](https://github.com/Kong/kong-operator/pull/3625/changes#diff-378016cc7ca143f74739d26a6eaa7c106fdbb32d312636827e3a426cedc5a894R354-R358) or see the actual code coverage of branches with your editor tooling.


pmalek (approved):
Nice 👍

PR #3635, opened 2026-03-18
feat: specify http client options for Konnect

pmalek:

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:

### DIFF
--- a/modules/manager/httpclient_konnect.go
+++ b/modules/manager/httpclient_konnect.go
@@ -1,0 +1,33 @@
+package manager
+
+import (
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/hashicorp/go-cleanhttp"
+	"golang.org/x/net/http2"
+)
+
+func httpClientForKonnect(c *Config) (*http.Client, error) {
+	// Clone the default transport to inherit all Go defaults (TLS settings,
+	// timeouts, proxy support, etc.) and only override what we need.
+	transport := cleanhttp.DefaultPooledTransport()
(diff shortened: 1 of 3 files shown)


modules/manager/httpclient_konnect.go
```diff
@@ -0,0 +1,33 @@
...
+	"net/http"
+	"time"
+
+	"golang.org/x/net/http2"
+)
+
+func httpClientForKonnect(c *Config) (*http.Client, error) {
+	// Clone the default transport to inherit all Go defaults (TLS settings,
+	// timeouts, proxy support, etc.) and only override what we need.
+	transport := http.DefaultTransport.(*http.Transport).Clone()
```
programmer04: We have "github.com/hashicorp/go-cleanhttp" already used in some part of the codebase. But it can be done this way too

programmer04 (dismissed):

modules/manager/httpclient_konnect.go
pmalek: Actually it makes more sense to use:

https://github.com/hashicorp/go-cleanhttp/blob/v0.5.2/cleanhttp.go#L36

```
MaxIdleConnsPerHost:   runtime.GOMAXPROCS(0) + 1
```

I made this to use cleanhttp for pooled transport (since we're mostly concerned about a single (load balanced through DNS) host.

Done in 37cfe2b48abea90e0491e294b619aca1c4bce84b

PTAL

programmer04 (approved):

PR #3639, opened 2026-03-18
fix(hybrid): fix hybrid gateway route status

pmalek:

Hybrid route status was using a broad managed-by label selector when building `Programmed` conditions, so an `HTTPRoute` could pick up Kong resources that actually belonged to a different route in the same namespace. In the intermittent `HTTPRouteWeight` failure, that let stale/unprogrammed resources from a neighboring route leak into the weighted route’s status and incorrectly report `KongServiceProgrammed=False`.

This mostly works around the fact that [label on routes is set only with GVK information](https://github.com/Kong/kong-operator/blob/5dce811c807ddb7a7da11ea65ee4e028ee2f094d/controller/hybridgateway/metadata/label.go#L17-L36).

This change keeps the existing list-by-label behavior, but filters the listed resources with the hybrid-routes ownership annotation before evaluating their Programmed state. That makes status computation consistent with orphan cleanup and prevents cross-route status contamination. A regression test was added to verify that unrelated route-owned resources are ignored.

**Which issue this PR fixes**

Fixes #

An alternative would be to add namespace and name information to route labels

**PR Readiness Checklist**:


alacuku (approved):
Nice Catch!

PR #3738, opened 2026-03-31
fix(admission): validate KongPlugin against gateways that expose the plugin

### LINKED ISSUE #3737: Admission: KongPlugin validation should not depend on arbitrary gateway client when multiple gateways exist
## Problem statement

When the controller discovers **multiple Kong Gateway Admin API clients** (multiple DataPlanes / gateway instances), the validating admission webhook validates `KongPlugin` (and related) configuration against **whichever client ends up first** in the internal client list.

Today `DefaultAdminAPIServicesProvider` uses a single “designated” Admin API client derived from `GatewayClients()[0]`. The slice returned by `GatewayClients()` comes from map values, so **order is not stable** across calls. As a result, validation can hit a gateway that **does not include** the plugin (or schema) even though another discovered gateway would — leading to **false rejections** or **flaky** behavior in multi-gateway clusters with differing plugin bundles.

This is the same class of problem discussed upstream for KIC:

- https://github.com/Kong/kubernetes-ingress-controller/issues/3363

## Expected behavior

`KongPlugin` admission validation should remain **backward compatible** for single-gateway setups, but in multi-gateway scenarios it should **not** fail solely because the arbitrarily chosen gateway lacks the plugin/schema when **another** discovered gateway could validate it (or validation should follow a clearly defined, deterministic policy).

## Steps to reproduce (conceptual)

1. Run Kong Operator / embedded ingress controller with **two (or more)** Kong Gateway instances discovered (different plugin bundles; e.g. one minimal, one with extra bundled/custom plugins).
2. Apply a `KongPlugin` that is only valid / present on **one** of those gateways.
3. Observe admission **reject** the resource depending on which Admin API client is selected first.

## Related

- Possibly related test / webhook area (different root cause, same component): https://github.com/Kong/kong-operator/issues/2867

## Suggested direction (optional)

- Probe plugin availability per gateway (e.g. schema endpoint) and validate only on compatible clients, with fallback to legacy single-client behavior when appropriate.

## Version

Observed on current `main` / embedded `ingress-controller` admission path (see `ingress-controller/internal/admission/`).

badam1:
# PR: Multi-gateway KongPlugin admission validation (#3737)

## What this PR does

When several Kong Gateway Admin API clients are discovered, `KongPlugin` admission used to validate through a single chosen client (`GatewayClients()[0]` via `DefaultAdminAPIServicesProvider`). That order is not stable, so validation could hit a gateway that does not expose a plugin/schema even when another discovered gateway does—causing false rejections or flaky behavior in multi-gateway setups with different plugin bundles.

This PR makes `KongPlugin` admission validation **gateway-aware**:

- Extends `DefaultAdminAPIServicesProvider` with an optional **`GetPluginsServices()`** path via **`MultiGatewayAdminAPIServicesProvider`**.
- Probes **`GetFullSchema(pluginName)`** on each discovered gateway.
- Validates only against gateways that expose the plugin schema.
- **Accepts** the resource if **any** matching gateway validates the plugin successfully.
- **Preserves** the legacy single-client path when **no** gateway exposes the plugin schema.

**Schema probing is order-independent:**

- **404 Not Found** → plugin not available on that gateway (skip).
- **Non-404** probe errors are recorded but **do not** abort early if another gateway exposes the schema.
- If **no** gateway exposes the schema: return the **first** non-404 probe error when present; otherwise fall back to the legacy single-client validation.

**Fixes #3737**

Related upstream discussion: [kubernetes-ingress-controller#3363](https://github.com/Kong/kubernetes-ingress-controller/issues/3363)

### Trade-offs

1. **First successful validation wins** — If several gateways expose the plugin, admission succeeds as soon as one validates the plugin. This favors availability in mixed multi-gateway deployments.

2. **Extra Admin API work at admission** — Probing `GetFullSchema()` per gateway and validating the matching subset adds some overhead vs. a single client, bounded by the number of discovered gateways.

3. **Scope** — Only **`KongPlugin` admission validation** is changed. Other admission paths that still use the single designated client are unchanged.

### Changelog

Release notes are under **Unreleased → Fixes** in `CHANGELOG.md`.

## PR readiness

- [x] `CHANGELOG.md` updated for significant changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -147,10 +147,11 @@
   `http,https`.
   [#3587](https://github.com/Kong/kong-operator/pull/3587)
 - Fix `KongPlugin` admission validation when multiple Kong Gateway Admin API
-  clients are discovered: probe plugin schema per gateway and validate only on
-  gateways that expose the plugin, falling back to the previous single-client
-  behavior when none match. Avoids false rejections when plugin bundles differ
-  across gateways.
+  clients are discovered: probe plugin schema on every gateway (order-independent),
+  validate only on gateways that expose the plugin, and fall back to the previous
+  single-client behavior when none match. Partial probe failures on one gateway do
+  not reject admission if another gateway exposes the plugin. Avoids false rejections
+  when plugin bundles differ across gateways.
   [#3737](https://github.com/Kong/kong-operator/issues/3737)
 
 ## [v2.1.3]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -146,6 +146,12 @@
 - Fix the on-prem translator to set `protocols` in translated Kong routes to
   `http,https`.
   [#3587](https://github.com/Kong/kong-operator/pull/3587)
+- Fix `KongPlugin` admission validation when multiple Kong Gateway Admin API
+  clients are discovered: probe plugin schema per gateway and validate only on
+  gateways that expose the plugin, falling back to the previous single-client
+  behavior when none match. Avoids false rejections when plugin bundles differ
+  across gateways.
+  [#3737](https://github.com/Kong/kong-operator/issues/3737)
 
 ## [v2.1.3]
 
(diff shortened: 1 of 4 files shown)


programmer04:
Hey, thank @badam1  for your valuable contribution. It will be included in the next release.

This PR has been cherry-picked (with commit attribution left) to
- https://github.com/Kong/kong-operator/pull/3754

It's done this way due too limitation of GitHub and our internal security policies that don't allow running our CI on forked repos

PR #3745, opened 2026-03-31
feat: rotate certificates issued by KO on expiration

### LINKED ISSUE #199: Rotate mTLS certificates
### Problem Statement

Certificates expire. Certificates do not work well when expired. Using mTLS requires functionality to replace certificates that are expiring, and when there is a new CA certificate (which will itself expire and be replaced at some point).

After replacing an expired certificate, the operator must also ensure the applications that use the certificate start using the updated certificate.

### Proposed Solution

Nothing firm. The operator side (watching for certificates that need updating and updating them) should be straightforward if a bit tedious in a Secret reconciler. Getting them to Pods without downtime is more of an open question, but the likely strategies are to use some update method that allows updates in place (see additional info) or one that restarts Pods while maintaining system availability.

### Additional information

AFAIK neither the controller nor Kong can use an updated certificate without restarting:
- The controller [mTLS implementation](https://github.com/Kong/kubernetes-ingress-controller/pull/2148/files) builds its admin API client at start time and bakes it into the manager. It should be possible to rewrite this to support generating a new client for each request, which will then start using updated files naturally.
- Kong loads its certificates through NGINX directives. NGINX will not reload its certificates after starting unless you intentionally run `kong reload` or send a SIGHUP to the NGINX master (PID 1). It should be possible to do this, but would require the operator enumerate data plane Pods, exec into each, and send a SIGHUP. This is a bit unorthodox.

### Acceptance Criteria

- [ ] https://github.com/Kong/gateway-operator/issues/185
- [ ] Operator-issued certificates have a short lifetime, for an arbitrary value of "short" (1 year?).
- [ ] When an mTLS certificate is about to expire, the operator issues a new certificate and updates its Secret.
- [ ] After the CA certificate changes, the operator issues new certificates and updates all mTLS certificate Secrets.
- [ ] The operator checks managed certificates at startup. If they are either expired or issued by a CA other than the current CA, the operator replaces them.
- [ ] After updating a Secret with a new certificate, the operator takes some action to make managed Pods use the new certificate in the updated Secret. When doing so, the operator uses some strategy to avoid downtime, e.g. by leveraging a PodDisruptionBudget or by updating Pods without restarting them and without otherwise affecting their operation.
- [ ] _Optional_ If the CA is expired, the operator raises an event or similar to indicate the problem.

programmer04:

Preserve old defaults regarding the validity of certificates issued by KO for internal purposes, and introduce a mechanism to rotate them near expiration that leverages the graceful recreation of a missing certificate secret (a new one is issued).

**Which issue this PR fixes**

Fixes #199, #1666

**PR Readiness Checklist**:

### DIFF
--- a/controller/secret_cert/controller.go
+++ b/controller/secret_cert/controller.go
@@ -1,0 +1,114 @@
+package secretcert
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	"github.com/kong/kong-operator/v2/controller/pkg/finalizer"
+	"github.com/kong/kong-operator/v2/controller/pkg/log"
+	"github.com/kong/kong-operator/v2/modules/manager/config"
+	"github.com/kong/kong-operator/v2/modules/manager/logging"
+	"github.com/kong/kong-operator/v2/pkg/consts"
+)
+
+// Reconciler reconciles TLS Secrets managed by DataPlane or ControlPlane.
+// Certs in these Secrets expires after a certain time, and the controller
+// is responsible for renewing them by deleting the expiring Secret,
+// which will trigger the creation of a new Secret with renewed certs by the
+// respective owner controllers (CP/DP).
+type Reconciler struct {
+	client.Client
+
+	ControllerOptions    controller.Options
+	LoggingMode          logging.Mode
+	CertExpirationMargin time.Duration
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *Reconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		WithOptions(r.ControllerOptions).
+		For(&corev1.Secret{},
+			builder.WithPredicates(
+				predicate.NewPredicateFuncs(secretMatchesFilter),
+			),
+		).
+		Complete(r)
+}
+
+// secretMatchesFilter returns true if the Secret has the required labels and type.
+func secretMatchesFilter(obj client.Object) bool {
+	labels := obj.GetLabels()
+	if labels[config.DefaultSecretLabelSelector] != config.LabelValueForSelectorTrue {
+		return false
+	}
+	if managedBy := labels[consts.GatewayOperatorManagedByLabel]; managedBy != consts.DataPlaneManagedLabelValue && managedBy != consts.ControlPlaneManagedLabelValue {
+		return false
+	}
+
+	secret, ok := obj.(*corev1.Secret)
+	if !ok {
+		return false
+	}
+
+	return secret.Type == corev1.SecretTypeTLS
+}
+
+// Reconcile handles certificate Secret reconciliation.
+func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	var cert corev1.Secret
+	if err := r.Get(ctx, req.NamespacedName, &cert); err != nil {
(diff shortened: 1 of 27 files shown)


controller/secret_cert/controller.go
```diff
@@ -0,0 +1,100 @@
...
+	// Remove the wait-for-owner finalizer if present so that the secret can be deleted.
+	if controllerutil.RemoveFinalizer(&cert, consts.DataPlaneOwnedWaitForOwnerFinalizer) {
+		if err := r.Update(ctx, &cert); err != nil {
+			return finalizer.HandlePatchOrUpdateError(err, logger)
+		}
+		log.Debug(logger, "removed wait-for-owner finalizer from secret")
+	}
+
+	if err := r.Delete(ctx, &cert); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to delete secret %s/%s: %w", cert.Namespace, cert.Name, err)
```
pmalek, line 109: Won't this cause any disruption for users?

AFAIK this will only cover the CP/DP mtls certs correct? I believe that it's worth adding a note about this somewhere otherwise this delete can raise some eyebrows like it did for me.

pmalek (commented):
Looks good 👍 

Left just one minor comment to clarify the purpose of the new controller.

controller/secret_cert/controller.go
programmer04: Added comment here and at the top. I've dested it, and disruption is negligible

pmalek: Couldn't we mark the `Secret` with an annotation and update it in place when it's marked this way rather than delete it? I'm worried about the downtime here.

programmer04: This way, I have now it's simpler, I rely on the logic which is already there, we do not touch with two controllers the same resource, I mean modifying. During testing, I haven't basicallty got a downtime, it was smooth. But I had a very small cluster...

But the most important thing is that the modification in place is not enough, see
> Kong loads its certificates through NGINX directives. NGINX will not reload its certificates after starting unless you intentionally run kong reload or send a SIGHUP to the NGINX master (PID 1). It should be possible to do this, but would require the operator enumerate data plane Pods, exec into each, and send a SIGHUP. This is a bit unorthodox.

from #199

pmalek: I understand the appeal of simplicity here and I agree with it 👍 I'm only concerned about users experience with the possible downtime.

Let's make sure we document this explicitly that downtime is expected and should be minimal between the removal of the Secret and creating a new one.

Input from @justindavies would also be beneficial here.

programmer04: Sure, it can be documented. Let's wait for input too. Do you have an idea for a place?

Note that before the changes in this PR, cert expiration breaks everything and requires manual intervention. The behavior is nowhere documented. Also, the hot reload is basically impossible to do in a sane way for Kong

pmalek: ACK.

I believe changelog for start but also perhaps a very short docs page about this. That page could then be expanded to cover cert-manager integration if/when that comes.

programmer04: Mentioned in `CHANGELOG.md`. AFAIK, for the next release cycle, probably cert-manager integration will be on our plate

justindavies: I'm ok with this, as discussed with @programmer04 we need to be clear how the user can check for the expiry, and what will happen during rotation

programmer04: A dedicated issue for @justindavies https://github.com/Kong/kong-operator/issues/3877 about docs

controller/secret_cert/controller.go
```diff
@@ -0,0 +1,105 @@
...
+
+	expiresAt, err := time.Parse(time.RFC3339, cert.Annotations[consts.CertExpiresAtAnnotation])
+	if err != nil {
+		log.Error(
+			logger,
+			err,
+			fmt.Sprintf("failed to parse %s annotation for secret %s/%s", consts.CertExpiresAtAnnotation, cert.Namespace, cert.Name),
+		)
+		return ctrl.Result{}, err
+	}
```
pmalek: If the `Secret` does not have the annotation, the parsing will fail with an error here right? (this can happen for instance before the full migration to the new setup which enforces the annotation)

Can't we check if the annotation is present and return without an error if it's absent? WDYT?

programmer04: Nice catch, I fixed it

pmalek (approved):

PR #3765, opened 2026-04-02
feat(eventgateway): add EventGateway DataPlane CRD

### LINKED ISSUE #3540: Add EventGatewayDataPlane CRD for deploying keg data planes
## Problem Statement

Add a new CRD EventGatewayDataPlane that manages the full lifecycle of Kong Event Gateway data plane (keg) deployments in Kubernetes. This is the Event Gateway equivalent of DataPlane + KonnectExtension for Kong Gateway.

The keg data plane requires careful orchestration to connect to Konnect:

* Bootstrap configuration with cluster ID and region from the control plane
* mTLS certificate provisioning and rotation for secure Konnect connectivity
* Proper networking setup for Kafka client access (LoadBalancer or TLSRoute)
* Runtime configuration (polling intervals, observability, graceful shutdown)

This CRD provides a single declarative resource that orchestrates all required components and automatically wires them together.

alacuku:

Introduces the `eventgateway.konghq.com/v1alpha1` API group with a new `DataPlane` kind that manages a keg binary Deployment connecting to Konnect via a referenced KonnectEventGateway resource.

**Which issue this PR fixes**

Fixes #3540

**PR Readiness Checklist**:


test/crdsvalidation/eventgateway.konghq.com/dataplane_test.go
```diff
@@ -0,0 +1,183 @@
...
+					Spec: eventgatewayv1alpha1.DataPlaneSpec{
+						KonnectEventGatewayRef: corev1.LocalObjectReference{Name: "my-event-gateway"},
+					},
+					// Non-zero status triggers the framework's status update, which causes
+					// the API server to apply the +kubebuilder:default on status.conditions.
+					Status: eventgatewayv1alpha1.DataPlaneStatus{
+						Replicas: 1,
+					},
+				},
+				// Update runs after the framework's status update+get, capturing t via closure.
```
pmalek:
```suggestion
				// Update runs after the framework's status update+get, capturing it via closure.
```

test/crdsvalidation/eventgateway.konghq.com/dataplane_test.go
```diff
@@ -0,0 +1,183 @@
...
+				TestObject: &eventgatewayv1alpha1.DataPlane{
+					ObjectMeta: common.CommonObjectMeta(ns.Name),
+					Spec: eventgatewayv1alpha1.DataPlaneSpec{
+						KonnectEventGatewayRef: corev1.LocalObjectReference{Name: "my-event-gateway"},
+					},
+					// Non-zero status triggers the framework's status update, which causes
+					// the API server to apply the +kubebuilder:default on status.conditions.
+					Status: eventgatewayv1alpha1.DataPlaneStatus{
+						Replicas: 1,
+					},
```
pmalek, line 372: 1. I'm actually quite surprised this works. Status won't (shouldn't?) get applied this way as [Create](https://github.com/Kong/kong-operator/blob/5dce811c807ddb7a7da11ea65ee4e028ee2f094d/test/crdsvalidation/common/testcase.go#L139-L140) won't fill in the status (usually).
2. We wouldn't want to use the `Update` here for assertion anyway. That's what `Assert` is for. https://github.com/Kong/kong-operator/blob/5dce811c807ddb7a7da11ea65ee4e028ee2f094d/test/crdsvalidation/common/testcase.go#L86-L88

In general I'd say we don't need to test the status defaults this way but if we really want I'd like to not hack this by setting something in the status. Is there a different way to achieve this? I generally thought that defaults would be applied regardless of status being non empty 🤔

api/eventgateway/v1alpha1/dataplane_types.go
```diff
@@ -0,0 +1,330 @@
...
+}
+
+// DataPlaneSpec defines the desired state of DataPlane.
+type DataPlaneSpec struct {
+	// KonnectEventGatewayRef references the KonnectEventGateway that this
+	// data plane connects to. The controller reads its status to obtain the
+	// Konnect gateway ID and region.
+	//
+	// +required
+	KonnectEventGatewayRef corev1.LocalObjectReference `json:"konnectEventGatewayRef"`
```
pmalek: This is more of a control plane rather than a gateway, right? Shouldn't we call it more like that?

api/eventgateway/v1alpha1/dataplane_types.go
```diff
@@ -0,0 +1,330 @@
...
+	// When a field is omitted the keg default applies.
+	//
+	// +optional
+	Config *Config `json:"config,omitempty"`
+}
+
+// DeploymentOptions specifies options for the Deployment managed
+// by the DataPlane controller.
+type DeploymentOptions struct {
+	operatorv1beta1.DeploymentOptions `json:",inline"`
```
pmalek: I'd suggest that we don't cross import the API groups, especially since the referred one is not "common".

A better course of action here would be to just copy paste this (only **if** we want to support all the spec fields that `DataPlane` from kong-operator API group supports).

---

Secondly, I can see strongly typed config in `Config` so if we want to allow users to specify both pods envs and strongly typed config I'd add comments about precedence and that specifying envs is for advanced use cases and is discouraged.

api/eventgateway/v1alpha1/dataplane_types.go
```diff
@@ -0,0 +1,330 @@
...
+	// For testing only, do not use in production.
+	// Corresponds to konnect.insecure_skip_verify / KONG_KONNECT_INSECURE_SKIP_VERIFY.
+	//
+	// +optional
+	InsecureSkipVerify *bool `json:"insecureSkipVerify,omitempty"`
+}
+
+// ObservabilityConfig configures logging, metrics, and tracing for KEG.
+type ObservabilityConfig struct {
+	// LogFlags sets the log level. Accepted values: trace, debug, info, warn, error.
```
pmalek: > Accepted values: trace, debug, info, warn, error.

Let's make sure to enforce this via kubebuilder enum annotation like we did it below for logformat.

pmalek (commented):
Looking good 👍 

Left minor comments.

Let's make sure to add a sample manifest showing exemplar usage of this CRD.

config/samples/eventgateway-dataplane.yaml
```diff
@@ -0,0 +1,16 @@
...
+  konnectEventGatewayRef:
+    name: my-event-gateway
+  network:
+    services:
+      kafka:
+        type: LoadBalancer
+        ports:
+          - name: kafka
+            port: 9092
+            targetPort: 9092
```
pmalek:
```suggestion
            targetPort: 9092

```

api/eventgateway/v1alpha1/dataplane_types.go
```diff
@@ -0,0 +1,330 @@
...
+// +genclient
+// +kubebuilder:object:root=true
+// +kubebuilder:object:generate=true
+// +kubebuilder:subresource:status
+// +kubebuilder:subresource:scale:specpath=.spec.deployment.replicas,statuspath=.status.replicas,selectorpath=.status.selector
+// +kubebuilder:resource:shortName=egdp,categories=kong,plural=dataplanes
+// +kubebuilder:printcolumn:name="Ready",description="The Resource is ready",type=string,JSONPath=`.status.conditions[?(@.type=='Ready')].status`
+// +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
+// +kong:channels=kong-operator
+type DataPlane struct {
```
mlavacca: I'd re-evaluate this name: I know it's a different group, but the collision with the `kong-operator.Dataplane` CRD will make the UX a little bit cumbersome in my opinion (`kubectl get dataplane` will be potentially confusing). Maybe `EventDataPlane` or something like that?

What do you think, does it make sense to you?

api/eventgateway/v1alpha1/dataplane_types.go
```diff
@@ -0,0 +1,330 @@
...
+	// +optional
+	Ports []ServicePort `json:"ports,omitempty"`
+}
+
+// ServicePort contains information on a service port.
+type ServicePort struct {
+	// The name of this port within the service.
+	//
+	// +optional
+	Name string `json:"name,omitempty"`
```
mlavacca:
```suggestion
	Name *string `json:"name,omitempty"`
```

optional field should be a pointer

test/crdsvalidation/eventgateway.konghq.com/dataplane_test.go
alacuku: Fixed it. The crdvalidation testcases already had the logic to update the status in order to get defaults. I just moved the `assert` function to run after the update logic.

api/eventgateway/v1alpha1/dataplane_types.go
```diff
@@ -0,0 +1,330 @@
...
+
+	// The port that will be exposed by this service.
+	Port int32 `json:"port"`
+
+	// Number or name of the port to access on the pods targeted by the service.
+	//
+	// +optional
+	TargetPort intstr.IntOrString `json:"targetPort,omitempty"`
```
mlavacca:
```suggestion
	TargetPort *intstr.IntOrString `json:"targetPort,omitempty"`
```

api/eventgateway/v1alpha1/dataplane_types.go
```diff
@@ -0,0 +1,330 @@
...
+
+	// The port on each node on which this service is exposed when type is
+	// NodePort or LoadBalancer.
+	//
+	// +optional
+	NodePort int32 `json:"nodePort,omitempty"`
```
mlavacca:
```suggestion
	NodePort *int32 `json:"nodePort,omitempty"`
```

api/eventgateway/v1alpha1/dataplane_types.go
```diff
@@ -0,0 +1,330 @@
...
+	// Network configures how the keg pod is exposed to Kafka clients.
+	//
+	// +optional
+	Network *NetworkOptions `json:"network,omitempty"`
+
+	// Config provides optional overrides for keg runtime settings.
+	// When a field is omitted the keg default applies.
+	//
+	// +optional
+	Config *Config `json:"config,omitempty"`
```
mlavacca: Can we use a more meaningful name here? `config` is pretty generic and low explanatory

api/eventgateway/v1alpha1/dataplane_types.go
alacuku: > Secondly, I can see strongly typed config in `Config` so if we want to allow users to specify both pods envs and strongly typed config I'd add comments about precedence and that specifying envs is for advanced use cases and is discouraged.

The `Config` section is needed to give users a single point of configuration of the dataplane. They can still override the configuration of the dataplane using the `env` variables in the `podTemplateSpec` but that is a collateral effect of exposing the `podTemplateSpec` in the API.

api/eventgateway/v1alpha1/dataplane_types.go
alacuku: Changed it to `controlPlaneRef`

api/eventgateway/v1alpha1/dataplane_types.go
```diff
@@ -0,0 +1,330 @@
...
+	// cert paths) are derived automatically and cannot be overridden here.
+	//
+	// +optional
+	Konnect *KonnectConfig `json:"konnect,omitempty"`
+
+	// ConfigPollInterval overrides how often keg polls Konnect for config changes.
+	// Corresponds to config_poll_interval / KEG__CONFIG_POLL_INTERVAL. Default: 5s.
+	//
+	// +optional
+	ConfigPollInterval *metav1.Duration `json:"configPollInterval,omitempty"`
```
mlavacca: Since Konnect gives a default, would it make sense to explicitly set it here as default as well? This way, in case Konnect defaults change in the future (it has already happened) we don't silently change the behavior.

api/eventgateway/v1alpha1/dataplane_types.go
```diff
@@ -0,0 +1,330 @@
...
+
+	// EnableDebugEndpoints enables the /debug/pprof/allocs endpoint.
+	// Corresponds to enable_debug_endpoints / KEG__ENABLE_DEBUG_ENDPOINTS. Default: false.
+	//
+	// +optional
+	EnableDebugEndpoints *bool `json:"enableDebugEndpoints,omitempty"`
```
mlavacca: Ditto about default

api/eventgateway/v1alpha1/dataplane_types.go
mlavacca: Also, if we decide to go with explicitly setting the defaults (see comments below), it makes sense to set defaults for the whole field.

api/eventgateway/v1alpha1/dataplane_types.go
```diff
@@ -0,0 +1,404 @@
...
+
+// DataPlaneSpec defines the desired state of DataPlane.
+type DataPlaneSpec struct {
+	// ControlPlaneRef references the KonnectEventGateway resource that acts as the
+	// control plane for this data plane. KonnectEventGateway is the Konnect-side
+	// resource that provisions the Event Gateway cluster. The controller reads its
+	// status to obtain the Konnect gateway ID and region.
+	//
+	// +required
+	ControlPlaneRef corev1.LocalObjectReference `json:"controlPlaneRef"`
```
pmalek: Can we add a union type field indicating the type of the reference? This way the API is easily extensible in the future if we ever need to add a different control plane reference type.

This would look similar to this in the manifest (I'm to comments on this):

```
controlPlaneRef:
    type: konnectNamespacedRef # naming is hard
    konnectNamespacedRef:
      name: my-event-gateway
      # namespace: ns
```

pmalek (commented):
I almost forgot: let's add this new API group to the paths checked by the kube api linter in https://github.com/Kong/kong-operator/blob/da077227d59658d63e3c372c835af4c508e893f1/Makefile#L1136-L1143

api/eventgateway/v1alpha1/dataplane_types.go
alacuku: I would keep it as it is until we have clear the path forward.

mlavacca (commented):
Looks great!

I have a broad comment, related to all the fields that have a default in Konnect: I'd consider explicitly setting the defaults in the CRD as well, instead of silently defaulting to Konnect. This way, if Konnect defaults change (they should not, but who knows) we don't silently change the behavior.

api/eventgateway/v1alpha1/dataplane_types.go
pmalek: If we ever need to change this:

```
controlPlaneRef:
  name: my-event-gateway
```

and add new types then we have to make a breaking change, add conversion webhooks etc. This is pretty high cost for us not making it a more structured field (which is essentially 0 cost now).

We can skip the `type` field now, and add it later when we add new types, this would not be breaking change.

So we can do something like:

```
controlPlaneRef:
  konnectNamespacedRef:
    name: my-event-gateway
    # namespace: ns
```

and have this be extensible in the future without breaking changes. WDYT?

pmalek:
> I'd consider explicitly setting the defaults in the CRD as well, instead of silently defaulting to Konnect. This way, if Konnect defaults change (they should not, but who knows) we don't silently change the behavior.

I was also thinking about this and originally wanted to make this comment but wasn't sure about it. I'm on board either way but setting the defaults in the CRD API but actually be the better approach here (it would ensure that a particular version of the API was the same expected outputs regardless of the external defaults).

BTW: it's KEG not Konnect defaults AFAIK

config/samples/eventgateway-dataplane.yaml
pmalek, line 1: General question: which image is going to run here? We don't seem to have any defaults defined at the CRD level. I assume that by allowing `podTemplateSpec` customization users can set the image that way but is this the way we want them to set the image?

api/eventgateway/v1alpha1/dataplane_types.go
alacuku: I think keeping DataPlane is the right call here. The two CRDs live in different API groups (eventgateway.konghq.com vs gateway-operator.konghq.com), so kubectl already handles disambiguation the standard Kubernetes way: `kubectl get dataplane.eventgateway.konghq.com`. Adding a prefix like Event would be redundant given the group name already carries that context.

api/eventgateway/v1alpha1/dataplane_types.go
alacuku: Any suggestion on the field name?

api/eventgateway/v1alpha1/dataplane_types.go
alacuku: Not sure I got your point. Having the `type` outside the `ControlPlaneRef` would not introduce a breaking change. And having the `ControlPlaneRef` a reference to a k8s object can accommodate any future `type`. For now it is just a reference to a localObjectreference, and for sure we are going to extend it to a name/namespaced reference. I don't see any advantage on having a `KonnectNamespacedRef` since it locks it to reference only objects related to `konnect`.

api/eventgateway/v1alpha1/dataplane_types.go
alacuku: I got your point, and currently defaults will be set by the operator instead of leaving it to be set by `konnect`. But still we can set defaults in the CRD.

config/samples/eventgateway-dataplane.yaml
alacuku: The default image will be set by the operator and yes the podTemplateSpec is the way to set the image.

alacuku:
> Looks great!
> 
> I have a broad comment, related to all the fields that have a default in Konnect: I'd consider explicitly setting the defaults in the CRD as well, instead of silently defaulting to Konnect. This way, if Konnect defaults change (they should not, but who knows) we don't silently change the behavior.

The configuration fields exposed in the CRD are not applied in Konnect but to the `keg binary`.  Anyway, I added defaults for those fields.

api/eventgateway/v1alpha1/dataplane_types.go
pmalek:
```
spec:
  type: <type>
  controlPlaneRef:
    name: my-event-gateway
```

This way the `type` is not bound to the `controlPlaneRef` in any way. There is no way for the user to know what is this `type` field. When it's done like this:

```
spec:
  controlPlaneRef:
    type: localReference
    localReference: # or a different name
      name: my-event-gateway
```

then you have the union type which indicates which "type" is "active".

---

> I don't see any advantage on having a `KonnectNamespacedRef` since it locks it to reference only objects related to konnect.

I suggested it this way because it indicates that the referred cp is a konnect cp (yes, the reference is local but the cp lives in Konnect). If we for example introduce locally deployed cp then we can change the type to be called `localNamespacedRef` (or anything else indicating that it lives locally in the cluster).

alacuku: I got your point now and it makes sense. Let me change it!

alacuku:
@pmalek @mlavacca , i addressed your comments. PTAL.

api/eventgateway/v1alpha1/dataplane_types.go
```diff
@@ -0,0 +1,467 @@
...
+
+// DataPlaneSpec defines the desired state of DataPlane.
+type DataPlaneSpec struct {
+	// ControlPlaneRef references the control plane this DataPlane connects to.
+	// The type field identifies which kind of control plane is being referenced.
+	// Currently only konnectNamespacedRef is supported, which references a
+	// KonnectEventGateway resource in the same namespace.
+	//
+	// +required
+	ControlPlaneRef ControlPlaneRef `json:"controlPlaneRef,omitempty"`
```
pmalek: Linter complains about this

api/eventgateway/v1alpha1/dataplane_types.go
```diff
@@ -0,0 +1,467 @@
...
+	// Corresponds to observability.metrics_rollup_allow_map /
+	// KEG__OBSERVABILITY__METRICS_ROLLUP_ALLOW_MAP.
+	//
+	// +optional
+	// +kubebuilder:default="messaging.operation.name=produce,fetch"
+	// +kubebuilder:validation:MaxLength=2048
+	MetricsRollupAllowMap *string `json:"metricsRollupAllowMap,omitempty"`
+
+	// PolicyErrorsInfoLogIntervalSeconds sets the interval for INFO-level logging of policy errors, in seconds.
+	// Set to 0 to disable. Default: 1.
```
pmalek:
```suggestion
	// Set to 0 to disable.
```

api/eventgateway/v1alpha1/dataplane_types.go
```diff
@@ -0,0 +1,467 @@
...
+	// Domain overrides the Konnect domain.
+	// Corresponds to konnect.domain / KONG_KONNECT_DOMAIN.
+	//
+	// +optional
+	// +kubebuilder:default="konghq.com"
+	// +kubebuilder:validation:MaxLength=253
+	Domain *string `json:"domain,omitempty"`
+
+	// APIRequestTimeoutSeconds overrides the Konnect API request timeout, in seconds.
+	// Corresponds to konnect.api_request_timeout / KONG_KONNECT_API_REQUEST_TIMEOUT. Default: 5.
```
pmalek:
```suggestion
	// Corresponds to konnect.api_request_timeout / KONG_KONNECT_API_REQUEST_TIMEOUT.
```

api/eventgateway/v1alpha1/dataplane_types.go
```diff
@@ -0,0 +1,467 @@
...
+type Config struct {
+	// Konnect provides optional overrides for the keg → Konnect connection
+	// parameters. All other connection values (region, gateway_cluster_id,
+	// cert paths) are derived automatically and cannot be overridden here.
+	//
+	// +optional
+	Konnect *KonnectConfig `json:"konnect,omitempty"`
+
+	// ConfigPollIntervalSeconds overrides how often keg polls Konnect for config changes, in seconds.
+	// Corresponds to config_poll_interval / KEG__CONFIG_POLL_INTERVAL. Default: 5.
```
pmalek:
```suggestion
	// Corresponds to config_poll_interval / KEG__CONFIG_POLL_INTERVAL.
```

api/eventgateway/v1alpha1/dataplane_types.go
```diff
@@ -0,0 +1,467 @@
...
+	// The port on each node on which this service is exposed when type is
+	// NodePort or LoadBalancer.
+	//
+	// +optional
+	NodePort *int32 `json:"nodePort,omitempty"`
+}
+
+// Config provides optional overrides for keg runtime settings.
+// All fields map 1-to-1 to keg configuration variables.
+type Config struct {
```
pmalek: This is good 👍 

The only comment I'd have here is about the naming: when we add any other root level types (CRD types) here, we couldn't now what this `Config` is for. Also using it in code doesn't imply what it's for:

```
import 	eventv1alpha1 "github.com/kong/kong-operator/v2/api/event-gateway/v1alpha1"
...

// How do we know that this is the config for Event Gateway DataPlane?
// What if we add a different type in eventv1alpha1?
var config eventv1alpha1.Config
```

I'd suggest to name it something along the lines of 


```suggestion
// DataPlaneConfiguration provides optional overrides for keg runtime settings.
// All fields map 1-to-1 to keg configuration variables.
type DataPlaneConfiguration struct {
```

api/eventgateway/v1alpha1/dataplane_types.go
pmalek: https://github.com/Kong/kong-operator/pull/3765/changes#r3032100000

test/crdsvalidation/eventgateway.konghq.com/dataplane_test.go
```diff
@@ -0,0 +1,493 @@
...
+							},
+						},
+					},
+				},
+				ExpectedErrorMessage: new("Unsupported value: \"verbose\""),
+			},
+		}.RunWithConfig(t, cfg, scheme)
+	})
+
+	t.Run("spec defaults applied when parent object is set", func(t *testing.T) {
```
pmalek: This is good and comprehensive 👍 

The only test case (I believe) is missing is one where `Config` is not set at all. We want to ensure that defaults are set when config is not provided at all.

So something like:

```
				TestObject: &eventgatewayv1alpha1.DataPlane{
					ObjectMeta: common.CommonObjectMeta(ns.Name),
					Spec: eventgatewayv1alpha1.DataPlaneSpec{
						ControlPlaneRef: eventgatewayv1alpha1.ControlPlaneRef{
							Type: eventgatewayv1alpha1.ControlPlaneRefTypeKonnectNamespacedRef,
							KonnectNamespacedRef: &eventgatewayv1alpha1.KonnectNamespacedRef{
								Name: "my-event-gateway",
							},
						},
					},
				},
```

alacuku: Yeah, that's how Kubernetes defaulting works. Defaults on fields are only applied when the parent object exists in the submitted manifest. Since config is a pointer (*Config), if you don't include it the API server sees null and never looks inside to apply the field defaults. Adding +kubebuilder:default={} on the field tells the API server to always initialize it to an empty object, which then triggers all the nested field defaults (enableDebugEndpoints: Disabled, configPollIntervalSeconds: 5, etc.) even when the user doesn't set config at all.
So the user experience would be: user submits without config, you get back the resource with config populated with all the defaults. It's still "optional" from the user's perspective, they just never have to think about it.
On the other hand even without defaults in the CR the operator still would apply defaults.

pmalek: > Adding +kubebuilder:default={} on the field tells the API server to always initialize it to an empty object, which then triggers all the nested field defaults (enableDebugEndpoints: Disabled, configPollIntervalSeconds: 5, etc.) even when the user doesn't set config at all.

Does it actually work this way? If so I believe we should do it. Otherwise, I believe there was a comment from @mlavacca which got lost that we should set the default annotation on the `Config` (or a renamed version of that field as discussed in other thread).

> On the other hand even without defaults in the CR the operator still would apply defaults.

I'd prefer that we rely on defaults in one place: if we decide that we're defining defaults in CRD, let's keep them here but not introduce a "fallback for defaults" in code. That can introduce inconsistencies in the future if we forget to update one or the other (and not have tests checking they are aligned).

alacuku:
Hey @pmalek and @mlavacca , after thinking about it more, I'd actually prefer not to add +kubebuilder:default={} on config and its sub-fields. Here's my reasoning:

The operator already knows the keg defaults and will apply them when constructing the Deployment regardless of what's in the spec. The CRD defaults would just mirror those values unnecessarily.

The downside of defaulting everything is that it clobbers the user's spec even when they never touched config. A user who submits the minimal manifest suddenly sees a fully-populated config block in their resource, with values they didn't set and may not expect. That makes the API noisier and harder to reason about, it becomes unclear what the user actually configured vs. what was silently injected.

The cleaner contract is: if config is nil, the operator uses its built-in defaults. If the user sets config, they're explicitly overriding specific fields. That way the spec reflects intent, not infrastructure defaults.

Happy to discuss further if there's a use case I'm missing!

test/crdsvalidation/eventgateway.konghq.com/dataplane_test.go
alacuku: Please see: https://github.com/Kong/kong-operator/pull/3765#issuecomment-4183060409

api/eventgateway/v1alpha1/dataplane_types.go
alacuku: See: https://github.com/Kong/kong-operator/pull/3765#issuecomment-4183060409

pmalek:
> Hey @pmalek and @mlavacca , after thinking about it more, I'd actually prefer not to add +kubebuilder:default={} on config and its sub-fields. Here's my reasoning:
> 
> The operator already knows the keg defaults and will apply them when constructing the Deployment regardless of what's in the spec. The CRD defaults would just mirror those values unnecessarily.
> 
> The downside of defaulting everything is that it clobbers the user's spec even when they never touched config. A user who submits the minimal manifest suddenly sees a fully-populated config block in their resource, with values they didn't set and may not expect. That makes the API noisier and harder to reason about, it becomes unclear what the user actually configured vs. what was silently injected.
> 
> The cleaner contract is: if config is nil, the operator uses its built-in defaults. If the user sets config, they're explicitly overriding specific fields. That way the spec reflects intent, not infrastructure defaults.
> 
> Happy to discuss further if there's a use case I'm missing!

I think there are no good or wrong approaches here. Looking at some of the operators on github it looks like some use the approach with defaults and some without.

One thing I wanted to note is that there are other APIs, also core APIs like `Pod` which have exactly this behavior as you noted: they have a lot of defaults, which causes the object to have a lot more fields set after being applied to the API server.

I believe what's important here for us is UX of the API and maintenance costs:

- From UX perspective I believe we should add the defaults at the CRD level. This can be read by users and by tools and be used on various different levels.
- From maintenance perspective, if we add the defaults in operator code we need to make sure comments on CRD fields are aligned with them. I'm not sure if there is an easy way to do that.
  - Additionally, I believe I mentioned it somewhere else in this PR comments is that having defaults on fields will make the API stable in a way that changing the operator version will not cause any changes to the underlying (managed) resources. Only changes to the CRDs in cluster would do that (but AFAIK this would not affect the pre-existing custom resources in the cluster).

alacuku:
I talked with @pmalek and we agreed on this approach: no +kubebuilder:default markers on config fields. The keg binary already sets its own defaults internally, so adding them at the CRD level would just duplicate information that already lives in the binary. If the binary's defaults ever change, we'd need to update the CRD too, which is an unnecessary maintenance burden. The comments on each field document what the keg default is, so users have the information they need without the API server silently injecting values they didn't set.

pmalek:
> The comments on each field document what the keg default is, so users have the information they need without the API server silently injecting values they didn't set.

And the CRD API will link to the underlying (KEG) docs page so that users can inspect the defaults there.

Additionally, default are not going to be set at the operator level as that would be tricky in upgrade scenarios (when KO upgrades, it carries with it an update to a default KEG version which would change defaults as well).

api/eventgateway/v1alpha1/dataplane_types.go
mlavacca: Yep, I am not saying it won't work; different groups make the trick here. What I am just saying is that `kubectl get dataplane` won't be enough anymore, the group will be required, and that just makes UX less smooth. 

However, since you are opinionated here, I am fine with keeping `DataPlane` only 👍🏻

api/eventgateway/v1alpha1/dataplane_types.go
mlavacca: As @pmalek suggested, I agree that `DataPlaneConfig` is a good candidate. Nothing fancy, just a bit more descriptive. 

Since we are setting defaults in the config struct, we need to set a default for this value here as well, so that in case it is not set by the user, the proper defaults are still set.

Makefile
```diff
@@ -1140,4 +1140,5 @@ lint.api: download.kube-api-linter
 		./api/konnect/v1alpha1/... \
 		./api/konnect/v1alpha2/... \
 		./api/common/v1alpha1/...  \
-		./api/x-konnect/...
+		./api/x-konnect/... \
+		./api/eventgateway/v1alpha1/... \
```
```suggestion
		./api/eventgateway/v1alpha1/...
```

api/eventgateway/v1alpha1/dataplane_types.go
alacuku: No defaults for the config struct so far. See latest comments, unless you have a strong use case around defaults.

mlavacca (dismissed):

pmalek (dismissed):

api/eventgateway/v1alpha1/kegdataplane_types.go
pmalek, line 1: Nit: now that the type is renamed we can rename the file as well to match it: `kegdataplane_types.go`

The kustomization seems to be already updated: https://github.com/Kong/kong-operator/pull/3765/changes#diff-4a47b4f788a351fb74ab951b33251242fd072022131e0a0931043110d0a5eb9bR31

test/integration/ko/dataplane_scale_subresource_test.go
```diff
@@ -26,7 +26,7 @@ func runKubectlScaleDataPlane(t *testing.T, namespacedName types.NamespacedName,
 	t.Helper()
 
 	// Execute kubectl scale command
-	cmd := "kubectl scale --namespace=" + namespacedName.Namespace + " dataplane/" + namespacedName.Name + " --replicas=" + fmt.Sprint(replicas)
+	cmd := "kubectl scale --namespace=" + namespacedName.Namespace + " dataplanes.gateway-operator.konghq.com/" + namespacedName.Name + " --replicas=" + fmt.Sprint(replicas)
```
pmalek: We should be able to revert this change now, right?

CHANGELOG.md
```diff
@@ -49,6 +49,9 @@
 - TLSRoute support: Add `TLSRoute` reconciler in hybrid gateway controllers
   and add translator of `TLSRoute`.
   [#3763](https://github.com/Kong/kong-operator/pull/3763)
+- EventGatewayDataPlane: introduces the eventgateway.konghq.com/v1alpha1 API group with
+  a new DataPlane kind that manages a keg binary Deployment connecting to Konnect
+  via a referenced KonnectEventGateway resource.
```
pmalek, line 54: This should be update to reflect the new name

test/integration/ko/dataplane_scale_subresource_test.go
alacuku: Yes, but it doesn't hurt having it more scoped when getting the dataplane. Do you want me to remove it?

CHANGELOG.md
alacuku: Done.

test/integration/ko/dataplane_scale_subresource_test.go
alacuku: Dropped the commit.

pmalek: This mimicks how users run these commands so it's better to keep it as it was 👍

pmalek (approved):

PR #3769, opened 2026-04-02
fix(gateway): patch controlplane when extensions differ

### LINKED ISSUE #3768: ControlPlane Extensions equality check is inverted, causing unnecessary patches
## Current Behavior

When reconciling the `ControlPlane`, the update condition checks:

`reflect.DeepEqual(controlPlane.Spec.Extensions, expectedExtensions)`

inside an `if` that is meant to detect drift.

Because of this, the reconciler enters the update path when `controlPlane.Spec.Extensions` is already equal to `expectedExtensions`. This can lead to:

* unnecessary patch attempts
* misleading `"controlplane config is out of date"` logs
* unnecessary reconcile noise

## Expected Behavior

The reconciler should only enter the update path when `Extensions` actually differ from the expected value.

The comparison should be negated:

`!reflect.DeepEqual(controlPlane.Spec.Extensions, expectedExtensions)`

## Steps To Reproduce

1. Reconcile a `ControlPlane` where `controlPlane.Spec.Extensions` is already equal to `expectedExtensions`
2. Observe that the update branch is still entered because the equality check is inverted
3. Check logs for:

   * `"controlplane config is out of date"`

## Operator Version

N/A

## `kubectl version`

N/A

## Additional Context

Current code:

```go
if !controlPlaneSpecDeepEqual(&controlPlane.Spec.ControlPlaneOptions, &expectedControlPlaneOptions) ||
    reflect.DeepEqual(controlPlane.Spec.Extensions, expectedExtensions) {
```

I believe the `Extensions` comparison should be:

```go
!reflect.DeepEqual(controlPlane.Spec.Extensions, expectedExtensions)
```

badam1:
## What this PR does / why we need it

This PR fixes a bug in `provisionControlPlane()` where `ControlPlane` reconciliation updates the resource only when `Extensions` are equal, instead of when they differ.

Before this change, extensions-only drift was ignored because the condition used:

```go
reflect.DeepEqual(controlPlane.Spec.Extensions, expectedExtensions)
```

inside the update check, which has the opposite meaning of what we want. As a result, if only `Extensions` changed, the `ControlPlane` spec was not patched and could remain stale.

This PR corrects that comparison and adds a unit test covering the extensions-only drift case to prevent regression.

## Which issue this PR fixes

Fixes #3768

## Special notes for your reviewer

* This is a small logic fix with a focused regression test.
* The new test verifies that `ControlPlane.Spec.Extensions` is updated even when `ControlPlaneOptions` stay unchanged.

## PR Readiness Checklist

* [ ] the `CHANGELOG.md` release notes have been updated to reflect significant changes


pmalek:
This has been cherry picked and merged in https://github.com/Kong/kong-operator/pull/4477.

Thanks for this contribution @badam1 🙇

PR #3777, opened 2026-04-02
feat(crd-from-oas): add reconciler code generation infrastructure with Portal as first entity

### LINKED ISSUE #3767: Reconciler code generation: tool infrastructure + Portal root entity
### Problem statement

According to the design document of #3356, we need to extend the `crd-from-oas` tool to generate reconciler wiring code (interface methods, watch options, cache indexes) and use Portal (`/v3/portals`) as the first test case.

This includes:
- Adding `reconciler` subsection parsing in `config.yaml`
- Generating `zz_generated_reconciler_funcs.go` (EntityType + auth accessor interfaces)
- Generating `zz_generated_watch_<entity>.go` (watch options for entity + KonnectAPIAuthConfiguration)
- Generating `zz_generated_<entity>.go` index files (cache indexes for auth ref queries)
- Supporting output to directories outside the API types directory (`RelativeDir`)
- Manual Portal entity wiring (constraints, ops stubs, controller setup, scheme registration)

### Acceptance criteria

- [ ] `crd-from-oas` generates reconciler funcs, watch, and index files for Portal based on `config.yaml`
- [ ] Portal is registered as a `SupportedKonnectEntityType` and wired into the controller setup
- [ ] CRD validation tests are updated for the new `KonnectConfiguration` field in Portal spec
- [ ] Unit tests pass (`make test.unit`)

### Out of scope

- SDK CRUD operations generation (`ops_<entity>.go`) — tracked separately
- Elimination of manual switch/case registration
- Non-root (child) entity support

tao12345666333:

**Which issue this PR fixes**

Fixes #3767

**PR Readiness Checklist**:


controller/konnect/zz_generated_watch_portal.go
```diff
@@ -0,0 +1,58 @@
...
+		auth, ok := obj.(*konnectv1alpha1.KonnectAPIAuthConfiguration)
+		if !ok {
+			return nil
+		}
+		var l xkonnectv1alpha1.PortalList
+		if err := cl.List(ctx, &l,
+			// TODO: change this when cross namespace refs are allowed.
+			client.InNamespace(auth.GetNamespace()),
+			client.MatchingFields{
+				index.IndexFieldPortalOnAPIAuthConfiguration: auth.Name,
```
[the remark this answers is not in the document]
tao12345666333: Great

api/x-konnect/v1alpha1/zz_generated_reconciler_funcs.go
pmalek, line 1: Can we extract this into a separate PR to make this a slightly smaller PR for easier review? WDYT?

config/crd/kong-operator/x-konnect.konghq.com_portals.yaml
```diff
@@ -295,7 +317,6 @@ spec:
                 description: |-
                   ID is the unique identifier of the Konnect entity as assigned by Konnect API.
                   If it's unset (empty string), it means the Konnect entity hasn't been created yet.
-                maxLength: 256
```
pmalek, line 298: Why are we removing this?

crd-from-oas/main.go
```diff
@@ -13,9 +13,10 @@ import (
 
 // EnvConfig holds the configuration for the generator, populated from environment variables.
 type EnvConfig struct {
-	InputFile  string `env:"INPUT_FILE" envDefault:"openapi.yaml"`
-	OutputDir  string `env:"OUTPUT_DIR" envDefault:"api/"`
-	ConfigFile string `env:"CONFIG_FILE,required"`
+	InputFile   string `env:"INPUT_FILE" envDefault:"openapi.yaml"`
+	OutputDir   string `env:"OUTPUT_DIR" envDefault:"api/"`
+	ConfigFile  string `env:"CONFIG_FILE,required"`
+	ProjectRoot string `env:"PROJECT_ROOT" envDefault:".."`
```
pmalek, line 21: What do we need this for? Can we add a comment for this?

hack/generators/conversion-webhook/main.go
```diff
@@ -61,8 +61,15 @@ func main() {
 	}
 	crdContent := out.String()
 
-	// TODO: This makes sure that temporary CRDs are not included in the chart.
-	crdContent = filterOutAPIGroup(crdContent, "x-konnect.konghq.com")
+	// Keep temporary x-konnect CRDs out of the chart, but allow the Portal CRD,
+	// which is required by the operator at startup.
```
pmalek, line 65: Let's keep a TODO here as we'll need to update this as we add more CRDs in x-konnect and then after we migrate all of those into the konnect group.

modules/manager/controller_setup.go
```diff
@@ -134,6 +135,7 @@ func SetupCacheIndexes(ctx context.Context, mgr manager.Manager, cfg Config) err
 			index.OptionsForKonnectCloudGatewayNetwork(),
 			index.OptionsForKonnectExtension(),
 			index.OptionsForKonnectCloudGatewayDataPlaneGroupConfiguration(cl),
+			index.OptionsForPortal(),
```
pmalek: Are we making this an auto generated list already? If not let's leave a comment with an issue link about this being changed in the future to an autogenerated list

modules/manager/controller_setup.go
```diff
@@ -744,6 +751,7 @@ func SetupControllers(mgr manager.Manager, c *Config, cpsMgr *multiinstance.Mana
 			newKonnectEntityController[configurationv1alpha1.KongDataPlaneClientCertificate](controllerFactory),
 			newKonnectEntityController[configurationv1alpha1.KongVault](controllerFactory),
 			newKonnectEntityController[configurationv1alpha1.KongSNI](controllerFactory),
+			newKonnectEntityController[xkonnectv1alpha1.Portal](controllerFactory),
```
pmalek: Similarly this: let's make sure this is captured with a TODO (to be migrated to an auto generated list).

config/crd/kong-operator/x-konnect.konghq.com_portals.yaml
tao12345666333: Good catch.

This came from switching Portal status to inline `konnectv1alpha2.KonnectEntityStatus`. That shared type currently does not carry the same `MaxLength` markers, so the generated CRD schema became weaker here.  I'll restore those constraints instead of removing them.

crd-from-oas/main.go
tao12345666333: This is needed because reconciler generation now writes some generated files outside the API types directory, specifically into `controller/konnect` and `internal/utils/index`.
I'll add a comment explaining that.

hack/generators/conversion-webhook/main.go
tao12345666333: OK

modules/manager/controller_setup.go
tao12345666333: ok

modules/manager/controller_setup.go
tao12345666333: OK， I will add a TODO link to #3785

charts/kong-operator/templates/cluster-role.yaml
```diff
@@ -523,3 +523,21 @@ rules:
...
+      - update
+      - watch
+  - apiGroups:
+      - x-konnect.konghq.com
+    resources:
+      - portals/finalizers
+      - portals/status
+    verbs:
+      - patch
+      - update
```
pmalek, line 543: Is this generated? Can we capture generation of this with an issue?

api/konnect/v1alpha2/konnect_entity_status.go
pmalek, line 1: These changes seem OK but I'm a little bit hesitant if we'd like to include this in this PR. Are we adding this because of kube-api-linter complaints?

crd-from-oas/config.yaml
```diff
@@ -22,6 +22,8 @@ apiGroupVersions:
             path: github.com/Kong/sdk-konnect-go/models/components.CreatePortal
           update:
             path: github.com/Kong/sdk-konnect-go/models/components.UpdatePortal
+        reconciler:
+          isRoot: true
```
pmalek, line 26: This is OK to do now but I believe we can easily infer this later on simply from the URL path: if there are 0 params (`{}` enclosed strings) then this is a root entity.

controller/konnect/constraints/constraints.go
```diff
@@ -59,7 +60,8 @@ type SupportedKonnectEntityType interface {
 		configurationv1alpha1.KongKeySet |
 		configurationv1alpha1.KongSNI |
 		configurationv1alpha1.KongDataPlaneClientCertificate |
-		konnectv1alpha1.MCPServer
+		konnectv1alpha1.MCPServer |
+		xkonnectv1alpha1.Portal
```
pmalek, line 66: Can we make place for a comment here and capture the generation of this list with an issue?

config/rbac/role/role.yaml
```diff
@@ -523,3 +523,21 @@ rules:
...
+  - update
+  - watch
+- apiGroups:
+  - x-konnect.konghq.com
+  resources:
+  - portals/finalizers
+  - portals/status
+  verbs:
+  - patch
+  - update
```
pmalek, line 543: Ditto about generating this

crd-from-oas/pkg/generator/reconciler.go
```diff
@@ -0,0 +1,313 @@
...
+			return b.For(&{{.APIGroupPackageAlias}}.{{.EntityName}}{})
+		},
+		func(b *ctrl.Builder) *ctrl.Builder {
+			return b.Watches(
+				&{{.APIAuthPackageAlias}}.KonnectAPIAuthConfiguration{},
+				handler.EnqueueRequestsFromMapFunc(
+					enqueue{{.EntityName}}ForKonnectAPIAuthConfiguration(cl),
+				),
+			)
+		},
```
pmalek, line 101: This is somewhat OK now but as soon as we generate functions for other types which do not have a direct reference to `KonnectAPIAuthConfiguration` this won't be true.

I believe this should only be done for root objects with have the reference, right?

api/konnect/v1alpha2/konnect_entity_status.go
tao12345666333: This was not added just to satisfy kube-api-linter.

When `Portal` switched to inline the shared `konnectv1alpha2.KonnectEntityStatus`, the generated CRD schema lost the existing `MaxLength` constraints that were previously present on the Portal status fields. 

https://github.com/Kong/kong-operator/pull/3777#discussion_r3052037502

Adding the markers to the shared type restores those constraints at the source of truth, instead of keeping stricter validation only in one generated CRD.

api/x-konnect/v1alpha1/zz_generated_reconciler_funcs.go
tao12345666333: Due to time constraints, I prefer to maintain the current state for now. In future PRs, I will pay attention to controlling the scope of changes to avoid putting pressure on the reviewer.

crd-from-oas/config.yaml
tao12345666333: Agreed that we should be able to infer this later from the path shape.

For this PR, I kept `reconciler.isRoot` explicit because the current generator logic uses it to decide which entities get root-specific auth-ref wiring and watch/index generation. I wanted to keep the initial implementation explicit and low-risk, and we can follow up by deriving it from the URL path once we add support for more entity shapes.

hack/generators/conversion-webhook/main.go
tao12345666333: Added a TODO there with an issue link

controller/konnect/constraints/constraints.go
tao12345666333: Added a TODO with an issue link here.

This list is still manually maintained today, and issue #3785 tracks eliminating these manual generated-entity registrations.

crd-from-oas/config.yaml
pmalek: Thanks for the explanation 👍 

Let's make sure that when we write similar features in the future to include a comment explaining this just as you explained here. 

Let's leave it as is.

charts/kong-operator/templates/cluster-role.yaml
tao12345666333: This RBAC is generated downstream, so I did not add the TODO in the YAML outputs.

Instead, I added the TODO at  `controller/konnect/reconciler_generic_rbac.go`, and opened #3833 to track generating RBAC markers for generated Konnect entities.

controller/konnect/constraints/constraints.go
pmalek: Thanks! 🙇 

Can we add a TODO comment with link to 3785 here?

charts/kong-operator/templates/cluster-role.yaml
pmalek: Oh, I must have missed it. Thanks! 🙇

crd-from-oas/pkg/generator/reconciler.go
tao12345666333: Good point.

I updated the generator so that `GetKonnectAPIAuthConfigurationRef()` is only generated for root entities (`isRoot: true`). That keeps the current Portal behavior, but avoids generating the wrong accessor for future child entities that inherit auth through a parent object.

pmalek: `GetKonnectAPIAuthConfigurationRef()` is indeed fixed and guarded with `isRoot` but the watch func above is not.

We should also update that, shouldn't we?

tao12345666333: The watch template itself is still root-oriented, but it is only generated for root entities.

`generateReconcilerFiles()` already guards watch/index generation with `reconciler.isRoot` before calling `generateWatch()` / `generateIndex()`, so non-root entities do not get this watch wiring today.

https://github.com/Kong/kong-operator/pull/3777/changes#diff-d9a5511f545aa8c96cbc5f1085b58e338e12d255ae6b793e9e5cdaf7bb5eac81R184

I kept the template root-specific for now, and we can generalize it later when we add Phase 2 child-entity watch generation.

pmalek: > The watch template itself is still root-oriented, but it is only generated for root entities.

Ok I missed that. Alright, let's keep it this way and this will need to be updated for non root types anyway. Thanks for explanation 👍

pmalek (approved):

PR #3847, opened 2026-04-09
feat(hybridgateway): add support for `konghq.com/protocol` service annotation

### LINKED ISSUE #3750: Hybrid Gateway - Honor backend protocol annotations when translating KongService resources
## Problem Statement

In the Hybrid Gateway translation path, generated `KongService` resources are always created with protocol `http`.

Today, `controller/hybridgateway/service/service.go` calls `WithProtocol("http")` unconditionally, and `controller/hybridgateway/builder/kongservice.go` only accepts `http` for translated `KongService` resources.

This means that when an `HTTPRoute` points to a backend that is actually served over HTTPS (for example `Service` on port `443`), the resulting Konnect `Service` is still translated as `http` and effectively defaults to `http/80`.

## Proposed Solution

We can add support for `konghq.com/protocol` when using Hybrid Gateway.

## Additional Information

This behavior appears to have been introduced intentionally in #2872 , where the PR description states that because this is `HTTPRoute` conversion, the protocol is always set to HTTP.

## Acceptance Criteria

- [ ] As a user, I can configure an HTTPRoute backend that is served over HTTPS and have the generated Konnect Service translated with the correct upstream protocol and port

sandromodarelli:

The hybrid gateway controller was hardcoding `protocol: http` when translating HTTPRoute rules into `KongService` resources, regardless of the actual protocol the backend Service speaks. This meant users could not configure HTTPS or gRPC upstreams via HTTPRoute — the generated Konnect Service always used `http`.

This PR adds support for the `konghq.com/protocol` annotation on backend Kubernetes Services in the hybrid gateway translation path, consistent with how the existing ingress-controller honors this annotation today.

Changes:
- **`metadata/annotation.go`**: Added `ExtractProtocol()` and `IsValidProtocol()` helpers, mirroring `ExtractProtocolName`/`ValidateProtocol` from the ingress-controller (those live under `internal/` and cannot be imported directly).
- **`builder/kongservice.go`**: Expanded `WithProtocol()` from only `http`/`tls` to the full set of SDK-supported protocols: `https`, `grpc`, `grpcs`, `ws`, `wss`, `tcp`, `tls_passthrough`, `udp`.
- **`service/service.go`**: Added `resolveProtocolFromBackendRefs()` that iterates the HTTPRoute rule's BackendRefs, fetches each referenced Kubernetes Service, and reads the `konghq.com/protocol` annotation. The first valid annotation found wins; invalid values are logged and ignored; missing annotation falls back to `http`.

**Which issue this PR fixes**

Fixes #3750

- The ingress-controller's `ExtractProtocolName` (`ingress-controller/internal/annotations`) and `ValidateProtocol` (`ingress-controller/internal/util`) are the reference implementations. They live under Go `internal/` packages and cannot be imported from the operator, so equivalent one-liner functions are added to the `metadata` package instead.
- TLSRoute is intentionally left unchanged (still defaults to `tls`) — that case has its own existing TODO about `tls_passthrough`.
- When a rule has multiple BackendRefs with conflicting annotations, "first valid wins" is used. All backends in a single rule share one `KongService` so they should agree; a future issue can add a warning for mismatches.

**PR Readiness Checklist**:


controller/hybridgateway/service/service.go
```diff
@@ -105,3 +107,52 @@ func ServiceForRule[
...
+		}
+
+		svc := &corev1.Service{}
+		if err := cl.Get(ctx, client.ObjectKey{Namespace: bRefNamespace, Name: string(bRef.Name)}, svc); err != nil {
+			log.Debug(logger, "Failed to fetch backend Service for protocol annotation check",
+				"service", fmt.Sprintf("%s/%s", bRefNamespace, bRef.Name), "error", err)
+			continue
+		}
+
+		protocol := metadata.ExtractProtocol(svc.GetAnnotations())
```
randmonkey: Should we run a `toLower` first to allow upper case version of protocols?

randmonkey (approved):

PR #3854, opened 2026-04-10
test(crd-from-oas): strengthen generated SDK ops assertions

### LINKED ISSUE #3774: Reconciler code generation: KonnectEventControlPlane Entity
### Problem statement

KonnectEventControlPlane is a root entity exposed via `/v1/event-gateways`. Per the design document (#3356), once the reconciler generation infrastructure is in place (#3767), adding a new root entity should involve:
1. Adding `reconciler: isRoot: true` to the entity's `config.yaml` entry
2. Implementing SDK CRUD operations (or generating them if ops generation is available)
3. Manual entity registration in constraints, ops, watch, controller setup, and scheme

This issue tracks adding KonnectEventControlPlane reconciler support.

### Acceptance criteria

- [ ] KonnectEventControlPlane has `reconciler` config in `config.yaml` and generated wiring files
- [ ] SDK CRUD operations (create, update, delete, getForUID) are implemented or generated
- [ ] Entity is registered in controller setup and reconciles correctly
- [ ] Tests are added (CRD validation, unit tests)

### Out of scope

- CRD generation (tracked separately under #2839 sub-issues)
- Envtest lifecycle tests (tracked separately)

tao12345666333:

To facilitate the review, this has been split from #3843

It comes from https://github.com/Kong/kong-operator/pull/3843#discussion_r3061391728

`testValuesForProperty()` decides what test value to use for each field in the generated SDK ops tests. It works fine for basic types like `string` or `*bool`, but when a field uses a named type from `$ref` (like LabelsUpdate or GatewayName), `goType()` returns the alias name instead of `string`. The old switch doesn't have a case for these, so it returns `("", "")` and the field gets quietly skipped, no test value, no assertion.

This PR adds handling for those `RefName` cases: 
Named string types: `GatewayName("test-value")`
Pointer to named string types: `new(MinRuntimeVersion("test-value"))`
Map types with string values: `Labels{"test-key": "test-value"}`

You can see the real effect in `portal_sdkops_test.go`: `Portal.labels (type LabelsUpdate) `was missing before, now it's covered.

**Which issue this PR fixes**

part of https://github.com/Kong/kong-operator/issues/3774

**PR Readiness Checklist**:


crd-from-oas/pkg/generator/generator.go
```diff
@@ -1160,16 +1160,30 @@ func testValuesForProperty(prop *parser.Property, goType string) (string, string
 	case "string":
 		return `"test-value"`, `"test-value"`
 	case "*string":
-		return `new("test-value")`, `"test-value"`
+		return `func(v string) *string { return &v }("test-value")`, `"test-value"`
 	case "bool":
 		return "true", "true"
 	case "*bool":
-		return `new(true)`, "true"
+		return `func(v bool) *bool { return &v }(true)`, "true"
```
pmalek: Can you add description to this PR why is this needed? It would seem to me that `new(<value>)` should work in most cases and I don't understand why we'd need a closure func 🤔

tao12345666333: Thank you for pointing that out. I have updated the description, and reverted this change to use `new(<value>)`

pmalek (approved):

PR #3870, opened 2026-04-13
chore(ci): adjust automatic GitHub secret scanning

programmer04:

As discussed during team sync 8.04.2026, avoid false positives with proper configuration as described in https://docs.github.com/en/code-security/how-tos/secure-your-secrets/customize-leak-detection/excluding-folders-and-files-from-secret-scanning


pmalek (approved):
🚀

tao12345666333 (approved):

Issue #3833, opened 2026-04-09
Reconciler code generation: generate RBAC markers for generated Konnect entities


tao12345666333:
## Problem Statement

The reconciler generation infrastructure added in #3767 generates wiring files for new Konnect entities, but RBAC still requires manual updates.

Today, when adding a generated entity such as `Portal`, we still need to manually add `+kubebuilder:rbac` markers in `controller/konnect/reconciler_generic_rbac.go`. Those markers are then propagated into:

- `config/rbac/role/role.yaml`
- `charts/kong-operator/templates/cluster-role.yaml`

This adds another manual follow-up step when enabling reconciliation for a generated entity, and it is easy to miss during review.

## Proposed Solution

Extend the reconciler code generation flow so that generated Konnect entities do not require manual RBAC marker additions in `controller/konnect/reconciler_generic_rbac.go`.

The source of truth should stay close to reconciler generation rather than in generated YAML outputs.

## Additional Information

Related issues:

- Epic: #2839
- Reconciler generation infrastructure: #3767
- Manual entity registration follow-up: #3785

This issue is intentionally scoped to RBAC generation only.
It does not cover SDK ops generation or the broader registration cleanup tracked elsewhere.

## Acceptance Criteria

- [ ] Generated Konnect entities do not require manual `+kubebuilder:rbac` marker additions in `controller/konnect/reconciler_generic_rbac.go`
- [ ] Generated RBAC covers the entity resource plus `/status` and `/finalizers` where needed
- [ ] `config/rbac/role/role.yaml` is produced correctly from the generated/source RBAC definition
- [ ] `charts/kong-operator/templates/cluster-role.yaml` continues to stay in sync with generated RBAC output
- [ ] ~~Documentation is updated with this new feature / enhancement~~
- [ ] Tests are added for this feature / enhancement
- [ ] Helm chart is updated to work with this feature / enhancement
- [ ] As a developer, I can add a generated Konnect entity without separately updating RBAC markers by hand

pmalek:
This is covered now by #3888.

PR #3888, opened 2026-04-14
feat(crd-from-oas): generate RBAC markers

pmalek:

**Which issue this PR fixes**

Part of #3833

**PR Readiness Checklist**:


crd-from-oas/pkg/generator/reconciler.go
```diff
@@ -112,10 +112,54 @@ func {{.EntityNameLowerCamel}}APIAuthConfigurationRef(object client.Object) []st
...
+
+// generateRBAC generates a Go file containing +kubebuilder:rbac marker comments
+// for the given entities. The file is placed in the controller/konnect package
+// so that controller-gen picks up the markers when generating RBAC manifests.
+func (g *Generator) generateRBAC(entityNames []string) (string, error) {
+	entities := make([]rbacEntity, 0, len(entityNames))
+	for _, entityName := range entityNames {
+		entities = append(entities, rbacEntity{
+			APIGroup:     g.config.APIGroup,
+			ResourceName: strings.ToLower(entityName) + "s",
```
tao12345666333: The only point I'm concerned about is this; here, the plural form is simply created by adding an "s," which may not be very safe. I checked that there might also be some resources like *Policy on Konnect (for example, eventgatewaylistenerpolicies), so I suggest using a method similar to https://github.com/gobuffalo/flect  here, which might be safer.

tao12345666333: `flect.Pluralize`

pmalek: Well spotted.

I believe we can just resort to `meta.UnsafeGuessKindToResource` which _is_ marked as broken upstream but should do the trick for us here.

This way we don't have to pull in a new dependency.

PTAL 5c7f487127b35132a69ab6e51418ba3aa38eff09

tao12345666333 (dismissed):

tao12345666333 (approved):

PR #3899, opened 2026-04-15
feat(ssa): add SSA engine and shared infra for KegDataPlane controller

alacuku:
**What this PR does / why we need it**:

Add the foundational building blocks needed by the upcoming KegDataPlane controller.

**Which issue this PR fixes**

Fixes #

Part of #3542 
**Special notes for your reviewer**:

**PR Readiness Checklist**:

Complete these before marking the PR as `ready to review`:

- [ ] the `CHANGELOG.md` release notes have been updated to reflect significant changes

Issue #3542, opened 2026-03-10
Add controller for EventGatewayDataPlane to manage keg deployments


alacuku:
## Problem Statement

Implement the controller reconciliation logic that translates EventGatewayDataPlane resources into running keg data plane deployments with proper Konnect connectivity.

Deploying keg manually requires coordinating multiple Kubernetes resources (Deployment, Service, ConfigMap, Secret) and complex bootstrapping (fetching cluster ID and region from Konnect, provisioning certificates, generating configuration YAML). The controller automates this entire workflow based on a single declarative resource

PR #3901, opened 2026-04-15
feat(eventgateway): add KegDataPlane controller implementation

alacuku:

Add the reconciler for the KegDataPlane custom resource, which manages
the lifecycle of a Kong Event Gateway data plane deployment.

The controller:
- Reconciles KegDataPlane objects by ensuring a Deployment, a Kafka
  Service, and an mTLS client certificate Secret.
- Resolves the referenced KonnectEventControlPlane and surfaces the
  resolution result as a status condition.
- Uses Server-Side Apply (SSA) for all owned resource mutations, with a
  diff-before-apply guard to prevent spurious updates.
- Applies status conditions via the status subresource using SMD-based
  comparison (ApplyStatusIfChanged).
- Records Kubernetes events via the new events.EventRecorder API.
- Watches KonnectEventControlPlane objects and enqueues the owning
  KegDataPlane when they change.

**Which issue this PR fixes**

Fixes #

Part of: #3542 

Unit tests will be implemented in another PR. Wait for it.
This PR depends on: #3899 

**PR Readiness Checklist**:

### DIFF
--- a/controller/eventgateway/dataplane/consts.go
+++ b/controller/eventgateway/dataplane/consts.go
@@ -1,0 +1,82 @@
+/*
+Copyright 2025 Kong, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package dataplane
+
+const (
+	// FieldManager is the field manager name used for Server-Side Apply operations.
+	FieldManager = "gateway-operator"
+
+	// ControllerName is the name used for logging and event recording.
+	ControllerName = "keg-dataplane"
+
+	// DefaultKafkaPort is the default port exposed by the Kafka listener.
+	DefaultKafkaPort int32 = 9092
+
+	// DefaultHealthPort is the default port used for the keg health endpoint.
+	DefaultHealthPort int32 = 8080
+
+	// KonnectCertVolumeName is the name of the volume that holds the Konnect mTLS certificate.
+	KonnectCertVolumeName = "konnect-cert"
+
+	// KonnectCertMountPath is the path where the Konnect certificate Secret is mounted in the keg container.
+	KonnectCertMountPath = "/var/konnect-client-certificate/"
+)
+
+// -----------------------------------------------------------------------------
+// Consts - KEG environment variable names
+// Konnect settings use the KONG_KONNECT_ prefix; all other keg settings use
+// the KEG__ prefix with double underscores for nesting.
+// -----------------------------------------------------------------------------
+
+const (
+	// EnvKonnectRegion is the keg environment variable for the Konnect region.
+	EnvKonnectRegion = "KONG_KONNECT_REGION"
+	// EnvKonnectGatewayClusterID is the keg environment variable for the Konnect gateway cluster ID.
+	EnvKonnectGatewayClusterID = "KONG_KONNECT_GATEWAY_CLUSTER_ID"
+	// EnvKonnectClientCertPath is the keg environment variable for the Konnect mTLS client certificate path.
+	EnvKonnectClientCertPath = "KONG_KONNECT_CLIENT_CERT_PATH"
+	// EnvKonnectClientKeyPath is the keg environment variable for the Konnect mTLS client key path.
+	EnvKonnectClientKeyPath = "KONG_KONNECT_CLIENT_KEY_PATH"
+	// EnvKonnectDomain is the keg environment variable for the Konnect domain override.
+	EnvKonnectDomain = "KONG_KONNECT_DOMAIN"
+	// EnvKonnectAPIRequestTimeout is the keg environment variable for the Konnect API request timeout.
+	EnvKonnectAPIRequestTimeout = "KONG_KONNECT_API_REQUEST_TIMEOUT"
+	// EnvKonnectInsecureSkipVerify is the keg environment variable to skip TLS verification.
+	// For testing only.
+	EnvKonnectInsecureSkipVerify = "KONG_KONNECT_INSECURE_SKIP_VERIFY"
+
+	// EnvConfigPollInterval is the keg environment variable for the config poll interval.
+	EnvConfigPollInterval = "KEG__CONFIG_POLL_INTERVAL"
+	// EnvEnableDebugEndpoints is the keg environment variable to enable debug endpoints.
+	EnvEnableDebugEndpoints = "KEG__ENABLE_DEBUG_ENDPOINTS"
+
+	// EnvObsLogFlags is the keg environment variable for log level flags.
+	EnvObsLogFlags = "KEG__OBSERVABILITY__LOG_FLAGS"
+	// EnvObsLogFormat is the keg environment variable for the log output format.
+	EnvObsLogFormat = "KEG__OBSERVABILITY__LOG_FORMAT"
+	// EnvObsMetricsRollupAllowMap is the keg environment variable for the metrics rollup allow map.
+	EnvObsMetricsRollupAllowMap = "KEG__OBSERVABILITY__METRICS_ROLLUP_ALLOW_MAP"
+	// EnvObsPolicyErrorsInfoLogInterval is the keg environment variable for the policy errors info log interval.
+	EnvObsPolicyErrorsInfoLogInterval = "KEG__OBSERVABILITY__POLICY_ERRORS_INFO_LOG_INTERVAL"
+
+	// EnvRuntimeHealthAddr is the keg environment variable for the health listener address and port.
+	EnvRuntimeHealthAddr = "KEG__RUNTIME__HEALTH_LISTENER_ADDRESS_PORT"
+	// EnvRuntimeDrainDuration is the keg environment variable for the drain duration.
+	EnvRuntimeDrainDuration = "KEG__RUNTIME__DRAIN_DURATION"
+	// EnvRuntimeShutdownTimeout is the keg environment variable for the graceful shutdown timeout.
+	EnvRuntimeShutdownTimeout = "KEG__RUNTIME__SHUTDOWN_TIMEOUT"
+)
--- a/controller/eventgateway/dataplane/consts.go
+++ b/controller/eventgateway/dataplane/consts.go
@@ -40,6 +40,7 @@ const (
 // Consts - KEG environment variable names
 // Konnect settings use the KONG_KONNECT_ prefix; all other keg settings use
 // the KEG__ prefix with double underscores for nesting.
+// For more info see https://developer.konghq.com/event-gateway/configuration/.
 // -----------------------------------------------------------------------------
 
 const (
--- a/controller/eventgateway/dataplane/controller.go
+++ b/controller/eventgateway/dataplane/controller.go
@@ -1,0 +1,131 @@
+/*
+Copyright 2025 Kong, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package dataplane
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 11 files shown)


controller/eventgateway/dataplane/consts.go
```diff
@@ -0,0 +1,82 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package dataplane
+
+const (
+	// FieldManager is the field manager name used for Server-Side Apply operations.
+	FieldManager = "gateway-operator"
```
pmalek, line 21: Could we use `kong-operator` or will it be a problem?

I can see hybrid gateway reconciler already uses `gateway-operator`.

controller/eventgateway/dataplane/consts.go
```diff
@@ -0,0 +1,82 @@
...
+	KonnectCertVolumeName = "konnect-cert"
+
+	// KonnectCertMountPath is the path where the Konnect certificate Secret is mounted in the keg container.
+	KonnectCertMountPath = "/var/konnect-client-certificate/"
+)
+
+// -----------------------------------------------------------------------------
+// Consts - KEG environment variable names
+// Konnect settings use the KONG_KONNECT_ prefix; all other keg settings use
+// the KEG__ prefix with double underscores for nesting.
```
pmalek, line 42: Can we add a link here to KEG documentation for reference?

controller/eventgateway/dataplane/controller.go
```diff
@@ -0,0 +1,131 @@
...
+	LoggingMode logging.Mode
+
+	ClusterCASecretName      string
+	ClusterCASecretNamespace string
+	SecretLabelSelector      string
+	CertTTL                  time.Duration
+
+	// typeConverter is initialised once during SetupWithManager from the API
+	// server's OpenAPI v3 schemas. It supports all types (core K8s + CRDs) and
+	// is used for both diff-before-apply and SMD-based
```
pmalek: I'd suggest to unroll the SMD abbreviation to prevent ambiguity.

controller/eventgateway/dataplane/watch.go
```diff
@@ -0,0 +1,56 @@
...
+
+	eventgatewayv1alpha1 "github.com/kong/kong-operator/v2/api/eventgateway/v1alpha1"
+	konnectv1alpha1 "github.com/kong/kong-operator/v2/api/konnect/v1alpha1"
+	"github.com/kong/kong-operator/v2/internal/utils/index"
+)
+
+// enqueueForKonnectEventGatewayRef returns a MapFunc that enqueues reconcile requests
+// for all DataPlanes in the same namespace whose
+// spec.konnectEventGatewayRef.name matches the changed KonnectEventGateway.
+func enqueueForKonnectEventGatewayRef(cl client.Client) handler.MapFunc {
```
pmalek, line 34: ~Gateway -> ControlPlane? In order to stay consistent with CRD naming, WDYT?~

Actually, let's discuss this here: https://kongstrong.slack.com/archives/C0APS0S3ELF/p1776249962474109

I believe we should reconsider changing the name from `KonnectEventControlPlane` to `KonnectEventGateway` so that it aligns with UI and API.

controller/eventgateway/dataplane/status.go
```diff
@@ -0,0 +1,105 @@
...
+	return nil
+}
+
+// applyStatus patches the KegDataPlane status subresource via SSA and joins
+// any error into *err so the caller's named return reflects the failure.
+func (r *Reconciler) applyStatus(
+	ctx context.Context,
+	logger logr.Logger,
+	egdp *eventgatewayv1alpha1.KegDataPlane,
+	err *error,
```
pmalek: I understand we need this "trick" to use `defer applyStatus` and set the error as well.

Can we at the very least add a comment about this + check if `err` is not `nil` so that this doesn't panic?

controller/eventgateway/dataplane/status.go
```diff
@@ -0,0 +1,105 @@
...
+		return err
+	}
+
+	egdp.Status.Replicas = deployment.Status.Replicas
+	egdp.Status.ReadyReplicas = deployment.Status.ReadyReplicas
+
+	// Mark Not Ready only when zero pods are serving traffic.
+	// During a rolling update some replicas are still ready, so we stay Ready
+	// throughout the rollout and only flip to False when there is nothing left to serve.
+	if deployment.Status.ReadyReplicas == 0 {
```
pmalek, line 70: What if user specifies that he wants 0 replicas? Shouldn't 0 `ReadyReplicas` be a status considered as "ready"?

controller/eventgateway/dataplane/owned_deployment.go
```diff
@@ -0,0 +1,367 @@
...
+						consts.GatewayOperatorManagedByLabel:     consts.DataPlaneManagedByLabelValue,
+						consts.GatewayOperatorManagedByNameLabel: egdp.Name,
+					},
+				},
+				Template: *egdp.Spec.Deployment.PodTemplateSpec,
+			},
+		}
+		var mergeErr error
+		u, mergeErr = controllerpkgssa.MergeObjects(tc, base, userDeployment)
+		if mergeErr != nil {
```
pmalek: Why not simply:


```suggestion
		u, err = controllerpkgssa.MergeObjects(tc, base, userDeployment)
		if err != nil {
```

controller/eventgateway/dataplane/owned_secret.go
```diff
@@ -0,0 +1,84 @@
...
+			certificatesv1.UsageKeyEncipherment,
+			certificatesv1.UsageDigitalSignature,
+			certificatesv1.UsageClientAuth,
+		},
+		r.Client,
+		matchingLabels,
+		r.CertTTL,
+	)
+	if err != nil {
+		apimeta.SetStatusCondition(&egdp.Status.Conditions, metav1.Condition{
```
pmalek, line 67: Just curious: why in some places we use `apimeta.SetStatusCondition` and in some our own `k8sutils.SetCondition`? 🤔

controller/eventgateway/dataplane/consts.go
alacuku: Keeping it as `gateway-operator`. We could consider to change the FieldManager to `kong-operator` and make it a shared const value for all our controllers.

controller/eventgateway/dataplane/status.go
alacuku: I don't think so. The dataplane is considered ready when at least a replica is serving traffic.

controller/eventgateway/dataplane/owned_secret.go
alacuku: Nice catch, switching to use the `apimeta.SetStatusCondition` where possible. 
`k8sutils.SetCondition` carries some legacy baggage from the `ControlPlane v1beta1 `days, it silently filters out a "Scheduled" condition and requires the ConditionsAware interface. Since this is a brand-new controller with none of that history, going straight to the upstream `apimeta.SetStatusCondition` felt cleaner. It also handles `LastTransitionTime ` for us automatically, which is one less thing to get wrong.

controller/eventgateway/dataplane/status.go
alacuku: I see your point, now applyStatus returns its own error rather than taking a *error pointer, avoiding a
nil-dereference if called with a nil pointer. The caller merges it into the named return.

controller/eventgateway/dataplane/consts.go
pmalek: 💯

controller/eventgateway/dataplane/watch.go
pmalek: Discussion on CRD renaming will continue in that thread.
Closing my comment here. If we change the name of the CRD it will happen in a separate PR.

controller/eventgateway/dataplane/status.go
pmalek: Ok, this is consistent with `DataPlane`. When replicas are set to 0, it is also marked as `Ready=False`.

controller/eventgateway/dataplane/controller.go
```diff
@@ -0,0 +1,132 @@
...
+	// Ensure mTLS client certificate secret and set certificate condition.
+	certResult, certSecret, err := r.ensureCertificateSecret(ctx, egdp)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// Requeue if the Secret was just created/updated so the Deployment
+	// picks up the correct Secret name on the next reconcile.
+	if certResult != op.Noop {
+		return ctrl.Result{Requeue: true}, nil
```
pmalek: `Requeue` is [deprecated](https://github.com/kubernetes-sigs/controller-runtime/blob/f9589b9f2b9dddf8532b432bb8315f2820ab9971/pkg/reconcile/reconcile.go#L41).

Some of our other APIs still use it but if we're adding a new one let's try not to go against upstream's guidelines and use `RequeueAfter` instead.

controller/eventgateway/dataplane/owned_deployment.go
```diff
@@ -0,0 +1,367 @@
...
+
+	container := corev1.Container{
+		Name:  consts.KEGContainerName,
+		Image: image,
+		Env:   envVars,
+		SecurityContext: &corev1.SecurityContext{
+			AllowPrivilegeEscalation: &allowPrivilegeEscalation,
+			ReadOnlyRootFilesystem:   &readOnlyRootFilesystem,
+			RunAsUser:                &runAsUser,
+			RunAsGroup:               &runAsGroup,
```
pmalek: With go 1.26 you can simply use:


```suggestion
	container := corev1.Container{
		Name:  consts.KEGContainerName,
		Image: image,
		Env:   envVars,
		SecurityContext: &corev1.SecurityContext{
			AllowPrivilegeEscalation: new(false),
			ReadOnlyRootFilesystem:   new(true),
			RunAsUser:                new(int64(65532)),
			RunAsGroup:               new(int64(65532)),
```

similarly in other places.

Less code, to look at.

controller/eventgateway/dataplane/owned_deployment.go
```diff
@@ -0,0 +1,367 @@
...
+	}
+
+	healthPort := intstr.FromInt32(DefaultHealthPort)
+
+	allowPrivilegeEscalation := false
+	readOnlyRootFilesystem := true
+	runAsUser := int64(65532)
+	runAsGroup := int64(65532)
+
+	container := corev1.Container{
```
pmalek, line 182: To make this code ever so slightly more lean, WDYT about extracting container generation to a dedicated func?

controller/eventgateway/dataplane/owned_deployment.go
```diff
@@ -0,0 +1,367 @@
...
+			},
+		},
+	}
+
+	var replicas *int32
+	if egdp.Spec.Deployment != nil {
+		replicas = egdp.Spec.Deployment.Replicas
+	}
+
+	d := &appsv1.Deployment{
```
pmalek, line 249: Ditto about extracting this to a dedicated func.

controller/eventgateway/dataplane/owned_deployment.go
```diff
@@ -0,0 +1,367 @@
...
+
+	k8sutils.SetOwnerForObject(d, egdp)
+	return d, nil
+}
+
+// buildKEGEnvVars builds the full list of keg environment variables from
+// KonnectEventGateway status and DataPlane spec.config.
+func buildKEGEnvVars(
+	egdp *eventgatewayv1alpha1.KegDataPlane,
+	keg *konnectv1alpha1.KonnectEventControlPlane,
```
pmalek, line 286: This is not really `keg` right? It's more a cp

controller/eventgateway/dataplane/owned_service.go
```diff
@@ -0,0 +1,175 @@
...
+			sp.NodePort = *p.NodePort
+		}
+		ports = append(ports, sp)
+	}
+
+	var extraLabels map[string]string
+	for k, v := range kafka.Labels {
+		if extraLabels == nil {
+			extraLabels = make(map[string]string)
+		}
```
pmalek:
```suggestion
	extraLabels := make(map[string]string)
	for k, v := range kafka.Labels {
```

controller/eventgateway/dataplane/controller.go
alacuku: Nice catch, just returning is enough since the watch on the owned secret triggers a reconcile.

controller/eventgateway/dataplane/owned_deployment.go
alacuku: Nice!

controller/eventgateway/dataplane/owned_deployment.go
alacuku: This code is already in a separate file, I would leave it as it is. No real value is added by extracting it.

controller/eventgateway/dataplane/owned_deployment.go
alacuku: Same as above.

controller/eventgateway/dataplane/owned_deployment.go
alacuku: It is `keg`. I mean the env variables are set for and used by `keg binary`.

controller/eventgateway/dataplane/owned_deployment.go
pmalek: Just saying, this func is already more than 100s lines of code.

pmalek (approved):

PR #3904, opened 2026-04-15
feat(mcp): support `KongRoute` and `KongService` ID

mlavacca:

In the MCPServer feature, we need to support setting KonnectID a priori, as that's supported by Konnect. With this change, we introduce the KonnectID in the `KongService` and `KongRoute` specs and set it when creating the CRs.

Alongside, the `MCPServer` naming changes and gets rid of the `MCPServer` name as that's a mutable field in Konnect. The resource name is now part of the status and printed as an additional column.

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:


api/konnect/v1alpha1/konnect_mcpserver_types.go
```diff
@@ -91,10 +92,10 @@ type MCPServerStatus struct {
 
 	konnectv1alpha2.KonnectEntityStatusWithControlPlaneRef `json:",inline"` //nolint:embeddedstructfieldcheck
 
-	// Workloads defines the observed state of the MCPServer workloads.
+	// KonnectSpec holds MCPServer-specific status fields related to its state on Konnect, such as the remote name and version.
 	//
 	// +optional
-	Workloads MCPServerWorkloadsStatus `json:"workloads,omitempty"`
+	KonnectSpec *MCPServerKonnectSpec `json:"konnectSpec,omitempty"`
```
randmonkey, line 98: Why did the PR changed `MCPServerWorkloadsStatus` to `MCPServerKonnectSpec`? And also should the change be in another PR that touches the changes on `MCPServer` CRD?

randmonkey (commented):
Generally LGTM, a small comment.
But I have a question why `MCPServer` requires configuring IDs of services and routes, and does that apply to other resources?

pmalek (changes requested):
This would make a precedent for setting Konnect ID in api spec fields. Isn't that a different way of supporting "mirroring"?

According to https://github.com/Kong/kong-operator/blob/54a954cfb4a048e9419ef26a83f97aa3188b66fc/controller/mcpserver/owned_kong_entities.go#L23-L26 we're not creating these objects but we're [obtaining them from Konnect](https://github.com/Kong/kong-operator/blob/54a954cfb4a048e9419ef26a83f97aa3188b66fc/controller/mcpserver/owned_kong_entities.go#L37-L40).

To me this looks like another opportunity to add support for mirroring for route and service.

This comes purely from a concern that there will be 2 ways of achieving the same thing depending on the type and what it supports rather than unifying the experience across types.

mlavacca:
> This would make a precedent for setting Konnect ID in api spec fields. Isn't that a different way of supporting "mirroring"?
> 
> According to
> 
> https://github.com/Kong/kong-operator/blob/54a954cfb4a048e9419ef26a83f97aa3188b66fc/controller/mcpserver/owned_kong_entities.go#L23-L26
> we're not creating these objects but we're [obtaining them from Konnect](https://github.com/Kong/kong-operator/blob/54a954cfb4a048e9419ef26a83f97aa3188b66fc/controller/mcpserver/owned_kong_entities.go#L37-L40).
> 
> To me this looks like another opportunity to add support for mirroring for route and service.
> 
> This comes purely from a concern that there will be 2 ways of achieving the same thing depending on the type and what it supports rather than unifying the experience across types.

Nope, this is NOT another way of creating a mirror entity. With this PR we are enabling the setting for the ID during the creation of routes and services. Konnect allows to create them by setting an arbitrary UUID, and in case a resource with the same UUID already exists, the creation fails (the exact same behavior that we have when creating a resource with a duplicated name).

This change is needed because in the `MCPServer` initiative, we get the UUID from Konnect to use for a `KongRoute` and a `KongService`, as Konnect needs to know what's the UUID of those resources beforehand.

pmalek:
> This change is needed because in the `MCPServer` initiative, we get the UUID from Konnect to use for a `KongRoute` and a `KongService`, as Konnect needs to know what's the UUID of those resources beforehand.

Alright, so essentially when MCP is created in Konnect API it preallocates IDs for route and service to use? And then a user (operator or other tool) is supposed to created these with those IDs specifically?

Can we document this explicitly as it's a rather non obvious behavior and https://github.com/Kong/kong-operator/blob/54a954cfb4a048e9419ef26a83f97aa3188b66fc/controller/mcpserver/owned_kong_entities.go#L23-L24 does not give it justice I believe:

> // ensureKongEntities fetches the Kong entities (services, routes) associated
> // with the given MCPServer from Konnect ...

mlavacca:
> > This change is needed because in the `MCPServer` initiative, we get the UUID from Konnect to use for a `KongRoute` and a `KongService`, as Konnect needs to know what's the UUID of those resources beforehand.
> 
> Alright, so essentially when MCP is created in Konnect API it preallocates IDs for route and service to use? And then a user (operator or other tool) is supposed to created these with those IDs specifically?
> 

Correct

> Can we document this explicitly as it's a rather non obvious behavior and
> 
> https://github.com/Kong/kong-operator/blob/54a954cfb4a048e9419ef26a83f97aa3188b66fc/controller/mcpserver/owned_kong_entities.go#L23-L24
> does not give it justice I believe:
> 
> > // ensureKongEntities fetches the Kong entities (services, routes) associated
> > // with the given MCPServer from Konnect ...

Yep, that makes sense. Improved godoc in 48af7ad2e15f44188c1858277f7a2a8a7b9904a3

api/konnect/v1alpha1/konnect_mcpserver_types.go
mlavacca: I've changed the name as we needed to accomodate different dataset and the old name was not suitable anymore. This API is in progress and this kind of changes are still possible before its release.

I've added a `CHANGELOG` entry related to this change as well.

pmalek (approved):

Issue #3773, opened 2026-04-02
Reconciler code generation: DCRProvider Entity


tao12345666333:
### Problem statement

DCRProvider is a root entity exposed via `/v2/dcr-providers`. Per the design document (#3356), once the reconciler generation infrastructure is in place (#3767), adding a new root entity should involve:
1. Adding `reconciler: isRoot: true` to the entity's `config.yaml` entry
2. Implementing SDK CRUD operations (or generating them if ops generation is available)
3. Manual entity registration in constraints, ops, watch, controller setup, and scheme

This issue tracks adding DCRProvider reconciler support.

### Acceptance criteria

- [ ] DCRProvider has `reconciler` config in `config.yaml` and generated wiring files
- [ ] SDK CRUD operations (create, update, delete, getForUID) are implemented or generated
- [ ] Entity is registered in controller setup and reconciles correctly
- [ ] Tests are added (CRD validation, unit tests)

### Out of scope

- CRD generation (tracked separately under #2839 sub-issues)
- Envtest lifecycle tests (tracked separately)

pmalek:
DCRProvider is in v2 dev portal API.

KO is working on implementing v3.

This should not be needed for the 2.2 milestone as far as the planning goes. Closing.

PR #3928, opened 2026-04-16
feat(crd-from-oas): Support root-union sdkops generation

tao12345666333:

This PR adds support for root-union (oneOf) schema in the crd-from-oas SDK ops code generator, which is required for entities like `DcrProvider` whose API spec uses a discriminated union at the root level.

This is needed for follow-up work on entities such as `DcrProvider`, where the CRD/API spec already models variant-specific config, but we still need generated SDK request conversion logic for create/update reconciliation paths.

**Which issue this PR fixes**

part of #3773 

The helper renaming is intentional to avoid generated symbol collisions across entities.

**PR Readiness Checklist**:


crd-from-oas/pkg/generator/generator.go
```diff
@@ -1048,6 +1074,82 @@ func (g *Generator) generateSDKOps(entityName string, schema *parser.Schema, ops
...
+
+	variants := make([]sdkOpsRootUnionVariant, 0, len(schema.OneOf))
+	for i, variant := range schema.OneOf {
+		variantRefName := variant.Name
+		if variant.RefName != "" {
+			variantRefName = variant.RefName
+		}
+		updateVariantTypeName := ""
+		updateConstructorName := ""
+		if configProp := findPropertyByName(variant.Properties, "dcr_config"); configProp != nil && configProp.RefName != "" {
```
pmalek: This has `dcr_config` hardcoded here so I understand this won't work for other types.

api/x-konnect/v1alpha1/zz_generated_portal_sdkops.go
pmalek, line 1: I'm having hard time understanding what's the result of this PR apart from these functions being renamed.

We already have DCR config with discriminated union types in CRD definition: https://github.com/Kong/kong-operator/blob/8927e15fa7c9666c5b563f9f5f78949c0ffe38a2/api/x-konnect/v1alpha1/zz_generated_dcrprovider_types.go#L87-L115.

🤔

crd-from-oas/pkg/generator/generator.go
tao12345666333: Good catch 💯 

I will modify this part to automatically derive the corresponding parameters, making it more versatile.

api/x-konnect/v1alpha1/zz_generated_portal_sdkops.go
tao12345666333: This is what I've prepared for #3929. It may seem like just a renaming, but by renaming it, I can make it support multiple different types of ops.

While CRDs do have definitions related to discriminate union types, reconcile part requires corresponding operations. I will revise the PR description to make it clearer.

pmalek: Oh OK that makes sense. The description will be useful.

crd-from-oas/pkg/generator/templates.go
```diff
@@ -403,6 +403,231 @@ func (s *{{$.EntityName}}APISpec) {{.MethodName}}() (*{{.ImportAlias}}.{{.TypeNa
...
+		configData, err := json.Marshal(configPayload)
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal {{$.EntityName}} {{.UpdatePayloadJSONName}} payload: %w", err)
+		}
+		var member {{$importAlias}}.{{.UpdateVariantTypeName}}
+		if err := json.Unmarshal(configData, &member); err != nil {
+			return nil, fmt.Errorf("failed to unmarshal into {{.UpdateVariantTypeName}}: %w", err)
+		}
+		unionValue := {{$importAlias}}.{{.UpdateConstructorName}}(member)
+		target.DcrConfig = &unionValue
```
pmalek: Shouldn't this hardcode be fixed as well?

tao12345666333: Thank you! I missed this one. Let me fix

tao12345666333: fixed https://github.com/Kong/kong-operator/pull/3928/changes/c022587f6c5e0efebea2925d94d50944624be465

pmalek (approved):

Issue #2630, opened 2025-11-17
Remove code from KonnectExtension reconciler which uses SDK to handle `KongDataPlaneClientCertificate`


pmalek:
### Problem statement

https://github.com/Kong/kong-operator/issues/1294 introduced `KongDataPlaneClientCertificate` to `KonnectExtension` reconciler.

It did so, in a non breaking manner:

- if users already had data plane client certificates in Konnect for a particular CP, it was reused
- if not, then `KongDataPlaneClientCertificate` has been created

The reason for this was to prevent breakage or re-creating the certificates during upgrades.

This task tracks removal of Konnect SDK usage in `KonnectExtension` reconciler in one of the next releases to allow users to start using the new reconciler logic (and thus managed `KongDataPlaneClientCertificate` CRs) and only then have the old logic (direct SDK manipulation) removed.

PR #3949, opened 2026-04-17
fix(konnect): adopt shared client-cert CR instead of hot-looping Konnect List API

### LINKED ISSUE #3940: KonnectExtensions sharing the same DP client certificate Secret trigger a hot loop of GET /dp-client-certificates calls to Konnect
## Current Behavior

When two (or more) `KonnectExtension` resources in the same namespace share the **same** client-certificate `Secret` (via `spec.clientAuth.certificateSecret.secretRef.name`), the operator enters a permanent hot loop that hammers the Konnect control-plane API:

```
GET /v2/control-planes/{cpID}/dp-client-certificates
```

One of the extensions (the one that "loses" the race to create the `KongDataPlaneClientCertificate` CR) performs this call on **every** reconciliation — by default every minute (`--konnect-sync-period`, `consts.DefaultKonnectSyncPeriod = time.Minute`). Over time this easily accumulates into a very large number of redundant requests against the Konnect API.

### Why this happens

In `controller/konnect/konnectextension_controller.go` (`Reconcile`, around lines 461–565 on `main`):

1. The reconciler lists in-cluster `KongDataPlaneClientCertificate`s using the index `IndexFieldKongDataPlaneClientCertificateOnKonnectExtensionOwner` filtered by `ext.Name`. The index (`internal/utils/index/kongdataplanecertificate.go`, `kongDataPlaneCertificateOnKonnectExtensionOwner`) only returns CRs whose `ownerReferences` include a `KonnectExtension` matching the current extension name.
2. `mappedIDs` is populated **only** from CRs owned by the currently-reconciling extension.
3. When `len(mappedIDs) == 0`, the reconciler falls back to calling `ops.ListKongDataPlaneClientCertificates(...)` (→ `sdk.ListDpClientCertificates(ctx, cpID)` in `controller/konnect/ops/ops_kongdataplaneclientcertificate.go`), i.e. `GET /v2/control-planes/{cpID}/dp-client-certificates`.
4. The reconciler then attempts `r.Create(ctx, &dpCert)` where the name is set to `certificateSecret.Name` (see `pkg/utils/konnect/resources/kongdataplaneclientcertificate.go`, `GenerateKongDataPlaneClientCertificate`, `ObjectMeta.Name: name`). Because both extensions share the same secret name in the same namespace, the `Create` returns `AlreadyExists`, which is silently swallowed (debug log only).
5. Reconciliation returns `ctrl.Result{Requeue: true, RequeueAfter: r.SyncPeriod}` (default 1 min).
6. Next reconcile: `mappedIDs` is still 0 for the "losing" extension (it never owns the CR), goto step 3.

Result: steady-state 1 call / minute / affected extension.

The comment on the fallback block explicitly flags this as temporary migration code that should go away as part of #2630:

```go
// If there are no mapped IDs from KongDataPlaneClientCertificates in cluster,
// then let's use the SDK to query Konnect directly to find any existing certificates.
// This is just to make sure that users migrating from older versions of the operator
// where the dataplane client certificates were not managed using
// KongDataPlaneClientCertificate CRs don't end up creating duplicate
// certificates in Konnect.
// TODO: https://github.com/Kong/kong-operator/issues/2630
// remove this block in future major release after several operator releases.
```

## Expected Behavior

Two `KonnectExtension`s sharing the same client-certificate `Secret` should be a supported configuration (valid when two different Konnect control planes use the same mTLS client cert for DP auth). In that case the operator must **not** call `ListDpClientCertificates` on every reconcile of the "losing" extension.

Any of the following would fix it:

1. **Primary:** complete #2630 and remove the SDK fallback entirely.
2. **Short-term:** broaden the in-cluster lookup so that an extension can match an existing `KongDataPlaneClientCertificate` by cert content / `controlPlaneRef`, not only by `ownerReference` to itself — and add itself as an additional owner (or reuse the CR via status/annotation) when found.
3. When `Create` returns `AlreadyExists`, don't just log-and-continue — fetch the existing CR, ensure it covers the current extension's CP (or adopt / attach an extra owner), and skip further Konnect List calls.
4. Generate a CR name that is unique per `(secretName, extensionName, cpRef)` instead of using `certificateSecret.Name` verbatim — avoids the name collision.

## Steps To Reproduce

1. In a namespace, provision two `KonnectGatewayControlPlane`s mirroring two distinct existing Konnect control planes: one with `clusterType: ControlPlane` (for a standalone `DataPlane`), one with `clusterType: K8SIngressController` (for a Gateway-API-driven deployment).
2. Provision a single `Secret` containing a valid `tls.crt` / `tls.key` client certificate.
3. Create two `KonnectExtension`s whose `spec.clientAuth.certificateSecret.secretRef.name` points to the **same** `Secret` (with `provisioning: Manual`), each referencing a different `KonnectGatewayControlPlane` from step 1.
4. Consume each `KonnectExtension` from a `DataPlane` / `Gateway` so both get reconciled normally.
5. Observe:
   - Only **one** `KongDataPlaneClientCertificate` gets created (named after the shared secret), with `ownerReferences` pointing at whichever extension reconciled first.
   - Both extensions still report `DataPlaneCertificateProvisioned = True` / `Ready = True`, so conditions give no signal that anything is wrong.
   - On the Konnect side (or via debug logs / tcpdump on the Konnect endpoint), a `GET /v2/control-planes/{cpID}/dp-client-certificates` request fires roughly every `--konnect-sync-period` (default 1 min) for the CP of the "losing" extension, indefinitely.

## Operator Version

Reproduced on `main`. The problematic code path was introduced in #2631 when the `KonnectExtension` reconciler was taught to manage `KongDataPlaneClientCertificate` CRs; related cleanup is tracked in #2630.

## `kubectl version`

N/A — not version-dependent; reproducible on any supported Kubernetes minor version.

sandromodarelli:

When two `KonnectExtension` resources in the same namespace reference the **same** client-certificate `Secret` (a valid configuration — two Konnect control planes sharing one client cert), the operator enters a permanent hot loop against Konnect's `GET /v2/control-planes/{cpID}/dp-client-certificates` API. Each reconcile of the "losing" extension falls back to the SDK List path (because its owner-filtered index lookup returns no `KongDataPlaneClientCertificate` CR) and retries `r.Create(...)` with a name equal to the Secret name, which collides with the CR already created by the "winning" extension. The `AlreadyExists` error is silently swallowed and the loop repeats every reconcile.

On top of this, both extensions fight over an annotation on the shared Secret (`konnect.konghq.com/certificate-ids`), re-enqueueing each other. In a live repro this amounted to ~25 `KonnectExtension` reconciles/sec and ~6–7 `ListDpClientCertificates` calls/sec per cluster — enough to produce millions of Konnect API calls over typical operating windows.

This PR replaces the two "Create + swallow AlreadyExists" sites in `controller/konnect/konnectextension_controller.go` with a helper that, on `AlreadyExists`:
1. fetches the existing CR,
2. verifies the cert content matches (otherwise surfaces a clear `DataPlaneCertificateProvisioned=False` condition with a new reason `NameConflict`, and stops tight-requeuing), and
3. adopts the CR by adding the reconciling `KonnectExtension` as an **additional, non-controller** ownerReference.

After the one-time adoption update, the owner index `kongDataPlaneCertificateOnKonnectExtensionOwner` returns the CR for both extensions, `mappedIDs` is populated on every subsequent reconcile, the SDK List fallback is never re-entered, and the annotation ping-pong disappears as a side-effect (both extensions compute identical `mappedIDs`). `mappedIDs` is also now sorted before joining into the annotation so its value is stable regardless of slice order.

### Upgrade safety

- **Single-extension installs (the common case):** `Create` never returns `AlreadyExists`, the new adoption path is never entered, behaviour is bit-for-bit identical.
- **Multi-extension installs sharing a Secret (the bug case):** on the losing extension's first post-upgrade reconcile, one adoption `Update` adds its ownerReference to the existing CR. No new `KongDataPlaneClientCertificate` CR is created, no new Konnect entity is created, no spec/status changes beyond one extra ownerReference.
- No CRD, API, or schema changes.

### End-to-end verification

Reproduced in a kind cluster with the customer-shaped manifests (two `KonnectExtension`s sharing one client-cert Secret, pointing at two different Konnect CPs). Before the fix, the operator averaged ~25 `KONNECTEXTENSION - reconciling` log lines/sec with ~6–7 `"DataPlane client certificate already exists"` per second (steady state). After rolling the patched image into the same cluster:

- The adoption helper logs `"adopted existing KongDataPlaneClientCertificate by adding this KonnectExtension as an owner"` exactly once.
- Zero subsequent `"already exists"` log lines.
- Reconciles drop to ~2/min (one per extension, driven by `--konnect-sync-period`).
- `KongDataPlaneClientCertificate/repro-dp-cert` carries both extensions in `ownerReferences` (first one `controller: true, blockOwnerDeletion: true`, second non-controller).
- Both extensions remain `Ready=True` / `DataPlaneCertificateProvisioned=True`.

### Tests

- New envtest in `test/envtest/dataplane_konnectextensions_test.go`: two `KonnectExtension`s sharing one Secret, two `DataPlane`s consuming them, two `KonnectGatewayControlPlane`s. Asserts exactly one `KongDataPlaneClientCertificate` is bound to the shared Secret and carries both extensions in `ownerReferences`.
- Full `TestDataPlaneKonnectExtension` suite (all sub-tests, both key types) passes.

**Which issue this PR fixes**

Fixes #3940

- The SDK List fallback itself is not removed — that is still tracked by #2630. This PR only stops the hot loop that results from its interaction with the name-collision on the generated CR.
- `errors.AsType[*apierrors.StatusError]` at the original Create sites is replaced with `apierrors.IsAlreadyExists`, which is the idiomatic check for that one condition.
- The new `DataPlaneCertificateProvisionedReasonNameConflict` condition reason is only surfaced when the existing CR's cert content genuinely does not match the one in the referenced Secret (e.g. a user-managed CR sharing the name).

**PR Readiness Checklist**:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -163,7 +163,7 @@
   adding itself as an additional ownerReference instead of retrying `Create`
   on every reconcile, eliminating redundant calls to Konnect's
   `dp-client-certificates` List API.
-  [#3940](https://github.com/Kong/kong-operator/issues/3940)
+  [#3949](https://github.com/Kong/kong-operator/pull/3949)
 - Fix `ResolvedRefs` status condition on `HTTPRoute` not being updated when a
   referenced `KongPlugin` is deleted in self-managed ControlPlane mode.
   [#3206](https://github.com/Kong/kong-operator/pull/3206)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -157,6 +157,13 @@
 
 ### Fixes
 
+- Fix a hot loop in the `KonnectExtension` reconciler when two
+  `KonnectExtension`s share the same client-certificate `Secret`: the losing
+  extension now adopts the existing `KongDataPlaneClientCertificate` CR by
+  adding itself as an additional ownerReference instead of retrying `Create`
+  on every reconcile, eliminating redundant calls to Konnect's
+  `dp-client-certificates` List API.
+  [#3940](https://github.com/Kong/kong-operator/issues/3940)
 - Fix `ResolvedRefs` status condition on `HTTPRoute` not being updated when a
   referenced `KongPlugin` is deleted in self-managed ControlPlane mode.
   [#3206](https://github.com/Kong/kong-operator/pull/3206)
--- a/api/konnect/v1alpha1/konnect_conditions.go
+++ b/api/konnect/v1alpha1/konnect_conditions.go
@@ -206,6 +206,12 @@ const (
 	// used with the DataPlaneCertificateProvisioned condition type indicating that the
 	// DataPlane client certificate creation in Konnect is in progress.
 	DataPlaneCertificateProvisionedReasonProvisioning = "Provisioning"
+	// DataPlaneCertificateProvisionedReasonNameConflict is the reason
+	// used with the DataPlaneCertificateProvisioned condition type indicating that a
+	// KongDataPlaneClientCertificate with the target name already exists in the namespace
+	// but carries a different certificate than the one referenced by this KonnectExtension,
+	// so it cannot be adopted.
+	DataPlaneCertificateProvisionedReasonNameConflict = "NameConflict"
 )
 
 const (
--- a/controller/konnect/konnectextension_controller.go
+++ b/controller/konnect/konnectextension_controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sort"
 	"strings"
 	"time"
 
@@ -501,6 +502,8 @@ func (r *KonnectExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		}
 		return "", false
 	})
+	// Sort so that the annotation value is stable regardless of list order.
+	sort.Strings(mappedIDs)
 
 	// update the secret annotation with the IDs of the mapped certificates
 	newMappedIDsStr := strings.Join(mappedIDs, ",")
@@ -562,6 +565,8 @@ func (r *KonnectExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
 			}
 			return "", false
 		})
+		// Sort so that the annotation value is stable regardless of list order.
+		sort.Strings(mappedIDs)
 	}
 
 	switch {
@@ -584,34 +589,14 @@ func (r *KonnectExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
 				},
 			)
 
-			err := controllerutil.SetOwnerReference(&ext, &dpCert, r.Scheme(), controllerutil.WithBlockOwnerDeletion(true))
-			if err != nil {
+			if err := controllerutil.SetOwnerReference(&ext, &dpCert, r.Scheme(), controllerutil.WithBlockOwnerDeletion(true)); err != nil {
 				return ctrl.Result{}, err
 			}
 
-			if err := r.Create(ctx, &dpCert); err != nil {
-				if errS, ok := errors.AsType[*apierrors.StatusError](err); ok {
-					if errS.ErrStatus.Reason == metav1.StatusReasonAlreadyExists {
-						log.Debug(logger, "DataPlane client certificate already exists", "name", dpCert.Name, "namespace", dpCert.Namespace)
-					}
-				} else {
-					certProvisionedCond.Status = metav1.ConditionFalse
-					certProvisionedCond.Reason = konnectv1alpha1.DataPlaneCertificateProvisionedReasonKonnectAPIOpFailed
-					certProvisionedCond.Message = err.Error()
-					if res, updated, err := patch.StatusWithConditions(
-						ctx,
-						r.Client,
-						&ext,
-						readyCondition,
-						certProvisionedCond,
-					); err != nil || updated || !res.IsZero() {
-						return res, err
-					}
-					// In case of an error when creating the object in Konnect, we requeue "immediately"
-					// so that the operation is retried without waiting for the resync period.
-					return ctrl.Result{RequeueAfter: ctrlconsts.RequeueWithoutBackoff}, nil
-				}
+			if res, err := r.createOrAdoptDataPlaneClientCertificate(ctx, logger, &ext, &dpCert, certDataStr, &certProvisionedCond, readyCondition); err != nil || !res.IsZero() {
+				return res, err
 			}
+
 			updated, res, err := patch.WithFinalizer(ctx, r.Client, certificateSecret, KonnectCleanupFinalizer)
 			if err != nil || !res.IsZero() {
 				return res, err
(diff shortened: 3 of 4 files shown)


pmalek (commented):
I'll review thoroughly later but I wanted to suggest we consider an alternative approach:

Rather than creating the `KongDataPlaneClientCertificate` matching the `Secret` name (which we now know can conflict if `Secret` is reused) why don't we match the `KongDataPlaneClientCertificate` name to `KonnectExtension`?

This would be a simpler approach to solve this issue as I look at it now ( I might be skipping some important details here ).

cc: @mlavacca WDYT?

sandromodarelli:
> I'll review thoroughly later but I wanted to suggest we consider an alternative approach:
> 
> Rather than creating the `KongDataPlaneClientCertificate` matching the `Secret` name (which we now know can conflict if `Secret` is reused) why don't we match the `KongDataPlaneClientCertificate` name to `KonnectExtension`?
> 
> This would be a simpler approach to solve this issue as I look at it now ( I might be skipping some important details here ).
> 
> cc: @mlavacca WDYT?

I was thinking the same but i'm afraid that this can cause a breaking change and a downtime during operator upgrades. that's why i got an approach like "adopt and existing secret"

pmalek:
> > I'll review thoroughly later but I wanted to suggest we consider an alternative approach:
> > Rather than creating the `KongDataPlaneClientCertificate` matching the `Secret` name (which we now know can conflict if `Secret` is reused) why don't we match the `KongDataPlaneClientCertificate` name to `KonnectExtension`?
> > This would be a simpler approach to solve this issue as I look at it now ( I might be skipping some important details here ).
> > cc: @mlavacca WDYT?
> 
> I was thinking the same but i'm afraid that this can cause a breaking change and a downtime during operator upgrades. that's why i got an approach like "adopt and existing secret"

I think that if we just change the it in Create calls, specifically:

- https://github.com/Kong/kong-operator/blob/e61eda4020ac23402f31df79a533e3f5ca47a427/controller/konnect/konnectextension_controller.go#L572-L592
- https://github.com/Kong/kong-operator/blob/e61eda4020ac23402f31df79a533e3f5ca47a427/controller/konnect/konnectextension_controller.go#L627-L653

then we should be good.

Listing both through SDK and through k8s client are done based on either:
- cp scoped listing and then filtering based on cert content for SDK: https://github.com/Kong/kong-operator/blob/e61eda4020ac23402f31df79a533e3f5ca47a427/controller/konnect/konnectextension_controller.go#L533-L564
- konnect extension name based index: https://github.com/Kong/kong-operator/blob/e61eda4020ac23402f31df79a533e3f5ca47a427/controller/konnect/konnectextension_controller.go#L462-L468

Because of this we should be good and there will not be any downtime or recreation of object during operator upgrade.

mlavacca:
> > > I'll review thoroughly later but I wanted to suggest we consider an alternative approach:
> > > Rather than creating the `KongDataPlaneClientCertificate` matching the `Secret` name (which we now know can conflict if `Secret` is reused) why don't we match the `KongDataPlaneClientCertificate` name to `KonnectExtension`?
> > > This would be a simpler approach to solve this issue as I look at it now ( I might be skipping some important details here ).
> > > cc: @mlavacca WDYT?
> > 
> > 
> > I was thinking the same but i'm afraid that this can cause a breaking change and a downtime during operator upgrades. that's why i got an approach like "adopt and existing secret"
> 
> I think that if we just change the it in Create calls, specifically:
> 
>     * https://github.com/Kong/kong-operator/blob/e61eda4020ac23402f31df79a533e3f5ca47a427/controller/konnect/konnectextension_controller.go#L572-L592
> 
>     * https://github.com/Kong/kong-operator/blob/e61eda4020ac23402f31df79a533e3f5ca47a427/controller/konnect/konnectextension_controller.go#L627-L653
> 
> 
> then we should be good.
> 
> Listing both through SDK and through k8s client are done based on either:
> 
>     * cp scoped listing and then filtering based on cert content for SDK: https://github.com/Kong/kong-operator/blob/e61eda4020ac23402f31df79a533e3f5ca47a427/controller/konnect/konnectextension_controller.go#L533-L564
> 
>     * konnect extension name based index: https://github.com/Kong/kong-operator/blob/e61eda4020ac23402f31df79a533e3f5ca47a427/controller/konnect/konnectextension_controller.go#L462-L468
> 
> 
> Because of this we should be good and there will not be any downtime or recreation of object during operator upgrade.

I lean toward this approach as well.
If we keep the listing by label and not get by name, we should not have any problem during upgrades, as it's backward compatible. We might just want to clean up old secrets if unused, but not necessarily.

sandromodarelli:
superseded by https://github.com/Kong/kong-operator/pull/3961

PR #4005, opened 2026-04-23
fix: break Gateway reconcile storm on multi-listener / multi-parentRef topology

### LINKED ISSUE #4004: Gateway reconcile storm: HTTPRoute 409 conflict retried with no backoff, cascades into Gateway re-enqueue loop
## Current Behavior

FTI-7465

When a Gateway has multiple listeners on the **same port** (e.g. three HTTPS listeners all on port 443) and one or more HTTPRoutes carry **multiple `parentRefs` pointing to distinct `sectionName`s of the same Gateway**, the operator enters a high-rate reconciliation loop that does not converge on its own.

Observable symptoms:

- `httproute-controller` reconciler accumulates a large `result="requeue"` counter in the controller-runtime metrics (`controller_runtime_reconcile_total`). In a local reproduction with 30 external HTTPRoute changes over 66 seconds, **+3,470 instant requeues** were observed within that window.
- `gateway` reconciler fires at hundreds of iterations per minute (same 66 s window: **+2,165 reconciles**).
- Repeated `"gateway accepted"` log lines emitted by `controller/gateway/controller.go` for the same Gateway object with distinct `reconcileID`s.
- ArgoCD (when managing these resources with `selfHeal: true`) oscillates between `Healthy`, `Progressing`, and `Degraded` because each reconciliation patches DataPlane/ControlPlane status fields.

Minimum topology that triggers the issue:

```yaml
# Gateway — three listeners, all port 443
listeners:
- name: https
  port: 443
  protocol: HTTPS
  hostname: "*.example.com"
- name: https-api
  port: 443
  protocol: HTTPS
  hostname: "api.example.com"
- name: https-ui
  port: 443
  protocol: HTTPS
  hostname: "ui.example.com"
---
# HTTPRoute — two parentRefs pointing to distinct sectionNames of the same Gateway
parentRefs:
- name: shared-gateway
  sectionName: https
- name: shared-gateway
  sectionName: https-api
```

## Expected Behavior

1. A 409 conflict on an HTTPRoute status write should be retried with **exponential backoff** (or use `Status().Patch()` targeting only the affected parent entry), not with `ctrl.Result{Requeue: true}` which re-enqueues instantly.
2. The Gateway controller's HTTPRoute watch should filter out **status-only** changes so that KIC's own status writes do not re-trigger Gateway reconciliation.
3. Each Gateway reconcile should reach a stable point where `gatewayStatusNeedsUpdate` returns `false` and no further patches are emitted, breaking the feedback cycle.

## Steps To Reproduce

1. Install Kong Operator 2.1.x on a Kubernetes cluster.
2. Create a `GatewayClass` + `GatewayConfiguration` (v2beta1).
3. Create a `Gateway` with **three listeners on port 443** with distinct hostnames (as above).
4. Create at least one HTTPRoute with **two `parentRefs`** pointing to two distinct `sectionName`s of the same Gateway.
5. Wait for initial reconciliation to complete, then apply a spec change to the HTTPRoute (or let the Kong Admin API config push cycle fire).
6. Observe `controller_runtime_reconcile_total{controller="httproute-controller",result="requeue"}` climbing rapidly.
7. Observe `controller_runtime_reconcile_total{controller="gateway",result="success"}` climbing at 10–100×  the expected idle rate.

## Root Cause Analysis

Three defects interact:

### 1. Primary: no-backoff 409 requeue — `ingress-controller/internal/controllers/gateway/handle_error.go:14`

```go
func handleUpdateError(err error, log logr.Logger, obj client.Object) (ctrl.Result, error) {
    if apierrors.IsConflict(err) {
        debug(log, obj, "Conflict found when updating, retrying")
        return ctrl.Result{Requeue: true}, nil  // instant requeue, no backoff
    }
    ...
}
```

`Requeue: true` with `nil` error causes controller-runtime to immediately re-enqueue the HTTPRoute. Under concurrent writers (multiple reconciler goroutines, multiple HTTPRoutes with multi-sectionName parentRefs), the conflict rate compounds and the requeue storm is self-amplifying.

**Fix**: replace with `ctrl.Result{RequeueAfter: <backoff>}` (or exponential backoff via `controller-runtime/pkg/ratelimiter`), or switch `Status().Update()` to `Status().Patch()` targeting only the specific parent entry in `status.parents[]`.

### 2. Amplifier: HTTPRoute watch fires on status-only changes — `controller/gateway/controller.go:125-127`

```go
Watches(
    &gatewayv1.HTTPRoute{},
    handler.EnqueueRequestsFromMapFunc(r.listGatewaysAttachedByHTTPRoute))
```

No `predicate.GenerationChangedPredicate` or equivalent is applied. Every successful HTTPRoute **status** write (by the embedded KIC) fires `listGatewaysAttachedByHTTPRoute`, which enqueues the attached Gateway for reconciliation. With 3 HTTPRoutes and continuous status writes, this produces a steady stream of Gateway enqueues entirely driven by the operator's own output.

**Fix**: add `predicate.GenerationChangedPredicate` or a custom predicate that ignores status-subresource-only updates (compare `Generation` or use `ResourceVersionChangedPredicate` scoped to spec).

### 3. Amplifier: silent port deduplication in DataPlane Service generation — `pkg/utils/kubernetes/resources/services.go:147`

```go
alreadyUsedPorts := make(map[int32]struct{})
for _, p := range dataplane.Spec.Network.Services.Ingress.Ports {
    if _, ok := alreadyUsedPorts[p.Port]; !ok {
        newPorts = append(newPorts, ...)
        alreadyUsedPorts[p.Port] = struct{}{}
    }
}
```

Three port-443 listeners produce three `DataPlaneServicePort` entries. The Service generator silently deduplicates by port number, keeping only the first. Listeners 2 and 3 are dropped without any warning condition on the Gateway. Any perturbation that causes the Service to be re-evaluated (ELB re-annotation, port reshuffle) can trigger DataPlane patches and further Gateway watch events.

**Fix**: deduplicate by `(port, protocol)` tuple and emit a `ListenerConditionConflicted=True` condition on listeners that collide, instead of silently dropping them.

## Operator Version

2.1.2 (chart 1.2.2). The `handle_error.go` pattern exists in the embedded KIC codebase.

## `kubectl version`

```
Client Version: v1.32.x
Server Version: v1.32.x (Kind / EKS)
```

sandromodarelli:

Breaks the self-sustaining Gateway reconcile storm that occurs when a Gateway has multiple listeners on the same port (e.g. three HTTPS listeners on 443) and HTTPRoutes carry multiple `sectionName`-scoped `parentRefs`.

The fix:

- The HTTPRoute watches on the Gateway controller had no predicate. The embedded KIC writes HTTPRoute parent statuses on every sync; each write re-enqueued all attached Gateways. Added `GenerationChangedPredicate` so only spec changes (generation bump) trigger Gateway reconciliation.

**Which issue this PR fixes**

Fixes #4004

`GenerationChangedPredicate` passes Create and Delete events unchanged — only Update events are filtered. The Gateway controller computes `attachedRoutes` and listener conditions from HTTPRoute spec (parentRefs), not from HTTPRoute status, so ignoring status-only updates is safe.

**PR Readiness Checklist**:


ingress-controller/internal/controllers/gateway/handle_error.go
```diff
@@ -9,12 +9,12 @@ import (
...
 // an object or its status.
-// If the error is a conflict, it will return a requeue result.
-// Otherwise, it will return the provided error.
+// If the error is a conflict, it returns the error so the controller-runtime
+// work queue rate limiter applies exponential backoff before the next retry.
+// Otherwise, it returns the provided error as-is.
 func handleUpdateError(err error, log logr.Logger, obj client.Object) (ctrl.Result, error) {
 	if apierrors.IsConflict(err) {
 		debug(log, obj, "Conflict found when updating, retrying")
-		return ctrl.Result{Requeue: true}, nil
```
pmalek, line 17: I understand the appeal of using the exponential backoff here but whenever possible, we've been trying to not return the errors on conflicts because those are actually not user facing errors: these are purely internal to the operators stemming from the fact that controller had stale view of the cluster.

I'd be happy to hear @Kong/k8s-maintainers 's thoughts on this.

I think we can do either: 
- The proposed approach (return error) and be OK with those being logged and users asking questions about "the errors"
- Use a constant time with `RequeueAfter`
- Use a combination of either of the above and change `Update`s into `Patch`es wherever `handleUpdateError` is used

randmonkey (dismissed):

ingress-controller/internal/controllers/gateway/handle_error.go
randmonkey: It could be better if the exponential backoff is used so I think the proposed approach is a preferred one. 
Using a constant `RequeueAfter` is hard to decide the interval. 
Changing `Update` to `Patch` looks like to be a long time improvement but not a quick fix, because in the `handleUpdateError` itself we cannot do it and it was in the legacy KIC code, it is not easy to change all `update`s to `patch`es.
Update: I asked Gemini that Returning `{ Requeue:true }, nil` also triggers the exponential backoff. But I have not confirmed it yet. If the requeueing logic is the same as an error is returned, the fix is not considered an effective one.

sandromodarelli: let me say that the immediate requeue is an amplifier of the main issue (solved by stop watching for the httproute's status) so for the scope of the PR i can revert this change. WDYT?

pmalek: Makes sense, let's do that 👍

programmer04 (dismissed):

ingress-controller/internal/controllers/gateway/handle_error.go
randmonkey: Let's merge this first and discuss how to configure the requeueing later.

sandromodarelli: i updated the PR reverting the requeue, updating the changelog and PR description accordingly

randmonkey (dismissed):

ingress-controller/internal/controllers/gateway/handle_error_test.go
```diff
@@ -0,0 +1,59 @@
...
+			inputErr:   nil,
+			wantResult: ctrl.Result{},
+			wantErr:    nil,
+		},
+		{
+			// A 409 Conflict must be returned as an error so the controller-runtime
+			// work queue rate limiter applies exponential backoff before the retry.
+			// Returning Requeue:true with nil bypasses the rate limiter and causes
+			// an instant retry storm under concurrent writers.
+			name:       "conflict error returned as error for rate-limited retry",
```
pmalek: This is not accurate now after the recent change in this PR.

pmalek (approved):

PR #3908, opened 2026-04-15
feat(crd-from-oas): Add KonnectEventDataplaneCertificate wiring

tao12345666333:
**What this PR does / why we need it**:

**Which issue this PR fixes**

Fixes https://github.com/Kong/kong-operator/issues/3890

**Special notes for your reviewer**:

**PR Readiness Checklist**:

Complete these before marking the PR as `ready to review`:

- [ ] the `CHANGELOG.md` release notes have been updated to reflect significant changes

PR #4010, opened 2026-04-23
perf: add eager initialization of generated type reference handlers

### LINKED ISSUE #3927: perf: Avoid per-reconciliation slice allocation in `generatedTypeReferenceHandlers()`
Currently generatedTypeReferenceHandlers() in controller/konnect/reconciler_generic_generated_refs.go allocates a new []generatedReferenceHandler slice on every reconciliation call.

With only one handler this is negligible, but as we add more generated types, we should cache the slice (e.g., build it once on struct initialization or use a package-level variable) to avoid unnecessary allocations per reconcile loop.

>This is very neat 👍 

>We might need to tweak this a little bit as we add more types (we don't want to allocate the slice of handlers for each reconciliation) but overall a nice concept which is extendible.

_Originally posted by @pmalek in https://github.com/Kong/kong-operator/pull/3908#discussion_r3094586342_

Hangovers:

This PR moves the initialization of generated reference handlers inside the `KonnectEntityReconciler` constructor function instead of the generation inside the handler in `handleGeneratedTypeReferences()` in `reconciler_generic_generated_refs.go`.
It also adds guardrails against `KonnectEntityReconciler` objects initialized using literals instead of `NewKonnectEntityReconciler()`.

Improvement suggested by @pmalek in PR #3908 

**Which issue this PR fixes**

fixes #3927 

**PR Readiness Checklist**:

### DIFF
--- a/controller/konnect/reconciler_generic.go
+++ b/controller/konnect/reconciler_generic.go
@@ -51,7 +51,8 @@ type KonnectEntityReconciler[T constraints.SupportedKonnectEntityType, TEnt cons
 	LoggingMode       logging.Mode
 	SyncPeriod        time.Duration
 
-	MetricRecorder metrics.Recorder
+	MetricRecorder       metrics.Recorder
+	generatedRefHandlers []generatedReferenceHandler[TEnt]
 }
 
 // KonnectEntityReconcilerOption is a functional option for the KonnectEntityReconciler.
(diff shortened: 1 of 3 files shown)


Hangovers:
The guardrail in the handler actually seem to be introducing a concurrency issue since multiple goroutines can run the check simultaneously and write to slice header. 

Since this is supposed to be a performance improvement and the code is providing a way to construct the object using a function it wouldn't make sense to make the current guardrail safe to be used in a concurrent environment;

I think that it might be better to drop the guardrail and add something like that to prevents misuses in the future:

```go
if r.generatedRefHandlers == nil {
		return fmt.Errorf("generated reference handlers are nil, make sure to use NewKonnectEntityReconciler to create the reconciler")
	}
```
in `SetupWithManager` function inside the `reconciler_generic.go`.

pmalek:
Thanks for sending the PR, much appreciated 🙇 

This has been reworked in https://github.com/Kong/kong-operator/issues/4043 in a bigger refactor of `KonnectAPIAuthConfiguration` reference handling.

At this point, this change won't be necessary anymore.

PR #4013, opened 2026-04-24
fix(controllers): do not enable TLSRoute controller and index when TLSRoute CRD absent

### LINKED ISSUE #3861: Disable TLSRoute controller and indices when `TLSRoute` CRD does not exist
## Problem Statement

When `TLSRoute` does not exist, the indices `TLSRouteOnBackendServices` and `TLSRouteOnGateways` cannot be initialized so the manager fail to start when `TLSRoute` CRD does not exist. This makes KO 2.2 not compatible with gateway API 1.4 and before. 

## Proposed Solution

Check existence of `TLSRoute` CRD and do not create TLSRoute controller and indices related to `TLSRoute` when the CRD does not exist. When `TLSRoute` CRD is added, a restart of KO is required to enable the controller of `TLSRoute`.

## Additional Information

In KIC, we have used a dynamic CRD controller that watches the CRDs and enable the `TLSRoute` controller when it is added. In KO we are not able to use the dynamic way since the indices can only be added on start of the manager so a restart is required.

## Acceptance Criteria

- [ ] controllers for `TLSRoute` and indices requiring `TLSRoute` are disabled when `TLSRoute` CRD does not exist

randmonkey:

Do not enable the `TLSRoute` controller and index when `TLSRoute` is absent

**Which issue this PR fixes**

Fixes #3861

**PR Readiness Checklist**:

### DIFF
--- a/modules/manager/controller_setup.go
+++ b/modules/manager/controller_setup.go
@@ -89,6 +89,15 @@ func (c *ControllerDef) MaybeSetupWithManager(ctx context.Context, mgr ctrl.Mana
 func SetupCacheIndexes(ctx context.Context, mgr manager.Manager, cfg Config) error {
 	var indexOptions []index.Option
 
+	crdChecker := k8sutils.CRDChecker{
+		Client: mgr.GetClient(),
+	}
+	tlsRouteGVR := schema.GroupVersionResource{
+		Group:    gatewayv1.GroupVersion.Group,
+		Version:  gatewayv1.GroupVersion.Version,
+		Resource: "tlsroutes",
+	}
+
 	if cfg.ControlPlaneControllerEnabled || cfg.GatewayControllerEnabled {
 		indexOptions = slices.Concat(indexOptions,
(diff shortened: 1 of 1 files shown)


pmalek (dismissed):

randmonkey:
@Kong/k8s-maintainers What do you think is the preferred way to go when `TLSRoute` is absent?
- Like this PR, do not enable it when it is absent
- Require it to start the controllers (This makes KO 2.2 not compatible with gateway API 1.4 and earlier.)

pmalek:
> @Kong/k8s-maintainers What do you think is the preferred way to go when `TLSRoute` is absent?
> 
> * Like this PR, do not enable it when it is absent
> * Require it to start the controllers (This makes KO 2.2 not compatible with gateway API 1.4 and earlier.)

The way it's done in this PR I believe makes sense. Potentially we could think about issuing a log when a CRD is missing like in here.

pmalek (approved):

PR #4048, opened 2026-04-28
fix: fix endless reconciliation on not Programmed referenced entities

pmalek:

Fix endless reconciliation when referenced Konnect entities are not yet programmed

- Stop explicitly requeueing dependent Konnect resources when their referenced entity exists but is not yet Programmed.
- Rely on reconciliation triggered by the referenced resource once it becomes ready, which prevents reconcile loops while still updating status conditions correctly.
- Align related handlers and tests with the new behavior, and relax KongTarget UID lookup to avoid over-filtering by upstream during target discovery.

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:

### DIFF
--- a/controller/konnect/reconciler_certificateref.go
+++ b/controller/konnect/reconciler_certificateref.go
@@ -91,7 +91,8 @@ func handleKongCertificateRef[T constraints.SupportedKonnectEntityType, TEnt con
 		); err != nil || !res.IsZero() {
 			return ctrl.Result{}, err
 		}
-		return ctrl.Result{Requeue: true}, nil
+		// Don't requeue. The referenced entity's changes will trigger the reconciliation.
+		return ctrl.Result{}, nil
 	}
 
 	// TODO: make this more generic.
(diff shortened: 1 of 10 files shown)


controller/konnect/ops/ops_kongtarget.go
```diff
@@ -202,9 +202,9 @@ func getKongTargetForUID(
 			// NOTE: only filter on object's UID.
 			// Other fields like name might have changed in the meantime but that's OK.
 			// Those will be enforced via subsequent updates.
-			Tags:                new(UIDLabelForObject(target)),
-			ControlPlaneID:      target.GetControlPlaneID(),
-			UpstreamIDForTarget: upstreamID,
+			Tags:           new(UIDLabelForObject(target)),
+			ControlPlaneID: target.GetControlPlaneID(),
+			// UpstreamIDForTarget: upstreamID,
```
randmonkey: Why is this line commented out but not deleted?

pmalek: Good catch, this was part of my tests.

Fixed in 734f0f25cd84a67611475abccad0f5175572f719

PTAL

alacuku (approved):

PR #4077, opened 2026-04-30
test(conformance): enable HTTPRouteHostnameIntersection for Hybrid Gateway

### LINKED ISSUE #3476: Hybrid Gateway conformance: enable HTTPRouteHostnameIntersection test
## Problem Statement

The `HTTPRouteHostnameIntersection` Gateway API conformance test is currently skipped for Hybrid Gateway mode.
This is a core HTTPRoute test that verifies hostname intersection behavior between Gateway listeners and HTTPRoutes.

Tracked as part of #3134 and #2551.

## Proposed Solution

Investigate and fix the root cause preventing this conformance test from passing with Hybrid Gateway,
then remove it from `skippedTestsForHybrid` in `test/conformance/conformance_test.go`.

## Additional Information

- Test definition: [gateway-api conformance/tests](https://github.com/kubernetes-sigs/gateway-api/tree/main/conformance/tests)
- Feature: `SupportHTTPRoute` (core)
- Skipped in: https://github.com/Kong/kong-operator/pull/2552

## Acceptance Criteria

- [ ] Root cause identified and documented
- [ ] Fix implemented
- [ ] `HTTPRouteHostnameIntersection` removed from `skippedTestsForHybrid`
- [ ] Conformance test passes for Hybrid Gateway in CI

tao12345666333:

fix `HTTPRoute` hostname intersection with `Gateway`  listeners to follow Gateway API semantics.

**Which issue this PR fixes**

Fixes #3476 

**PR Readiness Checklist**:

### DIFF
--- a/controller/hybridgateway/converter/common.go
+++ b/controller/hybridgateway/converter/common.go
@@ -82,6 +82,7 @@ func getHostnamesByParentRef[T gwtypes.SupportedRoute, TPtr gwtypes.SupportedRou
 
 	var err error
 	var hostnames []string
+	routeHostnames := routeHostNamesString(*route)
 
 	listeners, err := refs.GetListenersByParentRef(ctx, cl, route, pRef)
 	if err != nil {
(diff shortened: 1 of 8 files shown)


controller/hybridgateway/converter/common.go
```diff
@@ -109,20 +110,31 @@ func getHostnamesByParentRef[T gwtypes.SupportedRoute, TPtr gwtypes.SupportedRou
...
 				log.Trace(logger, "Found hostname intersection", "listenerHostname", *listener.Hostname, "routeHostname", host, "intersection", intersection)
 				hostnames = append(hostnames, intersection)
 			}
 		}
 	}
 
+	hostnames = lo.Uniq(hostnames)
+	if len(hostnames) == 0 {
+		log.Debug(logger, "No hostname intersection found for ParentRef")
+		return nil, nil
```
randmonkey, line 137: Why return `nil` instead of the original empty slice here?

controller/hybridgateway/utils/hostname.go
```diff
@@ -6,30 +6,83 @@ import "strings"
...
+
+func wildcardMatches(wildcard, hostname string) bool {
+	if wildcard == "*" {
+		return hostname != ""
+	}
+	if !strings.HasPrefix(wildcard, "*.") {
+		return false
+	}
+	suffix := wildcard[1:] // includes leading dot
+	if !strings.HasSuffix(hostname, suffix) {
```
randmonkey:
```suggestion
	return strings.HasSuffix(hostname, suffix) && len(hostname) > len(suffix)
```
This should be shorter and has better readaillity.

controller/hybridgateway/utils/hostname.go
```diff
@@ -6,30 +6,83 @@ import "strings"
...
+func wildcardOverlaps(a, b string) bool {
+	if a == "*" || b == "*" {
+		return true
+	}
+	suffixA := a[1:]
+	suffixB := b[1:]
+	return strings.HasSuffix(suffixA, suffixB) || strings.HasSuffix(suffixB, suffixA)
+}
+
+func moreSpecificWildcard(a, b string) string {
```
randmonkey, line 77: I believe there should be a more readable way to write this.

randmonkey (commented):
It could be confusing in the `HostnameIntersection` because it returns an empty string for no intersection (the case that the route should not match the listener) and empty string means a wildcard hostname `*` in gateway API spec.
How about making it clearer by changing the return value of `HostnameIntersection` to `string, bool` where the second return value is `false` if no intersection?

controller/hybridgateway/converter/common.go
tao12345666333: Since the `getHybridGatewayParents` use `hostnames == nil` as a condition, if we don't return `nil` the parent will not be skipped.

https://github.com/Kong/kong-operator/blob/b6ed4cab381e0c20322739a426251e0715eaf028/controller/hybridgateway/converter/common.go#L59-L62

The behavior here will change. 

I can add a comment here.

randmonkey (approved):

PR #2423, opened 2025-10-09
fix: proper deletion of `KonnectExtension` (and its dependent `Secret`)

programmer04:
**What this PR does / why we need it**:

Please start with reading issue #2414, which includes reproduction steps and some more context.

The root cause is that when `KonnectGatewayControlPlane` is deleted, `KonnectExtension` can't resolve `KonnectAPIAuthConfiguration` to use it to authenticate to Konnect. The chain of references looks like that

```mermaid
  graph TD;
      KonnectExtension-->KonnectGatewayControlPlane-->KonnectAPIAuthConfiguration;
```

When the "middleman"  `KonnectGatewayControlPlane` is missing during deletion of `KonnectExtension` (that is referencing that `KonnectGatewayControlPlane`), the `KonnectAPIAuthConfiguration` can't be resolved (reconciler can't access Konnect API). This PR ensures that `KonnectExtension` has the latest `KonnectAPIAuthConfiguration` name preserved in its status to use it when  `KonnectGatewayControlPlane` is gone.

It's safe to assume that `KonnectAPIAuthConfiguration` is available, because otherwise KO can't talk to Konnect. There is a dedicated issue to make it bulletproof - do not allow deletion before any dependent resources - [#2353](https://github.com/Kong/kong-operator/issues/2353)

**Which issue this PR fixes**

Fixes #2414

**Special notes for your reviewer**:

- a1e8d654fbc8f8be12853a743fbfbe1b8bd3ec00 introduces a test case that reproduces the bug, which ends with 
   ```
   konnect_extension_test.go:308: Deleting KonnectControlPlane
    konnect_entities_test.go:380: Deleting dc521abd-50d0-4074-be7d-7d0ee9e045a8/cp-521f598b and waiting for it gone
    konnect_entities_test.go:385: Waiting for *v1alpha2.KonnectGatewayControlPlane dc521abd-50d0-4074-be7d-7d0ee9e045a8/cp-521f598b to disappear
    konnect_extension_test.go:310: Deleting KonnectExtension
    konnect_entities_test.go:380: Deleting dc521abd-50d0-4074-be7d-7d0ee9e045a8/konnect-extension-dqx4f and waiting for it gone
    konnect_entities_test.go:385: Waiting for *v1alpha2.KonnectExtension dc521abd-50d0-4074-be7d-7d0ee9e045a8/konnect-extension-dqx4f to disappear
    konnect_entities_test.go:385:
[...]

PR #4085, opened 2026-05-04
fix(konnectExtension): remove dead code left over from getKonnectAPIAuthRefNN refactor

alacuku:

Replace the stale block with correct minimal error handling.

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:


controller/konnect/konnectextension_controller.go
```diff
@@ -398,22 +397,14 @@ func (r *KonnectExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
-				Name: ext.Status.Konnect.AuthRef.Name,
-				// ext.Status.Konnect.AuthRef.Namespace is never nil as enforced in the status update.
-				Namespace: *ext.Status.Konnect.AuthRef.Namespace,
-			}
-		} else {
-			// Requeue until the reference becomes valid.
+			// Cannot determine the KonnectAPIAuthConfiguration reference: either the
+			// KonnectExtension was never programmed (Status.Konnect is nil / AuthRef is nil)
+			// and the ControlPlane is also gone. There is nothing to clean up server-side;
+			// allow the finalizer-removal logic to proceed.
```
pmalek: > allow the finalizer-removal logic to proceed

What do we want to allow here? This changes logic to simply return so there's nothing that will happen afterwards if I'm reading this correctly.

controller/konnect/konnectextension_controller.go
```diff
@@ -398,22 +397,14 @@ func (r *KonnectExtensionReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
-			return ctrl.Result{}, err
-		}
 		if cleanup {
-			// In case if KonnectExtension is during deletion and respective KonnectGatewayControlPlane
-			// has been already deleted, take apiAuthRef from the status, because it contains the last
-			// known reference and it is needed to perform all reconciliation steps.
-			apiAuthRef = types.NamespacedName{
-				Name: ext.Status.Konnect.AuthRef.Name,
-				// ext.Status.Konnect.AuthRef.Namespace is never nil as enforced in the status update.
-				Namespace: *ext.Status.Konnect.AuthRef.Namespace,
```
pmalek, line 402: This was specifically added in https://github.com/Kong/kong-operator/pull/2423 and https://github.com/Kong/kong-operator/pull/2852.

Now looking at this I wonder if this was correct 🤔 Can you elaborate a bit more why you think this is not needed and why it's not correct?

alacuku: The same thing is done here: https://github.com/Kong/kong-operator/blob/b36a2a284cddd13cc2273be651c454e4a7422337/controller/konnect/konnectextension_controller_utils.go#L194-L196.

And if it happens that https://github.com/Kong/kong-operator/blob/b36a2a284cddd13cc2273be651c454e4a7422337/controller/konnect/konnectextension_controller_utils.go#L191 returns an error, we prevent de-referencing of the `nil` namespace in  `Namespace: *ext.Status.Konnect.AuthRef.Namespace` if the status in the extension has not been set.

controller/konnect/konnectextension_controller.go
alacuku: If you have a look at the code in the reconcile you will see that the subsequent steps need the `KonnectAPIAuth` to be in place. Why should we proceed at all if we can't find a KonnectAPIAuth reference?
We are in the `cleanup` branch, the KonnectExtension has been marked for deletion, and we can not cleanup anything since the `KonnectAPIAuth` is not found. So it makes sense to just return. The comment is misleading, a leftover from my previous work on this. Let me adjust it.

pmalek (approved):

tao12345666333 (approved):

tao12345666333 (approved):

programmer04 (approved):

Issue #4134, opened 2026-05-06
Reconciler code generation: cross namespace references handling using KonnectReferenceGrant


pmalek:
### Problem statement

Currently generated code does account for cross namespace references in several places, e.g.:

- https://github.com/Kong/kong-operator/blob/bbea967f9e9e8e499ae7b557c7eb36b688bfb7ab/controller/konnect/reconciler_apiauthref_portal.go#L35-L51
- https://github.com/Kong/kong-operator/blob/bbea967f9e9e8e499ae7b557c7eb36b688bfb7ab/controller/konnect/reconciler_apiauthref_eventgateway.go#L32-L37
- https://github.com/Kong/kong-operator/blob/53ae593d1cf09ab5b2aafc0779b71aaa4a6d1522/controller/konnect/reconciler_generic_handle_eventgatewayref.go#L146-L163

This issue tracks adding support for cross namespace references and granting them via `KongReferenceGrant`s.

### Acceptance criteria

- [ ] As a user I can define cross namespace references between the generated CRDs:
  - [ ] Portal and its configuration entities
  - [ ] EventGateway and its configuration entities

pmalek:
This has to be reopened as cross namespace references do not work for e.g. `KonnectEventGateway` https://github.com/Kong/kong-operator/blob/14cb50a43f0f4a8b7b4116897fea096c898d2da2/config/samples/kongreferencegrant_eventgatewayvirtualclusterconsumepolicy_to_eventgatewayvirtualcluster.yaml#L81-L96.

The backing type does not have the namespace field: https://github.com/Kong/kong-operator/blob/afbba16f7532fc8bbc6813eba49135e68f95c0af/api/konnect/v1alpha2/konnect_configuration.go#L3-L14.

This has not been caught because the test.samples target had a bug where it didn't report the errors.

https://github.com/Kong/kong-operator/pull/4838 aims to fix that but the cross namespace references remain something to fix for (possibly) 2.3.

---

What seems to be missing here is "just" the x-ns KonnectAPIAuthConfiguration references.

PR #4144, opened 2026-05-06
feat(crd-from-oas): refactor parent reference handling, generate SetParentID methods and status condition methods

pmalek:

**Which issue this PR fixes**

Part of https://github.com/Kong/kong-operator/issues/4134

**PR Readiness Checklist**:


controller/konnect/reconciler_apiauthref_parent.go
```diff
@@ -16,15 +16,8 @@ import (
 	konnectv1alpha2 "github.com/kong/kong-operator/v2/api/konnect/v1alpha2"
 )
 
-type parentRefAccessor interface {
-	client.Object
-	parentRefGetter
-}
-
 func getAPIAuthConfigurationRefFromParent[
```
alacuku, line 18: This function seems to miss unit tests.

controller/konnect/reconciler_generic_handle_parentref.go
```diff
@@ -92,3 +104,151 @@ func ensureKongReferenceGrantForParentRef[
...
+	ctx context.Context,
+	cl client.Client,
+	ref commonv1alpha1.ObjectRef,
+	namespace string,
+) (ParentTPtr, types.NamespacedName, error) {
+	switch ref.Type {
+	case commonv1alpha1.ObjectRefTypeNamespacedRef:
+		if ref.NamespacedRef == nil {
+			return nil, types.NamespacedName{}, fmt.Errorf("ref.namespacedRef is required when type is namespacedRef")
+		}
```
alacuku, line 227: What about cluster scoped resources? Can they act as parents for other resources?

controller/konnect/reconciler_generic_handle_parentref.go
```diff
@@ -92,3 +104,151 @@ func ensureKongReferenceGrantForParentRef[
...
+		nn := types.NamespacedName{
+			Name:      ref.NamespacedRef.Name,
+			Namespace: namespace,
+		}
+		if ref.NamespacedRef.Namespace != nil && *ref.NamespacedRef.Namespace != "" {
+			nn.Namespace = *ref.NamespacedRef.Namespace
+		}
```
alacuku, line 236: A comment to explain that if the namespace in the `parentRef` field is an empty string, than we are going to use the namespace of the resource that contains the `parentRef` would be much appreciated.

controller/konnect/reconciler_generic_handle_parentref.go
```diff
@@ -92,3 +104,151 @@ func ensureKongReferenceGrantForParentRef[
...
+		consts.ConditionReason(obj.GetStatusConditionReasonParentRefValid()),
+		fmt.Sprintf("Referenced %s %s is programmed", parentType, nn),
+	); errStatus != nil || !res.IsZero() {
+		return res, errStatus
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func getParentForRef[
```
alacuku, line 214: Let's unit test this function in isolation.

controller/konnect/reconciler_generic_handle_parentref.go
```diff
@@ -92,3 +104,151 @@ func ensureKongReferenceGrantForParentRef[
...
+	}
+
+	cond, ok := k8sutils.GetCondition(konnectv1alpha1.KonnectEntityProgrammedConditionType, parent)
+	if !ok || cond.Status != metav1.ConditionTrue {
+		if res, errStatus := patch.StatusWithCondition(
+			ctx, cl, obj,
+			consts.ConditionType(obj.GetStatusConditionTypeParentRefValid()),
+			metav1.ConditionFalse,
+			consts.ConditionReason(obj.GetStatusConditionReasonParentRefNotProgrammed()),
+			fmt.Sprintf("Referenced %s %s is not programmed yet", parentType, nn),
```
alacuku: Nit: For consistency with the logic just a few lines above, we should probably pull this message out into a `msg` variable rather than passing it inline.

alacuku (commented):
Left some comments.

controller/konnect/reconciler_generic_handle_parentref.go
pmalek: 👍 d737c2c43a0c8dd2a9632d9ec3fe908e85da78d0 PTAL

controller/konnect/reconciler_apiauthref_parent.go
pmalek: 👍 d737c2c43a0c8dd2a9632d9ec3fe908e85da78d0 PTAL

controller/konnect/reconciler_generic_handle_parentref.go
pmalek: Done. PTAL

controller/konnect/reconciler_generic_handle_parentref.go
pmalek: Currently we're not supporting any cluster scoped resources that could be parents (in generated or not generated CRDs).

When we get to a point where we need that it will require a whole refactor on its own hence I'd leave it as is for now. WDYT?

controller/konnect/reconciler_generic_handle_parentref.go
pmalek: 👍 cd39c0f59d4227166b2fc7b96811278280221918

controller/konnect/reconciler_generic_handle_parentref.go
alacuku: 👍

alacuku (approved):

PR #4182, opened 2026-05-08
fix: complete ListenerPolicy reconciliation generation

### LINKED ISSUE #4055: Reconciler code generation: ListenerPolicy Entity
## Problem Statement

ListenerPolicy CRD generation is tracked in #3931, but there is no corresponding reconciler-tracking issue yet.

Once the CRD exists, the resource still needs generated reconciliation support so it can be created, updated, deleted, and adopted through the Konnect controller stack.

This issue tracks adding reconciler generation support for ListenerPolicy.

## Proposed Solution

Add reconciler generation for ListenerPolicy by:

1. Adding the entity's `reconciler` and `ops` configuration to `crd-from-oas/config.yaml`
2. Generating the entity's watch, index, SDK ops, and controller setup wiring
3. Adding any type-specific parent or listener reference handling required by the Event Gateway API shape
4. Adding tests for the generated reconciler behavior

## Additional Information

- Parent issue: #2839
- CRD generation issue: #3931
- Similar Event Gateway reconciler issues: #3774, #3890

## Acceptance Criteria

- ~~Documentation is updated with this new feature / enhancement~~ (not user-facing yet)
- [ ] ListenerPolicy has reconciler config in `crd-from-oas/config.yaml`
- [ ] Generated reconciler wiring files are produced for ListenerPolicy
- [ ] SDK CRUD operations are implemented or generated for ListenerPolicy
- [ ] Any required Event Gateway reference handling is implemented or generated
- [ ] ListenerPolicy is registered in controller setup and reconciles correctly
- ~~Helm chart is updated to work with this feature / enhancement~~ (CRD work is tracked separately)
- [ ] As a user, I can declaratively manage ListenerPolicy resources in Konnect through Kong Operator

tao12345666333:

**Which issue this PR fixes**

Fixes #4055

**PR Readiness Checklist**:

### DIFF
--- a/controller/konnect/ops/ops_eventgatewaylistenerpolicy_manual.go
+++ b/controller/konnect/ops/ops_eventgatewaylistenerpolicy_manual.go
@@ -1,0 +1,81 @@
+package ops
+
+import (
+	"context"
+	"fmt"
+
+	sdkkonnectgo "github.com/Kong/sdk-konnect-go"
+	sdkkonnectops "github.com/Kong/sdk-konnect-go/models/operations"
+
+	konnectv1alpha1 "github.com/kong/kong-operator/v2/api/konnect/v1alpha1"
+)
+
+func getEventGatewayListenerPolicyForUID(
+	ctx context.Context,
+	sdk sdkkonnectgo.EventGatewayListenerPoliciesSDK,
+	obj *konnectv1alpha1.EventGatewayListenerPolicy,
+) (string, error) {
+	gatewayID := obj.GetGatewayID()
+	if gatewayID == "" {
+		return "", CantPerformOperationWithoutParentIDError{Entity: obj, Parent: "Gateway", Op: GetOp}
+	}
+	eventGatewayListenerID := obj.GetEventGatewayListenerID()
+	if eventGatewayListenerID == "" {
+		return "", CantPerformOperationWithoutParentIDError{Entity: obj, Parent: "EventGatewayListener", Op: GetOp}
+	}
+
+	policyType, policyName := eventGatewayListenerPolicyIdentity(obj)
+	if policyType == "" || policyName == "" {
+		return "", EntityWithMatchingUIDNotFoundError{Entity: obj}
+	}
+
+	resp, err := sdk.ListEventGatewayListenerPolicies(ctx, sdkkonnectops.ListEventGatewayListenerPoliciesRequest{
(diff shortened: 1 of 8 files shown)


controller/konnect/ops/ops_eventgatewaylistenerpolicy_manual.go
pmalek, line 1: Is the whole purpose of this file to work around the fact that `resp.GetListEventGatewayListenerPoliciesResponse()` doesn't have a `.Data` field?

If that's the case can we try to 
- either fix it upstream in the OAS
- or make a config knob to account for that and generate code which will handle it this way?

tao12345666333: That makes sense, and I looked in that direction as well.

This is not only about `resp.GetListEventGatewayListenerPoliciesResponse()` not having a `.Data` field. I checked the upstream Event Gateway OAS, and `ListEventGatewayListenerPoliciesResponse` is currently a type: `array`, so the bare-slice shape appears to be the actual spec/SDK output rather than a small operator-side issue we can fix independently upstream. 

https://github.com/Kong/sdk-konnect-go/blob/aebd95d73ac23096679296d1d221d496c3646fd0/models/operations/listeventgatewaylistenerpolicies.go#L40-L49

And this helper handles a bit more ListenerPolicy-specific logic: besides the bare-slice list response, `getForUID` also needs to derive identity from the root-union variant (name/type) and map the CRD-side type to the discriminator values used by the SDK/list response. The current generated `getForUID` path is still built around the paginated `resp.<field>.Data ` shape and does not express this union-based identity matching very well.

So for this PR I kept the hand-written helper as the smallest correct fix.

I do think your 'config knob/generator support' suggestion is a good direction. How about I treat it as a follow-up task instead of including it in the current PR?

pmalek: Fair enough. If that works best for you let's do a follow up issue for this 👍

pmalek (approved):

controller/konnect/ops/ops_eventgatewaylistenerpolicy_manual.go
tao12345666333: created an issue for tracking https://github.com/Kong/kong-operator/issues/4215

PR #4187, opened 2026-05-08
feat: handle parent refs generically

pmalek:

Fix generic parent ref handling by adding necessary functions to CRD types:

- `HasParent() bool` returning `true` for types that have parents
- `GetParentGVK() schema.GroupVersionKind` which are attached to all non root objects

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:

### DIFF
--- a/controller/konnect/reconciler_generic_handle_generated_refs.go
+++ b/controller/konnect/reconciler_generic_handle_generated_refs.go
@@ -6,7 +6,6 @@ package konnect
 
 import (
 	"context"
-	"errors"
 
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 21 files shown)


controller/konnect/reconciler_generic_handle_generated_refs.go
```diff
@@ -62,27 +70,25 @@ func (r *KonnectEntityReconciler[T, TEnt]) handleGeneratedTypeReferences(
...
 		}
 	}
 
+	res, err := handler.handleParentRef(ctx, r.Client, obj)
+
+	stop, res, err := handleRefResult(ctx, r.Client, ent, res, err)
+	if stop || err != nil {
+		return true, res, err
+	}
+
```
tao12345666333, line 101: I think this introduces a regression: we now consider unregistered generated parent classes as serious errors, whereas the previous implementation simply continued to the next handler. Since the generated parent class now includes `EventGatewayListener` will fail here unless that parent type is also registered and supported. WDYT?

pmalek: Agreed, 57bc406c8c8d806301b641f6e2177e5b1a00c81a fixes `EventGatewayListener` parent references.

As for the error itself: I think we're OK returning an error here for the unsupported parents as that will give us a clear signal what's the missing piece. The types that do not implement the `objectWithParentRef` interface will not even get to this place (as for that we return `false, ctrl.Result{}, nil` above).

tao12345666333 (approved):

PR #4199, opened 2026-05-08
ci(nightly): produce nightly and date tags on workflow_dispatch

tao12345666333:

When we trigger the nightly build manually, it will not publish the `nightly` tag. This PR will produce `nightly` and date tags for manually trigger

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:

### DIFF
--- a/.github/workflows/__build-workflow.yaml
+++ b/.github/workflows/__build-workflow.yaml
@@ -74,6 +74,13 @@ on:
           When set to true, specify the slack-webhook-url and slack-team-id secrets.
         default: false
         type: boolean
+      nightly-tags:
+        description: |
+          When true, also produce 'nightly' and YYYYMMDD tags on non-schedule events
+          (e.g. workflow_dispatch). Intended to let the nightly workflow's manual
+          dispatch produce the same tags as its scheduled run.
+        default: false
+        type: boolean
 
     outputs:
       full_tag:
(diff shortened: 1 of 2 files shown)


alacuku (approved):

.github/workflows/__build-workflow.yaml
```diff
@@ -387,11 +398,12 @@ jobs:
...
             ${{ inputs.registry }}/${{ inputs.image-name }}:${{ needs.semver.outputs.fullversion }}-arm64
 
-      # When building on schedule, `steps.meta.outputs.tags` contains multiple entries, so it cannot be used
+      # When building on schedule (or on workflow_dispatch with nightly-tags enabled),
+      # `steps.meta.outputs.tags` contains multiple entries, so it cannot be used
       # directly in sources. Instead, the sources are constructed using the `inputs.registry`, `inputs.image-name` and
       # the current date.
       - uses: int128/docker-manifest-create-action@fa55f72001a6c74b0f4997dca65c70d334905180 # v2.20.0
-        if: ${{ inputs.tag == '' && github.event_name == 'schedule' }}
+        if: ${{ inputs.tag == '' && (github.event_name == 'schedule' || (github.event_name == 'workflow_dispatch' && inputs.nightly-tags)) }}
```
pmalek, line 406: 🤔 Is there a reason we can't just do 


```suggestion
        if: ${{ inputs.tag == '' }}
```

and not add the workflow param?

tao12345666333: The reason I added the `nightly-tags` parameter here is to avoid generating logs related to nightly/YYMMDD during the execution of other workflows, even though they won't actually be pushed to DockerHub.

But I also agree with your suggestion that if we directly remove the param conditions, it can be more concise here.

pmalek: OK that makes sense. Let's proceed as you proposed (no changes needed). 👍

PR #4208, opened 2026-05-11
chore(log): emit error logs when dataplane contains operator managed env vars

### LINKED ISSUE #193: Warnings for overriding controller-managed ENV vars
## Problem Statement

It's possible for end-users to override `ENV` variables for `ControlPlane` and `DataPlane` that are _intended_ to only be provided by the controller, when using `GatewayConfiguration`. Examples include `CONTROLLER_PUBLISH_SERVICE` which the `Gateway` controller expects to manage itself. The purpose of this task is to emit warnings when this happens, so that the user can come to some sort of awareness that this may not be what they intended, or that it may have adverse effects.

## Anything else ?

Generate both logs and events for this.

## Acceptance Criteria

- [ ] when manually setting values in the `GatewayConfiguration` for fields that are not expected to be managed by end-users, warnings are emitted that help the user understand that what they're doing may have unintended side effects, as the controller intends to manage these.
- [ ] a note in docs is added about this

## Notes

Extra context: previously in Kong/gateway-operator-archive#109 we actually suggested blocking these configurations _outright_. Technically we could still do this in the future, but the team decided that for the early alpha phase we would evaluate simply emitting warnings about this.

randmonkey:

Emit error log when `DataPlane` specifies operator managed env variables.

Currently includes:
- `KONG_CLUSTER_CERT`
- `KONG_CLUSTER_CERT_KEY`

**Which issue this PR fixes**

Fixes #193

**PR Readiness Checklist**:

### DIFF
--- a/controller/dataplane/owned_deployment.go
+++ b/controller/dataplane/owned_deployment.go
@@ -299,8 +299,14 @@ func podTemplateSpecHasRestartAnnotation(template *corev1.PodTemplateSpec) (stri
 	return v, ok && v != ""
 }
 
-// envVarExistsInPodTemplateSpec checks if an environment variable with the given name exists in the proxy container of the PodTemplateSpec in the DataPlane.
-func envVarExistsInPodTemplateSpec(ctx context.Context, envVarName string, dataplane *operatorv1beta1.DataPlane, cl client.Client) (bool, error) {
+// envVarExistsInPodTemplateSpec checks if an environment variable with the given name
+// exists in the proxy container of the PodTemplateSpec in the DataPlane.
+func envVarExistsInPodTemplateSpec(
+	ctx context.Context,
+	envVarName string,
+	dataplane *operatorv1beta1.DataPlane,
+	cl client.Client,
+) (bool, error) {
 	templateSpec := dataplane.Spec.Deployment.PodTemplateSpec
 	if templateSpec == nil {
 		return false, nil
--- a/controller/dataplane/owned_deployment.go
+++ b/controller/dataplane/owned_deployment.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/go-logr/logr"
 	"github.com/google/go-cmp/cmp"
-	"github.com/samber/lo"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


controller/dataplane/owned_deployment.go
```diff
@@ -274,6 +296,21 @@ func podTemplateSpecHasRestartAnnotation(template *corev1.PodTemplateSpec) (stri
...
+	for _, container := range template.Spec.Containers {
+		if lo.ContainsBy(container.Env, func(env corev1.EnvVar) bool {
+			return env.Name == envVarName
+		}) {
+			return true
+		}
+	}
+	return false
+}
+
```
pmalek: We can already use `GetPodContainerByName` and `GetEnvValueFromContainer` for this. Why implement yet another way of doing this same thing 🤔 ?

https://github.com/Kong/kong-operator/blob/5dce811c807ddb7a7da11ea65ee4e028ee2f094d/controller/gatewayclass/controller_reconciler_utils.go#L120-L127

controller/dataplane/owned_deployment.go
```diff
@@ -38,6 +39,17 @@ import (
...
+// environment variable keys that are generated from the DataPlane configuration and need to be preserved across user patches.
+const (
+	envKongClusterCert    = "KONG_CLUSTER_CERT"
+	envKongClusterCertKey = "KONG_CLUSTER_CERT_KEY"
+)
+
+var operatorManangedEnvVars = []string{
+	envKongClusterCert,
+	envKongClusterCertKey,
+}
```
pmalek, line 50: Can we make this a `map` so that we don't have to iterate over the slice every time?

controller/dataplane/owned_deployment.go
randmonkey: Copilot did not find it :(


Updated to use the function in `k8sutils`.

controller/dataplane/owned_deployment.go
randmonkey: This does not make sense because we are checking a single env var in cotnainer's envs but not dumping all the envs in the container. So we always need to iterate over the env vars.

pmalek (approved):
This is good 👍 

We can also think about adding an event (not sure what would be a good name for the event type ) but this is already an improvement.

Issue #1743, opened 2025-06-16
☔ Refactor: Use managed fields and server-side apply to track fields owned by operators


alacuku:
## Problem Statement
The current implementation manages resources (deployments, services, etc..) by starting from defaults and applying user customizations via `StrategicMergePatch`. To avoid unnecessary diffs, we fill resources with defaults using internal Kubernetes APIs. However,` mutating webhooks` and other `controllers` can still introduce changes, causing perpetual reconciliations. Some resources use a spec hash to mitigate this.
See:  #128, #500, #239, #888

## Proposed Solution

* Use the structured-merge-diff library for merging and diffing Kubernetes resources, matching Kubernetes Server-Side Apply logic.
* Create and patch managed resources using Server-Side Apply, supporting field ownership.
* Leverage managed fields to clearly delineate field ownership by the gateway-operator.

## Acceptance Criteria
- [ ] Managed resources are created and updated using Server-Side Apply;
- [ ] Field ownership is tracked and respected using managed fields;
- [ ] The operator uses `structured-merde-diff` for merging and diffing resource specs;
- [ ] Reconciliation logic only considers fields owned by the operator, ignoring changes form other controllers, users or webhooks;

PR #4211, opened 2026-05-11
fix(dataplane): support ignoring certain annotations in comparing deployments for DataPlanes

### LINKED ISSUE #4071: Reconciliation of deployments owned by `DataPlane` is never diverged if deployment can be changed by other controllers
## Current Behavior

A customer found that the `DataPlane` never gets `Ready` condition set and `status.replicas` and `status.readyReplicas` is always 0. During the investigation, we found that the `spec.podTemplate` of the created `Deployment` has extra annotations. Then the reconciliation of the `Deployment` is never diverged because it is always different from the generated one.

## Expected Behavior

Can configure ignored fields in comparing of deployments then the reconciliation can diverge in the case.

## Steps To Reproduce

1. Add a mutating webhook that adds an annotation in `spec.podTemplate` of `Deployment`s
2. Create a `Gateway` or `DataPlane`
3. The `Deployment` is created but the `DataPlane` can never set the `Ready` condition

## Operator Version

2.1.2

randmonkey:

Added `ignoredAnnotations` to ignore certain annotation keys in comparing the existing `Deployment` and the desired `Deployment` in reconciling `DataPlane`.

This fixes the issue where there are mutating webhooks or other controllers in the cluster to modify the `Deployment`s, the reconciliation on deployments for `DataPlane`s can never diverge.

**Which issue this PR fixes**

Fixes #4071

**PR Readiness Checklist**:


api/gateway-operator/v1beta1/shared_types.go
```diff
@@ -33,6 +33,12 @@ type DeploymentOptions struct {
 	//
 	// +optional
 	PodTemplateSpec *corev1.PodTemplateSpec `json:"podTemplateSpec,omitempty"`
+
+	// IgnoredAnnotations is a list of annotation keys that the operator will ignore when comparing the spec of the managed Deployment with the desired spec.
+	//  This is useful for annotations that are added by other controllers or processes and should not trigger a rollout when they change.
+	//
+	// +optional
+	IgnoredAnnotations []string `json:"ignoredAnnotations,omitempty"`
```
randmonkey, line 41: Should we add a `MaxItems` and check of format of annotation keys here?

randmonkey:
REVIEW: Should we configure the ignored annotations in `DataPlane` CRD level or KO config level?

pmalek:
I believe the `enforce-config` CLI flag should help mitigate this: https://github.com/Kong/kong-operator/blob/7f394c5a064d8cc323dd75e933e6d47320d08a00/modules/cli/cli.go#L52C1-L53C169.

WDYT?

pmalek:
Based on https://github.com/Kong/kong-operator/pull/4211#issuecomment-4646717244 and discussion on team sync, closing this as enforce-config flag should help here.

Sidenote: there's a parallel work stream (lead by @alacuku) which introduces SSA in controllers: #1743, which would solve this type of issues.

PR #4235, opened 2026-05-12
feat(konnect): support cross-namespace certificateRef in KongSNI

### LINKED ISSUE #4152: CrossNamespace: enable cross-namespace KongSNI → KongCertificate
## Problem Statement

`KongSNI.spec.certificateRef` only supports same-namespace references because the field
is typed as `commonv1alpha1.NameRef`, which has no `namespace` field. The reconciler
(`reconciler_certificateref.go`) carries a `// TODO: handle cross namespace refs`
marker, confirming that cross-namespace support was deferred.

Users cannot place `KongSNI` and `KongCertificate` in separate namespaces, even when
they have a valid `KongReferenceGrant`. This is inconsistent with `KongRoute → KongService`
and `KongCACertificate → Secret`, which both support cross-namespace references via grants.

## Proposed Solution

1. Change `KongSNI.spec.certificateRef` from `NameRef` to `NamespacedRef` so a namespace
   can be specified. Run `make generate` to regenerate CRDs and deepcopy.
2. In `reconciler_certificateref.go`, when `certificateRef.namespace` is set and differs
   from the SNI's namespace, call `crossnamespace.CheckKongReferenceGrantForResource` and
   surface `ResolvedRefs=False/RefNotPermitted` if no grant exists, mirroring the pattern
   used in `reconciler_secretref.go`.
3. In `watch_kongsni.go`, add a `KongReferenceGrant` watch so grant create/update/delete
   re-queues affected SNIs.

## Acceptance Criteria

- [x] As a user, I can place a `KongSNI` and its referenced `KongCertificate` in different namespaces, and the SNI reaches `Programmed=True` when a valid `KongReferenceGrant` exists.
- [x] As a user, when no `KongReferenceGrant` exists, the `KongSNI` surfaces `ResolvedRefs=False/RefNotPermitted` instead of silently failing.
- [x] As a user, creating or deleting a `KongReferenceGrant` triggers immediate re-reconciliation of the affected `KongSNI`.

alacuku:

Allow KongSNI.spec.certificateRef to reference a KongCertificate in a different namespace via KongReferenceGrant.

API changes:
- Change KongSNI.spec.certificateRef from NameRef to NamespacedRef, adding an optional namespace field.
- Add KongSNI to the allowed "from" kinds and KongCertificate to the allowed "to" kinds in KongReferenceGrant webhook validation markers.

Controller changes:
- handleKongCertificateRef: when certificateRef.namespace differs from the SNI namespace, check KongReferenceGrant in the target namespace. Set ResolvedRefs=False/RefNotPermitted and requeue (without backoff) when no grant exists; set ResolvedRefs=True when granted.
- GetAPIAuthRefNN (reconciler_apiauthref.go): use certNamespace (derived from certificateRef.namespace) instead of ent.GetNamespace() when looking up the KongCertificate and resolving its CP auth ref, so the correct namespace is used for cross-namespace cert refs.
- watch_kongsni.go / index/sni.go: index and watch KongSNI by certificateRef using the namespace-aware NamespacedRef field.

Tests:
- reconciler_certificateref_test.go: add two unit test cases covering the cross-namespace path (grant missing → RefNotPermitted, grant present → ResolvedRefs=True + KongCertificateRefValid=True).

Generated: deepcopy, CRD manifests, Helm chart golden files, docs.

**Which issue this PR fixes**

Fixes #4152 

E2E tests implemented in https://github.com/Kong/kong-operator/pull/4239

**PR Readiness Checklist**:

### DIFF
--- a/controller/konnect/reconciler_certificateref.go
+++ b/controller/konnect/reconciler_certificateref.go
@@ -12,26 +12,29 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
 
+	"github.com/kong/kong-operator/v2/api/common/consts"
 	commonv1alpha1 "github.com/kong/kong-operator/v2/api/common/v1alpha1"
 	configurationv1alpha1 "github.com/kong/kong-operator/v2/api/configuration/v1alpha1"
 	konnectv1alpha1 "github.com/kong/kong-operator/v2/api/konnect/v1alpha1"
 	konnectv1alpha2 "github.com/kong/kong-operator/v2/api/konnect/v1alpha2"
+	ctrlconsts "github.com/kong/kong-operator/v2/controller/consts"
 	"github.com/kong/kong-operator/v2/controller/konnect/constraints"
 	"github.com/kong/kong-operator/v2/controller/pkg/controlplane"
 	"github.com/kong/kong-operator/v2/controller/pkg/patch"
+	"github.com/kong/kong-operator/v2/internal/utils/crossnamespace"
 	k8sutils "github.com/kong/kong-operator/v2/pkg/utils/kubernetes"
 )
 
 // getKongCertificateRef gets the reference of KongCertificate.
 func getKongCertificateRef[T constraints.SupportedKonnectEntityType, TEnt constraints.EntityType[T]](
 	e TEnt,
-) mo.Option[commonv1alpha1.NameRef] {
+) mo.Option[commonv1alpha1.NamespacedRef] {
 	switch e := any(e).(type) {
 	case *configurationv1alpha1.KongSNI:
 		// Since certificateRef is required for KongSNI, we directly return spec.CertificateRef here.
 		return mo.Some(e.Spec.CertificateRef)
 	default:
-		return mo.None[commonv1alpha1.NameRef]()
+		return mo.None[commonv1alpha1.NamespacedRef]()
 	}
 }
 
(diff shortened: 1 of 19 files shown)


pmalek (dismissed):

controller/konnect/reconciler_certificateref.go
```diff
@@ -45,11 +48,53 @@ func handleKongCertificateRef[T constraints.SupportedKonnectEntityType, TEnt con
...
+			ctx, cl, ent,
+			consts.ConditionType(configurationv1alpha1.KongReferenceGrantConditionTypeResolvedRefs),
+			metav1.ConditionTrue,
+			configurationv1alpha1.KongReferenceGrantReasonResolvedRefs,
+			"KongReferenceGrants allow access to KongCertificate",
+		); errStatus != nil || !res.IsZero() {
+			return res, errStatus
+		}
+	}
+
```
tao12345666333, line 93: Looks like `ResolvedRefs` can become stale here. If `certificateRef` changes from cross-namespace back to same-namespace, we stop setting the condition but never remove the old one, so status may remain incorrect. Should we clear it in the non-cross-namespace path?

alacuku: Nice catch!

Fixed in a separate commit. PTAL @tao12345666333

pmalek (dismissed):

controller/konnect/reconciler_certificateref.go
alacuku: The fix added an else branch in handleKongCertificateRef and corrected
patch.StatusWithoutCondition to clear a stale ResolvedRefs=False/RefNotPermitted
condition when spec.certificateRef switches from cross-namespace to same-namespace.

Dropping this for now: KongSNI.spec.certificateRef is immutable once
Programmed=True, so the window where this can happen is extremely narrow
(only before the resource is first programmed). The risk of shipping a
known-problematic StatusWithoutCondition fix without broader test coverage
outweighs the benefit for this rare scenario.

A tracking issue will be opened to revisit the StatusWithoutCondition
correctness and the stale-condition clearing in a dedicated PR.

pmalek (approved):

controller/konnect/reconciler_certificateref.go
alacuku: Tracking issue: https://github.com/Kong/kong-operator/issues/4248

PR #4240, opened 2026-05-12
feat(konnect): support `KongCertificate` and `KongCACertificate` refs in `KongService`

### LINKED ISSUE #864: Konnect entities: Implement CA Certificate refs for `KongService` and `KongUpstream`
### Problem statement

Service and Upstream types in Konnect APIs contain a `ca_ceritificates` field which is described as:

```
Array of CA Certificate object UUIDs that are used to build the trust store while verifying upstream server's TLS certificate. If set to null when Nginx default is respected. If default CA list in Nginx are not specified and TLS verification is enabled, then handshake with upstream server will always fail (because no CA are trusted).
```

Given that the operator works in Kubernetes environment, we'd probably want to implement that as a list of references to in cluster resources instead of a verbatim list of UUIDs.

### Acceptance criteria

- [ ] As a user I can set the `ca_certificates` field on `KongService` and `KongUpstream` objects when working with Konnect entities.

sandromodarelli:

Adds `clientCertificateRef` and `caCertificateRefs` fields to `KongService`, allowing users to reference a `KongCertificate` (client TLS cert) and a list of `KongCACertificates` (CA chain) by namespaced ref. Cross-namespace references require a `KongReferenceGrant` in the target namespace, following the same pattern as `KongRoute → KongService`.

Resolved Konnect IDs are stored in `KongService.status.konnect` and mapped into the Konnect API `clientCertificate` and `caCertificates` fields on Create/Update. As a side effect, `KongSNI` gains cross-namespace cert ref support via the shared handler.

**Which issue this PR fixes**

Partially fixes #864 

**Changes**:
- `KongServiceAPISpec`: add `clientCertificateRef` (`NamespacedRef`) and `caCertificateRefs` (`[]NamespacedRef`, `+listType=atomic`)
- New `KonnectEntityStatusWithControlPlaneAndCertificateAndCACertificatesRefs` status struct
- `KongReferenceGrant.spec.to[*].kind` enum: add `KongCertificate` + `KongCACertificate`
- `handleKongCertificateRef`: extended for `KongService` with cross-NS grant check
- `handleKongCACertificateRefs` (new): validates list of CA cert refs; stale IDs cleared when refs removed
- SDK mapping in `kongServiceToSDKServiceInput`
- Indexers and watchers for `KongCertificate` / `KongCACertificate` → `KongService`
- Unit tests, CRD validation tests, sample manifest, regenerated CRDs + chart goldens

`CACertificateIDs` written to status is not assertable at unit-test level (fake client `MergeFrom` patch only carries condition diffs). e2e coverage is the authoritative path for that field.

**PR Readiness Checklist**:


api/konnect/v1alpha2/konnect_entity_status.go
```diff
@@ -212,6 +212,38 @@ type KonnectEntityStatusWithControlPlaneAndCertificateRefs struct {
...
 
+// KonnectEntityStatusWithControlPlaneAndCertificateAndCACertificatesRefs represents the status of a Konnect entity
+// with references to a ControlPlane, a (client) Certificate, and a list of CA Certificates.
+type KonnectEntityStatusWithControlPlaneAndCertificateAndCACertificatesRefs struct {
+	KonnectEntityStatus `json:",inline"`
+
+	// ControlPlaneID is the Konnect ID of the ControlPlane this entity is associated with.
+	//
+	// +optional
+	ControlPlaneID string `json:"controlPlaneID,omitempty"`
```
pmalek, line 223: Couldn't we inline `KonnectEntityStatusWithControlPlaneRef` instead of this?

controller/konnect/reconciler_generic.go
```diff
@@ -405,6 +418,55 @@ func (r *KonnectEntityReconciler[T, TEnt]) Reconcile(
 		return patchWithProgrammedStatusConditionBasedOnOtherConditions(ctx, r.Client, ent)
 	}
 
+	// If a type has KongCACertificate refs (KongService), handle them.
+	res, err = handleKongCACertificateRefs(ctx, r.Client, ent)
+	if err != nil {
+		if errDel, ok := errors.AsType[ReferencedKongCACertificateIsBeingDeletedError](err); ok &&
```
pmalek: In hindsight, using type specific errors here might not have been necessary.

At this point I'm fine either way but I'd prefer if we tried to use `ReferencedObjectIsBeingDeletedError` instead. There's nothing specific which would require the error type to be that specific.

controller/konnect/reconciler_cacertificatesrefs.go
```diff
@@ -0,0 +1,188 @@
...
+		konnectv1alpha1.KongCACertificateRefsValidConditionType,
+		metav1.ConditionTrue,
+		konnectv1alpha1.KongCACertificateRefsReasonValid,
+		"All referenced KongCACertificates are programmed",
+	); errStatus != nil || !res.IsZero() {
+		return res, errStatus
+	}
+
+	// Set CACertificateIDs after the patch so the server response cannot overwrite it.
+	svc.Status.Konnect.CACertificateIDs = collectedIDs
```
pmalek: Just FYI I believe you can use `SetStatusWithConditionIfDifferent` + `ApplyStatusPatchIfNotEmpty` for that.

api/konnect/v1alpha2/konnect_entity_status.go
sandromodarelli: here i'm following the same pattern as for other entities (like consumers here in the same file). if we plan to change this, i would do it in another issue/PR

pmalek: Agreed, I'm fine with that given the current pattern. Let's continue as proposed in the pr

pmalek (approved):

PR #4236, opened 2026-05-12
feat(crd-from-oas): add PortalCustomDomain CRD

pmalek:
**What this PR does / why we need it**:

**Which issue this PR fixes**

Fixes https://github.com/Kong/kong-operator/issues/3664

**Special notes for your reviewer**:

**PR Readiness Checklist**:

Complete these before marking the PR as `ready to review`:

- [ ] the `CHANGELOG.md` release notes have been updated to reflect significant changes

PR #4252, opened 2026-05-12
chore(crd-from-oas): fix matching cel expressions against produced Go type shema and not against OpenAPI spec

pmalek:

**Which issue this PR fixes**

Follow up from https://github.com/Kong/kong-operator/pull/4236

**PR Readiness Checklist**:

### DIFF
--- a/crd-from-oas/pkg/generator/generator.go
+++ b/crd-from-oas/pkg/generator/generator.go
@@ -39,7 +39,7 @@ type Config struct {
 	// Defaults to true.
 	GenerateGroupVersionInfo bool
 	// APIGroupPackagePath is the full Go import path for the generated API types package
-	// (e.g. "github.com/kong/kong-operator/v2/api/x-konnect/v1alpha1").
+	// (e.g. "github.com/kong/kong-operator/v2/api/konnect/v1alpha1").
 	APIGroupPackagePath string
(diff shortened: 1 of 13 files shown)


crd-from-oas/pkg/generator/generator.go
```diff
@@ -222,155 +225,177 @@ func (g *Generator) Generate(parsed *parser.ParsedSpec) ([]GeneratedFile, error)
...
+				if variantSchema == nil {
+					continue
+				}
+				out[fixInitialisms(variant.refName)] = variantCursor
+				if variantCursor != nil {
+					if err := g.collectSchemaCursors(entityName, variantPath, variantSchema, variantCursor, schemas, out); err != nil {
+						return err
+					}
+				}
+			}
```
tao12345666333, line 410: Here we use the `out[]=..`. However, since it is performed in a range within the map, the order is also uncertain, which may lead to overwriting instead of merging. WDYT?

pmalek: Both `schema.Properties` and `buildUnionVariants()` return value are slices so the order is enforced.

I've added overwrite detection in ff3a38b84d91e94620be210725e433b645cbb649 though to make sure this wouldn't happen.

PTAL

tao12345666333 (approved):

PR #4261, opened 2026-05-13
feat(crd-from-oas): implement cross object reference handling

### LINKED ISSUE #4069: Konnect CRDs: EventGatewayVirtualCluster should use an in cluster reference for EventGatewayBackendCluster
### Problem statement

`EventGatewayVirtualCluster` requires a reference to `EventGatewayBackendCluster`. Currently the API allows specifying either an id or a name.

The ID reference is correct, but name reference should use an in cluster reference to a `EventGatewayBackendCluster` object

```
kind: EventGatewayVirtualCluster
metadata:
  name: cp-event-1-virtual-cluster-1
  namespace: default
spec:
  apiSpec:
    name: cp-event-1-virtual-cluster-1
    description: Test Event Gateway Virtual Cluster
    destination:
      name: cp-event-1-backend-cluster-1
      # optionally use an id reference
      # id: 11111111111111111111111111
    ...
```

### Acceptance criteria

- [ ] As a user I can reference in cluster `EventGatewayBackendCluster` objects from `EventGatewayVirtualCluster`s.

pmalek:

Implement cross object reference handling for example between `EventGatewayVirtualCluster` and `EventGatewayBackendCluster`.

Adapt `EventGatewayVirtualCluster` schema to this change by removing the `gatewayRef` and `spec.apiSpec.destination` field and replacing it with `eventGatewayBackendClusterRef`.

The referenced gateway is inferred via from event gateway referenced by the `EventGatewayBackendCluster`.

**Which issue this PR fixes**

Fixes https://github.com/Kong/kong-operator/issues/4069

**PR Readiness Checklist**:


tao12345666333 (commented):
The `gatewayRef` and `spec.apiSpec.destination` of `EventGatewayVirtualCluster` allow `type: konnectID` in the CRD, but the actual reconciler only supports `namespacedRef`. Is this expected? Or should we handle it in a later PR?

https://github.com/Kong/kong-operator/blob/7df87e00e6430ac163306deb56b474b397b24cfb/controller/konnect/reconciler_generic_handle_parentref.go#L257-L259

pmalek:
> The `gatewayRef` and `spec.apiSpec.destination` of `EventGatewayVirtualCluster` allow `type: konnectID` in the CRD, but the actual reconciler only supports `namespacedRef`. Is this expected? Or should we handle it in a later PR?
> 
> https://github.com/Kong/kong-operator/blob/7df87e00e6430ac163306deb56b474b397b24cfb/controller/konnect/reconciler_generic_handle_parentref.go#L257-L259

Yup, it's good you're calling this out. I plan to disable this via CEL ( to keep the same underlying reference type). Konnect ID will be either supported through mirror or otherwise but this is not the goal for 2.2.

There's another issue with this PR which I'm working on: we have both the event gateway ref and backend cluster ref. There should only be the latter of the 2 since current approach would allow users to configure virtual cluster with 2 different event gateways (which would fail).

pmalek:
> > The `gatewayRef` and `spec.apiSpec.destination` of `EventGatewayVirtualCluster` allow `type: konnectID` in the CRD, but the actual reconciler only supports `namespacedRef`. Is this expected? Or should we handle it in a later PR?
> > https://github.com/Kong/kong-operator/blob/7df87e00e6430ac163306deb56b474b397b24cfb/controller/konnect/reconciler_generic_handle_parentref.go#L257-L259
> 
> Yup, it's good you're calling this out. I plan to disable this via CEL ( to keep the same underlying reference type). Konnect ID will be either supported through mirror or otherwise but this is not the goal for 2.2.
> 
> There's another issue with this PR which I'm working on: we have both the event gateway ref and backend cluster ref. There should only be the latter of the 2 since current approach would allow users to configure virtual cluster with 2 different event gateways (which would fail).

Updated. PTAL

controller/konnect/zz_generated_reconciler_authref.go
```diff
@@ -42,7 +47,31 @@ func getAPIAuthRef[
...
+	ctx context.Context,
+	cl client.Client,
+	obj eventGatewayBackendClusterRefAccessor,
+) (types.NamespacedName, error) {
+	bcRef := obj.GetEventGatewayBackendClusterRef()
+	if bcRef.Type != commonv1alpha1.ObjectRefTypeNamespacedRef || bcRef.NamespacedRef == nil {
+		return types.NamespacedName{},
+			fmt.Errorf("invalid EventGatewayBackendCluster reference: must be a NamespacedRef with a non-nil NamespacedRef field")
+	}
+	nn := types.NamespacedName{Name: bcRef.NamespacedRef.Name, Namespace: obj.GetNamespace()}
```
tao12345666333: I have a concern: since we use `Namespace: obj.GetNamespace()`, will the cross-namespace parent reference not function properly?

tao12345666333: How about `bcRef.NamespacedRef.Namespace`

pmalek: I've made this code return an error when cross-namespace ref is used and added TODOs for handling that. 

This will be done in a dedicated issue, in https://github.com/Kong/kong-operator/issues/4134

pmalek: Updated the PR via 17d573f9a4f349e60cb95cba138a170911a3cc8c. PTAL

tao12345666333 (approved):

PR #4263, opened 2026-05-13
feat(konnect): support cross-namespace upstreamRef in KongTarget

### LINKED ISSUE #4153: CrossNamespace: enable cross-namespace KongTarget → KongUpstream
## Problem Statement

`KongTarget.spec.upstreamRef` only supports same-namespace references because the field
is typed as `commonv1alpha1.NameRef`, which has no `namespace` field. Both
`watch_kongtarget.go` and `watch_kongupstream.go` carry `// TODO: change this when cross
namespace refs are allowed` markers, confirming that cross-namespace support was deferred.

Users cannot place `KongTarget` and `KongUpstream` in separate namespaces, even when
they have a valid `KongReferenceGrant`. This is inconsistent with `KongRoute → KongService`,
which supports cross-namespace references via grants.

## Proposed Solution

1. Change `KongTarget.spec.upstreamRef` from `NameRef` to `NamespacedRef` so a namespace
   can be specified. Run `make generate` to regenerate CRDs and deepcopy.
2. In `reconciler_upstreamref.go`, when `upstreamRef.namespace` is set and differs from
   the target's namespace, call `crossnamespace.CheckKongReferenceGrantForResource` and
   surface `ResolvedRefs=False/RefNotPermitted` if no grant exists, mirroring the pattern
   used in `reconciler_serviceref.go`.
3. In `watch_kongtarget.go`, update the enqueue logic to handle cross-namespace upstream
   references and remove the same-namespace restriction.
4. In `watch_kongupstream.go`, add a `KongReferenceGrant` watch and update the enqueue
   logic so targets in other namespaces are re-queued when the upstream or a related
   grant changes.

## Acceptance Criteria

- [ ] As a user, I can place a `KongTarget` and its referenced `KongUpstream` in different namespaces, and the target reaches `Programmed=True` when a valid `KongReferenceGrant` exists.
- [ ] As a user, when no `KongReferenceGrant` exists, the `KongTarget` surfaces `ResolvedRefs=False/RefNotPermitted` instead of silently failing.
- [ ] As a user, creating or deleting a `KongReferenceGrant` triggers immediate re-reconciliation of the affected `KongTarget`.

alacuku:

Change KongTarget.spec.upstreamRef from NameRef to NamespacedRef so that a KongTarget can reference a KongUpstream in a different namespace.

When the namespace field is set and differs from the KongTarget's namespace, the reconciler checks for a KongReferenceGrant in the upstream's namespace that allows the KongTarget to reference it. If no grant exists, ResolvedRefs=False/RefNotPermitted is surfaced on the KongTarget. Once a valid grant is present the condition is cleared and reconciliation proceeds normally.

Changes:
- api: change KongTarget.spec.upstreamRef type NameRef → NamespacedRef; regenerate CRD and deepcopy
- reconciler_upstreamref: check KongReferenceGrant for cross-namespace refs; set/clear ResolvedRefs condition accordingly
- reconciler_apiauthref: resolve upstream namespace from upstreamRef.Namespace instead of always using the KongTarget's namespace
- watch_kongtarget: update predicate and enqueue function to handle cross-namespace upstream lookups; add KongReferenceGrant watch to trigger re-reconciliation
- index/kongtarget: include upstream namespace in index key
- hybridgateway/builder, test helpers, CRD validation test: update NameRef → NamespacedRef

**Which issue this PR fixes**

Fixes #4153 

E2E tests: https://github.com/Kong/kong-operator/pull/4264

**PR Readiness Checklist**:

### DIFF
--- a/controller/konnect/reconciler_upstreamref.go
+++ b/controller/konnect/reconciler_upstreamref.go
@@ -2,6 +2,7 @@ package konnect
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/samber/mo"
@@ -16,22 +17,24 @@ import (
 	configurationv1alpha1 "github.com/kong/kong-operator/v2/api/configuration/v1alpha1"
 	konnectv1alpha1 "github.com/kong/kong-operator/v2/api/konnect/v1alpha1"
 	konnectv1alpha2 "github.com/kong/kong-operator/v2/api/konnect/v1alpha2"
+	ctrlconsts "github.com/kong/kong-operator/v2/controller/consts"
 	"github.com/kong/kong-operator/v2/controller/konnect/constraints"
 	"github.com/kong/kong-operator/v2/controller/pkg/controlplane"
 	"github.com/kong/kong-operator/v2/controller/pkg/patch"
+	"github.com/kong/kong-operator/v2/internal/utils/crossnamespace"
 	k8sutils "github.com/kong/kong-operator/v2/pkg/utils/kubernetes"
 )
 
 // getKongUpstreamRef gets the reference of KongUpstream.
 func getKongUpstreamRef[T constraints.SupportedKonnectEntityType, TEnt constraints.EntityType[T]](
 	e TEnt,
-) mo.Option[commonv1alpha1.NameRef] {
+) mo.Option[commonv1alpha1.NamespacedRef] {
 	switch e := any(e).(type) {
 	case *configurationv1alpha1.KongTarget:
 		// Since upstreamRef is required for KongTarget, we directly return spec.UpstreamRef here.
 		return mo.Some(e.Spec.UpstreamRef)
 	default:
-		return mo.None[commonv1alpha1.NameRef]()
+		return mo.None[commonv1alpha1.NamespacedRef]()
 	}
 }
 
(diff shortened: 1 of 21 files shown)


controller/konnect/reconciler_upstreamref.go
```diff
@@ -41,30 +44,64 @@ func handleKongUpstreamRef[T constraints.SupportedKonnectEntityType, TEnt constr
...
 
-	kongUpstream := &configurationv1alpha1.KongUpstream{}
+	// Snapshot current status. All condition writes below are in-memory only.
+	// A single status patch is issued on exit regardless of exit path.
+	old := ent.DeepCopyObject().(TEnt)
+	defer func() {
+		if _, patchErr := patch.ApplyStatusPatchIfNotEmpty(ctx, cl, ctrllog.FromContext(ctx), ent, old); patchErr != nil {
+			err = errors.Join(err, patchErr)
+		}
+	}()
```
pmalek, line 60: If we're using a named return value and a defer, shouldn't we return `err` in all `return` paths?

alacuku: When returning an explicit error it is assigned to the named error: https://go.dev/ref/spec#Return_statements

pmalek (approved):
👍 

One small nit is that I'd suggest to add a sample manifest for this (specifying all 3: `KongUpstream`, `KongTarget` and `KongReferenceGrant`.)

alacuku:
Adding the manifest here: https://github.com/Kong/kong-operator/pull/4264

PR #4265, opened 2026-05-13
feat(crd-from-oas): Support generated getForUID for ListenerPolicy

### LINKED ISSUE #4215: Reconciler code generation: support generated getForUID for ListenerPolicy without manual helper
## Problem Statement
`EventGatewayListenerPolicy` currently requires a hand-written `ops_eventgatewaylistenerpolicy_manual.go` helper for `getForUID`.

This is not only because the SDK list response is returned as a bare slice rather than the usual paginated `resp.<field>.Data` shape. The generated `getForUID` path also does not currently express the identity matching needed for ListenerPolicy's root-union spec shape, where matching needs to account for the selected variant and its discriminator/name fields.

As a result, ListenerPolicy reconciliation support is not yet fully declarative in the generator/config layer.

## Proposed Solution
Extend generated `getForUID` support so that `EventGatewayListenerPolicy` can use generated logic instead of a hand-written helper.

For ListenerPolicy specifically, this likely requires:
- allowing generated `getForUID` code to iterate list response items that are returned as a bare array rather than `resp.<field>.Data`
- supporting declarative matching for root-union-backed entities, so identity can be derived from the selected variant (for example name + discriminator mapping)
- updating generator/config plumbing so `EventGatewayListenerPolicy` no longer needs `skipGetForUID: true`

## Additional Information
- Follow-up from PR #4182
- Related issue: #4055
- Related generator work: #3784, #3857
- Current workaround:
  - `crd-from-oas/config.yaml` sets `skipGetForUID: true`
  - `controller/konnect/ops/ops_eventgatewaylistenerpolicy_manual.go` provides the manual implementation
- Upstream Event Gateway OAS currently defines `ListEventGatewayListenerPoliciesResponse` as `application/json` `type: array`, so this is not just a local `.Data` wrapper mismatch

## Acceptance Criteria
- [ ] Generated `getForUID` can be configured to work with bare-slice / non-paginated list response shapes
- [ ] Generated `getForUID` can express ListenerPolicy-style matching without a hand-written helper
- [ ] `EventGatewayListenerPolicy` no longer requires `skipGetForUID: true`
- [ ] `ops_eventgatewaylistenerpolicy_manual.go` can be removed
- [ ] Tests are added for the new generator/config behavior
- [ ] Documentation is updated if needed for the new generator config

tao12345666333:

**Which issue this PR fixes**

Fixes https://github.com/Kong/kong-operator/issues/4215

**PR Readiness Checklist**:


pmalek (dismissed):

crd-from-oas/config.yaml
```diff
@@ -79,9 +79,23 @@ apiGroupVersions:
...
+                  responseTypeValue: tls_server
+                  matchFields:
+                    - objectField: Name
+                      responseField: GetName()
+                - typeValue: forwardToVirtualCluster
+                  variantField: ForwardToVirtualClust
+                  responseTypeValue: forward_to_virtual_cluster
+                  matchFields:
+                    - objectField: Name
+                      responseField: GetName()
```
pmalek, line 98: This is rather complicated to specify for my taste.

If possible, let's add crd-from-oas/README.md notes on how to use this.

For posterity I just wanted to note that these getForUID functions (which currently do not have much to do with UID) are all temporary until we get proper "managed_by" solution implemented in Konnect (through metadata, tags or something else). At that point we'll be able to simply infer the matching UID through those.

tao12345666333: Yes, let me add some notes, thank you!

tao12345666333: Added https://github.com/Kong/kong-operator/pull/4265/commits/a352f4c50ec0afd85a85174e000f49850be1a253

controller/konnect/ops/zz_generated_ops_eventgatewaylistenerpolicy.go
```diff
@@ -100,3 +100,86 @@ func deleteEventGatewayListenerPolicy(
...
+			switch id := any(entry.GetID()).(type) {
+			case string:
+				if id != "" {
+					return id, nil
+				}
+			case *string:
+				if id != nil && *id != "" {
+					return *id, nil
+				}
+			}
```
pmalek, line 178: For this we could implement a function to get a `string` from either a `string` or `*string`.

Can be done in a separate PR but that could be useful and would decrease the code at call sites.

pmalek (approved):

controller/konnect/ops/zz_generated_ops_eventgatewaylistenerpolicy.go
tao12345666333: I agree. I will create a sperate PR for this.

PR #4305, opened 2026-05-14
feat(konnect): support `KongCertificate` ref in `KongUpstream`

### LINKED ISSUE #864: Konnect entities: Implement CA Certificate refs for `KongService` and `KongUpstream`

sandromodarelli:

Adds `spec.clientCertificateRef` to `KongUpstream`, allowing a declarative reference to a `KongCertificate` CR instead of hand-pasting a raw Konnect ID via the deprecated `spec.client_certificate` field.

- Cross-namespace refs validated via `KongReferenceGrant`
- `client_certificate` and `clientCertificateRef` are mutually exclusive (CEL rule)
- `client_certificate` kept functional but marked deprecated
- Status type swapped to `KonnectEntityStatusWithControlPlaneAndCertificateRefs` (consistent with KongSNI/KongService)

Follows the same pattern as #4240 (KongService cert refs).

**Which issue this PR fixes**

Fixes #864 

Status type change on `KongUpstream` is a Go-level breaking change but JSON wire-compatible (same field name `konnect`, same sub-fields plus optional `certificateID`).

**PR Readiness Checklist**:


api/configuration/v1alpha1/kongupstream_types.go
```diff
@@ -78,11 +78,18 @@ type KongUpstreamSpec struct {
...
 // +kubebuilder:validation:XValidation:rule="!has(self.adopt) ? true : (has(self.controlPlaneRef) && self.controlPlaneRef.type == 'konnectNamespacedRef')", message="spec.adopt is allowed only when controlPlaneRef is konnectNamespacedRef"
 // +kubebuilder:validation:XValidation:rule="(has(oldSelf.adopt) && has(self.adopt)) || (!has(oldSelf.adopt) && !has(self.adopt))", message="Cannot set or unset spec.adopt in updates"
+// +kubebuilder:validation:XValidation:rule="!(has(self.client_certificate) && has(self.clientCertificateRef))", message="client_certificate and clientCertificateRef are mutually exclusive"
 type KongUpstreamAPISpec struct {
 	// Which load balancing algorithm to use.
 	Algorithm *sdkkonnectcomp.UpstreamAlgorithm `default:"round-robin" json:"algorithm,omitempty"`
-	// If set, the certificate to be used as client certificate while TLS handshaking to the upstream server.
+	// Deprecated: use clientCertificateRef instead. If set, the certificate to be used as client certificate while TLS handshaking to the upstream server.
```
pmalek, line 85: Sadly https://github.com/kubernetes/kubernetes/issues/131817 is not done yet and https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/#failure-policy can only specify `Fail` or `Ignore`.

pmalek (commented):
Just 1 Q: can we add a sample manifest with the new field so show how it can be used ( + test the sample on each PR on CI)?

api/configuration/v1alpha1/kongupstream_types.go
sandromodarelli: i added a validating admission policy for the deprecation notice as we did for other reosurces. i didn't modified the comment on API, it's always useful to have it field description

config/default/validating_policies/konnect/kongupstream_client_certificate_deprecation_policy.yaml
```diff
@@ -0,0 +1,34 @@
...
+---
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingAdmissionPolicyBinding
+metadata:
+  name: "binding-client-certificate.kongupstream.configuration.konghq.com"
+spec:
+  policyName: "client-certificate.kongupstream.configuration.konghq.com"
+  validationActions:
+    - Warn
+    - Audit
```
pmalek, line 34: I forgot about these. Good call to add the VAP like this. 👍 🚀

pmalek (approved):

PR #4352, opened 2026-05-19
feat(hybridgateway): support konghq.com/client-cert service annotation

### LINKED ISSUE #4353: Hybridgateway: support `konghq.com/client-cert` service annotation
## Problem Statement

`controller/hybridgateway/` doesn't honor the `konghq.com/client-cert` annotation on backend Kubernetes Services. The embedded ingress-controller already supports it, so KIC and HybridGateway translation diverge.

## Proposed Solution

Mirror the pattern introduced by #3847 (`konghq.com/protocol`): extract the annotation from each `BackendRef`'s Service and set `KongService.Spec.clientCertificateRef` so it refers to a brand new `KongCertificate` that refers to the secret mentioned in the annotation.

## Additional Information

- Reference extractor: `ingress-controller/internal/annotations/annotations.go`
- Target field: `api/configuration/v1alpha1/kongservice_types.go` `KongServiceAPISpec.clientCertificateRef`

## Acceptance Criteria

- [x] As a user, I can set `konghq.com/client-cert` on a backend Service and see it propagated to the generated `KongService`
- [x] Unit tests are added

sandromodarelli:

Adds support for the `konghq.com/client-cert` Service annotation in the hybridgateway translator, bringing it to parity with the embedded ingress controller.

When a Service used as a `backendRef` in an `HTTPRoute` or `TLSRoute` carries `konghq.com/client-cert: <secret-name>`, the hybridgateway controller now:

1. Produces a `KongCertificate` of type `secretRef` referencing the Secret, placed in the Gateway namespace.
2. Wires it into the generated `KongService` via `spec.clientCertificateRef`.
3. When the Gateway namespace and the Secret namespace differ, creates a `KongReferenceGrant` in the Secret's namespace permitting the cross-namespace reference.
4. Reconciles and cleans up all three resources every loop using the existing `hybrid-routes` annotation machinery and orphan cleanup.

Business rules match the ingress controller:
- **First-wins** across `backendRefs` of the rule.
- Annotation absent or empty → skipped silently.
- Non-TLS effective protocol (`http`, `grpc`, `tcp`, `tls_passthrough`, `udp`, `ws`) → cert skipped, `KongService` produced without `clientCertificateRef`.
- Secret not found → cert skipped, error logged, `KongService` produced without `clientCertificateRef`.

**Which issue this PR fixes**

Fixes #4353 

Key files:
- `controller/hybridgateway/service/client_cert.go` — `resolveClientCertFromBackendRefs`, `buildClientCertificate`, `buildClientCertReferenceGrant`, `isNonTLSProtocol`
- `controller/hybridgateway/service/service.go` — `ServiceForRule` now returns `(KongService, KongCertificate, KongReferenceGrant, error)`
- `controller/hybridgateway/converter/http_route.go` + `tls_route.go` — updated call sites, `outputStore` ordering (grant → cert → service), `expectedGVKs` cleanup ordering
- `controller/hybridgateway/reconciler_utils.go` — `KongService` apply is gated on `KongCertificate.Programmed=True` via `stopAtKind`
- `controller/hybridgateway/route/status.go` — `KongReferenceGrant` excluded from programmed-condition checks (passive policy object, no Konnect sync)
- `controller/hybridgateway/watch/` — Secret → route requeue, `KongCertificate`/`KongReferenceGrant` → route requeue
- `test/e2e/chainsaw/hybridgateway/httproute-service-annotations/` — chainsaw e2e test

**PR Readiness Checklist**:


controller/hybridgateway/builder/kongservice_test.go
```diff
@@ -509,3 +509,40 @@ func TestKongServiceBuilder_MultipleErrors(t *testing.T) {
...
+		{
+			name:        "valid name sets NamespacedRef",
+			input:       "my-cert",
+			expectedRef: &commonv1alpha1.NamespacedRef{Name: "my-cert"},
+		},
+		{
+			name:        "second call overwrites first",
+			input:       "other-cert",
+			expectedRef: &commonv1alpha1.NamespacedRef{Name: "other-cert"},
+		},
```
programmer04:
```suggestion
```

This testcase is redundant, the name is misleading, each of the test cases is a separate case

controller/hybridgateway/watch/mapfuncs_secret.go
```diff
@@ -0,0 +1,66 @@
...
+		var routeRequests []reconcile.Request
+		var err error
+		switch routeKind {
+		case kindHTTPRoute:
+			routeRequests, err = listHTTPRoutesForService(ctx, cl, svc.Namespace, svc.Name)
+		case kindTLSRoute:
+			routeRequests, err = listTLSRoutesForService(ctx, cl, svc.Namespace, svc.Name)
+		}
+		if err != nil {
+			continue
```
programmer04, line 61: Maybe it would be beneficial to log an error so as not to be silent in case something is going wrong? Or even just breaking the loop and returning...

controller/hybridgateway/controller.go
```diff
@@ -52,6 +52,7 @@ const (
 //+kubebuilder:rbac:groups=configuration.konghq.com,resources=kongpluginbindings/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=configuration.konghq.com,resources=kongcertificates,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=configuration.konghq.com,resources=kongcertificates/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=configuration.konghq.com,resources=kongreferencegrants,verbs=get;list;watch;create;update;patch;delete
```
programmer04, line 55: There is such an entry here `controller/gateway/controller_rbac.go ` should we have it here too

controller/hybridgateway/watch/mapfuncs_secret.go
sandromodarelli: unfortunately there's no logger already open in the chain (this map is used to setup the watcher). also all the other mappers are not logging errors. about breaking the loop, i wouldn't do it: other mappers are returning nil in case of error and since this function needs to return a list of elements, i think it's ok to just skip the element and proceed with other ones

programmer04 (approved):

controller/hybridgateway/controller.go
sandromodarelli: i don't have a strong opinion. i would prefer to maintain it because that resource is managed by both controllers for 2 different goals

PR #4357, opened 2026-05-19
chore: get rid of `golang.org/x/net` and simplify `go.mod`

programmer04:

Parameters can be configured directly https://pkg.go.dev/net/http#Transport.HTTP2 in newer versions of Go and   `ReadIdleTimeout` seems to map to stdlib `SendPingTimeout` [see the source](https://github.com/golang/go/blob/93da30397d6c1596e3c67df70df9e2ccde378995/src/net/http/internal/http2/transport.go#L631-L632)

### DIFF
--- a/modules/manager/httpclient_konnect.go
+++ b/modules/manager/httpclient_konnect.go
@@ -1,33 +1,26 @@
 package manager
 
 import (
-	"fmt"
 	"net/http"
 	"time"
 
 	"github.com/hashicorp/go-cleanhttp"
-	"golang.org/x/net/http2"
 )
 
-func httpClientForKonnect(c *Config) (*http.Client, error) {
-	// Clone the default transport to inherit all Go defaults (TLS settings,
-	// timeouts, proxy support, etc.) and only override what we need.
+func httpClientForKonnect(c *Config) *http.Client {
 	transport := cleanhttp.DefaultPooledTransport()
 	transport.IdleConnTimeout = 2 * c.KonnectSyncPeriod
-
 	// Configure HTTP/2 health checks to detect stalled connections.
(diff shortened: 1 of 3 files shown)


modules/manager/httpclient_konnect.go
```diff
@@ -1,33 +1,26 @@
...
 	transport := cleanhttp.DefaultPooledTransport()
 	transport.IdleConnTimeout = 2 * c.KonnectSyncPeriod
-
 	// Configure HTTP/2 health checks to detect stalled connections.
-	// When no frames are received for ReadIdleTimeout, a PING is sent.
-	// If no PING response within PingTimeout, the connection is closed
+	// When no frames are received for SendPingTimeout, a PING is sent.
+	// If no PING response is received within PingTimeout, the connection is closed
 	// and removed from the pool, forcing new requests to open a fresh connection.
-	h2Transport, err := http2.ConfigureTransports(transport)
```
pmalek, line 22: `http2.ConfigureTransports` seems to do a bit more than set the timeouts as we do here: 

- https://cs.opensource.google/go/x/net/+/refs/tags/v0.54.0:http2/transport_common.go;l=34
- https://cs.opensource.google/go/x/net/+/refs/tags/v0.54.0:http2/transport.go;l=108-162

are we sure we want to replace it like this and this will be equivalent to previous version?

programmer04: https://github.com/golang/go/blob/2dc996f71b0ebafb77e64433e58333e049488a3c/src/net/http/transport.go#L434 calls https://github.com/golang/go/blob/2dc996f71b0ebafb77e64433e58333e049488a3c/src/net/http/h2_bundle.go#L7526-L7578, which equals the aforementioned

pmalek (approved):

PR #4408, opened 2026-05-26
feat: support namespace selectors in KongReferenceGrant

### LINKED ISSUE #4407: Support namespace label selectors in KongReferenceGrant
## Problem

KongReferenceGrant currently requires every allowed source namespace to be listed explicitly via `spec.from[].namespace`. This works for small setups, but it becomes repetitive and operationally fragile in multi-tenant clusters where many namespaces share the same trust boundary.

For example, an operator may want a `KongPlugin`, `KongService`, `KongCertificate`, or `KonnectGatewayControlPlane` in one namespace to be referenceable by all tenant namespaces labeled with something like `konghq.com/cross-namespace-access=true`, without editing the `KongReferenceGrant` every time a tenant namespace is added or removed.

## Proposed Solution

Add a Kubernetes-native namespace selector option to `KongReferenceGrant` source entries, for example:

```yaml
kind: KongReferenceGrant
metadata:
  name: allow-tenant-namespaces
  namespace: shared-kong-config
spec:
  from:
  - group: configuration.konghq.com
    kind: KongPluginBinding
    namespaceSelector:
      matchLabels:
        konghq.com/cross-namespace-access: "true"
  to:
  - group: configuration.konghq.com
    kind: KongPlugin
```

Suggested API behavior:

- Support either `spec.from[].namespace` or `spec.from[].namespaceSelector`.
- Reject entries that set both fields.
- Keep current exact namespace matching behavior unchanged for existing manifests.
- Evaluate `namespaceSelector` against Kubernetes `Namespace` labels.
- Reconcile affected resources when relevant `Namespace` labels change, so grants are applied and revoked without waiting for unrelated resource updates.

## Alternatives Considered

- Continue requiring explicit namespace entries. This is simple and secure, but does not scale well when namespaces are created dynamically.
- Use a wildcard namespace value. A label selector is more Kubernetes-native and gives administrators better control over trust boundaries.

## Security Considerations

Namespace selector grants would make `Namespace` label ownership security-sensitive. Documentation should call out that users who can modify `Namespace` labels may be able to affect cross-namespace reference permissions.

The implementation should also ensure permission revocation happens promptly when labels are removed.

## Context

Today the API exposes `spec.from[].namespace` as a single namespace string, and runtime matching checks it by exact equality. There is no selector/tag-based source namespace grant mechanism.

jerry153fish:
## Summary

- add spec.from[].namespaceSelector to KongReferenceGrant for label-selected source namespaces
- validate namespace and namespaceSelector as mutually exclusive, reject empty selectors, and keep KongVault cluster-scoped behavior unchanged
- evaluate namespace selectors against Kubernetes Namespace labels and requeue selected namespaces when KongReferenceGrant changes
- regenerate deepcopy, CRDs, chart CRD, and API docs

Fixes #4407

## Tests

- GOCACHE=/tmp/kong-operator-go-build go test ./internal/utils/crossnamespace ./controller/konnect
- git diff --check


jerry153fish:
@pmalek some review? thanks

pmalek:
As commented in https://github.com/Kong/kong-operator/issues/4407: the conclusion for this will be to close this PR at the current stage.

I'm still open to discuss adding similar functionality but through different means.

Feel free to drop a comment or create a new issue with new proposal.

And again: thanks for taking the time to create this PR and issue describing the problem.

PR #4419, opened 2026-05-27
chore(deps): get rid of `phayes/freeport`

programmer04:

Get rid of the dependency that imports a single function to discover a free port with the help of the operating system. It's a well-established way, so it won't change. Keeping `phayes/freeport` as a dependency is redundant (having it poses a risk of a potential supply-chain attack).


ingress-controller/test/helpers/ports.go
```diff
@@ -51,3 +50,18 @@ func GetFreePort(t *testing.T) int {
 
 // userPorts keeps track of ports that were used in the current test run.
 var usedPorts sync.Map
+
+// getFreePort asks the kernel for a free open port that is ready to use.
+func getFreePort() (int, error) {
```
pmalek: Nit: this could be called `getFreeTCPPort()`

pmalek (approved):

PR #4445, opened 2026-05-29
fix(gateway): Check existence of TLSRoute CRD

randmonkey:

Check if gateway v1 `TLSRoute` CRD exists before starting watch on `TLSRoute` for gateway controller to prevent errors when CRD does not exist. This keeps compatability with Gateway API 1.4.

**Which issue this PR fixes**

**PR Readiness Checklist**:

### DIFF
--- a/controller/gateway/controller.go
+++ b/controller/gateway/controller.go
@@ -16,6 +16,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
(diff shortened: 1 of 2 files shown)


controller/gateway/controller.go
```diff
@@ -83,7 +84,7 @@ const provisionDataPlaneFailRetryAfter = 5 * time.Second
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *Reconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager) error {
-	builder := ctrl.NewControllerManagedBy(mgr).
+	blder := ctrl.NewControllerManagedBy(mgr).
```
alacuku, line 87: Could we hold off on refactoring the unrelated variable names for this PR? Keeping the scope tight helps us maintain a smaller change surface and keeps the real `git blame` intact. Let's only change what's absolutely needed for this fix.

randmonkey: The rename is required because it shadows the `builder` as the package name. So we need to rename either one (package name or variable name) to make it compile.
If it is required to separate the two changes, please look at https://github.com/Kong/kong-operator/pull/4446 first.

alacuku (approved):

pmalek (approved):

PR #4469, opened 2026-06-01
chore: bump github.com/moby/spdystream

pmalek:

Bumping `[github.com/moby/spdystream](https://pkg.go.dev/github.com/moby/spdystream)` to mitigate:

```
- GO-2026-4958 (aka CVE-2026-35469, GHSA-pc3f-x583-g7j2)

	The SPDY/3 frame parser in spdystream does not validate attacker-controlled counts and lengths before allocating memory. A remote peer that can send SPDY frames to a service using spdystream can cause the process to allocate gigabytes of memory with a small number of malformed control frames, leading to an out-of-memory crash.
	
	Three allocation paths in the receive side are affected:
	1. SETTINGS entry count: The SETTINGS frame reader reads a 32-bit numSettings from the payload and allocates a slice of that size without checking it against the declared frame length.
	2. Header count: parseHeaderValueBlock reads a 32-bit numHeaders from the decompressed header block and allocates an http.Header map of that size with no upper bound.
	3. Header field size: Individual header name and value lengths are read as 32-bit integers and used directly as allocation sizes with no validation.
	
	Because SPDY header blocks are zlib-compressed, a small on-the-wire payload can decompress into attacker-controlled bytes that the parser interprets as 32-bit counts and lengths. A single crafted frame is enough to exhaust process memory.
```

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -236,7 +236,7 @@ require (
 	github.com/moby/docker-image-spec v1.3.1 // indirect
 	github.com/moby/go-archive v0.2.0 // indirect
 	github.com/moby/patternmatcher v0.6.1 // indirect
-	github.com/moby/spdystream v0.5.0 // indirect
+	github.com/moby/spdystream v0.5.1 // indirect
 	github.com/moby/sys/sequential v0.6.0 // indirect
 	github.com/moby/sys/user v0.4.0 // indirect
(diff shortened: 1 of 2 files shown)


pmalek:
Handled in https://github.com/Kong/kong-operator/pull/4468.

PR #4509, opened 2026-06-03
fix(hybridgateway): merge duplicate KongTargets when backendRefs resolve to the same endpoint

### LINKED ISSUE #4288: Konnect entities reconciler: no requeue for Hybrid Gateway configuration like plugins
### Problem statement

https://github.com/Kong/kong-operator/pull/3088 tried to address the problem of no requeues for Hybrid Gateway configuration like plugins where objects are re-created with the same hash and we need to requeue as the controller tries to create an object which still exists in Konnect (this happens when e.g. `HTTPRoute` is changed in place).

This has a tight relationship with requeue which on the other hand **should not** happen, e.g. when an invalid Service is attempted to be created and thus causing a 400. When this is constantly retries it causes unnecessary load on Konnect APIs.

While the former can receive errors like:

```
{
	"code": 3,
	"details": [
		{
			"@type": "type.googleapis.com/kong.admin.model.v1.ErrorDetail",
			"field": "hash",
			"messages": [
				"unique-plugin-per-entity (type: unique) constraint failed"
			],
			"type": "ERROR_TYPE_REFERENCE"
		}
	],
	"message": "unique-plugin-per-entity (type: unique) constraint failed"
}
```

from Konnect, the latter receives:

```
{
	"code": 3,
	"details": [
		{
			"@type": "type.googleapis.com/kong.admin.model.v1.ErrorDetail",
			"field": "name",
			"messages": [
				"name (type: unique) constraint failed"
			],
			"type": "ERROR_TYPE_REFERENCE"
		}
	],
	"message": "name (type: unique) constraint failed"
}
```

The decision whether to requeue (ignore the error) or not is taking place in:

- https://github.com/Kong/kong-operator/blob/89dc30da073eccb04db0a09b76d61c48a42bf866/controller/konnect/ops/ops_errors.go#L473
- and specifically in https://github.com/Kong/kong-operator/blob/89dc30da073eccb04db0a09b76d61c48a42bf866/controller/konnect/ops/ops_errors.go#L215-L249

### Acceptance criteria

- [ ] `ERROR_TYPE_REFERENCE` errors are retried only when necessary and when retry will help resolve an issue.

### LINKED ISSUE #4508: KongTarget stuck in Programmed=False when multiple backendRef Services select the same pods
## Current Behavior

When an HTTPRoute (or TLSRoute) rule references multiple Services that resolve to the same underlying pod IP and port, the Kong Operator attempts to create one KongTarget per backendRef for that endpoint, resulting in multiple KongTargets with identical target addresses pointing to the same upstream. Konnect enforces uniqueness on upstream/target pairs, so all but the first creation fail with a `400 target (type: unique)` constraint failed error. The affected KongTargets are left permanently stuck in `Programmed=False`. Unlike other target errors, this does not self-resolve on subsequent reconciliations.

## Expected Behavior

The operator should produce a single KongTarget per unique endpoint address, with the weights from all contributing backendRefs combined. All KongTargets should reach Programmed=True.

## Steps To Reproduce

1. Create a Deployment with a single ReplicaSet.
2. Create two or more Services that all select the same pods (e.g. an Argo Rollouts blue-green activeService and previewService after promotion, where both point to the stable ReplicaSet).
3. Create an HTTPRoute with those Services as backendRefs in the same rule.
4. Observe that the first KongTarget is created successfully while subsequent ones for the same endpoint fail with a 400 error and remain Programmed=False.

## Operator Version

KO 2.1 and main.

alacuku:

```
When multiple backendRefs in an HTTPRoute or TLSRoute rule resolve to the
same pod IP and port, the operator was creating one KongTarget per backendRef
per endpoint, resulting in duplicate target addresses for the same upstream.
Konnect rejects these with a 400 uniqueness constraint error, leaving the
affected KongTargets permanently stuck in Programmed=False.

The operator now creates one KongTarget per unique endpoint address across
all backendRefs in a rule. Endpoints shared across multiple backendRefs are
merged into a single KongTarget whose weight is the sum of all contributing
per-endpoint weights. The KongTarget name still encodes the contributing
backendRef for freshly minted targets.

On upgrade, existing clusters may already have KongTargets whose names were
minted under an older scheme. To avoid creating a duplicate Spec.Target for
the same upstream on the next reconcile, the operator looks up existing
KongTargets for that upstream by address and reuses their name. A fresh name
is only minted when no target exists yet for that address.

Adds unit tests covering the merge, weight-summing, and all lifecycle
transitions (overlap introduced, overlap removed, endpoints diverged).
Adds Chainsaw E2E tests for both HTTPRoute and TLSRoute covering the same
scenarios end-to-end against a real Konnect control plane
```

Furthermore there is a bonus commit that should fix: https://github.com/Kong/kong-operator/issues/4288:
```
fix(konnect): requeue HybridGateway-managed resources on reference-only 400s with fixed delay
400 errors from Konnect where all details are ERROR_TYPE_REFERENCE are
transient: a unique-constraint conflict because an old entity hasn't been
cleaned up yet, or a referenced entity hasn't propagated yet. Previously
these either fell through to controller-runtime's exponential backoff
(causing minute-long convergence delays) or were swallowed by
ErrorIsCreateConflict and never retried.

Introduce ReferenceError wrapping in IgnoreUnrecoverableAPIErr, checked
before the conflict/bad-request suppression block so ERROR_TYPE_REFERENCE
unique-constraint errors are not silently dropped. In the reconciler, a
ReferenceError triggers a 500ms fixed requeue only for resources annotated
with the HybridGateway HTTPRoute or TLSRoute ownership annotations — for
user-created resources the same error shape may indicate a permanent
misconfiguration, so it is suppressed instead.
```

**Which issue this PR fixes**

Fixes #4508 

**PR Readiness Checklist**:


randmonkey (dismissed):
/lgtm

randmonkey:
Put to patch release

controller/hybridgateway/converter/http_route_converter_test.go
```diff
@@ -286,13 +286,31 @@ func TestHTTPRouteConverter_Translate(t *testing.T) {
...
 	}
 
+	// translateAndFindTarget runs Translate on a fresh converter built from objects and
+	// returns the first KongTarget in the output store.
+	translateAndFindTarget := func(t *testing.T, objects []client.Object) *configurationv1alpha1.KongTarget {
+		t.Helper()
+		route := newHTTPRouteForTranslation([]string{"api.example.com"}, []gwtypes.HTTPBackendRef{newBackendRef("")}, nil)
+		cl := fake.NewClientBuilder().WithScheme(scheme.Get()).WithObjects(objects...).Build()
+		conv := newHTTPRouteConverter(route, cl, false, "").(*httpRouteConverter)
+		_, err := conv.Translate(context.Background(), logr.Discard())
```
pmalek:
```suggestion
		_, err := conv.Translate(t.Context(), logr.Discard())
```

controller/hybridgateway/namegen/name_test.go
```diff
@@ -612,68 +612,91 @@ func TestNewKongPluginBindingName(t *testing.T) {
...
-						Namespace: func() *gatewayv1.Namespace { ns := gatewayv1.Namespace("api-ns"); return &ns }(),
-						Port:      func() *gatewayv1.PortNumber { p := gatewayv1.PortNumber(9090); return &p }(),
-					},
-					Weight: func() *int32 { w := int32(100); return &w }(),
+	httpBR := func(name string) *gwtypes.HTTPBackendRef {
+		return &gwtypes.HTTPBackendRef{
+			BackendRef: gatewayv1.BackendRef{
+				BackendObjectReference: gatewayv1.BackendObjectReference{
+					Name: gatewayv1.ObjectName(name),
+					Port: func() *gatewayv1.PortNumber { p := gatewayv1.PortNumber(8080); return &p }(),
```
pmalek, line 620: You should be able to do:


```suggestion
					Port: new(gatewayv1.PortNumber(8080)),
```

(unless we're aiming for automatic backport to work with 2.2's go 1.25)

controller/hybridgateway/target/target.go
```diff
@@ -428,8 +432,78 @@ func recalculateWeightsAcrossBackendRefs[T gwtypes.SupportedBackendRef](validBac
...
+		programmed bool
+	}
+	chosen := make(map[string]candidate, len(list.Items))
+	for i := range list.Items {
+		t := &list.Items[i]
+		// Scope to the upstream we are reconciling: the same address under a different upstream is a
+		// distinct target and must keep its own name.
+		if t.Spec.UpstreamRef.Name != upstreamName {
+			continue
+		}
```
WeichengWang1, line 480: I was concerning if we should skip those targets during deletion

controller/konnect/reconciler_generic.go
```diff
@@ -696,6 +697,21 @@ func (r *KonnectEntityReconciler[T, TEnt]) Reconcile(ctx context.Context, ent TE
...
+			// resources the same error shape can represent a permanent misconfiguration, so
+			// we suppress it instead of requeueing to avoid hammering the Konnect API.
+			case okReferenceErr:
+				annotations := ent.GetAnnotations()
+				if _, hasHTTPRoute := annotations[consts.GatewayOperatorHybridRoutesHTTPRouteAnnotation]; hasHTTPRoute {
+					return ctrl.Result{RequeueAfter: referenceErr.RetryAfter}, nil
+				}
+				if _, hasTLSRoute := annotations[consts.GatewayOperatorHybridRoutesTLSRouteAnnotation]; hasTLSRoute {
+					return ctrl.Result{RequeueAfter: referenceErr.RetryAfter}, nil
+				}
```
pmalek: I fear this will require us to manually maintain this list and add other route's annotations.

Can't we use the gateway annotation here (which is also attached to hybrid gateway resources): `gateway-operator.konghq.com/hybrid-gateways`?

---

BTW we could use `isHybridGatewayManaged()` here right?

controller/konnect/ops/ops_errors.go
```diff
@@ -454,8 +461,34 @@ func handleDeleteError[
...
+	// so the reconciler can use it to set RequeueAfter.
+	if retryAfter, isRateLimited := GetRetryAfterFromRateLimitError(err); isRateLimited {
+		logger.Info("rate limited by Konnect API", "retry_after", retryAfter.String())
+		return RateLimitError{
+			Err:        err,
+			RetryAfter: retryAfter,
+		}
+	}
+
+	// If the error is a 400 whose details are all ERROR_TYPE_REFERENCE it is transient
```
pmalek: > If the error is a 400 whose details are all ERROR_TYPE_REFERENCE it is transient

This is not 100% true. When user created the resources himself this will not be a transient error.

controller/konnect/reconciler_generic.go
```diff
@@ -896,3 +924,15 @@ func patchWithProgrammedStatusConditionBasedOnOtherConditions[
...
+// isHybridGatewayManaged returns true if the entity was generated by the
+// HybridGateway controller on behalf of an HTTPRoute or TLSRoute, as indicated
+// by the presence of the corresponding ownership annotations.
+func isHybridGatewayManaged(obj client.Object) bool {
+	annotations := obj.GetAnnotations()
+	if _, ok := annotations[consts.GatewayOperatorHybridRoutesHTTPRouteAnnotation]; ok {
+		return true
+	}
+	_, ok := annotations[consts.GatewayOperatorHybridRoutesTLSRouteAnnotation]
+	return ok
```
pmalek, line 929: Same comment about `gateway-operator.konghq.com/hybrid-gateways` annotation.

controller/hybridgateway/target/target.go
alacuku: Resources in the desired set are always skipped during the cleanup.

controller/hybridgateway/namegen/name_test.go
alacuku: We are going to backport to 2.2 and 2.1. Since our policy is to not bump the go version there we need to be compatible.

controller/konnect/ops/ops_errors.go
alacuku: Well the error itself can be considered transient, but not for some resources that we know that are user managed. Depends how you want to see it.

pmalek: Can you elaborate? IIUC it's not about the type of the resource but about the context. If a user has created 2 plugins of the same type and bound it to the same resource then this error is not transient but the same error in hybridgateway can be transient when the old plugin has not been cleaned up yet.

Is that correct?

controller/konnect/reconciler_generic.go
alacuku: > I fear this will require us to manually maintain this list and add other route's annotations.
> 
Switching to use `hybrid-gateways` annotation that is more generic.
> Can't we use the gateway annotation here (which is also attached to hybrid gateway resources): `gateway-operator.konghq.com/hybrid-gateways`?
> 

Oh yes, clearly i missed to refactor all the places after I introduced the helper function! Nice catch!

> BTW we could use `isHybridGatewayManaged()` here right?

controller/hybridgateway/converter/http_route_converter_test.go
alacuku: PTAL: 4ab5b82a91614b80aa806e40f01d50c931d786fa

controller/konnect/reconciler_generic.go
alacuku: Changed: 4ab5b82a91614b80aa806e40f01d50c931d786fa

controller/konnect/reconciler_generic.go
alacuku: See: 4ab5b82a91614b80aa806e40f01d50c931d786fa

controller/konnect/ops/ops_errors.go
alacuku: Changed: 3312793f305f5eedc06e07d5009f03982e781554

controller/hybridgateway/target/target.go
```diff
@@ -428,8 +432,78 @@ func recalculateWeightsAcrossBackendRefs[T gwtypes.SupportedBackendRef](validBac
...
+		switch {
+		case !ok:
+			chosen[t.Spec.Target] = candidate{name: t.Name, programmed: programmed}
+		case programmed && !cur.programmed:
+			// Prefer a Programmed target over a non-programmed one for the same address.
+			chosen[t.Spec.Target] = candidate{name: t.Name, programmed: programmed}
+		case programmed == cur.programmed && t.Name < cur.name:
+			// Deterministic tie-break when both have the same Programmed status.
+			chosen[t.Spec.Target] = candidate{name: t.Name, programmed: programmed}
+		}
```
pmalek, line 492: All of these `switch` branches are identical, is that expected? If so, why differentiate between them here?

alacuku: As the comments say the switch selects the best fit among the existing targets. Where do you see the identical branches?

alacuku: The switch body has been de duplicated in 5f27c0240eba827924716f22b215a871203df94b

pmalek (approved):

PR #4521, opened 2026-06-04
fix: fix routes become unaccepted and removed from dataplane unexpectly

### LINKED ISSUE #4507: Routes incorrectly marked  Accepted=False/NoMatchingParent and removed from dataplane during transient listener state change
### TL,DR
If *_routes are reconciled in the window that gateway's listeners' statuses are wiped(for example caused by adding new Routes), these routes will be removed from kong store and become unprogrammed. 
If the next reconcilation of *_routes caused by status update still in the window that gateway's listeners' statuses are wiped, *_routes won't be recovered until further enqueue.
The case is time sensitive.

## Expected Behavior
Adding/removing routes against an operator-managed Gateway does not produce transient Accepted=False/NoMatchingParent and Existing routes are not removed from the Kong dataplane during this period.

## Steps To Reproduce
When a second UDPRoute is attached to the same listener as an existing UDPRoute on an operator-managed on-prem Gateway, the second route's parent status sometimes gets stuck at Accepted=False, reason=NoMatchingParent and is removed from the dataplane, even though the listener correctly reports attachedRoutes=2 and the second route's parentRef is identical to the first. 

This is not UDPRoute-specific(actually not a very good example since a udp listener shouldn't attach more than 1 routes(btw it's not forbidden by gateway api community), but just take it as an example here) — the same code path affects HTTPRoute, TLSRoute, TCPRoute, and GRPCRoute. Existing healthy routes can be transiently removed from the dataplane (traffic-affecting) whenever the operator's gateway controller re-renders listener status.

### Flow
1. controller/gateway/controller.go calls initListenersStatus  — wipes g.Status.Listeners to empty Conditions: [].
2. initProgrammedAndListenersStatus — finds no Programmed condition on wiped slice → sets Listener.Programmed=False, Reason=Pending.
3. setAcceptedAndAttachedRoutes → countAttachedUDPRoutes counts both routes → AttachedRoutes=2. This makes gatewayStatusNeedsUpdate return true.
4. then patches the full regressed status to the API. Early-returns before setProgrammed(True).
5. KIC's UDPRoute reconciler enqueued. getSupportedGatewayForRoute reads gateway from cache.
6. listenerProgrammedInStatus returns errListenerNotProgrammedYet. → continue. allowedByListenerName=false.
7. matched=false. Switch picks parentRef.SectionName != nil && !allowedByListenerName → reason=NoMatchingParent.
8. ensureGatewayReferenceStatusAdded writes Accepted=False/NoMatchingParent.
9. isGatewayProgrammed checks Gateway.Status.Conditions[Programmed] only. Gateway-level stays True during regression window. Listener-level Programmed=False slips through.
10. isRouteAccepted(false, actually checked if there are matched programmed listener, not status condition of Gateway CR) -> DataplaneClient.DeleteObject(udproute) — route removed from kong store. Traffic stops. and writes Accepted=False/NoMatchingParent status. Status write -> next reconciliation.
11. If listener still not Programmed (cache lag, or gateway controller in another wipe cycle, or gateway's listener status not recovered), getSupportedGatewayForRoute returns same NoMatchingParent.
- No status write → no new UpdateEvent → no further enqueue from self watch.
- Routes stuck in UnProgrammed and removed from kong store.
```yaml
  listeners:
  - attachedRoutes: 1
    conditions:
    - lastTransitionTime: "2026-06-03T15:36:28Z"
      message: ""
      observedGeneration: 1
      reason: Accepted
      status: "True"
      type: Accepted
    - lastTransitionTime: "2026-06-03T15:36:28Z"
      message: ""
      observedGeneration: 1
      reason: Programmed
      status: "True"
      type: Programmed

  listeners:
  - attachedRoutes: 2
    conditions:
    - lastTransitionTime: "2026-06-03T15:36:42Z"
      message: ""
      observedGeneration: 1
      reason: Accepted
      status: "True"
      type: Accepted
    - lastTransitionTime: "2026-06-03T15:36:42Z"
      message: ""
      observedGeneration: 1
      reason: Pending
      status: "False"
      type: Programmed

  listeners:
  - attachedRoutes: 2
    conditions:
    - lastTransitionTime: "2026-06-03T15:36:42Z"
      message: ""
      observedGeneration: 1
      reason: Accepted
      status: "True"
      type: Accepted
    - lastTransitionTime: "2026-06-03T15:36:42Z"
      message: ""
      observedGeneration: 1
      reason: Programmed
      status: "True"
      type: Programmed
```

## Possible Solution
I don't have enough context about wiping listener conditions on every reconciliation, if this is by design, maybe we can gate route reconciliations on listener level programmed.

WeichengWang1:

**Which issue this PR fixes**

Fixes #4507 
In this pr, besides checking programmed in the gateway level, we added checking if there are any programmed listener could attach to the route. If there are listeners attachable but none programmed, return requeue.
So when gateway's listeners' statuses are wiped, *route_controller will return requeue instead of continuing the reconciliation and then causing DataplaneClient.DeleteObject

p.s
if wiping listener conditions on every reconciliation in gateway controller is not by design, we don't have to fix the issue in the way of this pr.

**PR Readiness Checklist**:

### DIFF
--- a/ingress-controller/internal/controllers/gateway/grpcroute_controller.go
+++ b/ingress-controller/internal/controllers/gateway/grpcroute_controller.go
@@ -329,7 +329,7 @@ func (r *GRPCRouteReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	// requeue the object and wait until all supported gateways are ready.
 	debug(log, grpcroute, "Checking if the grpcroute's gateways are ready")
 	for _, gateway := range gateways {
-		if !isGatewayProgrammed(gateway.gateway) {
+		if !isGatewayProgrammedForRoute(grpcroute, gateway) {
 			debug(log, grpcroute, "Gateway for route was not ready, waiting")
 			return ctrl.Result{Requeue: true}, nil
 		}
--- a/ingress-controller/internal/controllers/gateway/grpcroute_controller.go
+++ b/ingress-controller/internal/controllers/gateway/grpcroute_controller.go
@@ -329,7 +329,7 @@ func (r *GRPCRouteReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	// requeue the object and wait until all supported gateways are ready.
 	debug(log, grpcroute, "Checking if the grpcroute's gateways are ready")
 	for _, gateway := range gateways {
-		if !isGatewayProgrammedForRoute(grpcroute, gateway) {
+		if !isGatewayProgrammedForRoute(gateway) {
 			debug(log, grpcroute, "Gateway for route was not ready, waiting")
 			return ctrl.Result{Requeue: true}, nil
 		}
--- a/ingress-controller/internal/controllers/gateway/httproute_controller.go
+++ b/ingress-controller/internal/controllers/gateway/httproute_controller.go
@@ -452,7 +452,7 @@ func (r *HTTPRouteReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	// requeue the object and wait until all supported gateways are ready.
 	debug(log, httproute, "Checking if the httproute's gateways are ready")
 	for _, gateway := range gateways {
-		if !isGatewayProgrammedForRoute(httproute, gateway) {
+		if !isGatewayProgrammedForRoute(gateway) {
 			debug(log, httproute, "Gateway for route was not ready, waiting")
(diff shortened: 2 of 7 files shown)


randmonkey:
@WeichengWang1 decides to put the fix to KO 2.3.

pmalek (approved):

PR #4530, opened 2026-06-05
feat(hybridgateway): treat malformed annotations as errors

### LINKED ISSUE #4233: Hybridgateway: treat malformed annotations as errors.
## Problem Statement
As discussed in #4143 & #4214 , silent defaulting hides intent and leads to objects that appear healthy but behave unexpectedly.  
Warning msg is added to log, and defaulting logic is kept currectly.
In upcoming releases from KO 2.3, malformed annos should be treated as errors and such resource creation should be blocked.

...

## Proposed Solution
Besides log malformed err in controller/hybridgateway/service/service.go & controller/hybridgateway/kongroute/route.go, return these err back to func translate, finally HybridGatewayReconciler will emit TranslationFailed events.

We should also treat these errs in HybridGatewayReconciler specially, since we don't want reconciling these CRs with malformed annos continuously. We should recognize these errs in HybridGatewayReconciler and return ctrl.Result{}, nil.
When the malformed annos are modified, reconciliations will be triggered.

~~To block CRs with malformed annos be created, we can add CEL in CRDs,~~
~~Discussion:~~
~~Is it still necessary if we treated malformed annos as err already?~~
we don't own those crds

...

## Additional Information

...

## Acceptance Criteria

- [ ] As a user, I can ...

WeichengWang1:

**Which issue this PR fixes**

Fixes #4233 

This PR makes HybridGateway treat malformed konghq.com/* annotations as invalid configuration instead of silently ignoring them and defaulting values.

Why this is needed: 
Silent defaulting can hide user intent: Routes and backend Services may look healthy while Kong is programmed with defaults the user did not expect. Since Kong Operator does not own the Gateway API CRDs, this cannot be enforced with CRD validation, so the operator needs to surface the problem in status and stop applying invalid configuration until the annotation is fixed.

How it is implemented:
 - Adds a sentinel ErrMalformedAnnotation error and returns it when annotation parsing fails.
 - Converts malformed HTTPRoute annotations like konghq.com/strip-path and konghq.com/preserve-host into errors.
 - Converts malformed backend Service annotations like tls-verify, tls-verify-depth, timeouts, and retries into errors.
 - Adds KongConfigurationValid=False with reason InvalidKongConfiguration on affected HTTPRoute/TLSRoute status entries.
 - Keeps Accepted=True; the issue is reported separately as implementation-specific Kong configuration validity.
 - Omits/removes KongConfigurationValid once annotations are fixed.
 - Stops state enforcement while malformed configuration exists, then resumes when annotation changes trigger reconciliation.
 - Updates unit and chainsaw e2e tests for malformed route and Service annotations, including condition removal after fixes.

**PR Readiness Checklist**:


randmonkey:
Merge after KO 2.2.0 release.

alacuku (commented):
Can we update the e2e chainsaw tests too?

controller/hybridgateway/converter/converter.go
```diff
@@ -26,6 +26,8 @@ type APIConverter[t RootObject] interface {
 	GetOutputStore(ctx context.Context, logger logr.Logger) ([]unstructured.Unstructured, error)
 	// UpdateRootObjectStatus updates the status for the root object.
 	UpdateRootObjectStatus(ctx context.Context, logger logr.Logger) (updated bool, stop bool, err error)
+	// SetRootAcceptedFalse sets Accepted=False for the root object's supported parent references.
+	SetRootAcceptedFalse(ctx context.Context, logger logr.Logger, reason, message string) error
```
alacuku: SetRootAcceptedFalse should not be on APIConverter. Status is already fully
managed by UpdateRootObjectStatus; adding a second entry point that writes status
partially makes the overall status of a resource hard to reason about and maintain.
I would move this logic in the existing `UpdateRootObjectStatus`

controller/hybridgateway/converter/http_route.go
```diff
@@ -296,6 +295,35 @@ func (c *httpRouteConverter) UpdateRootObjectStatus(ctx context.Context, logger
...
+func (c *httpRouteConverter) SetRootAcceptedFalse(ctx context.Context, logger logr.Logger, reason, message string) error {
+	logger = logger.WithValues("phase", "httproute-status")
+	supportedParentRefs, err := getHybridGatewayParents(ctx, logger, c.Client, c.route)
+	if err != nil {
+		return err
+	}
+
+	updated := false
+	condition := route.SetConditionMeta(metav1.Condition{
+		Type:    string(gwtypes.RouteConditionAccepted),
```
alacuku: `Accepted=False/UnsupportedValue `is a best-effort workaround: Kong annotations are implementation-specific and have no relation to the Gateway API at all, so they cannot be correctly represented by Accepted, which is strictly about gateway-listener attachment. UnsupportedValue compounds the mismatch, it is defined for unrecognised fields in the route spec, not for annotation parse failures. 
https://github.com/kubernetes-sigs/gateway-api/blob/b6fb8e1548a25cbd1da8dec28a0ae87e1bd21f6f/apis/v1beta1/shared_types.go#L34-L53

A dedicated implementation-specific condition should be introduced to surface this failure clearly so users know exactly what to fix.

alacuku (commented):
Left a couple of comments.

controller/hybridgateway/converter/converter.go
WeichengWang1: thanks! 
moved, added a validation check process in UpdateRootObjectStatus and update condition KongConfigurationValid=False with reason InvalidKongConfiguration if invalid.

controller/hybridgateway/converter/http_route.go
WeichengWang1: Thank you!
KongConfigurationValid is introduced.

randmonkey (approved):

PR #4558, opened 2026-06-09
refactor(hybridgateway): generalize route status updates

### LINKED ISSUE #2267: Hybrid gateways - Implement `GRPCRoute` status controller
## Problem Statement

Similarly to what we have done for `HTTPRoute`, we need to implement the status controller for `GRPCRoute`. All the underlying resource will have their programmed condition in the `GRPCRoute` status, and the goal of this controller is to gather all that information and summarize it into the resource's status.

During this effort, we should evaluate if we can make the code already written for `HTTPRoute`s generic enough to be shared among the two different routes.

## Acceptance Criteria

- [ ] The `GRPCRoute` status is updated with conditions coming from all the different underlying resources.

WeichengWang1:

**Which issue this PR fixes**

Fixes #2267 

This PR refactors hybrid gateway route status handling so status updates are no longer duplicated in each route converter. It moves the shared logic into controller/hybridgateway/route/status.go via a generic UpdateRouteStatus helper and has both HTTPRoute and TLSRoute converters call it.

Issue #2267 asks for GRPCRoute status support and specifically notes that existing HTTPRoute status logic should be evaluated for reuse. This PR prepares for that by generalizing route status reconciliation across supported route types instead of keeping route-specific copies.

**PR Readiness Checklist**:


internal/types/gatewaytypes_interfaces.go
```diff
@@ -4,7 +4,7 @@ import "sigs.k8s.io/controller-runtime/pkg/client"
 
 // SupportedRoute defines a supported route type.
 type SupportedRoute interface {
-	HTTPRoute | TLSRoute
+	HTTPRoute | TLSRoute | GRPCRoute
```
tao12345666333: Although you have added GRPCRoute here, the other parts downstream still only support HTTPRoute and TLSRoute.

https://github.com/Kong/kong-operator/blob/99199e3a8a702f9ab68d2faa87fe68b438ccfe1c/controller/hybridgateway/service/service.go#L88-L108

I think if this PR is just a `refactor`, its changes should be kept within a certain scope. If you want to add other features at the same time, then it would be more appropriate to change the PR title to `feat`.

WeichengWang1: This pr is for enforceStatus controller/hybridgateway/reconciler_utils.go:310, so translate and other phases in hybridgateway reconciliation is not included.
But yes, keeping it within a refactor scope makes sense, I'll separate it.

controller/hybridgateway/converter/route_status.go
alacuku, line 1: Let's put this logic in the route package https://github.com/Kong/kong-operator/pull/4558/changes#diff-5f3cf8034ec167a12633b66d7140da014ea46ac2e672cc483473a232e4472a78

controller/hybridgateway/route/status.go
alacuku, line 1: Can you please increase the test coverage for this file? `func UpdateRouteStatus[T gwtypes.SupportedRoute, TPtr gwtypes.SupportedRoutePtr[T]]` has 0 coverage.

WeichengWang1: Sure thing

tao12345666333 (dismissed):

controller/hybridgateway/route/status_test.go
```diff
@@ -936,6 +941,286 @@ func Test_BuildProgrammedCondition(t *testing.T) {
 	}
 }
 
+func Test_UpdateRouteStatus(t *testing.T) {
+	ctx := context.Background()
```
alacuku: Let's go with `t.Context()`

alacuku (dismissed):
Left a minor comment

/lgtm

alacuku (approved):

PR #4577, opened 2026-06-11
fix(hybridgateway): prevent traffic drops when HTTPRoute spec changes rotate resource names

alacuku:

When an HTTPRoute is updated in a way that changes the desired `KongUpstream`/`KongService`/`KongRoute`
names (e.g. adding, removing, or reordering `backendRef` entries), the reconciler would immediately
delete the old resources as orphans even though the new resources may not yet be live in Konnect. This
opened a window where Kong had no valid upstream to route traffic to.

Two complementary fixes land together:

### 1. Enforce-time ordering gate , `KongService` waits for its upstream and targets

**`reconciler_utils.go`**

Before creating a `KongService`, the reconciler now verifies:

- Its `KongUpstream` is present in the cluster **and** Programmed in Konnect — prevents Konnect from
  holding a service whose host resolves to nothing (NXDOMAIN).
- Every desired `KongTarget` for that upstream is also Programmed — ensures the upstream has healthy
  backends before the service is created and the route is pointed at it.

Two new helpers (`desiredHasUpstreamNamed`, `upstreamTargetsProgrammed`) drive this logic. The gate
only applies when the host refers to a *desired* upstream, not an external hostname, to avoid an
infinite wait on legitimate external-host services.

### 2. Cleanup-time readiness gate — defer orphan deletion until new chain is live

**`converter.go` · `http_route.go` · `controller.go`**

A new optional interface `DesiredStateReadinessChecker` lets converters signal when orphan cleanup
is safe. The reconciler checks it before running `cleanOrphanedResources` — if not ready, it requeues
instead of deleting.

`httpRouteConverter` implements the interface with a single rule: **every desired `KongRoute` must
report `status.konnect.serviceID` matching the Konnect ID of its desired `KongService`**.

The `Programmed` condition is deliberately *not* used — after a `serviceRef` repoint the route keeps
reporting `Programmed=True` against the *old* service until Konnect reprograms it, so that condition
is a false green. Only the Konnect service ID confirms the route has truly switched and the old chain
(`KongTargets` / `KongUpstream` / `KongService`) is safe to remove.

**Which issue this PR fixes**

Fixes #https://github.com/Kong/kong-operator/issues/4606

**PR Readiness Checklist**:

### DIFF
--- a/controller/hybridgateway/converter/http_route.go
+++ b/controller/hybridgateway/converter/http_route.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/go-logr/logr"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -15,6 +16,7 @@ import (
 	commonv1alpha1 "github.com/kong/kong-operator/v2/api/common/v1alpha1"
 	configurationv1 "github.com/kong/kong-operator/v2/api/configuration/v1"
 	configurationv1alpha1 "github.com/kong/kong-operator/v2/api/configuration/v1alpha1"
+	konnectv1alpha1 "github.com/kong/kong-operator/v2/api/konnect/v1alpha1"
 	hybridgatewayerrors "github.com/kong/kong-operator/v2/controller/hybridgateway/errors"
 	"github.com/kong/kong-operator/v2/controller/hybridgateway/kongroute"
 	"github.com/kong/kong-operator/v2/controller/hybridgateway/metadata"
@@ -32,7 +34,10 @@ import (
 	"github.com/kong/kong-operator/v2/pkg/vars"
 )
 
-var _ APIConverter[gwtypes.HTTPRoute] = &httpRouteConverter{}
+var (
+	_ APIConverter[gwtypes.HTTPRoute] = &httpRouteConverter{}
+	_ DesiredStateReadinessChecker    = &httpRouteConverter{}
+)
 
 // httpRouteConverter is a concrete implementation of the APIConverter interface for HTTPRoute.
 type httpRouteConverter struct {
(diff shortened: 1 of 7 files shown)


controller/hybridgateway/reconciler_utils_test.go
```diff
@@ -476,264 +475,300 @@ func TestEnforceState_CoreAndErrorPaths(t *testing.T) {
...
-		{Group: "configuration.konghq.com", Version: "v1alpha1", Kind: "KongRoute"},
-		{Group: "configuration.konghq.com", Version: "v1alpha1", Kind: "KongService"},
+	ctx := t.Context()
+	logger := logr.Discard()
+
+	routeGVK := schema.GroupVersionKind{Group: "configuration.konghq.com", Version: "v1alpha1", Kind: "KongRoute"}
+	serviceGVK := schema.GroupVersionKind{Group: "configuration.konghq.com", Version: "v1alpha1", Kind: "KongService"}
+
+	root := &gwtypes.HTTPRoute{
+		TypeMeta:   metav1.TypeMeta{APIVersion: gatewayv1.GroupVersion.String(), Kind: "HTTPRoute"},
```
pmalek, line 485: FYI for this we have https://github.com/Kong/kong-operator/blob/d36d0cf3a7f8d99fa8fac4648d5550c64723eea1/ingress-controller/internal/gatewayapi/typemeta.go#L20-L23.

No need to change it in this PR but we can use it in the long term to minimize duplication.

controller/hybridgateway/converter/http_route.go
```diff
@@ -296,6 +301,85 @@ func (c *httpRouteConverter) UpdateRootObjectStatus(ctx context.Context, logger
...
+			return false, nil
+		}
+
+		var desiredServiceID string
+		if svc.Status.Konnect != nil {
+			desiredServiceID = svc.Status.Konnect.ID
+		}
+		if desiredServiceID == "" || boundServiceID != desiredServiceID {
+			log.Debug(logger, "KongRoute not yet bound to its referenced KongService in Konnect, deferring orphan cleanup",
+				"route", client.ObjectKeyFromObject(r), "service", serviceRefName, "boundServiceID", boundServiceID, "desiredServiceID", desiredServiceID)
```
pmalek: Nit: Consider structuring these logs pairwise: `key: value` so that it's easily readable. The way it's done now, creates long lines with several entries which makes it hard to follow.

controller/hybridgateway/reconciler_utils.go
```diff
@@ -290,6 +324,49 @@ func enforceState[t converter.RootObject](ctx context.Context, cl client.Client,
...
 
+// desiredHasUpstreamNamed reports whether the desired objects contain a KongUpstream with the given name.
+// Used to decide whether a KongService's host refers to a managed upstream (and should therefore be gated
+// on it) versus an external hostname (which must not gate).
+func desiredHasUpstreamNamed(desiredObjects []unstructured.Unstructured, name string) bool {
+	for i := range desiredObjects {
+		if desiredObjects[i].GetKind() == "KongUpstream" && desiredObjects[i].GetName() == name {
+			return true
+		}
+	}
```
pmalek: This has an O(n^2) complexity being called in a `for` loop in `enforceState`. 

Consider creating a lookup map for `desiredObject` to prevent this.

controller/hybridgateway/reconciler_utils.go
```diff
@@ -290,6 +324,49 @@ func enforceState[t converter.RootObject](ctx context.Context, cl client.Client,
...
+	return false
+}
+
+// upstreamTargetsProgrammed reports whether every desired KongTarget belonging to the named KongUpstream
+// is present in the cluster and Programmed. It is used to hold a KongRoute's serviceRef switch until the
+// upstream it is about to point at actually has healthy targets, so a rollout never repoints a route onto
+// a freshly-created, still-empty upstream (which would drop traffic). An upstream with no desired targets
+// is considered ready (there is nothing to wait for).
+func upstreamTargetsProgrammed(ctx context.Context, cl client.Client, desiredObjects []unstructured.Unstructured, upstreamName string) (bool, error) {
+	for i := range desiredObjects {
```
pmalek: Same thing about O(n^2).

Consider adding a lookup map.

controller/hybridgateway/reconciler_utils_test.go
alacuku: The type is declared `internal` to `ingress-controller`.

controller/hybridgateway/converter/http_route.go
alacuku: e2d1494f649a6610a5d27bb75d781f2b178a004b

controller/hybridgateway/reconciler_utils.go
alacuku: e2d1494f649a6610a5d27bb75d781f2b178a004b

controller/hybridgateway/reconciler_utils.go
alacuku: e2d1494f649a6610a5d27bb75d781f2b178a004b

pmalek (approved):

PR #4578, opened 2026-06-11
fix: fix reconcile compare bugs

### LINKED ISSUE #3770: Reconcile compare guards trigger redundant updates and can miss Gateway address-only status changes
## Current Behavior

Several reconcile compare/update guards behave incorrectly:

1. `controller/cpextensions`:
   Prometheus plugin drift checks compare generated values against pointers to actual fields, which can trigger redundant `Update()` calls even when nothing changed.

2. `controller/gateway`:
   In hybrid mode, `dataPlaneSpecDeepEqual()` ignores `KonnectExtension` only on one side of comparison, which can report false drift.

3. `ingress-controller/internal/controllers/gateway`:
   `updateAddressesAndListenersStatus()` may miss address-only status changes after Gateway is already `Programmed`, leaving `status.addresses` stale.

## Expected Behavior

- Reconcile updates should happen only on real drift.
- Hybrid DataPlane spec comparison should ignore `KonnectExtension` consistently on both current and expected sides.
- Programmed Gateways should still update status when addresses change (even if listeners are unchanged).

## Steps To Reproduce

1. For cpextensions:
   - Reconcile a Service/ControlPlane/DataPlaneMetricsExtension where generated Prometheus plugin spec already matches existing plugin.
   - Observe repeated plugin update attempts/churn.

2. For gateway hybrid compare:
   - Use hybrid Gateway setup with Konnect extension references.
   - Trigger reconcile without meaningful DataPlane option changes.
   - Observe false out-of-date detection due to asymmetric extension filtering.

3. For ingress-controller Gateway status:
   - Reconcile unmanaged Gateway to `Programmed=True`.
   - Change publish service/LB address while listeners remain unchanged.
   - Observe `Gateway.status.addresses` not being updated.

## Operator Version

`main` (observed on current HEAD during local validation)

## `kubectl version`

N/A (logic-level issue reproducible in unit tests and controller code paths)

pmalek:

**Which issue this PR fixes**

Fixes https://github.com/Kong/kong-operator/issues/3770

**PR Readiness Checklist**:

### DIFF
--- a/controller/cpextensions/controller.go
+++ b/controller/cpextensions/controller.go
@@ -355,14 +355,14 @@ func (r *Reconciler) ensurePrometheusPlugin(
 	}
 
 	// If it exists, ensure it's up to date.
-	if !cmp.Equal(generatedPlugin.Config, &prometheusPluginActual.Config) ||
-		!cmp.Equal(generatedPlugin.Annotations, &prometheusPluginActual.Annotations) ||
-		!cmp.Equal(generatedPlugin.OwnerReferences, &prometheusPluginActual.OwnerReferences) ||
-		!cmp.Equal(generatedPlugin.Labels, &prometheusPluginActual.Labels) ||
-		!cmp.Equal(generatedPlugin.Finalizers, &prometheusPluginActual.Finalizers) ||
-		!cmp.Equal(generatedPlugin.InstanceName, &prometheusPluginActual.InstanceName) ||
-		!cmp.Equal(generatedPlugin.PluginName, &prometheusPluginActual.PluginName) ||
-		!cmp.Equal(generatedPlugin.Disabled, &prometheusPluginActual.Disabled) {
+	if !cmp.Equal(generatedPlugin.Config, prometheusPluginActual.Config) ||
+		!cmp.Equal(generatedPlugin.Annotations, prometheusPluginActual.Annotations) ||
+		!cmp.Equal(generatedPlugin.OwnerReferences, prometheusPluginActual.OwnerReferences) ||
+		!cmp.Equal(generatedPlugin.Labels, prometheusPluginActual.Labels) ||
+		!cmp.Equal(generatedPlugin.Finalizers, prometheusPluginActual.Finalizers) ||
+		!cmp.Equal(generatedPlugin.InstanceName, prometheusPluginActual.InstanceName) ||
+		!cmp.Equal(generatedPlugin.PluginName, prometheusPluginActual.PluginName) ||
+		!cmp.Equal(generatedPlugin.Disabled, prometheusPluginActual.Disabled) {
 		prometheusPluginActual.Config = generatedPlugin.Config
 		prometheusPluginActual.Disabled = generatedPlugin.Disabled
 		prometheusPluginActual.Annotations = generatedPlugin.Annotations
--- a/controller/cpextensions/controller.go
+++ b/controller/cpextensions/controller.go
@@ -355,24 +355,24 @@ func (r *Reconciler) ensurePrometheusPlugin(
 	}
 
 	// If it exists, ensure it's up to date.
-	if !cmp.Equal(generatedPlugin.Config, prometheusPluginActual.Config) ||
-		!cmp.Equal(generatedPlugin.Annotations, prometheusPluginActual.Annotations) ||
-		!cmp.Equal(generatedPlugin.OwnerReferences, prometheusPluginActual.OwnerReferences) ||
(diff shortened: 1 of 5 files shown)


controller/gateway/controller.go
```diff
@@ -1041,18 +1041,22 @@ func (r *Reconciler) patchStatus(ctx context.Context, gateway, oldGateway *gwtyp
...
+	specExpectedExtensions := specExpected.Extensions
 	// For Hybrid gateways ignore KonnectExtension in the comparison,
 	// it's managed later (separately) by the Gateway controller.
 	if isHybrid {
 		specCurrentExtensions = lo.Filter(specCurrentExtensions, func(e commonv1alpha1.ExtensionRef, _ int) bool {
 			return e.Group != konnectv1alpha2.SchemeGroupVersion.Group || e.Kind != konnectv1alpha2.KonnectExtensionKind
 		})
+		specExpectedExtensions = lo.Filter(specExpectedExtensions, func(e commonv1alpha1.ExtensionRef, _ int) bool {
+			return e.Group != konnectv1alpha2.SchemeGroupVersion.Group || e.Kind != konnectv1alpha2.KonnectExtensionKind
+		})
```
programmer04: Maybe put
```
filter := func(e commonv1alpha1.ExtensionRef, _ int) bool {
			return e.Group != konnectv1alpha2.SchemeGroupVersion.Group || e.Kind != konnectv1alpha2.KonnectExtensionKind
}
```
to emphasise that in both cases is filtered with the same criteria instead of repeating. WDYT?

controller/cpextensions/controller.go
```diff
@@ -355,14 +355,14 @@ func (r *Reconciler) ensurePrometheusPlugin(
...
-		!cmp.Equal(generatedPlugin.PluginName, &prometheusPluginActual.PluginName) ||
-		!cmp.Equal(generatedPlugin.Disabled, &prometheusPluginActual.Disabled) {
+	if !cmp.Equal(generatedPlugin.Config, prometheusPluginActual.Config) ||
+		!cmp.Equal(generatedPlugin.Annotations, prometheusPluginActual.Annotations) ||
+		!cmp.Equal(generatedPlugin.OwnerReferences, prometheusPluginActual.OwnerReferences) ||
+		!cmp.Equal(generatedPlugin.Labels, prometheusPluginActual.Labels) ||
+		!cmp.Equal(generatedPlugin.Finalizers, prometheusPluginActual.Finalizers) ||
+		!cmp.Equal(generatedPlugin.InstanceName, prometheusPluginActual.InstanceName) ||
+		!cmp.Equal(generatedPlugin.PluginName, prometheusPluginActual.PluginName) ||
+		!cmp.Equal(generatedPlugin.Disabled, prometheusPluginActual.Disabled) {
```
programmer04: Could you make the same order in this condition and all of the assignments below, it'll make code easier to read

ingress-controller/internal/controllers/gateway/gateway_controller.go
```diff
@@ -889,36 +889,39 @@ func (r *GatewayReconciler) determineListenersFromDataPlane(
...
+		if !programmed {
+			programmedCondition := metav1.Condition{
+				Type:               string(gatewayapi.GatewayConditionProgrammed),
+				Status:             metav1.ConditionTrue,
+				ObservedGeneration: gateway.Generation,
+				LastTransitionTime: metav1.Now(),
+				Reason:             string(gatewayapi.GatewayReasonProgrammed),
+			}
+			setGatewayCondition(gateway, programmedCondition)
+			info(log, gateway, "Gateway programmed status updated")
```
programmer04: It seems not to be always true, because listeners and addresses may be updated too, but it's a minor thing I think

controller/cpextensions/controller.go
pmalek: Sure, 38c5ee3fd4913e34d45c23da4d573a0062ec5a2f.

PTAL

controller/gateway/controller.go
pmalek: Good idea 👍 

cdc0f3e76fced6c0c71675e85086e9bb5cbc78f0

ingress-controller/internal/controllers/gateway/gateway_controller.go
pmalek: Dropped the "programmed" from the log.

2f620102b364e0e74ea8d5a2457f8d28de3d8158

PTAL

programmer04 (approved):

PR #4563, opened 2026-06-09
feat: enable HTTPRouteHeaderMatching conformance test for on-prem traditional router

tao12345666333:
**What this PR does / why we need it**:

Enables the Gateway API `HTTPRouteHeaderMatching` conformance test for the on-prem `traditional_compatible` router.

The traditional router now assigns `regex_priority` to generated Kong routes so overlapping `HTTPRoute` matches follow Gateway API precedence. Priority is based on a compact rank of the match's precedence class—hostname specificity, path type and length, method presence, effective header count, and effective query parameter count—plus rule/match order to break ties between equally specific matches within the same `HTTPRoute`.

Each precedence class is assigned its own range:

`classRank * traditionalMatchPriorityClassSize + withinRouteOffset`

Adding or removing an `HTTPRoute` whose precedence classes are already present does not renumber routes generated from unrelated `HTTPRoute`s. Introducing or removing a precedence class can renumber higher-ranked classes. Equally specific matches from different `HTTPRoute`s share the same priority, so their relative order remains unspecified.

Prioritized header-only matches receive a scoped synthetic catch-all regex path so Kong's traditional router applies `regex_priority`.

**Which issue this PR fixes**

Fixes https://github.com/Kong/kong-operator/issues/4454

**Special notes for your reviewer**:

- Header-only matches are kept separate from path-and-header matches so they do not inherit unrelated path priority.
- The synthetic catch-all regex path is limited to prioritized header-only routes. Matches without headers, such as method-only or query-only matches, are unchanged.
- This PR affects only the on-prem translation path; hybrid gateway behavior is not changed.

**PR Readiness Checklist**:

- [x] the `CHANGELOG.md` release notes have been updated to reflect significant changes

PR #4597, opened 2026-06-12
fix: normalize HTTPRoute header and query matches

### LINKED ISSUE #4598: Align traditional HTTPRoute translation with HTTPRouteHeaderMatching semantics
## Problem Statement

Expression routes already satisfy the Gateway API `HTTPRouteHeaderMatching` conformance behavior.

This issue tracks the remaining work for the traditional router path that was split out of PR #4563 and is now covered by PR #4597.

The goal is to make traditional HTTPRoute translation follow the same effective matching semantics for header/query handling and match precedence so the traditional router can pass `HTTPRouteHeaderMatching` as well.

## Proposed Solution

Update the traditional HTTPRoute translation path to:
- normalize HTTP header names according to Gateway API semantics
- preserve first-wins behavior for equivalent header matches in traditional route generation
- use effective header/query match counts when deriving traditional route priority
- keep expression route behavior unchanged and out of scope for this issue

## Additional Information

This work was split from PR #4563 to keep the scope of the traditional router changes focused and reviewable.

This issue is intentionally scoped to the traditional router path only.

## Acceptance Criteria

- [ ] ~~Documentation is updated with this new feature / enhancement~~ (translator behavior only)
- [ ] ~~Helm chart is updated to work with this feature / enhancement~~ (no chart change)
- [ ] As a user, HTTPRoute translation in traditional router mode matches the expected `HTTPRouteHeaderMatching` semantics.

tao12345666333:

According to the semantics of the Gateway API, header names should be case insensitive.

https://gateway-api.sigs.k8s.io/reference/api-spec/1.5/spec/#httpheadermatch

> Name is the name of the HTTP Header to be matched. Name matching MUST be
case-insensitive. (See [https://tools.ietf.org/html/rfc7230#section-3.2)](https://tools.ietf.org/html/rfc7230#section-3.2%29).
If multiple entries specify equivalent header names, only the first
entry with an equivalent name MUST be considered for a match. Subsequent
entries with an equivalent header name MUST be ignored. Due to the
case-insensitivity of header names, “foo” and “Foo” are considered
equivalent.

**Which issue this PR fixes**

Fixes #4598 

I actually discovered this issue while handling #4563. I think it is relatively independent and can be addressed as a separate PR.

**PR Readiness Checklist**:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,10 @@
   enforce-time gate delays `KongService` creation until its `KongUpstream` and
   all desired `KongTarget`s are Programmed.
   [#4577](https://github.com/Kong/kong-operator/pull/4577)
+- HTTPRoute: traditional route translation now treats header match names
+  case-insensitively and ignores later equivalent duplicates, aligning with
+  Gateway API matching semantics.
+  [#4597](https://github.com/Kong/kong-operator/pull/4597)
 - Hybridgateway: release Gateway API route finalizers once generated Kong
   resource delete requests have been issued, so immediate same-name route
   re-creates are not blocked by child resource finalizers.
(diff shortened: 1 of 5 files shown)


programmer04 (commented):
Maybe it would be worth mentioning this in the changelog, wdyt?

tao12345666333:
@programmer04 Thank you! Added https://github.com/Kong/kong-operator/pull/4597/changes/a82c7bbd3c2c8776efbc33f301b6aa19e9c3661e

programmer04 (dismissed):

programmer04 (approved):

PR #4599, opened 2026-06-12
chore: make KonnectOptions immutable in GatewayConfiguration to match KonnectGatewayControlPane

### LINKED ISSUE #2639: HybridGateway - enforce the generated `KonnectGatewayControlPlane` spec
## Problem Statement

In the `Gateway` Controller, we generate a `KonnectGatewayControlPlane` when the related `GatewayConfiguration` is configured for Konnect. This `KonnectGatewayControlPlane` is never updated and the desired spec is never enforced, leading to a potentially inconsistent state. We need to leverage the structured merge diff to be sure that the desired spec is enforced.

## Acceptance Criteria

- [ ] When the `KonnectGatewayControlPlane` is managed by the `GatewayConfiguration`, we need to keep the desired spec up to date and revert the undesired changes.

programmer04:

Active reconciliation is not a viable solution. `KonnectGatewayControlPlane` enforces its own immutability constraints `spec.konnect.authRef` is immutable once the entity is valid, and others are unconditionally immutable. Any attempt to patch these fields on a live CP would be rejected by the API server. Instead of reconciling divergence reactively, prevent it at the source: add CEL validation rules to `KonnectOptions` in `GatewayConfiguration` and reconcile only what can be reconciled. 

**Which issue this PR fixes**

Closes https://github.com/Kong/kong-operator/issues/2639

**PR Readiness Checklist**:

### DIFF
--- a/controller/gateway/controller_reconciler_utils.go
+++ b/controller/gateway/controller_reconciler_utils.go
@@ -176,15 +176,10 @@ func (r *Reconciler) enforceKonnectGatewayControlPlaneSpec(
 	old := kgcp.DeepCopy()
 	updated := false
 
-	// Enforce the auth ref only when APIAuthValid=True and Programmed=False.
-	// The field is immutable once APIAuthValid becomes True (enforced by the CEL
-	// validation rule on the CRD), so patching is only meaningful in the window
-	// between APIAuthValid being set True and the resource becoming Programmed.
-	apiAuthValidCond, hasAPIAuthValid := k8sutils.GetCondition(
-		konnectv1alpha1.KonnectEntityAPIAuthConfigurationValidConditionType, kgcp,
-	)
 	if gatewayConfig.Spec.Konnect.APIAuthConfigurationRef != nil &&
-		hasAPIAuthValid && apiAuthValidCond.Status == metav1.ConditionTrue &&
+		!k8sutils.HasConditionTrue(
+			konnectv1alpha1.KonnectEntityAPIAuthConfigurationValidConditionType, kgcp,
+		) &&
 		!k8sutils.IsProgrammed(kgcp) {
 		desired := *gatewayConfig.Spec.Konnect.APIAuthConfigurationRef
 		if kgcp.Spec.KonnectConfiguration.APIAuthConfigurationRef != desired {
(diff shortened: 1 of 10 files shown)


controller/gateway/controller_reconciler_utils.go
```diff
@@ -155,6 +155,47 @@ func (r *Reconciler) createKonnectGatewayControlPlane(
...
+		return false, nil
+	}
+
+	old := kgcp.DeepCopy()
+	updated := false
+
+	// Enforce the auth ref only when the control plane is not yet Programmed
+	// or APIAuthValid (the field is immutable once those conditions are True).
+	if gatewayConfig.Spec.Konnect.APIAuthConfigurationRef != nil &&
+		!k8sutils.IsProgrammed(kgcp) {
```
tao12345666333, line 183: `KonnectGatewayControlPlane` has two  immutability guards for `spec.konnect.authRef`
https://github.com/Kong/kong-operator/blob/8ce55706f9af2c8d89082c8be9eed808de1be14c/api/konnect/v1alpha2/konnect_gateway_controlplane_types.go#L24-L25

Here we only checked the `Programmed=True`, I think it would be better to consider `APIAuthValid=True && Programmed=False` as well

controller/gateway/controller_reconciler_utils.go
```diff
@@ -155,6 +156,53 @@ func (r *Reconciler) createKonnectGatewayControlPlane(
...
+
+	// Enforce the auth ref only when APIAuthValid=True and Programmed=False.
+	// The field is immutable once APIAuthValid becomes True (enforced by the CEL
+	// validation rule on the CRD), so patching is only meaningful in the window
+	// between APIAuthValid being set True and the resource becoming Programmed.
+	apiAuthValidCond, hasAPIAuthValid := k8sutils.GetCondition(
+		konnectv1alpha1.KonnectEntityAPIAuthConfigurationValidConditionType, kgcp,
+	)
+	if gatewayConfig.Spec.Konnect.APIAuthConfigurationRef != nil &&
+		hasAPIAuthValid && apiAuthValidCond.Status == metav1.ConditionTrue &&
```
tao12345666333:
```suggestion
		!k8sutils.HasConditionTrue(konnectv1alpha1.KonnectEntityAPIAuthConfigurationValidConditionType, kgcp) &&
```



I think we should reverse the logic here, as well as the logic in the corresponding test case.

https://github.com/Kong/kong-operator/blob/8ce55706f9af2c8d89082c8be9eed808de1be14c/api/konnect/v1alpha2/konnect_gateway_controlplane_types.go#L24-L25

programmer04: I've adjusted PTAL fc44122734f201b554b0049afd729d9e3a17b910

tao12345666333 (approved):

PR #4628, opened 2026-06-16
chore(crds): use crdify to detect breaking changes

### LINKED ISSUE #3676: Konnect CRDs: add kubernetes-sigs/crdify or similar to detect breaking changes to CRDs
### Problem statement

Since Konnect CRDs are going to be generated now, we need to be able to detect breaking changes.

One way to do this is to use https://github.com/kubernetes-sigs/crdify.

This task tracks adding crdify to CI and making it block the merge when breaking changes are detected.

### Acceptance criteria

- [ ] Breaking changes to generated CRDs are detected in CI with `crdify`

tao12345666333:

**Which issue this PR fixes**

Fixes #3676

**PR Readiness Checklist**:


tao12345666333:
Running job log

https://github.com/Kong/kong-operator/actions/runs/27630481280/job/81703991271?pr=4628

.github/workflows/__verify_crd_breaking_changes.yaml
```diff
@@ -0,0 +1,38 @@
+name: verify CRD breaking changes
+
+on:
+  workflow_call: {}
```
pmalek: Do we actually need to run this separately and through the workflow_call?

I was thinking that we could put this under https://github.com/Kong/kong-operator/blob/99199e3a8a702f9ab68d2faa87fe68b438ccfe1c/.github/workflows/__verify.yaml, WDYT?

scripts/verify-crd-breaking-changes.sh
```diff
@@ -0,0 +1,134 @@
...
+		continue
+	fi
+
+	old_file="${tmpdir}/${basename}"
+	git show "${base_revision}:${base_file}" > "${old_file}"
+
+	echo "Checking ${basename}"
+	if ! "${CRDIFY_BIN}" "file://${old_file}" "file://${current_file}"; then
+		status=1
+	fi
```
pmalek, line 134: We have `set -o errexit` set on the top, so I think that if this fails the script will stop instead of continuing, no?

scripts/verify-crd-from-oas-breaking-changes.sh
```diff
@@ -0,0 +1,134 @@
...
+readonly RELATIVE_CRD_DIR
+
+CURRENT_CRD_DIR="${SCRIPT_ROOT}/${RELATIVE_CRD_DIR}"
+readonly CURRENT_CRD_DIR
+
+is_crd_from_oas_generated_basename() {
+	local basename="${1}"
+
+	case "${basename}" in
+		konnect.konghq.com_*.yaml | configuration.konghq.com_eventgateway*.yaml)
```
pmalek: This comment is related to other parts as well: in general, this workflow and makefile target will work with all the CRDs, not only the generated ones, no?

I believe that should be the case but just checking because I believe the description should be updated to reflect that.

.github/workflows/__verify_crd_breaking_changes.yaml
tao12345666333: Yes, we don't need a separate workflow; we can move it into the verify workflow. 

I created a separate workflow here because the checkout process uses the `fetch-depth: 0` configuration. I don't want it to affect other jobs, which don't need to know about the full git history. WDYT?

scripts/verify-crd-breaking-changes.sh
tao12345666333: No, this is an exception in the bash shell.

Even if we set `set -o errexit`, it won't exit as long as it's used in an `if` conditional statement.


Many years ago, I spent some time [summarizing some exceptions](https://moelove.info/2021/11/21/Shell-%E8%84%9A%E6%9C%AC%E9%81%BF%E5%9D%91%E6%8C%87%E5%8D%97%E4%B8%80/) in shell scripting.It has too many general patterns and exceptions. 🤣

scripts/verify-crd-from-oas-breaking-changes.sh
tao12345666333: I'm currently limiting it to checking CRDs generated via crd-from-oas, but you're right, other CRDs can also be checked with it. Let me update this.

.github/workflows/tests.yaml
```diff
@@ -165,6 +165,12 @@ jobs:
     uses: ./.github/workflows/__verify.yaml
     secrets: inherit
 
+  verify-crd-breaking-changes:
+    needs: [check-docs-only]
+    if: ${{ needs.check-docs-only.outputs.docs_only != 'true' }}
+    uses: ./.github/workflows/__verify_crd_breaking_changes.yaml
+    secrets: inherit
```
pmalek: This is still in place (not under `verify` workflow).

scripts/verify-crd-from-oas-breaking-changes.sh
pmalek: I believe this still needs an update

.github/workflows/tests.yaml
tao12345666333: updated https://github.com/Kong/kong-operator/pull/4628/changes/c70f5cc6c65cee394b7aff395eb8032956d31fcd#diff-384f6ab587cbb0d154a26508bfbbd148fe52253cf386a525c114b56520504a27R36

scripts/verify-crd-from-oas-breaking-changes.sh
tao12345666333: updated https://github.com/Kong/kong-operator/pull/4628/commits/c70f5cc6c65cee394b7aff395eb8032956d31fcd PTAL, thank you!

pmalek (approved):

PR #4636, opened 2026-06-17
test(conformance): wait for resource cleanup to converge

### LINKED ISSUE #4635: HybridGateway conformance test flakiness
## Current Behavior

The Gateway API conformance suite (hybrid gateway type) intermittently fails
tests that reuse resource names, because a test can start before the previous
test's same-named resources have finished terminating.

Conformance runs top-level tests sequentially, and several reuse the same
resource names. For example the `reference-grant` HTTPRoute is created by both
`HTTPRouteInvalidReferenceGrant` and `HTTPRouteReferenceGrant`.

A test's per-manifest cleanup only issues `Delete` and returns without waiting.
With the hybrid gateway type, whose Konnect finalizers make teardown slower, the
previous test's HTTPRoute can still be terminating when the next test applies its
manifests. The conformance applier then updates the dying object, which is
finalized away moments later, and the test fails with:

```
Updating reference-grant HTTPRoute
error fetching HTTPRoute: httproutes.gateway.networking.k8s.io "reference-grant" not found
error waiting for HTTPRoute to have parents matching expectations
```

In a full run these tests are far apart so it usually passes, but it is flaky and
reliably reproducible when they run back to back, for example when narrowed with
a `go test -run` filter.

~~Analyzing CI failures such as
[this run](https://github.com/Kong/kong-operator/actions/runs/27617761035/job/81658098270)
points to the same behavior. A previous test leaves Kong entities behind in
Konnect, and because those entities share the same name and spec, the next test
cannot recreate them and fails with constraint errors returned by Konnect.~~

## Expected Behavior

A test should not start while a prior same-named resource is still finalizing.

alacuku:

Gateway API conformance runs top-level tests sequentially and several reuse the same resource names. For example the "reference-grant" HTTPRoute is created by both HTTPRouteInvalidReferenceGrant and HTTPRouteReferenceGrant. A test's per-manifest cleanup only issues the Delete and returns without waiting, so with the hybrid gateway type, whose Konnect finalizers make teardown slower, a deleted resource can still be terminating when a later same-named test applies its manifests, then disappear mid-test. The applier logs "Updating reference-grant HTTPRoute" and the test then fails with "error fetching HTTPRoute ... not found". This is reliably hit when such tests run back to back, for example when narrowed via a -run filter.

Add a post-test suite Hook that waits until the just-deleted HTTPRoutes and ReferenceGrants in the conformance namespaces are fully gone before the next test runs. It returns immediately when nothing is terminating, so it costs effectively nothing in the common case, and it is bounded so a stuck finalizer cannot hang the suite.

Additionally, during suite teardown keep the controller manager alive until the conformance namespaces are fully terminated. Deleting Gateways and KonnectGatewayControlPlanes cascades to Konnect entity CRs such as KongService, KongRoute, KongUpstream and KongTarget, which carry their own finalizers. Previously the manager could be stopped before removing them, leaving the entities dangling in the cluster and in Konnect. A namespace cannot finish terminating while it still holds finalizer-bearing resources, so waiting for the conformance namespaces to disappear guarantees every operator-owned entity has been reconciled away first.

**Which issue this PR fixes**

Fixes #4635

**PR Readiness Checklist**:

### DIFF
--- a/test/conformance/conformance_test.go
+++ b/test/conformance/conformance_test.go
@@ -3,6 +3,7 @@ package conformance
 import (
 	"context"
 	"fmt"
+	"strings"
 	"testing"
 	"time"
 
@@ -14,7 +15,10 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
+	gatewayv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
+	gatewayv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
 	"sigs.k8s.io/gateway-api/conformance"
 	conformancev1 "sigs.k8s.io/gateway-api/conformance/apis/v1"
 	conformanceconfig "sigs.k8s.io/gateway-api/conformance/utils/config"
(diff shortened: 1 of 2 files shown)


programmer04 (approved):
🆒

tao12345666333:
I will try to backport this to release/2.2.x, If the issue of conformance failure cannot be resolved, I will cancel the label.

PR #4650, opened 2026-06-19
fix(konnect): prevent orphaned and duplicate Konnect entities on create/delete race

### LINKED ISSUE #4649: Race between Konnect creation, status persistence, and deletion orphans/duplicates Konnect entities
## Current Behavior

Creating an entity in Konnect happens in two steps that are not atomic: first the
entity is created in Konnect, then its Konnect ID is persisted to the resource
status. There is a window between these two steps, and there is a race between
that window and the deletion of the resource.

If the resource is deleted (or the operator is restarted, or the status update
fails) after the entity has been created in Konnect but before its ID has been
persisted to the status, the operator loses track of the entity. On deletion it
has no Konnect ID to work with and does not know which Konnect entity to remove,
so the entity is left behind in Konnect (orphaned) and never cleaned up.

The same window can also lead to duplicate entities being created in Konnect: a
resource that has already been created, but whose status has not yet caught up
to show the Konnect ID, can be reconciled again and trigger a second create.

## Expected Behavior

Every entity that is created in Konnect is reliably tracked and cleaned up:

- If a resource is deleted, its corresponding Konnect entity is always removed,
  even if the operator was restarted or interrupted before the Konnect ID was
  written to the status.
- A resource that has already been created in Konnect is never created a second
  time, even when the operator is acting on a stale cached view of the resource.

No leftover or duplicate entities should remain in Konnect.

## Steps To Reproduce

Running the conformance test suite leaves leftover entities in Konnect after the
tests complete.

alacuku:

Creating a Konnect entity and persisting its Konnect ID to the resource status are two non-atomic steps. If the resource was deleted, the operator restarted, or the status write failed within that window, the operator lost track of the entity: it had no Konnect ID and could neither clean it up (orphan) nor avoid re-creating it from a stale cached status (duplicate).

- Add the cleanup finalizer before creating the entity in Konnect, so cleanup is always possible regardless of when the create is interrupted.
- Keep the freshly created Konnect ID in an in-memory store until the cached status reflects it, so the create/status-persist window is bridged.
- On deletion, recover a missing Konnect ID from the in-memory store, falling back to probing Konnect by Kubernetes UID (or by name for the tag-less Cloud Gateway types) so entities are cleaned up even after an operator restart.
- Use the in-memory store as the authoritative "already created" signal to avoid creating a duplicate when the cached status has not yet caught up.
- Add unit tests for the store, the de-dup guard, the delete-time recovery and the Konnect ID probe, plus an envtest asserting the finalizer is present even when the Konnect create fails.

**Which issue this PR fixes**

Fixes #4649 

**PR Readiness Checklist**:

### DIFF
--- a/controller/konnect/ops/ops.go
+++ b/controller/konnect/ops/ops.go
@@ -140,62 +140,7 @@ func Create[
 		// If there was a conflict on the create request, we can assume the entity already exists.
 		// We'll get its Konnect ID by listing all entities of its type filtered by the Kubernetes object UID.
 		var id string
-		switch ent := any(e).(type) {
-		case *konnectv1alpha2.KonnectGatewayControlPlane:
-			id, errGet = getControlPlaneForUID(ctx, sdk.GetControlPlaneSDK(), sdk.GetControlPlaneGroupSDK(), cl, ent)
-		case *konnectv1alpha1.KonnectCloudGatewayNetwork:
-			// NOTE: since Cloud Gateways resource do not support labels/tags,
-			// we can't reliably get the Konnect ID for a Cloud Gateway Network
-			// given a K8s object UID.
-			// For now this code uses a list, using a name filter, to get the Konnect ID.
-			id, err = getKonnectNetworkMatchingSpecName(ctx, sdk.GetCloudGatewaysSDK(), ent)
-		case *konnectv1alpha1.KonnectCloudGatewayDataPlaneGroupConfiguration:
-			// TODO: can't get the ID for a DataPlaneGroupConfiguration
-			// as this resource type does not support labels/tags.
-		case *konnectv1alpha1.KonnectCloudGatewayTransitGateway:
-			id, err = getKonnectTransitGatewayMatchingSpecName(ctx, sdk.GetCloudGatewaysSDK(), ent)
-		case *configurationv1alpha1.KongService:
-			id, errGet = getKongServiceForUID(ctx, sdk.GetServicesSDK(), ent)
-		case *configurationv1alpha1.KongRoute:
-			id, errGet = getKongRouteForUID(ctx, sdk.GetRoutesSDK(), ent)
-		case *configurationv1alpha1.KongSNI:
-			id, errGet = getKongSNIForUID(ctx, sdk.GetSNIsSDK(), ent)
-		case *configurationv1.KongConsumer:
-			id, errGet = getKongConsumerForUID(ctx, sdk.GetConsumersSDK(), ent)
-		case *configurationv1beta1.KongConsumerGroup:
-			id, errGet = getKongConsumerGroupForUID(ctx, sdk.GetConsumerGroupsSDK(), ent)
-		case *configurationv1alpha1.KongKeySet:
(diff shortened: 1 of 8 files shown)


controller/konnect/ops/ops.go
```diff
@@ -266,21 +211,121 @@ func Create[
...
+	TEnt constraints.EntityType[T],
+](
+	ctx context.Context,
+	sdk sdkops.SDKWrapper,
+	cl client.Client,
+	e TEnt,
+) (string, error) {
+	switch ent := any(e).(type) {
+	case *konnectv1alpha2.KonnectGatewayControlPlane:
+		return getControlPlaneForUID(ctx, sdk.GetControlPlaneSDK(), sdk.GetControlPlaneGroupSDK(), cl, ent)
```
tao12345666333, line 235: I believe there is a special scenario here; unlike other functions that only perform query operations, this one also includes `setGroupMembers()` operations, which leads to the possibility of issues arising.

https://github.com/Kong/kong-operator/blob/fb9ede2af4ad175af71e3aa3e1ffdf055bb25662/controller/konnect/ops/ops_controlplane.go#L268-L283

It may still return a `KonnectEntityCreatedButRelationsFailedError` error after obtaining the id, but the subsequent process does not handle that error. This may lead to a continuous return of errors, causing the resource to remain in a terminating state.

tao12345666333 (commented):
Overall LGTM, leaving one comment.

controller/konnect/ops/ops.go
alacuku: Nice catch, fixed in: d1782c70e7d23691f400e7ba40e0bf05817faaf5

tao12345666333 (approved):

controller/konnect/ops/ops.go
```diff
@@ -266,21 +211,128 @@ func Create[
...
-			)
+		// The Konnect ID was never persisted to the status. This can happen if the
+		// operator created the entity in Konnect but was interrupted before writing
+		// the ID to the status (and the in-memory record was lost, e.g. on restart).
+		// Probe Konnect to recover the ID so the entity can still be cleaned up.
+		id, errGet := getKonnectIDForUID(ctx, sdk, cl, ent)
+		if errGet != nil {
+			_, uidNotFound := errors.AsType[EntityWithMatchingUIDNotFoundError](errGet)
+			_, idNotFound := errors.AsType[EntityWithMatchingIDNotFoundError](errGet)
+			_, noConflictHandling := errors.AsType[ConflictOnCreateButNoConflifctHandlingImplementedError](errGet)
```
pmalek, line 306:
```suggestion
			_, noConflictHandling := errors.AsType[ConflictOnCreateButNoConflictHandlingImplementedError](errGet)
```

pmalek (approved):
Minor typo, overall this is a very good change 👍 🚀

controller/konnect/ops/ops.go
tao12345666333: good catch!

alacuku: Good catch, note this typo is pre-existing, not introduced by this PR. Let's handle this in another PR.

PR #4663, opened 2026-06-23
Fix(controller): Copy old object before removing condition in `StatusWithoutCondition`

### LINKED ISSUE #4248: Stale cross-namespace conditions are never cleared when using patch.StatusWithoutCondition
## Current Behavior

[patch.StatusWithoutCondition ](https://github.com/Kong/kong-operator/blob/aa35235a486554a9ea062a1bd39fc6e5b203ea3c/controller/pkg/patch/statuscondition.go#L95)is supposed to remove a condition from an entity's status by issuing a merge patch. However it has a bug that makes it a complete no-op: it captures old via DeepCopy after calling RemoveCondition, which mutates ent in place. The resulting merge patch is always empty, the condition is never removed from the API server.

Any place in the reconciler that relies on StatusWithoutCondition to clear a stale condition is silently broken. 

## Expected Behavior

The stale condition should be removed.

## Steps To Reproduce

 1. Create a KonnectAPIAuthConfiguration (konnect-auth) in svc-namespace with a
    valid Konnect PAT and server URL
 2. Create a KonnectGatewayControlPlane (test-cp) in svc-namespace referencing
    konnect-auth (same-namespace). Wait for it to be Programmed=True
 3. Create a KonnectGatewayControlPlane (test-cp) in cp-namespace referencing
    konnect-auth in svc-namespace (cross-namespace). Create a KongReferenceGrant
    in svc-namespace allowing cp-namespace KonnectGatewayControlPlanes to use
    KonnectAPIAuthConfiguration. Wait for it to be Programmed=True
 4. Create a KongService (stale-svc) in svc-namespace with controlPlaneRef pointing
    to test-cp in cp-namespace (cross-namespace). Do NOT create a KongReferenceGrant
    for the KongService → KonnectGatewayControlPlane ref. The operator sets
    ResolvedRefs=False/RefNotPermitted on the KongService
 5. Verify the False condition is set:
      kubectl get kongservice stale-svc -n svc-namespace \
        -o jsonpath='{.status.conditions[?(@.type=="ResolvedRefs")]}'
     
 6. Switch the KongService controlPlaneRef to same-namespace test-cp (this is
    allowed because Programmed=False, so the field is still mutable):
      kubectl patch kongservice stale-svc -n svc-namespace --type=merge \
        -p '{"spec":{"controlPlaneRef":{"type":"konnectNamespacedRef","konnectNamespacedRef":{"name":"test-cp"}}}}
 7. Observe that ResolvedRefs=False is NOT cleared (the bug):
      kubectl get kongservice stale-svc -n svc-namespace \
        -o jsonpath='{.status.conditions[?(@.type=="ResolvedRefs")]}'

randmonkey:

Fixes the `StatusWithoutCondition` in the utils so that the not required `ResolvedRef` condition in Konnect entities when a cross-namespace reference dissappears can be correctly removed.

**Which issue this PR fixes**

Fixes #4248

**PR Readiness Checklist**:

### DIFF
--- a/controller/konnect/reconciler_controlplaneref.go
+++ b/controller/konnect/reconciler_controlplaneref.go
@@ -137,6 +137,34 @@ func handleControlPlaneRef[T constraints.SupportedKonnectEntityType, TEnt constr
 	return ctrl.Result{}, nil
 }
 
+// entityHasCrossNamespaceRefs reports whether ent has any cross-namespace
+// reference (Secret, KongCACertificate, KongCertificate/ClientCertificate) whose
+// shared ResolvedRefs condition is owned by a dedicated handler. The
+// control-plane-ref handler must not remove the ResolvedRefs condition while any
+// such reference exists, otherwise it would clobber what those handlers set
+// (they run later in Reconcile).
+func entityHasCrossNamespaceRefs[
+	T constraints.SupportedKonnectEntityType, TEnt constraints.EntityType[T],
+](ent TEnt) bool {
+	ns := ent.GetNamespace()
+	isCross := func(n *string) bool { return n != nil && *n != "" && *n != ns }
+
+	for _, r := range getSecretRefs(ent) {
+		if isCross(r.Namespace) {
+			return true
+		}
+	}
+	for _, r := range getKongCACertificateRefs(ent) {
+		if isCross(r.Namespace) {
+			return true
+		}
+	}
+	if r, ok := getKongCertificateRef(ent).Get(); ok && isCross(r.Namespace) {
+		return true
+	}
+	return false
+}
+
 func conditionMessageReferenceKonnectAPIAuthConfigurationInvalid(apiAuthRef types.NamespacedName) string {
 	return fmt.Sprintf("referenced KonnectAPIAuthConfiguration %s is invalid", apiAuthRef)
 }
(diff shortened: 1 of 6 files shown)


pmalek:
@randmonkey There are failing tests in the pipeline. Can you take care of these before reviewing?

randmonkey:
> @randmonkey There are failing tests in the pipeline. Can you take care of these before reviewing?

https://github.com/Kong/kong-operator/pull/4663/commits/3a7022e23e69f0d8932871bb60746baab9bdbb7a that updates the logic of handling the `ResolvedRef` condition. It only deletes the condition if there are no cross-namespace reference left.

controller/pkg/patch/statuscondition.go
```diff
@@ -101,11 +101,11 @@ func StatusWithoutCondition[T interface {
 	ent T,
 	conditionType string,
 ) (res ctrl.Result, err error) {
+	old := ent.DeepCopyObject().(T)
 	if !k8sutils.RemoveCondition(kcfgconsts.ConditionType(conditionType), ent) {
 		return ctrl.Result{}, nil
 	}
```
WeichengWang1, line 111: there will be an extra deepcopy if there is no such condition.
how about check if the condition exist first and then deepcopy if existed?

controller/konnect/reconciler_controlplaneref.go
```diff
@@ -162,11 +190,17 @@ func ensureKongReferenceGrant[
 	if cpRef.Type != commonv1alpha1.ControlPlaneRefKonnectNamespacedRef ||
 		toNamespace == "" ||
 		toNamespace == fromNamespace {
-		if res, errStatus := patch.StatusWithoutCondition(
-			ctx, cl, ent,
-			configurationv1alpha1.KongReferenceGrantConditionTypeResolvedRefs,
-		); errStatus != nil || !res.IsZero() {
-			return res, errStatus
+		// The ControlPlane ref doesn't require a grant. Only remove the shared
+		// ResolvedRefs condition if no other cross-namespace reference of this
```
WeichengWang1, line 194: maybe in the future we can have a central ResolvedRefs computation cuz ResolvedRefs condition is shared and we have to check if another owner exist when we process it.

not blocking, just a minor comment

randmonkey: https://github.com/Kong/kong-operator/issues/4711

controller/pkg/patch/statuscondition.go
randmonkey: Done.

WeichengWang1 (approved):
/lgtm

PR #4664, opened 2026-06-23
fix(konnect): defer reconcile after recovering pending Konnect ID

alacuku:

When the informer cache serves a stale, ID-less status after a create, reconcilePendingKonnectID restores the ID from the in-memory store and patches the status. The patch refreshes ent from the server response, so continuing the loop would issue an enforcement Konnect call (e.g. ops.Update / UpsertPlugin) on an object stitched together from a stale cache plus the targeted ID patch — an extra, timing-dependent API call.

Stop the reconcile pass once the ID is written (op.Updated) and let the status-write watch event drive a fresh reconcile from a consistent cache, matching the create path's convention. The bridge store entry is purged on that subsequent pass. This keeps Konnect call counts deterministic.

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:

### DIFF
--- a/controller/konnect/reconciler_generic.go
+++ b/controller/konnect/reconciler_generic.go
@@ -807,16 +807,23 @@ func (r *KonnectEntityReconciler[T, TEnt]) Reconcile(ctx context.Context, ent TE
 // again on that basis would produce a duplicate Konnect entity. The in-memory
 // store records the Konnect ID as soon as the entity is created, so:
 //   - if the cached status lacks the ID but the store has it, restore and persist
-//     it. ApplyStatusPatchIfNotEmpty writes through to the API server and refreshes
-//     ent from the response, so the rest of the loop runs on a consistent object.
-//     Re-persisting is idempotent and also heals the case where the post-create
-//     status write failed.
+//     it. When the status is actually patched (op.Updated), reconciliation stops
+//     for this pass so the watch event from the status write triggers a fresh
+//     reconcile with a fully-consistent cached object. The status patch refreshes
+//     ent from the server response, so continuing here would run the rest of the
+//     loop and issue an enforcement Konnect update (e.g. ops.Update) on an object
+//     stitched together from a stale cache plus the targeted ID patch — an extra,
+//     timing-dependent API call. Deferring to a fresh reconcile matches the
+//     create path's convention (the status write drives the next pass) and keeps
+//     the behaviour deterministic. The store entry is purged in that fresh
+//     reconcile (see the block below). Re-persisting is idempotent and also heals
+//     the case where the post-create status write failed.
 //   - once the status carries the ID, the bridge entry has served its purpose and
 //     is dropped.
 //
 // It returns stop=true when the caller should return the provided result/error (a
-// conflict requeue or a persist error); otherwise reconciliation continues with a
-// consistent ent.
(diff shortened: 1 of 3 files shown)


programmer04 (approved):

ingress-controller/test/kongintegration/containers/kong.go
pmalek, line 1: Why do we need this change? The commit message says that it fixes remote docker host use cases but I've been running docker in lima for a while and it's been working for me.

Is there a particular test which should fail in those setups which is fixed now?

controller/konnect/reconciler_generic.go
```diff
@@ -826,12 +833,23 @@ func (r *KonnectEntityReconciler[T, TEnt]) reconcilePendingKonnectID(
...
+			if err != nil {
 				if apierrors.IsConflict(err) {
 					return ctrl.Result{Requeue: true}, true, nil
 				}
 				return ctrl.Result{}, true, fmt.Errorf("failed to persist recovered Konnect ID for %s: %w", pendingKey, err)
 			}
+			if result == op.Updated {
+				// ID was recovered and written to the cluster. Stop this reconcile
+				// pass so the status-update watch event triggers a fresh reconcile
+				// from a consistent cache. Continuing would run the enforcement
```
pmalek, line 846: > from a consistent cache

The triggered (subsequent) reconcile **can** still have an outdated view on the cluster (stale cache).

https://github.com/kubernetes-sigs/controller-runtime/issues/3320 will help in this.

ingress-controller/test/kongintegration/containers/kong.go
alacuku: When running docker in a minikube cluster i get errors like: 
```
=== RUN   TestControlPlaneReferenceHandling
    control_plane_reference_test.go:33: starting envtest environment for test TestControlPlaneReferenceHandling...
    control_plane_reference_test.go:33: envtest environment (v1.36.0) started at https://127.0.0.1:55968/
    kong.go:130: failed validating Kong version: making HTTP request: Get "http://localhost:56181/": dial tcp [::1]:56181: connect: connection refused
```

controller/konnect/reconciler_generic.go
alacuku: This change is only here to keep the envtests happy. They use sdk.Mock to assert the exact number of Konnect API requests, and after the deduplication fix those counts stopped being deterministic. Depending on how stale the cache is, the reconciler can fire an extra (idempotent) Konnect call during the create race, so the strict mock assertions started failing on and off. The early return just stops the dedup path before that extra call so the request count stays predictable.

Honestly though, this approach is too drastic. It adds production control flow (an extra reconcile pass plus some redundant status patches while the cache catches up) purely to hide a harmless, idempotent Konnect call, and it doesn't fix any real bug. So I'm not going down this road. I'll make the tests robust instead by letting them tolerate that optional enforcement call (assert on the Create/Delete counts that actually matter and treat the enforcement Get/Update as optional) rather than bending the reconciler just to keep a strict count happy.

I'm going to close this PR and split it up in two different PRs. Thanks for pointing out the `stale cache` issue..

ingress-controller/test/kongintegration/containers/kong.go
alacuku: https://github.com/Kong/kong-operator/pull/4672

PR #4673, opened 2026-06-24
Conformance: enable GRPCRoute conformance tests for on-prem

### LINKED ISSUE #2345: conformance: `GRPCRouteHeaderMatching`, `GRPCExactMethodMatching`, `GRPCRouteListenerHostnameMatching`
## Problem Statement

Tests
- `GRPCRouteHeaderMatching`
- `GRPCExactMethodMatching`

are skipped because to proxy different gRPC services and route requests based on the Header or Method, it is necessary to create separate catch-all routes for them. However, Kong does not define priority behavior in this situation unless priorities are manually added.

For test `GRPCRouteListenerHostnameMatching`, Kong's router requires `priority` to be specified for routes. We cannot provide that for routes that are part of the conformance suite.

Learn more 
- https://github.com/Kong/kubernetes-ingress-controller/issues/6144

### LINKED ISSUE #4621: conformance: GRPCExactMethodMatching
# Summary
`GRPCExactMethodMatching` conformance test fails. Two distinct problems.

## Problem 1 — exact matchType not effective (FIXED, validated)
`EchoTwo` calls were routed to `Echo` because the generated Kong path
regex was unanchored, so `Echo` prefix-matched `EchoTwo`. Fix — anchor the
method (and quote literals) for Exact matches:

After fixing it in my local env, 
```go
	if matchType == gatewayapi.GRPCMethodMatchExact {
		if matchService != nil {
			service = regexp.QuoteMeta(service)
		}
		if matchMethod != nil {
			method = regexp.QuoteMeta(method) + "$"
		}
	}
```
After this, Echo and EchoTwo route correctly and pass.

## Problem 2 — unmatched method returns NotFound(5), test expects Unimplemented(12) (OPEN)

The conformance GRPCRoute defines rules only for Echo and EchoTwo.
EchoThree matches no rule → Kong no-route → HTTP 404. The test expects
the gRPC code Unimplemented(12); Kong yields NotFound(5).

In Kong, https://github.com/Kong/kong/blame/master/kong/pdk/response.lua#L92 translate HTTPCode 404 to GRPCStatus NotFound(5)
While in grpc, https://github.com/grpc/grpc-go/blob/master/internal/transport/http_util.go#L79 HTTPCode 404 was translated to Unimplemented(12), same in Envoy (https://github.com/envoyproxy/envoy/blob/main/source/common/grpc/status.cc#L16)

## Others
Btw, should we create a catch-all route to respond to gRPC Reflection Requests? Conformance tests don't rely on Reflection.

# Possible Solution
if catch-all route to respond to gRPC Reflection Requests is necessary (not a requirement of gateway api):
EchoThree will routed by the catch-all route and responded with Unimplemented(12), GRPCExactMethodMatching should work, but problems described in https://github.com/Kong/kubernetes-ingress-controller/issues/6144 need to be solved.

if catch-all route to respond to gRPC Reflection Requests is not necessary:
## Fix Kong
Change HTTP_TO_GRPC_STATUS[404] 5→12
Needs Kong release; out of operator's direct control; not sure any further effects
##  Operator catch-all  + terminate→12
Per (host, service) with method matches, emit a low-priority route (`~/<service>/.+`) + request-termination yielding grpc-status 12; specific routes get higher  regex_priority

### LINKED ISSUE #4710: conformance: enable GRPCRouteWeight - on-prem
GRPCRouteWeight is skipped now. We should try to enable it after GRPCRoute implemented for on-prem.

WeichengWang1:

Enables four previously-skipped GRPCRoute Gateway API conformance tests for on-prem gateways (GRPCExactMethodMatching, GRPCRouteHeaderMatching, GRPCRouteListenerHostnameMatching, GRPCRouteWeight) and fixes the underlying routing bugs that kept them skipped.

**Changes**
1. Exact gRPC method matching is now anchored (subtranslator/grpcroute.go)
The generated Kong path regex for Exact method matches was unanchored and did not escape literals, so .../Echo also matched .../EchoTwo (prefix collision). Method literals are now run through regexp.QuoteMeta() and anchored with a trailing $.
  - Golden: ~/grpcbin.GRPCBin/Get → ~/grpcbin\.GRPCBin/Get$
2. Expressions router inherits listener hostnames for GRPCRoutes (subtranslator/grpcroute_atc.go, grpcroute.go, translate_grpcroute.go)
With the expressions router, SplitGRPCRoute read only spec.hostnames. A GRPCRoute that omits hostnames and relies on its attached Gateway listener hostnames produced routes with no host predicate, collapsing all traffic onto a single backend. Hostname resolution (spec hostnames, else listener hostnames honoring sectionName) is now factored into a shared getEffectiveHostnamesForGRPCRoute helper used by both the traditional and expressions paths.
3. Fix data race on the supported-features slices (pkg/gatewayapi/supportedfeatures.go)
GetSupportedFeatures returned the package-level slices directly; the GatewayClass reconciler then sorted them in place (slices.Sort) while the conformance suite read the same backing array — a data race flagged by -race. The function now returns slices.Clone(...).

**Which issue this PR fixes**

Fixes #4621 partially, resolved problem 1 in the issue, problem 2 should be resolved by kong, and this pr will then ready to review
Fixes #2345,  with the expressions router, a `GRPCRoute` that specifies no `spec.hostnames` now correctly inherits the hostnames from the listeners of its attached Gateways (honoring `sectionName`). 
Fixes #4710 

`GRPCExactMethodMatching` `GRPCRouteHeaderMatching` `GRPCRouteListenerHostnameMatching` are still skipped until FTI-7663 released.

**PR Readiness Checklist**:


ingress-controller/internal/dataplane/translator/subtranslator/grpcroute.go
```diff
@@ -35,6 +36,38 @@ func getGRPCMatchDefaults() (
...
+	}
+	if matchService == nil {
+		service = serviceMap[matchType]
+	} else {
+		service = *matchService
+	}
+
+	if matchType == gatewayapi.GRPCMethodMatchExact {
+		if matchService != nil {
+			service = regexp.QuoteMeta(service)
```
randmonkey, line 70: Could we add the comments to tell us what chars in regexp escape chars in `Service` or `Method` so this is required? (Like `.` ?)

randmonkey (dismissed):
lgtm, a minor comment.

ingress-controller/internal/dataplane/translator/subtranslator/grpcroute.go
WeichengWang1: done

randmonkey (approved):

PR #4675, opened 2026-06-24
test(crdify): add a ack label to bypass failed job

tao12345666333:

There is a scenario where, when the description is modified, it **may** not be a breaking change, so we can choose to ignore it.

https://github.com/Kong/kong-operator/actions/runs/28083840132/job/83144971760?pr=4665#step:11:251

```
v1alpha1 - ^.spec.apiSpec.schemaValidation.config.validateKey - description - ERROR - description changed : "If true, validate the record key.\n\nRequires a minimum runtime version of 1.2." -> ""
```

**Which issue this PR fixes**

Fixes #

I'll add a `ack_crd_breaking_change` label. If we encounter a CRD breaking change in CI that is within the expected scope and we have added the corresponding handling, then we only need to add the `ack_crd_breaking_change` label.

**PR Readiness Checklist**:


scripts/crdify-config.yaml
```diff
@@ -0,0 +1,4 @@
+# Description-only changes do not affect CRD validation compatibility.
+validations:
+  - name: description
+    enforcement: None
```
pmalek: Can we ensure that this will not allow removal/renaming of description field?

.github/workflows/ack-crd-breaking-change.yaml
```diff
@@ -0,0 +1,40 @@
+name: ack CRD breaking change
+
+on:
+  pull_request_target:
```
pmalek: We do not want the `pull_request_target` as that would cause the secrets to be exposed to CI runs on PRs from forks.

.github/workflows/ack-crd-breaking-change.yaml
```diff
@@ -0,0 +1,40 @@
...
+          if [[ -z "${run_id}" ]]; then
+            echo "No failed tests workflow run found for ${HEAD_SHA}."
+            exit 0
+          fi
+
+          echo "Rerunning failed jobs in tests workflow run ${run_id}."
+          gh api \
+            --method POST \
+            -H "Accept: application/vnd.github+json" \
+            "repos/${GITHUB_REPOSITORY}/actions/runs/${run_id}/rerun-failed-jobs"
```
pmalek: Instead of all of this, can't we just rerun the last `crd-breaking-changes` job which failed?

.github/workflows/__verify.yaml
```diff
@@ -93,4 +96,43 @@ jobs:
...
+      if: ${{ matrix.name == 'crd-breaking-changes' }}
+      env:
+        GH_TOKEN: ${{ github.token }}
+        PULL_REQUEST_NUMBER: ${{ github.event.pull_request.number || '' }}
+        MERGE_GROUP_HEAD_REF: ${{ github.event.merge_group.head_ref || '' }}
+      run: |
+        set -euo pipefail
+
+        ack="false"
+        pr_numbers="${PULL_REQUEST_NUMBER}"
```
pmalek, line 111: So is this plural or singular always? I'm confused which of those is possible.

.github/workflows/__verify.yaml
```diff
@@ -93,4 +96,43 @@ jobs:
...
+          if [[ -z "${pr_numbers}" ]]; then
+            echo "Could not determine pull requests for merge group; CRD breaking changes will not be acknowledged automatically."
+          fi
+        fi
+
+        for pr_number in ${pr_numbers}; do
+          if gh api \
+            -H "Accept: application/vnd.github+json" \
+            "repos/${GITHUB_REPOSITORY}/issues/${pr_number}/labels" \
+            --jq '.[].name' | grep -Fxq "${ACK_CRD_BREAKING_CHANGE_LABEL}"; then
```
pmalek: No need for grep:


```suggestion
            --jq '.[].name | select (. == env.ACK_CRD_BREAKING_CHANGE_LABEL)'; then
```

.github/workflows/__verify.yaml
```diff
@@ -93,4 +96,43 @@ jobs:
...
+        ack="false"
+        pr_numbers="${PULL_REQUEST_NUMBER}"
+
+        if [[ "${GITHUB_EVENT_NAME}" == "merge_group" ]]; then
+          pr_numbers="$(gh api \
+            -H "Accept: application/vnd.github+json" \
+            "repos/${GITHUB_REPOSITORY}/commits/${GITHUB_SHA}/pulls" \
+            --jq '.[].number' || true)"
+
+          if [[ -z "${pr_numbers}" && "${MERGE_GROUP_HEAD_REF}" =~ (^|/)pr-([0-9]+)- ]]; then
```
pmalek, line 122: You might want to sprinkle some comments here. It's not obvious what's happening here.

.github/workflows/ack-crd-breaking-change.yaml
tao12345666333: Yes, I know this, please see https://github.com/Kong/kong-operator/pull/4647#issuecomment-4788324704

The `action/checkout` v7 can now avoid this security issue.  https://github.blog/changelog/2026-06-18-safer-pull_request_target-defaults-for-github-actions-checkout/

pmalek: Ok, I see. Can you further explain why we want `pull_request_target` here though?

tao12345666333: 🥹 I was too focused on the `pull_request_target` has been safe, so I forgot that I could directly use on label. I will update it!

```
on:
  pull_request:
    types:
      - labeled
```

https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#label

.github/workflows/__verify.yaml
tao12345666333: Updated the comments

.github/workflows/__verify.yaml
tao12345666333: Added

.github/workflows/__verify.yaml
tao12345666333: Cool,  I used "any" as a substitute, which can output true/false values.

```
--jq 'any(.[].name; . == env.ACK_CRD_BREAKING_CHANGE_LABEL)')"
```

scripts/crdify-config.yaml
tao12345666333: crdify itself cannot achieve this; it is just a simple judgment. However, I have added some additional judgment logic in `scripts/verify-crd-breaking-changes.sh` to handle this situation.

.github/workflows/ack-crd-breaking-change.yaml
tao12345666333: cool, updated https://github.com/Kong/kong-operator/pull/4675/changes/bc0317f6bf51eef52c748b7c97e531d879273fc5

scripts/verify-crd-breaking-changes.sh
```diff
@@ -46,6 +46,70 @@ if [[ -z "${CRDIFY_BIN}" ]]; then
...
+	local old_crd_file="${1}"
+	local new_crd_file="${2}"
+	local old_description_paths="${3}"
+	local new_description_paths="${4}"
+	local removed_description_paths="${5}"
+
+	collect_description_paths "${old_crd_file}" > "${old_description_paths}"
+	collect_description_paths "${new_crd_file}" > "${new_description_paths}"
+
+	comm -23 "${old_description_paths}" "${new_description_paths}" > "${removed_description_paths}"
```
pmalek: `comm -23`? This sounds a little bit obscure. Can we document usage of such commands?

scripts/crdify-config.yaml
pmalek: Sorry for confusion around this.

Now reading the code which handles this special case I believe it's not worth it.

Let's revert the description handling code and take small steps. Adding the workflow for now should be enough, we can then observe how that's working for us. How does that sound?

scripts/verify-crd-breaking-changes.sh
tao12345666333: yes,  `comm` means compare two files.  

https://linux.die.net/man/1/comm

scripts/crdify-config.yaml
tao12345666333: SGTM, let's go

tao12345666333: updated https://github.com/Kong/kong-operator/pull/4675/commits/7077ac532fc27ee084f1b8eb3603d3c24490b03e

scripts/verify-crd-breaking-changes.sh
tao12345666333: Since we only need to acknowledge the workflow, this piece of code has been deleted.

pmalek (approved):

pmalek:
👍 

Let's make sure to communicate this with the team so that everyone knows how to use this.

PR #4679, opened 2026-06-24
fix(multiinstance): shut down diagnostics server on context cancellation

alacuku:

The multi-instance diagnostics server started an HTTP server with
ListenAndServe, but did not shut it down when the parent context was cancelled.
That left the listener blocked in Accept and could keep Start waiting forever on
its errgroup after manager shutdown.

Create the HTTP server outside the serving goroutine, shut it down when the
context is cancelled, and treat http.ErrServerClosed as a clean exit.

Update envtests to match the production shutdown contract: diagnostics server
shutdown now returns nil on normal context cancellation, and the manager leak test
waits for m.Run(ctx) to return before running goleak. The leak test also ignores
goroutines that existed before the test started, so it only checks goroutines
created by this test.

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:

### DIFF
--- a/ingress-controller/pkg/manager/multiinstance/diagnostics.go
+++ b/ingress-controller/pkg/manager/multiinstance/diagnostics.go
@@ -2,6 +2,7 @@ package multiinstance
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"net/http/pprof"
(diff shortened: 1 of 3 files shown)


ingress-controller/pkg/manager/multiinstance/diagnostics.go
```diff
@@ -63,14 +64,23 @@ func NewDiagnosticsServer(listenerPort int, opts ...DiagnosticsServerOption) *Di
...
+		err := server.ListenAndServe()
+		if errors.Is(err, http.ErrServerClosed) {
+			return nil
 		}
-		return server.ListenAndServe()
+		return err
+	})
+	errg.Go(func() error {
+		<-ctx.Done()
+		return server.Shutdown(context.Background())
```
pmalek: How about adding a timeout to the context provided to `Shutdown()`? Otherwise we can hang forever, no?

pmalek (dismissed):

ingress-controller/pkg/manager/multiinstance/diagnostics.go
alacuku: Nice catch!

alacuku: a76c6d0bf51d63e51dcbfbd80701e650557bbb94

pmalek (approved):

Issue #4661, opened 2026-06-23
Conformance: HTTPRouteWeight failed after bumping gateway api to v1.6.0-rc.1


WeichengWang1:
## Current Behavior
`TestGatewayConformance/HTTPRouteWeight/Requests_should_have_a_distribution_that_matches_the_weight` failed,
Traffic distribution test failed (10/10): backend "infra-backend-v1" weighted traffic of 1 not within tolerance 0.7 (+/-0.050000),  expect traffic to hit backend "infra-backend-v2" - but none was received

## Expected Behavior
test passed

### Diff
related difference in gateway api is: https://github.com/kubernetes-sigs/gateway-api/pull/4740/changes, which removed random delay.

### Possible cause
~is it possible that Konnect is not synced with Dataplane?~
KO sets programmed status after calling Konnect succeeded, when dataplane is not synced actually

HTTPRouteWeight executed after HTTPRouteSimpleSameNamespace

HTTPRouteWeight started, warm-up passed instead of 404 cuz configs of HTTPRouteSimpleSameNamespace not cleaned up in Dataplane yet.
then 500 requests sent to backend "infra-backend-v1"(configs of HTTPRouteWeight actually not synced to Dataplane yet), test failed.

10 re-runs in 4.02 seconds.

https://github.com/Kong/kong-operator/actions/runs/27743401555/job/82076001728?pr=4639
searching `Traffic distribution test failed`

<img width="2392" height="816" alt="Image" src="https://github.com/user-attachments/assets/21c14143-23f8-4497-ad4a-66dc090b786d" />

if `HTTPRouteBackendTimeout` `HTTPRouteRequestTimeout` not skipped, ~maybe~ `HTTPRouteWeight` won't be affected.

WeichengWang1:
For on-prem, HTTPRouteWeight is flaky, https://github.com/Kong/kong-operator/actions/runs/28232485271/job/83639433186?pr=4668

Because in conformance test, Programmed condition is not waited.  It waits for:
 1. GatewayAndHTTPRoutesMustBeAccepted(...) → route parent status with Accepted=True
 2. HTTPRouteMustHaveResolvedRefsConditionsTrue(...) → ResolvedRefs=True
So at the moment requests sent, the weight route configs probably not synced to dataplane.
The sync period from KIC's memory to dataplane is 3s.

Skip HTTPWeight for all conformance test until this problem solved.

WeichengWang1:
completed by #4714

PR #4714, opened 2026-06-30
conformance: enable HTTPRouteBackendTimeout and HTTPRouteWeight

### LINKED ISSUE #59: conformance: `HTTPRouteBackendTimeout` feature
## Problem Statement

KIC `ControlPlane` is already conformant with the `HTTPRouteBackendTimeout` extended feature; we need to enable such a feature and pass the related conformance tests.

## Acceptance Criteria

- [ ] The feature `HTTPRouteBackendTimeout` is enabled and the related conformance tests are successfully run.

WeichengWang1:

Enables the Gateway API HTTPRouteBackendTimeout conformance feature and makes the underlying backendRequest timeout translation backward-compatible and granular.

Background: KIC issues [#4914](https://github.com/Kong/kubernetes-ingress-controller/issues/4914), [#5478](https://github.com/Kong/kubernetes-ingress-controller/issues/5478), [#5451](https://github.com/Kong/kubernetes-ingress-controller/issues/5451)

#4661 can be solved if `HTTPRouteBackendTimeout` is not skipped.

**Main Changes**:

- Remove the uniform-timeout restriction in webhook (KIC [#5451](https://github.com/Kong/kubernetes-ingress-controller/issues/5451)). Per-rule timeouts are now supported, since rules with different timeouts translate to separate Kong services.
- Split a Kong service by timeout only when needed. Per-timeout Kong services exist only when its rules carry more than one distinct backendRequest timeout. When a split is needed, services are suffixed with .timeout.<ms>
- Map a 0s timeout to Kong's max (maxKongServiceTimeout = 2147483646), per the Gateway API "no timeout" semantics.

**Which issue this PR fixes**

Fixes #59 

1. KongService naming

Split only when needed to avoid renaming/recreating existing services on upgrade.

- On-prem (embedded KIC), combined mode — the base service name does not embed the timeout. Splitting a backend group into per-timeout services only when the group contains more than one distinct timeout. A single/uniform timeout keeps the original base name(without timeout); a genuine conflict yields …svc.<backends>.timeout.<ms>. 
- Hybrid — service names are hash-based and computed per rule (namegen), so there's no sibling-rule visibility at naming time. The effective timeout is therefore folded into the name hash whenever it is set. (unlike on-prem, without timeout and with uniform timeout -> different name) But this is correct and safe on upgrade because hybrid never applied backendRequest before.

2. Redundant (but harmless) KongServices when a konghq.com/*-timeout annotation is set

The konghq.com/connect|read|write-timeout annotation overrides the route's backendRequest timeout. The split decision (naming) is made from backendRequest and happens before the annotation override is applied. So if two rules share a backend with different backendRequest timeouts (→ split into two KongServices) and that backend Service carries a timeout annotation, both services get the annotation's value and end up with an identical effective timeout — i.e., two distinct KongService objects that behave the same.

Eliminating the redundancy would require keying the split on the post-override effective timeout, which means resolving backend Service annotations at naming time (cross-namespace lookups, per-backend). This is the accepted trade-off. Same behavior in both modes.

**PR Readiness Checklist**:


WeichengWang1:
After enabling `HTTPRouteBackendTimeout`, `HTTPRouteWeight` can pass.
But `HTTPRouteBackendTimeout` failed.

For on-prem, different backendtimeouts for same backend in different rules are not allowed.

For hybird mode, `.Timeouts.BackendRequest` is never read, In `controller/hybridgateway/service/service.go`, the KongService timeouts come only from backend Service annotations:
```go
connectTimeout, _ := resolveConnectTimeoutFromBackendRefs(...)  // -> metadata.ExtractConnectTimeout(svc.GetAnnotations())
readTimeout,    _ := resolveReadTimeoutFromBackendRefs(...)      // konghq.com/... annotations on the backend Service
writeTimeout,   _ := resolveWriteTimeoutFromBackendRefs(...)
```
For on-prem, `.Timeouts.BackendRequest` is read first and finally overwritten by annos if set.

Possible solution:
respect `.Timeouts.BackendRequest`, but `konghq.com/...` annotations have higher priority

ingress-controller/internal/dataplane/translator/subtranslator/httproute.go
```diff
@@ -449,6 +465,27 @@ func groupRulesByBackendRefs(ruleEntries []httpRouteRuleMeta) map[string][]httpR
...
+func splitServiceGroupsByBackendRequestTimeout(groups map[string][]httpRouteRuleMeta) map[string][]httpRouteRuleMeta {
+	result := make(map[string][]httpRouteRuleMeta, len(groups))
+	for serviceName, rulesMeta := range groups {
+		rulesByTimeout := groupSliceByKeyFn(rulesMeta, httpRouteRuleMeta.getBackendRequestTimeoutKey)
+		if len(rulesByTimeout) <= 1 {
+			result[serviceName] = rulesMeta
+			continue
+		}
+		for timeoutKey, rules := range rulesByTimeout {
+			result[serviceName+backendRequestTimeoutServiceNameSuffix(timeoutKey)] = rules
```
tao12345666333: `getKongServiceNameByBackendRefs()` trims the backend-derived base name to `maxKongServiceNameLength`, but this split path appends `.timeout.<ms>` after that check. 

A base name that is valid but close to 511 chars can therefore become overlong once conflicting `backendRequest` timeouts force a split, producing an invalid Kong service/host name. We can enforce the length limit on the final name as well, e.g by applying the existing trim/hash logic after suffixing, or by reserving suffix space before the initial trim. WDYT?

controller/hybridgateway/namegen/name.go
```diff
@@ -203,6 +229,20 @@ func hashForHTTPRouteRuleServiceLikeName(route *gwtypes.HTTPRoute, rule gatewayv
 		hash = utils.Hash32(zeroBackendRuleIdentity)
 	}
 
+	// Fold the backendRequest timeout into the hash so rules sharing the same backends but
+	// requesting different timeouts map to distinct KongServices (a KongService can only carry a
+	// single timeout). Rules without a backendRequest timeout keep the original hash, so enabling
+	// the feature does not rename existing KongServices.
+	if timeoutMS := BackendRequestTimeoutMilliseconds(rule); timeoutMS != nil {
```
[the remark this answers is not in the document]
WeichengWang1: 1. do we want to break 1:1 mapping between KongService and KongUpstream?
2. delete+recreate KongUpstream doesn't really matter, since KongServices have to been deleted and recreated.

controller/hybridgateway/namegen/name.go
```diff
@@ -11,6 +12,31 @@ import (
 	gwtypes "github.com/kong/kong-operator/v2/internal/types"
 )
 
+// MaxKongServiceTimeout is the largest timeout Kong accepts for Service connect, read, and
+// write timeouts. Kong has no "disable" value, so a zero-duration backendRequest timeout
+// (which the Gateway API defines as "no timeout") is mapped to this as the closest emulation.
+const MaxKongServiceTimeout int64 = 2147483646
```
tao12345666333: We have defined maxKongServiceTimeout in two places; perhaps we should unify them to avoid inconsistencies?

https://github.com/Kong/kong-operator/pull/4714/changes#diff-e2d9e167009c611773a991cfaa3b376ad43b0b9bd21e9cf94b82ff76801f1450R22

tao12345666333 (dismissed):
LGTM

tao12345666333 (approved):

PR #4723, opened 2026-07-01
fix(versions): tag parser accepts registry references with a port

### LINKED ISSUE #4724: DataPlane image validation rejects references with a port
## Current Behavior

`internal/versions.FromImage` uses `strings.Split(image, ":")` and then requires exactly two segments. That check fails for any image reference whose registry host carries a port, e.g.

    registry.example.com:5000/kong/kong-gateway:3.10

The split returns three segments (`["registry.example.com", "5000/kong/kong-gateway", "3.10"]`) and `FromImage` returns

    expected "<image>:<tag>" format, got: registry.example.com:5000/kong/kong-gateway:3.10

This bubbles up through `IsDataPlaneImageVersionSupported` (see [controller/dataplane/owned_deployment.go:197](https://github.com/Kong/kong-operator/blob/main/controller/dataplane/owned_deployment.go#L197)) so the DataPlane controller fails to build the Deployment. On the cluster the DataPlane CR sits at

    Ready: False   reason: DependenciesNotReady

with `.status.replicas: 0` and no Deployment beneath it. The Gateway never reaches `Programmed=True`, and Kong is unreachable — but the only log line pointing at the actual cause is a single "Reconciler error" in the KGO manager log; nothing surfaces on the DataPlane or Gateway CR.

## Expected Behavior

Image references with a port in the registry host should parse the same way `docker`, `containerd`, and `crictl` parse them: the tag, when present, is whatever follows the **last** `:` in the string (after any `@sha256:...` digest has been stripped).

## Steps To Reproduce

1. Install Kong Operator 2.2.0.
2. Create a `Gateway` + `GatewayConfiguration` whose data plane image points at a registry with a port:

   ```yaml
   kind: GatewayConfiguration
   metadata:
     name: kong-config
     namespace: kong-demo
   spec:
     dataPlaneOptions:
       deployment:
         podTemplateSpec:
           spec:
             containers:
               - name: proxy
                 image: registry.example.com:5000/kong/kong-gateway:3.10
   ```

3. Observe: `kubectl -n kong-demo get dataplane` — `Ready=False`, `.status.replicas: 0`, no Deployment created.
4. `kubectl -n <kong-ns> logs deploy/<kgo>-controller-manager | grep "expected <image>:<tag>"` — the error.

The same reference works fine as an OLM operator image (kubelet + containerd pull it correctly), so this is purely a KGO-side parser limitation.

## Operator Version

kong-operator 2.2.0 (`main` at 28fb8febd shows the same code)

## Root cause and fix

`internal/versions/validation.go:45` uses `strings.Split(image, ":")` and then requires `len == 2`. `strings.LastIndex(image, ":")` handles host-with-port correctly. PR: #4723.

luohua13:
Fixes #4724

## Current behavior

`internal/versions.FromImage` splits image references on every `:` and requires the result to have exactly two segments. That check fails for any reference whose registry host carries a port, e.g.

```
registry.example.com:5000/kong/kong-gateway:3.10
```

The split returns three segments (`["registry.example.com", "5000/kong/kong-gateway", "3.10"]`), and the caller [`IsDataPlaneImageVersionSupported`](https://github.com/Kong/kong-operator/blob/main/controller/dataplane/owned_deployment.go#L197) surfaces this as `expected "<image>:<tag>" format, got: <full-ref>`.

The DataPlane controller then fails to build a Deployment, silently: the DataPlane CR sits at `Ready=False reason=DependenciesNotReady` with `.status.replicas: 0`, no Deployment is created, and the Gateway never reaches `Programmed=True`. The only pointer to the actual cause is a single manager log line — nothing surfaces on the CR conditions.

The same reference works fine as an OLM operator image (kubelet + containerd pull it correctly), so this is purely a KGO-side parser limitation.

## Fix

Use `strings.LastIndex(image, ":")` — the tag, when present, always follows the last `:` (any digest was already stripped earlier via the `@sha256:` handling). If what follows the last `:` contains `/`, the last `:` was the host-port separator and the reference has no tag; treat that as the existing "no tag" error.

## Tests

New cases in `internal/versions/validation_test.go`:

- host with port + tag (`registry.example.com:5000/kong/kong-gateway:3.3.0`)
- host with port + tag containing dots (`...:5000/...:3.10`)
- host with port + tag + digest (digest stripped first, then last-`:` split)
- host with port + no tag (must error)
- no colon at all (must error, unchanged)
- trailing colon with empty tag (must error, unchanged)

Existing cases (semver-with-flavour, enterprise 4-segment, digest-only) unchanged.

## Operator version

kong-operator 2.2.0 (`main` at 28fb8febd shows the same code path).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,13 @@
 
 ### Fixes
 
+- Version parser: accept DataPlane image references whose registry host contains
+  a port (e.g. `registry.example.com:5000/kong/kong-gateway:3.10`). The tag is
+  now split at the last `:` instead of every `:`, so a host-port colon is no
+  longer misread as the tag separator. Without this fix, DataPlane Deployment
+  provisioning silently failed with
+  `expected "<image>:<tag>" format, got: <full-ref>`.
+  [#XXXX](https://github.com/Kong/kong-operator/pull/XXXX)
 - Konnect: prevent orphaned and duplicate Konnect entities when reconciliation
   races with the cached client. The cleanup finalizer is now added before the entity
   is created in Konnect, and the freshly created Konnect ID is kept in an in-memory
--- a/internal/versions/validation.go
+++ b/internal/versions/validation.go
@@ -42,12 +42,25 @@ func FromImage(image string) (semver.Version, error) {
 		image = image[:idx]
 	}
 
-	splitImage := strings.Split(image, ":")
-	if len(splitImage) != 2 {
+	// Use the last ':' as the tag separator. Splitting on every ':' breaks
+	// references whose registry host carries a port, e.g.
+	//   registry.example.com:5000/foo/bar:3.10
+	// where the first ':' is part of the host ("<host>:<port>") rather than
+	// a tag separator. The tag, if present, always follows the LAST ':'
+	// (any digest was already stripped above).
+	idx := strings.LastIndex(image, ":")
+	if idx == -1 {
+		return semver.Version{}, fmt.Errorf(`%w, got: %s`, ErrExpectedSemverVersion, image)
+	}
+	tag := image[idx+1:]
+	// If what looks like a tag actually contains '/', we split on a port
+	// separator in the registry host and the reference has no tag at all
+	// (e.g. "registry:5000/foo/bar" — LastIndex found ':' inside the host).
+	if tag == "" || strings.Contains(tag, "/") {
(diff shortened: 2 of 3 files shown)


programmer04:
Thanks for sending the PR, much appreciated 🙇 

Due to limitations of our process and CI setups, we can't directly accept contributions. So your commit has been cherry-picked for the new PR

- https://github.com/Kong/kong-operator/pull/5036

that will get merged and released 🎉

PR #4755, opened 2026-07-02
fix(gateway): preserve user KONG_STREAM_LISTEN address and options

sandromodarelli:

The gateway controller generated `KONG_STREAM_LISTEN` for TLS listeners and force-wrote it as `0.0.0.0:<port> ssl reuseport`, overwriting any bind address or listen options set on the `GatewayConfiguration` DataPlane pod template (breaking IPv6 stream listeners).

It now enforces only the Kong port and the protocol token (`ssl` for TLS), preserving the user-configured bind address (e.g. `[::]`) and listen options (`reuseport`, `backlog=...`). Defaults remain `0.0.0.0`/`reuseport` when unset. The protocol token is resolved per listener protocol so TCP/UDP stream listeners can reuse the same rendering path later.

**Which issue this PR fixes**

Fixes FTI-7669

Behavior:
- no user env → `0.0.0.0:PORT ssl reuseport` (unchanged)
- `[::]:PORT ssl reuseport` → preserved (IPv6)
- `[::]:PORT ssl` → `reuseport` dropped
- extra options like `backlog=...` preserved

**PR Readiness Checklist**:


controller/gateway/controller_reconciler_utils.go
```diff
@@ -1566,13 +1585,58 @@ func conditionMessage(oldStr, newStr string) string {
...
+// TCP -> none. Rendering is otherwise identical across protocols, so future TCPRoute
+// and UDPRoute support only needs to collect their ports (see streamListenPort) —
+// no separate render path is required.
+func streamListenProtocolTokens(p gatewayv1.ProtocolType) []string {
+	switch p {
+	case gatewayv1.TLSProtocolType:
+		return []string{"ssl"}
+	case gatewayv1.UDPProtocolType:
+		return []string{"udp"}
+	default: // TCP and anything else: no extra token.
```
randmonkey, line 1636: Could we explicitly separate `TCP` and "anything else" for clarity?

randmonkey (commented):
So this PR extracts the address and options of the first TLS listen ports. This may have issues when the user want to use the same port for different addresses in `KONG_STREAM_LISTEN`. (for example, using dual stack for v4 + v6)

randmonkey (approved):

PR #4795, opened 2026-07-06
feat(ssa): share a live-rebuilding TypeConverter across SSA controllers

### LINKED ISSUE #4607: One-shot SSA TypeConverter races CRD OpenAPI publication, permanently breaking status applies
## Current Behavior

The SSA `TypeConverter` used by our controllers (`controller/pkg/ssa`) is built
once at controller startup from the API server's aggregated `/openapi/v3`
document, and any requested GroupVersion whose schema isn't published yet is
silently skipped. Because a freshly-installed CRD is servable over REST *before*
its OpenAPI v3 schema is published, the operator can permanently end up with a
converter that has no type for one of our own CRDs. Every Server-Side Apply
against that type then fails for the operator's entire lifetime.

Surfaced via failing `eventgateway` Chainsaw e2e tests, but the bug is general
and affects every controller that uses `controller/pkg/ssa`.

- Both `eventgateway` Chainsaw e2e suites (`proxy-roundtrip`, `sni-routing`)
  fail: the `KegDataPlane` never gets a `status`, so the `Ready=True` assert
  times out after 5 minutes (`status: Required value: field not found in the
  input object`).
- Operator logs show, on **every** reconcile (32/32 in the captured run, 0
  successes):

  ```
  failed to patch KegDataPlane status: failed to convert existing status to
  TypedValue: no corresponding type for eventgateway.konghq.com/v1alpha1,
  Kind=KegDataPlane
  ```

## Expected Behavior

The operator should reconcile its own CRDs reliably regardless of CRD/operator
install ordering or OpenAPI publication timing. Once a CRD is servable, Server-
Side Apply against that type (including status) must succeed — the operator
should never be left permanently unable to apply a type whose schema it needs,
and such a condition must surface as a clear error rather than silent, repeated
apply failures.

alacuku:

Replace the per-controller (DataPlane, MCPServer) one-shot TypeConverter, built once from OpenAPI v3 discovery at SetupWithManager time, with a single TypeConverterProvider shared across controllers. CRD schemas are now built in-process from the live CRD objects (apiserver-style, no OpenAPI publication debounce); built-in core/apps schemas are still fetched once via discovery. A new crdschema controller watches CustomResourceDefinitions in the configured groups and rebuilds the converter in place, atomically swapping it, whenever a relevant CRD changes.

- Wire the shared provider through modules/manager/{run,controller_setup}.go, built only when a SSA-using controller (KEG DataPlane, MCPServer) is enabled, with a readyz check so the manager reports ready only once the converter has completed its initial build.
- Add unit tests for the provider (rebuild caching/concurrency, error paths) and the crdschema reconciler (Provider narrowed to an interface for testability), plus dedicated and extended envtest coverage exercising the real wiring end-to-end against a live API server.
- Regenerate RBAC/chart snapshots for the new customresourcedefinitions get/list/watch permission.

**Which issue this PR fixes**

Fixes #4607 

**PR Readiness Checklist**:


modules/manager/run.go
```diff
@@ -321,7 +323,32 @@ func Run(
...
+	// Build the shared SSA TypeConverterProvider when any SSA-using controller is
+	// active.  Built on every replica (not leader-gated) so the converter is
+	// ready before reconcile loops start.
+	//
+	// eventGatewayCRDGroups lists the CRD groups whose types are passed to
+	// ApplyIfChanged / ApplyStatusIfChanged by the EventGateway DataPlane
+	// controller; it lives here rather than in the generic controller/pkg/ssa
+	// package because only this package knows which controllers are enabled.
+	var ssaProvider *controllerpkgssa.TypeConverterProvider
+	if cfg.KEGDataPlaneControllerEnabled || cfg.FeatureGates.Enabled(FeatureGateMCPServer) {
```
pmalek: Can we refactor this if condition into a named func which will be appropriately named `IsSSAProviderNeeded` or similar?

modules/manager/run.go
```diff
@@ -321,7 +323,32 @@ func Run(
...
+		eventGatewayCRDGroups := map[string]struct{}{
+			"eventgateway.konghq.com":  {},
+			"configuration.konghq.com": {},
+			"konnect.konghq.com":       {},
+		}
+		var err error
+		ssaProvider, err = controllerpkgssa.NewTypeConverterProvider(ctx, setupLog.WithName("ssa-crd-schema"), mgr, eventGatewayCRDGroups)
```
pmalek: Could we also extract this into a func?

The comment about the event gateway API groups could be placed there (minimizing the code added here.

Also the comment "eventGatewayCRDGroups lists the CRD groups whose types are passed to ..." could explain why we're only concerned with event gateway API groups. Why not others? That's not immediately clear for the reader.

modules/manager/controller_setup.go
```diff
@@ -705,13 +716,21 @@ func SetupControllers(mgr manager.Manager, c *Config, cpsMgr *multiinstance.Mana
...
 				CertTTL:                  c.CertTTL,
+				TypeConverter:            ssaProvider,
 			},
 		},
+		// CRD schema reconciler: rebuilds the shared SSA TypeConverter when
+		// relevant CRDs change. Enabled whenever at least one SSA-using
+		// controller is active.
+		{
+			Enabled:    ssaProvider != nil,
+			Controller: &crdschema.Reconciler{Client: mgr.GetClient(), LoggingMode: c.LoggingMode, Provider: ssaProvider},
```
pmalek: One field per line just like in initializations above:


```suggestion
			Controller: &crdschema.Reconciler{
				Client: mgr.GetClient(), 
				LoggingMode: c.LoggingMode, 
				Provider: ssaProvider,
			},
```

modules/manager/run.go
alacuku: PTAL: f157e05fa1fc943de92693cafa7824900ea4b31e

modules/manager/run.go
alacuku: PTAL: f157e05fa1fc943de92693cafa7824900ea4b31e

modules/manager/controller_setup.go
alacuku: PTAL: f157e05fa1fc943de92693cafa7824900ea4b31e

modules/manager/ssa_provider_test.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+}
+
+func TestBuildSSAProvider_error(t *testing.T) {
+	mgr, err := ctrl.NewManager(&rest.Config{Host: "http://127.0.0.1:0"}, ctrlmgr.Options{Scheme: managerscheme.Get()})
+	require.NoError(t, err)
+
+	_, err = buildSSAProvider(t.Context(), logr.Discard(), mgr)
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "failed to build initial SSA TypeConverter")
```
pmalek, line 65: Can we make this clearer why this _should_ fail?

controller/crdschema/reconciler.go
```diff
@@ -0,0 +1,97 @@
...
+func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := log.GetLogger(ctx, "ssa-crd-schema", r.LoggingMode)
+	log.Trace(logger, "rebuilding SSA TypeConverter", "trigger", req.Name)
+
+	var relevant []*apiextensionsv1.CustomResourceDefinition
+	for group := range r.Provider.CRDGroups() {
+		list := &apiextensionsv1.CustomResourceDefinitionList{}
+		if err := r.List(ctx, list, client.MatchingFields{index.IndexFieldCRDOnGroup: group}); err != nil {
+			return ctrl.Result{}, err
+		}
```
pmalek: Nit:

```suggestion
		var list apiextensionsv1.CustomResourceDefinitionList
		listOpts := client.MatchingFields{index.IndexFieldCRDOnGroup: group}
		if err := r.List(ctx, &list, listOpts); err != nil {
			return ctrl.Result{}, err
		}
```

pmalek (commented):
LGTM, minor nits.

modules/manager/ssa_provider_test.go
alacuku: PTAL: 7dd23a9ab8f59ef00b4cddb4a18a1182679def09

controller/crdschema/reconciler.go
alacuku: PTAL: 7dd23a9ab8f59ef00b4cddb4a18a1182679def09

sandromodarelli (approved):

PR #4824, opened 2026-07-08
feat(crd): add CEL rules to enfore parentRef to be immutable after programmed

### LINKED ISSUE #4820: Enforce AI Gateway control plane to be immutable after programmed for `AIGateway*` CRDs
## Problem Statement

Similar to API gateway entities, the reference to AI gateway control planes should be immutable after they have "programmed" condition set to true, meaning that they are already applied on Konnect. 

## Proposed Solution

Add CEL `// +kubebuilder:validation:XValidation:rule="(!has(self.spec.aiGatewayRef)) ? true : (!has(self.status) || !self.status.conditions.exists(c, c.type == 'Programmed' && c.status == 'True')) ? true : oldSelf.spec.aiGatewayRef == self.spec.aiGatewayRef", message="spec.aiGatewayRef is immutable when an entity is already Programmed"` in CELs of the AI gateway entities with the AI gateway CP ref and generate these CRDs again. 

## Additional Information

...

## Acceptance Criteria

- [ ] CELs for validating AI gateway CP immutable is added for all the AIGateway CRDs
- [ ] Tests are added to verify the validation works

randmonkey:

Add CEL to validate that the CRDs cannot their change parent References after they are programmed.
Affects the generated `EventGateway*` and `AIGateway*` CRDs.

**Which issue this PR fixes**

Fixes #4820

The intension here is to add the check of parent ref immutablility for AIGateway related resources. But `EventGateway*` are also affected. Should we also make the change?

**PR Readiness Checklist**:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,10 +82,11 @@
   ```
 
 - All the CRDs generated from Konnect OpenAPISpec with a parent reference now
-  has a CEL to validate that the parent reference is immutable when a resouce
-  has `Programmed` condition set to `True` in its status, meaning that it is
-  already configured in Konnect.
-  This breaks the already released EventGateway CRDs, including:
+  have a CEL validation rule which ensures that the parent reference is immutable
+  when a resource has `Programmed` condition set to `True` in its status,
+  meaning that it is already configured in Konnect.
+  This prevents the unsupported scenarios through breaking changes in the already
+  released EventGateway CRDs (Tech Preview), including
   - `EventGatewayVirtualCluster`
   - `EventGatewayVirtualClusterConsumePolicy`
   - `EventGatewayVirtualClusterPolicy`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,12 +80,12 @@
     name: my-backend-cluster-tls
     key: tls.crt # Key name depends on the type of referenced Secret.
   ```
-  
+
 - All the CRDs generated from Konnect OpenAPISpec with a parent reference now
   has a CEL to validate that the parent reference is immutable when a resouce
   has `Programmed` condition set to `True` in its status, meaning that it is
   already configured in Konnect.
-  This breaks the already released EventGateway CRDs, incluiding:
+  This breaks the already released EventGateway CRDs, including:
   - `EventGatewayVirtualCluster`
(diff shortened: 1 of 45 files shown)


CHANGELOG.md
```diff
@@ -81,6 +81,17 @@
     key: tls.crt # Key name depends on the type of referenced Secret.
   ```
 
+- All the CRDs generated from Konnect OpenAPISpec with a parent reference now
+  has a CEL to validate that the parent reference is immutable when a resouce
+  has `Programmed` condition set to `True` in its status, meaning that it is
+  already configured in Konnect.
+  This breaks the already released EventGateway CRDs, including:
```
```suggestion
  This prevents the unsupported scenario through changes in the already released EventGateway CRDs, including:
```

pmalek (commented):
Minor nit but overall looks good.

I'd suggest we add (this can be done in a separate PR) a suite of CRD validation tests which ensures this works.

We already have something like this for CP refs (https://github.com/Kong/kong-operator/blob/032be3e1e8064d1fe2a56045b9090e5014144617/test/crdsvalidation/common/suite_crd_ref_change.go#L61) but it will require some adjustments to make it generic (or to write a new one based on it).

That could use the shared `GetParentRef() commonv1alpha1.ObjectRef` function.

pmalek:
> Minor nit but overall looks good.
> 
> I'd suggest we add (this can be done in a separate PR) a suite of CRD validation tests which ensures this works.
> 
> We already have something like this for CP refs (
> 
> https://github.com/Kong/kong-operator/blob/032be3e1e8064d1fe2a56045b9090e5014144617/test/crdsvalidation/common/suite_crd_ref_change.go#L61
> 
> ) but it will require some adjustments to make it generic (or to write a new one based on it).
> That could use the shared `GetParentRef() commonv1alpha1.ObjectRef` function.

Added an issue for this https://github.com/Kong/kong-operator/issues/4930 so that we can merge the change here and work on test suite separately.

pmalek:
@randmonkey PTAL I rebased and fixed the CEL expression which was discovered in tests.

randmonkey:
> @randmonkey PTAL I rebased and fixed the CEL expression which was discovered in tests.

LGTM but I cannot approve it as the person who submitted the PR.

WeichengWang1 (approved):

PR #4871, opened 2026-07-13
fix(reconcilers): Add watch for referneced secrets in generated reconcilers and fix `secretRefValid` condition handling

### LINKED ISSUE #4870: Updates of referenced secrets does not trigger reconciliation and `SecretRefValid` condition is not set back to `True` when once being set to `False` for generated reconcilers
## Current Behavior

Currently,  the generated reconciler for the Konnect CRDs with reference to `Secret` does not enqueue the object when the referenced secret is changed. 
Also, the generated reconcile sets the `secretRefValid` condition to `False` when the referenced secret is not found or does not have the required key. But when the secret is found, the condition is not set to `True`.

## Expected Behavior

The `secretRefValid` condition is set to `True` when the secret is found

## Steps To Reproduce

1. Create a Konnect CRD with secret reference but do not create the referenced secret (e.g: `AIGatewayModelProvider`) 
2. The `secretRefValid` condition is set to `False` in its status
3. Create the referenced secret
4. The `AIGatewayModelProvider` is not reconciled immediately after the secret is created.
5. The `AIGatewayModelProvider` gets reconciled after a Konnect refresh loop and its secretRefValid` condition is still `False` but `Programmed` is set to `True`.

## Operator Version

`main` & KO 2.2.2

randmonkey:

- Add watch of `secret`s in generated reconcilers and enqueue them when referenced secret is changed
- Fix handling of `secretRefValid` and set to `True` if secret reference is resolved

**Which issue this PR fixes**

Fixes #4870

**PR Readiness Checklist**:

### DIFF
--- a/controller/hybridgateway/controller.go
+++ b/controller/hybridgateway/controller.go
@@ -438,9 +438,9 @@ func (r *HybridGatewayReconciler[t, tPtr]) cleanupGeneratedResources(
 	logger logr.Logger,
 	conv converter.APIConverter[t],
 ) (bool, error) {
-	// Use the existing cleanup logic but do not wait for generated resources to
-	// fully disappear before releasing the root finalizer. Generated resources
-	// have their own cleanup/finalizer flows; keeping the Gateway API object in
-	// deletion until every child finishes can block immediate same-name re-creates.
-	return cleanOrphanedResources[t, tPtr](ctx, r.Client, logger, conv, orphanCleanupOptions{waitForDeletes: false})
+	// On the deletion path we create a fresh converter but do not run Translate(),
+	// so cleanOrphanedResources() sees no desired output objects. That makes all
+	// owned generated resources orphan candidates, and we wait for their deletion
+	// before releasing the root finalizer.
+	return cleanOrphanedResources[t, tPtr](ctx, r.Client, logger, conv, orphanCleanupOptions{waitForDeletes: true})
 }
--- a/controller/hybridgateway/reconciler_utils_test.go
+++ b/controller/hybridgateway/reconciler_utils_test.go
@@ -546,7 +546,6 @@ func TestCleanOrphanedResources(t *testing.T) {
 		gvks                 []schema.GroupVersionKind
 		desired              []unstructured.Unstructured // what the converter reports as desired
 		existing             []unstructured.Unstructured // what's in the cluster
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -134,6 +134,10 @@
 
 ### Fixes
 
+- Konnect: add the watch of referenced secrets in the reconcilers for generated
+  CRDs with reference to `Secret`s and set `secretRefValid` condition to `True`
+  when the references to secrets are resolved.
+  [#4871](https://github.com/Kong/kong-operator/pull/4871)
```
pmalek: I think we can leave this out because this PR fixes something that hasn't been released yet.

controller/konnect/watch.go
```diff
@@ -434,3 +435,105 @@ func enqueueObjectsForKongReferenceGrant[
...
+// api/configuration/v1alpha1, ...) generates its own identically-shaped
+// SensitiveDataSecretRef type, so a single generic method-return-type
+// constraint can't span all of them — dispatch by type switch instead,
+// mirroring getSecretRefs in reconciler_secretref.go. Add a case here when a
+// new API group-version starts generating entities with secretReferences.
+func secretRefsForSensitiveData(obj any) ([]commonv1alpha1.NamespacedRef, bool) {
+	switch g := obj.(type) {
+	case interface {
+		GetSensitiveDataSecretRefs() []konnectv1alpha1.SensitiveDataSecretRef
+	}:
```
pmalek, line 453: Can we put this interface (and the one below) outside of the `switch` statement so that it's a bit more readable?

We could also consider adding (generating) a method on those types to return `[]commonv1alpha1.NamespacedRef` so that we don't have to do this here. WDYT?

randmonkey: Looks like `konnectv1alpha1.SensitiveDataSecretRef` and `configurationv1alpha1.SensitiveDataSecretRef` are identical structs. Should we consider to move them to some common packages (say `commonv1alpha1`?)

pmalek: We could do either of these: either add a function to return `[]commonv1alpha1.NamespacedRef` or use a common type for secret ref. I'm fine with leaving the types as is (there's no harm in duplicating API types as can evolve independently then).

pmalek:
@randmonkey What's the reason behind closing this PR?

PR #4877, opened 2026-07-13
[Backport release/2.2.x]Preserve only one CA certificate if they have dupe IDs

randmonkey:

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:

### DIFF
--- a/ingress-controller/internal/dataplane/translator/translate_cacerts.go
+++ b/ingress-controller/internal/dataplane/translator/translate_cacerts.go
@@ -48,14 +48,52 @@ func (t *Translator) getCACerts() []kong.CACertificate {
 		return intermediateT{obj: configmap.DeepCopy(), data: configmap.Data}
 	})...)
 
-	caCerts := make([]kong.CACertificate, 0, len(caCertSecrets)+len(caCertConfigMaps))
+	// pickWinner returns the preferred and rejected intermediate for a shared ID.
+	// Priority: earliest creation timestamp → lexicographically smallest namespace → smallest name.
+	pickWinner := func(a, b intermediateT) (winner, loser intermediateT) {
+		aTime, bTime := a.obj.GetCreationTimestamp(), b.obj.GetCreationTimestamp()
+		if aTime.Before(&bTime) {
+			return a, b
+		}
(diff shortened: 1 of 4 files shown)


programmer04 (dismissed):

ingress-controller/internal/dataplane/translator/translate_cacerts.go
```diff
@@ -48,14 +48,52 @@ func (t *Translator) getCACerts() []kong.CACertificate {
...
+	// (breaking ties by namespace then name). Non-winners are reported as translation failures immediately.
+	winners := make(map[string]intermediateT)
 	for _, caCertData := range caCertsData {
-		secretID, ok := caCertData.data["id"]
+		id, ok := caCertData.data["id"]
 		if !ok {
 			t.registerTranslationFailure("invalid secret CA certificate: missing 'id' field in data", caCertData.obj)
 			continue
 		}
+		existing, seen := winners[id]
```
tao12345666333, line 84: This first pass picks a winner using only creation time / namespace / name, before checking whether that object actually contains a usable CA certificate. If the oldest object for a duplicated id is malformed (missing cert, non-CA PEM, expired cert, etc) and a newer duplicate is valid, the newer one is immediately marked as a duplicate here and never gets a chance to translate. 
The second pass then rejects the chosen winner, so we end up publishing no CA certificate for that id even though a valid replacement exists. Consider validating duplicate candidates before finalizing the winner, or letting a later valid duplicate replace an earlier invalid one. WDYT?

randmonkey: This is not considered as a better solution because filtering malformed certificates in the phase may shadow that there are CA certs with dupe IDs. Also, when the choosen oldest one is invalid, user can get from event that the older one does not have a CA cert and the newer one has dupe IDs, which is better for them to fix. 
If we choose the newer one and marking the older one as invalid in the case, after the user configured a valid CA, the case becomes that the older one is configured and newer one is marked as duplicate IDs, it would be surprising.

tao12345666333 (approved):

Issue #4842, opened 2026-07-08
Try to deduplicate code for KEG and AIGateway DataPlane reconcilers


programmer04:
## Problem Statement

> Prerequisite: both reconcilers should be covered with extensive integration and e2e tests before attempting the refactoring to ensure that nothing is broken in the process 

Those two controllers share a lot of similarities. Just compare
- `controller/aigateway/dataplane`
- `controller/eventgateway/dataplane`

thus probably some logic, helpers, etc. can be made generic and deduplicated to increase maintainability of the code and pay off in the future for similar controllers.

## Acceptance Criteria

- [ ] What can be refactored and deduplicated is deduplicated
- [ ] Tests are added/adjusted

PR #4918, opened 2026-07-14
fix(aigateway,eventgateway): always aggregate Ready condition via defer

alacuku:

ensureReadyStatus was only called on the happy path, so early-return branches (e.g. KonnectAIGateway not found) never set Ready=False.

Move the call into a defer so it runs regardless of reconcile outcome. Add a condition-scan fast-path at the top of ensureReadyStatus: if any non-Ready condition is already False, set Ready=False/DependenciesNotReady immediately without fetching the Deployment.

Apply the same fix symmetrically to KegDataPlane. Add unit tests for both new code paths.

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:


controller/aigateway/dataplane/controller.go
```diff
@@ -81,7 +81,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, aigwdp *aigatewayv1alpha1.AI
 
 	log.Trace(logger, "reconciling AIGatewayDataPlane resource")
 
-	defer func() { err = errors.Join(err, r.applyStatus(ctx, logger, aigwdp)) }()
+	defer func() {
+		err = errors.Join(err, ensureReadyStatus(ctx, r.Client, aigwdp))
+		err = errors.Join(err, r.applyStatus(ctx, logger, aigwdp))
```
programmer04, line 86: I wonder if it makes sense to combine them into something like `r.computeAndSetStatus(ctx, logger, aigwdp)`, but this way is also ok

programmer04 (approved):

controller/aigateway/dataplane/controller.go
alacuku: We have a tracking issue for this: https://github.com/Kong/kong-operator/issues/4842

PR #4924, opened 2026-07-14
feat(hybridgateway): konghq.com/tags propagation to Konnect entities

### LINKED ISSUE #4891: `konghq.com/tags` annotation on `HTTPRoute` is not propagated to generated `KongRoute.spec.tags`
## Current Behavior

When using Kong Operator in Hybrid Gateway mode, the `konghq.com/tags` annotation applied to a native Gateway API `HTTPRoute` is not propagated to the generated `KongRoute.spec.tags`.

This prevents users from managing only vendor-neutral Gateway API resources while still taking advantage of Kong entity tags for organization, ownership, search, filtering, and Konnect.

The generated `KongRoute` is created successfully, but the `KongRoute.spec.tags` key is not created. The `konghq.com/tags` annotation remains only on the `HTTPRoute`.

## Expected Behavior
Applying
```
metadata:
  annotations:
    konghq.com/tags: "tag1key:tag1value,tag2key:tag2value"
``` 
to an `HTTPRoute` should result in the generated `KongRoute` containing
```
spec:
  tags:
    - tag1key:tag1value
    - tag2key:tag2value
```
This would be consistent with the goal of allowing application teams to manage only Gateway API resources while the Operator generates Kong-specific resources.

## Steps To Reproduce
Create an `HTTPRoute` like:
```
kind: HTTPRoute
metadata:
  name: example
  annotations:
    konghq.com/tags: "tag1key:tag1value,tag2key:tag2value"
spec:
  parentRefs:
    - name: kong
  rules:
    - matches:
        - path:
            type: PathPrefix
            value: /
      backendRefs:
        - name: example-service
          port: 80
```
After reconciliation, observe that the generated `KongRoute` does not contain the expected `spec.tags`.

## Operator Version
Kong Operator: v2.2.0

## `kubectl version`
Client Version: v1.34.6-eks-bbe087e
Kustomize Version: v5.7.1
Server Version: v1.34.9-eks-8f14419

sandromodarelli:

Propagates the `konghq.com/tags` annotation onto the Kong CRDs generated by the Konnect-hybrid gateway translator, so user tags flow through to Konnect entities.

Tag source per generated entity (mirrors where each entity's other Kong annotations are read from):

| Set `konghq.com/tags` on | Tagged Konnect entity |
|---|---|
| backend `Service` | KongService, KongUpstream, KongTarget |
| `HTTPRoute` / `TLSRoute` | KongRoute |
| listener / client-cert `Secret` | KongCertificate, KongSNI |

Building blocks: `metadata.ExtractTags` parser, `utils.TagsFromBackendRefs` resolver (first-wins across a rule's backendRefs), and `WithSpecTags` on the six taggable builders. Tags are set on the CRD `spec.tags`; the existing `controller/konnect/ops` layer already merges `spec.tags` + annotation + k8s-metadata, so no changes there.

**Which issue this PR fixes**

Fixes #4891

- **Certificates are Secret-only.** KongCertificate/KongSNI tags come solely from the TLS Secret — tags on the Gateway listener or backend Service do not leak onto them. This matches on-prem KIC behavior (`GenerateTagsForObject(secret)`). Negative tests assert the isolation.
- **First-wins vs per-target split.** For a rule with multiple weighted backend Services carrying different tags: the shared KongService/KongUpstream take the first tagged Service's tags (consistent with how `protocol`/`host-header` already resolve), while each KongTarget takes its own backing Service's tags.
- **KongPlugin is intentionally out of scope** — KongPlugin/KongPluginBinding have no `spec.tags` field.

**PR Readiness Checklist**:


controller/hybridgateway/metadata/tags.go
programmer04, line 1: We already have a helper here `pkg/metadata/tags.go`

programmer04 (approved):

pmalek:
I believe we should make this known to users through a docs page (explicitly marked as 2.2+).

Not sure what the best place for this would be though. Perhaps a new sub page under https://developer.konghq.com/index/operator/#konnect--key-concepts?

cc: @justindavies

PR #5028, opened 2026-07-23
feat(UDPRoute): enhance translator, set ResolvedRefs and enable ReferenceGrant for UDPRoute

WeichengWang1:
**What this PR does / why we need it**:

The second part of #4429 
- enhance UDPRoute translator per GEP-2645#multiple-udproutes-attaching-to-the-same-listener
  - New translate_l4route_conflict.go: generic GEP-2645 L4 arbitration (shared by UDPRoute + TCPRoute via type constraint). Single winner per listener — oldest CreationTimestamp, ties broken by namespace/name. Only the winner is rendered to the dataplane.
  - translate_udproute.go rewritten: validate → group by listener → pick winner → render only winners. Arbitration losers still report Accepted=True (attached per spec); only their dataplane config is suppressed.
- set ResolvedRefs and enable ReferenceGrant for UDPRoute

**Which issue this PR fixes**

Fixes #

**Special notes for your reviewer**:

- GEP-2645. A UDPRoute that loses listener arbitration is still "attached" — it gets Accepted=True and is counted in Listener.Status.AttachedRoutes. Only its Kong config is suppressed. This split (status layer vs translation layer) is intentional; see the comments on countAttachedUDPRoutes and ingressRulesFromUDPRoutes.
  - The Programmed condition will be True for UDPRoutes loses listener arbitration, since it's determined by `r.DataplaneClient.KubernetesObjectIsConfigured`, but note that `KongClient.Update()` will skip object report updates when there is no config change, so the Programmed condition will first be `Unknown` and finally be `True` after other resources trigger config changes.
  - Will be covered by e2e tests in following PRs.
- Conflict logic is generalized to cover TCPRoute too (constraint *UDPRoute | *TCPRoute), though TCPRoute remains disabled in supportedRoutesByProtocol for now.

**PR Readiness Checklist**:

Complete these before marking the PR as `ready to review`:

- [ ] the `CHANGELOG.md` release notes have been updated to reflect significant changes

PR #5067, opened 2026-07-27
feat: keep listener predicates in translator same with controller

### LINKED ISSUE #5060: keep listener predicates in translator same with controller
## Problem Statement

In #5028, we introduced translate_l4_route, which requires query listeners referenced by routes. 
But translator query resources from CacheStores, running in different goroutine with kic, store in translator doesn't contain GatewayClass and Namespace.

In #5028, the predicates refactored, reusable checks extracted to `listener_attach.go`, in translator, only checked Programmed + AllowedRoutes(Kind/All/Same), while GatewayClass-ownership and Selector-namespace stayed out.

This issue aims to make the listener predicates in translator same with controller.

...

## Proposed Solution

Add GatewayClass and Namespace to translator's CacheStore.
Add checks in `listener_attach.go`.
Reuse it in both kic and translator.

...

## Additional Information

...

## Acceptance Criteria

- [ ] As a user, I can ...

WeichengWang1:

#5028 added L4Route (UDPRoute/TCPRoute) support to the dataplane translator, including listener-attachment arbitration. That arbitration only checked AllowedRoutes Kind and a subset of Namespaces.From (All/Same) — it had no way to check GatewayClass controller-ownership or evaluate Namespaces.From: Selector, because the translator's store.CacheStores never cached GatewayClass or Namespace objects (both TODOs left in the code from #5028).

Net effect: the translator's listener predicates diverged from what the KIC controller (getSupportedGatewayForRoute / routeMatchesListenerAllowedRoutes) actually enforces:
- A Gateway whose GatewayClass is controlled by a different ingress controller could still have its L4 routes arbitrated and pushed to Kong.
- A listener restricting AllowedRoutes.Namespaces via label Selector was always treated as "not attached" in the translator, rather than being evaluated.

This PR closes that gap so the translator's predicates match the controller's exactly, per #5060.

**Changes**

Cache: 
- added GatewayClass and Namespace to the translator's CacheStores (regenerated via the cache-stores generator), plus Storer.GetGatewayClass/GetNamespace.

Cache population:
- GatewayClassReconciler now pushes every observed GatewayClass into the dataplane cache (UpdateObject/DeleteObject), unconditionally — no ownership-gated watch predicate, so a GatewayClass reassigned away from this controller is reflected immediately rather than going stale.
- New CoreV1NamespaceReconciler, gated on GatewayAPIGatewayController && (GatewayAPITCPRouteController || GatewayAPIUDPRouteController) since Namespace caching is only needed for L4Route arbitration.

Shared predicates (ingress-controller/internal/gatewayapi/listener_attach.go):
- ListenerAllowsNamespace now fully resolves AllowedRoutes.Namespaces.From: Selector via an injected NamespaceGetter, returning (bool, error) instead of the old (ok, handled) shape.
- New GatewayClassControlledBy(gatewayClass, controllerName) helper.
- Both the KIC controller (route_utils.go, gateway_utils.go) and the translator (translate_l4route_conflict.go) now call these same shared functions instead of each having its own (partially duplicated, partially incomplete) logic.

**Which issue this PR fixes**

Fixes #5060 

**PR Readiness Checklist**:


tao12345666333 (approved):
LGTM.

Two non-blocking nits:

1. Changelog is empty for this change
2. Could we add envtest  case for the township-reassignment path?

WeichengWang1:
> LGTM.
> 
> Two non-blocking nits:
> 
> 1. Changelog is empty for this change
> 2. Could we add envtest  case for the township-reassignment path?

1. this pr is not related to user's behavior, imo, changelog is not necessary.
2. sure, will draft a pr towards it.

PR #5095, opened 2026-07-29
chore: bump Kong Gateway to 3.15 and adjust test suites

programmer04:

For Kong Gateway 3.15, a license is required not only for particular features, but also to access the admin API. Thus, it has to be injected basically for all test suites. License helpers are now consolidated 

**Which issue this PR fixes**

Replaces #5098, #5083 and #4765

**PR Readiness Checklist**:

### DIFF
--- a/ingress-controller/internal/admission/validation/gateway/httproute.go
+++ b/ingress-controller/internal/admission/validation/gateway/httproute.go
@@ -11,6 +11,7 @@ import (
 	"github.com/kong/go-kong/kong"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
 
 	"github.com/kong/kong-operator/v2/ingress-controller/internal/admission/validation"
 	gatewaycontroller "github.com/kong/kong-operator/v2/ingress-controller/internal/controllers/gateway"
@@ -260,9 +261,11 @@ func validateWithKongGateway(
 			route.Override(logr.Discard(), kongVersion)
 			ok, msg, err := routesValidator.Validate(ctx, &route.Route)
 			if err != nil {
-				return false, fmt.Sprintf("Unable to validate HTTPRoute schema: %s", err.Error())
-			}
-			if !ok {
+				// Validation against the Kong Gateway Admin API is best-effort: a failure here
+				// (e.g. DataPlane with license not yet provisioned) shouldn't block the HTTPRoute.
+				ctrllog.FromContext(ctx).Info(fmt.Sprintf("Unable to validate HTTPRoute schema due to: %s, allowing", err))
+				continue
+			} else if !ok {
 				errMsgs = append(errMsgs, msg)
 			}
 		}
--- a/test/integration/ko/suite_test.go
+++ b/test/integration/ko/suite_test.go
@@ -7,5 +7,5 @@ import (
 )
 
 func TestMain(m *testing.M) {
-	integration.Suite(m)
+	integration.Suite(m, integration.WithKongLicense())
 }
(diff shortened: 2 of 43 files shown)


test/test_dependencies.yaml
```diff
@@ -7,26 +7,22 @@ integration:
   helm:
     # renovate: datasource=helm depName=kong registryUrl=https://charts.konghq.com versioning=helm
     kong: '3.4.1'
-  # renovate: datasource=docker depName=kong versioning=docker
-  kong-oss: '3.9.3'
```
pmalek, line 11: Why would we want to remove this? We can still test OSS even if the license is injected into the env.

Without this we'll not be testing Kong OSS anymore.

test/test_dependencies.yaml
```diff
@@ -7,26 +7,22 @@ integration:
...
   # renovate: datasource=docker depName=kong/kong-gateway versioning=docker
-  kong-ee: '3.14.0.9'
+  kong-ee: '3.15.0.2'
 
 kongintegration:
-  # renovate: datasource=docker depName=kong versioning=docker
-  kong-oss: '3.9.3'
```
pmalek, line 17: Ditt here, what's the reason for removing? The maintenance cost of keeping separate EE and OSS golden files?

programmer04: Mostly yes, since 3.9 is in setting mode til end of the year and to avoid basically a double run of CI (we did it partially, so the coverage wasn't full). To streamline our experience and be sure that we develop against the newest one, and to avoid guarding some test cases by skipping for non-enterprise (since they were always skipped, probably by mistake - I discovered bugs in them). 

I tried to check what we promise in terms of Kong Gateway support, but we don't have anything there https://developer.konghq.com/operator/reference/version-compatibility/ 🤔

pmalek: I'd double check this with @justindavies.

Removing 3.9 from testing matrix should be a conscious decision we make and I'm not sure if we want to make it.

programmer04: I brought it back in a similar capacity to what it used to be, at least until the end of the year

pmalek:
@programmer04 Can you resolve conflicts so that we can review and merge this?

programmer04:
@pmalek it's rebased now, PTAL

ingress-controller/internal/admission/validation/gateway/httproute.go
```diff
@@ -260,9 +261,11 @@ func validateWithKongGateway(
 			route.Override(logr.Discard(), kongVersion)
 			ok, msg, err := routesValidator.Validate(ctx, &route.Route)
 			if err != nil {
-				return false, fmt.Sprintf("Unable to validate HTTPRoute schema: %s", err.Error())
-			}
-			if !ok {
+				// Validation against the Kong Gateway Admin API is best-effort: a failure here
+				// (e.g. DataPlane with license not yet provisioned) shouldn't block the HTTPRoute.
+				ctrllog.FromContext(ctx).Info(fmt.Sprintf("Unable to validate HTTPRoute schema due to: %s, allowing", err))
```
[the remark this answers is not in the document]
pmalek: This is quite important I think. 

I understand the decision here to not block admission when Kong is unavailable but I'd make it clear in the changelog about the changed behavior.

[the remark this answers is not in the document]
programmer04: You're right, it's basically a fix, since the webhook has always declared to provide best-effort validation. I updated changelog

test/integration/ko/suite_test.go
```diff
@@ -7,5 +7,5 @@ import (
 )
 
 func TestMain(m *testing.M) {
-	integration.Suite(m)
+	integration.Suite(m, integration.WithKongLicense())
```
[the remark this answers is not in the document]
pmalek: We could address this but it's low value because we inject license everywhere.

It's only an issue for PRs from forks which is already broken as of today.

pmalek (approved):

PR #5117, opened 2026-07-31
feat: add option to disable probes for DataPlane containers and initContainers

### LINKED ISSUE #5113: Support explicitly disabling the DataPlane proxy readiness probe
## Problem Statement

Kong Operator always adds a readiness probe to the DataPlane `proxy` container. Users can customize the probe through `PodTemplateSpec`, but there is no supported way to explicitly remove it.

Omitting the probe or setting `readinessProbe: null` is treated as unspecified, so the operator restores its default probe.
Some environments use an external readiness condition as the authoritative rollout signal. Our use case is AWS Load Balancer Controller Pod Readiness Gates:

- The AWS controller injects a target-health readiness gate into the Pod.
- A replacement Pod remains unready until AWS confirms it is a healthy NLB target.
- The NLB health check calls Kong’s `/status/ready` endpoint.
- The Deployment uses `maxUnavailable: 0`, so the old Pod must remain available until its replacement becomes NLB healthy.

Kubernetes combines container readiness and Pod readiness gates using AND semantics. The hardcoded container readiness probe therefore remains an additional readiness condition that users cannot disable, even when external target health is intended to be the authoritative rollout signal.

This request is **not** asking to remove or change the default readiness probe globally. Existing behavior should remain unchanged unless users explicitly opt out.

Related issues:

- #1117 added/fixed support for customizing readiness-probe values.
- #1235 established `/status/ready` as the appropriate default for Konnect-managed DataPlanes.

Neither provides a way to explicitly disable the probe.

## Proposed Solution

Add an explicit, opt-in DataPlane deployment setting that disables the readiness probe for the `proxy` container.

One possible API is:

```yaml
spec:
  deployment:
    readinessProbe:
      enabled: false
```

Alternatively, an enum could make the intent clearer:

```yaml
spec:
  deployment:
    readinessProbePolicy: Disabled
```

We are open to another API shape that better fits the operator’s conventions. The important requirement is distinguishing between:

1. The probe was not specified, so the operator should apply its existing default.
2. The probe was explicitly disabled, so the generated Deployment must not contain a readiness probe for the `proxy` container.

The explicitly disabled state should take precedence over:

- The operator’s default `/status` probe.
- Konnect-specific `/status/ready` defaulting.
- `PodTemplateSpec` merge and default processing.

## Backward Compatibility and Rollout Safety

An operator release containing this capability must remain safe to deploy before the option is enabled.

- Existing DataPlanes must retain their current readiness probes.
- Merely upgrading the operator must not change or roll existing DataPlane Deployments.
- Enabling or disabling the new setting is expected to change the Pod template and cause one normal controlled Deployment rollout.
- Existing custom readiness-probe configurations must continue working.

Documentation should identify this as an advanced option and explain that disabling the container readiness probe transfers readiness responsibility to Pod readiness gates or another external mechanism.

## Acceptance Criteria

- [ ] Existing DataPlane and GatewayConfiguration resources retain their current readiness probes by default.
- [ ] A user can explicitly disable the DataPlane `proxy` container readiness probe.
- [ ] When disabled, the generated Deployment and resulting Pods contain no `readinessProbe` for the `proxy` container.
- [ ] Konnect-specific defaulting does not restore the probe when explicitly disabled.
- [ ] Existing custom readiness-probe configurations continue working unchanged.
- [ ] The behavior works for directly created DataPlanes and DataPlane options supplied through `GatewayConfiguration`.
- [ ] Upgrading to a version containing the feature causes no DataPlane rollout unless the setting is explicitly changed.
- [ ] Tests cover default, custom, and disabled modes.
- [ ] CRDs and Helm chart CRDs are updated where required.
- [ ] Documentation explains the operational consequences and intended readiness-gate use case.

pmalek:

**Which issue this PR fixes**

Fixes https://github.com/Kong/kong-operator/issues/5113

**PR Readiness Checklist**:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -110,6 +110,8 @@
 - Hybridgateway: add controller and translator support for Gateway API `TCPRoute` resources.
   [#4727](https://github.com/Kong/kong-operator/pull/4727)
   [#5018](https://github.com/Kong/kong-operator/pull/5018)
+- Added option to disable `DataPlane`'s probes by specifying it as `{}`.
+  [#5117](https://github.com/Kong/kong-operator/pull/5117)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -111,6 +111,9 @@
   [#4727](https://github.com/Kong/kong-operator/pull/4727)
   [#5018](https://github.com/Kong/kong-operator/pull/5018)
 - Added option to disable `DataPlane`'s probes by specifying it as `{}`.
+  Using `{}` previously caused the operator to use default values for the probes
+  (no effect), which was not the intended behavior.
+  Now, the operator will not set any probes on the `DataPlane` whenever `{}` is specified.
   [#5117](https://github.com/Kong/kong-operator/pull/5117)
 
 ### Changed
--- a/controller/dataplane/owned_deployment_test.go
+++ b/controller/dataplane/owned_deployment_test.go
@@ -18,6 +18,7 @@ import (
 
 	operatorv1beta1 "github.com/kong/kong-operator/v2/api/gateway-operator/v1beta1"
 	"github.com/kong/kong-operator/v2/controller/pkg/op"
+	"github.com/kong/kong-operator/v2/modules/manager/scheme"
 	"github.com/kong/kong-operator/v2/pkg/consts"
 	k8sresources "github.com/kong/kong-operator/v2/pkg/utils/kubernetes/resources"
 )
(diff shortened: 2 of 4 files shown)


pmalek:
/muthur review this pr, take into account potential issues with unwanted DataPlane rollouts (which shouldn't happen)

CHANGELOG.md
```diff
@@ -110,6 +110,8 @@
 - Hybridgateway: add controller and translator support for Gateway API `TCPRoute` resources.
   [#4727](https://github.com/Kong/kong-operator/pull/4727)
   [#5018](https://github.com/Kong/kong-operator/pull/5018)
+- Added option to disable `DataPlane`'s probes by specifying it as `{}`.
```
randmonkey, line 113: Should this be considered as a breaking change as specifying probes to `{}` has different behaviour after the change?

pmalek:
/muthur re-review this PR, takes into account previous findings and the fixes that were pushed in this branch for them

CHANGELOG.md
pmalek: Even though users could use `{}` currently - which would have no effect - this was never mentioned in docs and guides.

I've added a note in f0c4440370452b2ccedc2e63d9458b0d3d05201b to explain this change a bit more but I wouldn't mark this as a breaking change.

randmonkey (approved):

PR #5136, opened 2026-08-05
fix: fix kongroute created without updating kongservice's hybrid-route annotation

### LINKED ISSUE #5133: KongService shared by multi Routes may stuck in terminating under a race window
## Current Behavior

The shared KongService became stuck terminating, while the new route remains in Konnect. In conformance this caused GRPCRoute traffic to route to the wrong backend or fail during cleanup.

This issue is not specific to GRPCRoute and not deterministic, happening under a race condition.

KongRoute creation and KongService annotation updates are both enforced by `controller/hybridgateway/reconciler_utils.go#enforceState`
`desiredObjects` is from converter.outputStore, `outputStore` is from converter.translate, which appends roughly with order: KongUpstream, KongService, KongRoute, KongTarget **per rule**

So with multiple backend/rule groups you can get:
```
v1 upstream
v1 service
v1 route         <- created, stopAtKind=KongRoute
v1 targets
v2 upstream      <- skipped
v2 service       <- the shared KongService, skipped, KongService != stopAtKind
v2 route         <- processed because KongRoute = stopAtKind, and KongRoute created
v2 targets
```
After v1 route created, stopAtKind set to KongRoute, so v2 service skipped, and v2 route created, then Konnect controller create route referencing the shared service via calling sdk-konnect.

In conclusion, Hybrid *Route can create a KongRoute that references a shared KongService before the shared KongService's live hybrid-routes-*Route annotation is updated to include the new *Route. This allows another *Route’s orphan cleanup to incorrectly consider the shared KongService orphaned and delete it.

CI example:
[ci job](https://github.com/Kong/kong-operator/actions/runs/30873910592/job/91923704503?pr=5120)

```
exact-matching previously created shared KongService v2
  annotation: gateway-conformance-infra/exact-matching

grpc-header-matching reconciliation starts
  desired KongService v2 annotation is computed as:
    exact-matching,grpc-header-matching

during enforceState():
  KongRoute v1 is created -> stopAtKind = KongRoute
  later KongService v2 is skipped because currentKind = KongRoute
  later KongRoute v2 is still created because currentKind = KongRoute

result:
  live KongService v2 annotation still only contains exact-matching
  Konnect has new KongRoute v2 referencing that shared KongService

exact-matching orphan cleanup runs
  removes exact-matching from the service annotation
  sees no routes left
  deletes KongService v2

Konnect rejects KongService deletion:
  constraint failed (type: foreign) as route references this value
```

Root cause:  
`enforceState()` uses `stopAtKind` to avoid moving past a newly-created object kind before dependencies are ready. This assumes the desired object list is effectively dependency-ordered by kind. However, hybrid Route converters emit resources per rule/backend group, not globally by dependency kind. After one `KongRoute` is created, `stopAtKind=KongRoute` can cause a later shared `KongService` annotation update to be skipped, while a later `KongRoute` is still allowed because it has the same kind as `stopAtKind`. If that service already exists and is Programmed from another Route, the later KongRoute can be created and synced to Konnect before the live service annotation records the new owner.

## Expected Behavior

a KongRoute should not be created unless the live referenced shared KongService annotation already includes the owning Route, or cleanup must otherwise detect live route references and not delete it.

WeichengWang1:

As described in #5133, Hybrid *Route can create a KongRoute that references a shared KongService before the shared KongService's live hybrid-routes-*Route annotation is updated to include the new *Route, which allows another *Route’s orphan cleanup to incorrectly consider the shared KongService orphaned and delete it.

This PR checked KongService before apply KongRoute, skip if KongService is terminating, or ensure annotation if KongService's hybrid-route annotation doesn't contain the Route.

**Which issue this PR fixes**

Fixes #5133 

**PR Readiness Checklist**:

### DIFF
--- a/controller/hybridgateway/reconciler_utils.go
+++ b/controller/hybridgateway/reconciler_utils.go
@@ -211,11 +211,25 @@ func enforceState[t converter.RootObject](ctx context.Context, cl client.Client,
 					stopAtKind = "KongService"
 					continue
 				}
+				if !svc.GetDeletionTimestamp().IsZero() {
+					log.Debug(logger, "Service is being deleted for route, waiting", "service", svcName)
+					objectsSkipped++
+					stopAtKind = "KongService"
+					continue
+				}
 				if !k8sutils.HasConditionTrue(konnectv1alpha1.KonnectEntityProgrammedConditionType, &svc) {
 					log.Debug(logger, "Service not Programmed yet for route, waiting", "service", svcName)
 					objectsSkipped++
 					stopAtKind = "KongService"
 					continue
+				}
+				if routeAnnotationKey != "" && !hybridRouteAnnotationContains(svc.GetAnnotations(), routeAnnotationKey, routeRef) {
+					if err := ensureHybridRouteAnnotationOnKongService(ctx, cl, &svc, routeAnnotationKey, routeRef); err != nil {
+						return false, false, fmt.Errorf("failed to update referenced KongService %s/%s hybrid-routes annotation before applying KongRoute %s: %w",
+							svc.GetNamespace(), svc.GetName(), desired.GetName(), err)
+					}
+					log.Debug(logger, "Updated referenced KongService hybrid-routes annotation before applying KongRoute",
+						"service", svcName, "route", desired.GetName())
 				}
 			}
 		case "KongPluginBinding":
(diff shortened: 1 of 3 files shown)


controller/hybridgateway/reconciler_utils.go
```diff
@@ -385,6 +399,39 @@ func mergeHybridRouteAnnotation(desired, existing *unstructured.Unstructured, an
...
+	annotationKey, routeRef string,
+) error {
+	base := svc.DeepCopy()
+	anns := svc.GetAnnotations()
+	if anns == nil {
+		anns = map[string]string{}
+	}
+
+	current := anns[annotationKey]
+	if hybridRouteAnnotationContains(anns, annotationKey, routeRef) {
```
randmonkey, line 419: Do we have a test case to confirm that the `ensureHybridRouteAnnotationOnKongService` is idempotent? 
Also I think some comment here should make the process clearer.

randmonkey (commented):
Mostly LGTM, one comment remaining.

controller/hybridgateway/reconciler_utils.go
WeichengWang1: updated in 1e0d65bddc2bee45d125cfafaf12587c24903298

randmonkey (approved):

PR #5193, opened 2026-08-06
KonnectConfigStore: implement the delete op

### LINKED ISSUE #5154: Konnect Vault: Add KonnectConfigStore CRD and reconciliation
## Problem Statement

Konnect Config Stores can back the Konnect vault backend, allowing sensitive values such as TLS private keys to live outside Kubernetes and be referenced by vault URI.

Today, Kong Operator does not provide a CRD for managing Konnect Config Store containers. Users must create the Config Store outside of Kubernetes and manually copy the generated Konnect Config Store ID into `KongVault.spec.config.config_store_id`.

This prevents a fully declarative GitOps workflow for the Config Store container and makes the setup harder to automate.

Parent epic: #5153

## Proposed Solution

Add a `KonnectConfigStore` CRD and reconciliation support for managing the existence of a Konnect Config Store container declaratively.

The CRD should manage the Config Store container itself, not the secret entries or secret values stored inside it.

Suggested API shape:

```yaml
kind: KonnectConfigStore
metadata:
  name: tls-cert-keys
  namespace: kong
spec:
  apiSpec:
    name: tls-cert-keys
  controlPlaneRef:
    type: namespacedRef
    namespacedRef:
      name: my-cp
      namespace: kong
```

The controller should:

- Resolve `spec.controlPlaneRef`.
- Reconcile the existence of the Konnect Config Store container through the Konnect Config Stores API.
- Store the resulting Konnect Config Store ID in status.
- Report readiness through standard Konnect entity conditions.
- Delete the Konnect Config Store container when the CRD is deleted.
- Never create, update, read, delete, or persist Config Store secret entries or secret values.

This task intentionally does not cover wiring `KongVault` to a `KonnectConfigStore`. That should be handled separately.

Secret values are intentionally out of scope. They must be written out-of-band through the Konnect API, Terraform, an external-secrets-style sync, or another user-managed mechanism.

## Acceptance Criteria

- [x] A `KonnectConfigStore` CRD is added. https://github.com/Kong/kong-operator/pull/5165
- [x] `KonnectConfigStore` is namespaced.
- [x] `KonnectConfigStore` can reference a Konnect Gateway Control Plane.
- [x] The controller creates the Konnect Config Store container when the CRD is created.
- [x] The controller deletes the Konnect Config Store container when the CRD is deleted.
- [x] `status.id` exposes the generated Konnect Config Store ID.
- [x] `status.controlPlaneID.id` records the resolved Control Plane ID.
- [x] Standard `Programmed` and reference-related conditions are reported. https://github.com/Kong/kong-operator/pull/5187
- [x] The CRD does not accept or store Config Store secret entries or secret values. https://github.com/Kong/kong-operator/pull/5165
- [x] The controller does not call Config Store secret-entry APIs. https://github.com/Kong/kong-operator/pull/5187
- [x] Tests are added for the new API and reconciliation behavior. https://github.com/Kong/kong-operator/pull/5193 https://github.com/Kong/kong-operator/pull/5202
- [ ] Documentation is updated with this new feature / enhancement. API references and samples are available; end-to-end user documentation is tracked in #5169.
- [x] Helm chart CRD templates are regenerated/updated to include the new CRD. https://github.com/Kong/kong-operator/pull/5165
- [x] As a user, I can declare a Konnect Config Store using a Kubernetes CRD and see its Konnect ID in status after reconciliation. https://github.com/Kong/kong-operator/pull/5193

### LINKED ISSUE #5191: KonnectConfigStore: implement the delete op
## Problem Statement

#5187 introduced the `KonnectConfigStore` CRD and wired up its `create` and `update` ops in `crd-from-oas/config.yaml`. The `delete` op was deliberately left out.

If I add `delete: {}` directly, it will fail with the following error.

```
controller/konnect/ops/conditions.go:1: : # github.com/kong/kong-operator/v2/controller/konnect/ops [github.com/kong/kong-operator/v2/controller/konnect/ops.test]
controller/konnect/ops/zz_generated_ops_konnect_configstore.go:78:39: cannot use parentID (variable of type string) as operations.DeleteConfigStoreRequest value in argument to sdk.DeleteConfigStore
controller/konnect/ops/zz_generated_ops_konnect_configstore.go:78:49: cannot use id (variable of type string) as operations.Option value in argument to sdk.DeleteConfigStore (typecheck)
package ops
1 issues:
* typecheck: 1
INFO File cache stats: 1 entries of total size 5.9KiB 
INFO Memory: 1082 samples, avg is 6963.9MB, max is 10429.3MB 
INFO Execution took 1m56.718143542s               
make: *** [generate.lint-fix] Error 1
```

Another thing about the delete API:

`DELETE /v2/control-planes/{controlPlaneId}/config-stores/{configStoreId}` takes a `force` query parameter that **defaults to false**, meaning: refuse to delete the Config Store while it still has secrets linked to it (400). The generator passes `nil` for every optional query parameter, i.e the default.

In the [PRD](https://docs.google.com/document/d/1IDezHAAssWI1dZ55TYwLmhH0oPf8pd6iSLhwVDwgpyE/edit?tab=t.0), some specific boundaries have been defined, and I think we can directly determine how to implement it based on the PRD.

> (_"FR-7 (MUST NOT): The Config Store CRD MUST NOT accept or store secret values; it manages the store container only."_) 

Per FR-7 the CRD never accepts or stores secret values, secrets are written out of band. So "the store has secrets" is the steady state, not an edge case. With force left at its default(`false value`), every delete would return 400; the ops layer only swallows 404, so a 400 becomes a retried failure and we are back at the stuck finalizer

## Proposed Solution

**Generator changes**:

  1. Support wrapped delete requests for single-parent entities , honour `ops.delete.path` the way update does.
  2. Allow generated delete ops to pass non-nil values for delete query parameters, so force can be set at all. (just a field)

Then add the delete stanza for `KonnectConfigStore` and regenerate.

  **Pass**` force=true`. Rationale:

  - Leaving it at the default makes deletion fail in the normal case and wedges the finalizer, violating FR-12(FR-12 (MUST): All resources in the chain MUST delete cleanly regardless of deletion order, without stuck finalizers requiring manual intervention.).
  - FR-7 constrains the CRD's API surface — it must not accept or store secret values. FR-5 explicitly puts creating and deleting the store
  container in scope. Konnect cascading to the container's contents on delete is container lifecycle, and matches ordinary Kubernetes
  semantics for deleting an owner.
  - Per NFR-4 this stays purely additive: no new API field now. If an opt-out is ever needed, adding an optional field later is backward
  compatible.

  **Do not block deletion on in-use references**. The real hazard is a user deleting only the Config Store CR while a KongVault /
  KongCertificate still references it. The instinctive guard — refuse to delete while referenced — would violate FR-12. The PRD puts this on
  the reference layer instead: FR-10 requires a KongVault pointing at an unavailable Config Store to surface that in its status and
  reconcile automatically once it is ready again.

  Document that deleting a KonnectConfigStore CR deletes every secret in that store in Konnect (NFR-2). This is the one genuinely surprising
  consequence and it must be explicit in the docs.

## Additional Information

...

## Acceptance Criteria

- [ ] As a user, I can ...

tao12345666333:

**Which issue this PR fixes**

Fixes https://github.com/Kong/kong-operator/issues/5191

**PR Readiness Checklist**:

### DIFF
--- a/controller/konnect/ops/zz_generated_ops_delete.go
+++ b/controller/konnect/ops/zz_generated_ops_delete.go
@@ -65,6 +65,8 @@ func DeleteGeneratedOps[
 		return deleteEventGatewayVirtualClusterProducePolicy(ctx, sdk.GetEventGatewayVirtualClusterProducePoliciesSDK(), ent)
 	case *konnectv1alpha1.KonnectAIGateway:
 		return deleteKonnectAIGateway(ctx, sdk.GetAIGatewaysSDK(), ent)
+	case *konnectv1alpha1.KonnectConfigStore:
+		return deleteKonnectConfigStore(ctx, sdk.GetConfigStoresSDK(), ent)
 	case *konnectv1alpha1.KonnectEventGateway:
 		return deleteKonnectEventGateway(ctx, sdk.GetEventGatewaysSDK(), ent)
 	case *konnectv1alpha1.Portal:
--- a/controller/konnect/ops/zz_generated_ops_konnect_configstore.go
+++ b/controller/konnect/ops/zz_generated_ops_konnect_configstore.go
@@ -64,6 +64,28 @@ func updateKonnectConfigStore(
 	return nil
 }
 
+func deleteKonnectConfigStore(
+	ctx context.Context,
+	sdk sdkkonnectgo.ConfigStoresSDK,
+	obj *konnectv1alpha1.KonnectConfigStore,
+) error {
+	parentID := obj.GetControlPlaneID()
+	if parentID == "" {
+		return CantPerformOperationWithoutParentIDError{Entity: obj, Parent: "KonnectGatewayControlPlane", Op: DeleteOp}
+	}
+	id := obj.GetKonnectStatus().GetKonnectID()
+
+	_, err := sdk.DeleteConfigStore(ctx, sdkkonnectops.DeleteConfigStoreRequest{
+		ControlPlaneID: parentID,
+		ConfigStoreID:  id,
+		Force:          sdkkonnectops.ForceTrue.ToPointer(),
+	})
+	if errWrap := wrapErrIfKonnectOpFailed(err, DeleteOp, obj); errWrap != nil {
+		return handleDeleteError(ctx, errWrap, obj)
+	}
+	return nil
+}
+
 func getKonnectConfigStoreForUID(
 	ctx context.Context,
 	sdk sdkkonnectgo.ConfigStoresSDK,
--- a/controller/konnect/ops/zz_generated_ops_konnect_configstore_test.go
+++ b/controller/konnect/ops/zz_generated_ops_konnect_configstore_test.go
@@ -140,3 +140,55 @@ func TestUpdateKonnectConfigStore_PropagatesSDKError(t *testing.T) {
 	err = updateKonnectConfigStore(ctx, sdk, obj)
 	require.ErrorContains(t, err, sdkErr.Error())
 }
+
+func TestDeleteKonnectConfigStore_UsesGeneratedSDKOps(t *testing.T) {
+	t.Parallel()
+
+	ctx := t.Context()
+	sdk := mocks.NewMockConfigStoresSDK(t)
+	obj := testGeneratedKonnectConfigStoreForSDKOps()
+	parentID := "parentID-1"
+	obj.SetControlPlaneID(parentID)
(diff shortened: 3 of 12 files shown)


tao12345666333:
The CI failure is due to an incident with GitHub Actions.
https://www.githubstatus.com/incidents/qcvjkzcs7j74

randmonkey (approved):

pmalek (approved):

PR #5207, opened 2026-08-07
fix(crd-from-oas): make OpenAPI-derived parent refs immutable

tao12345666333:

Our hand-written Konnect entities (KongService, KongRoute, etc.) all carry this on the type:

spec.<parent>Ref is immutable when an entity is already Programmed  e.g https://github.com/Kong/kong-operator/blob/a2a8594c6f38099eae02cf8faeb5ef64681f2542/api/configuration/v1alpha1/kongroute_types.go#L41

But the crd-from-oas only emits that rule for entities that configure a parentRef override in config.yaml. Entities whose parent is derived from an OpenAPI path parameter don't get it, the two template branches emit the same spec field, but only one of them gets the validation.

This PR fixes this inconsistent behavior.

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:


alacuku (approved):

pmalek (approved):

pmalek (commented):
We could also add tests for these when possible using [`NewCRDValidationTestCasesGroupParentRefChange`](https://github.com/Kong/kong-operator/blob/7a03a4be82d8ced768f7eb0ff4ebd01f6c4b12c3/test/crdsvalidation/common/suite_crd_parent_ref_change.go#L14) which is already used in generated CRDs for AIGateway.

tao12345666333:
Thank you for reviewing, I will address it separately in a subsequent PR.

PR #5223, opened 2026-08-10
test/e2e: add chainsaw coverage for AIGatewayModel route-model selectors

### LINKED ISSUE #5216: E2E Test: add coverage for AIGatewayModel route-model selectors (headers/body/path)
## Problem Statement

Add e2e coverage validating that `AIGatewayModel` correctly routes AI Gateway traffic based on each supported route-model selector type: `headers`, `body`, and `path`.

Today these selector types are only exercised at the CRD/unit level. We want confidence that, for each selector type, a real request through an `AIGatewayDataPlane` is matched to the correct model target based on the configured selector, and that non-matching requests are rejected.

alacuku:

Add a new e2e suite (aigateway/model-route-selectors) covering the `headers`, `body`, and `path` selector types on AIGatewayModel's config.route.model. Provisions a shared provider and one AIGatewayModel per selector type, then drives real chat-completions traffic to verify each selector type routes requests to the correct target based on its configured alias values, and that a non-matching alias is rejected.

Extend the shared aigw_chat_completion.sh script to support this suite's traffic verification steps.

**Which issue this PR fixes**

Fixes #5216 

The `path` selector type is not supported by Konnect yet, so step 9 (Verify-path-selector-routes-matching-alias) is expected to fail until Konnect ships support for it. Kept in as real coverage rather than removed/skipped.

**PR Readiness Checklist**:

### DIFF
--- a/test/e2e/chainsaw/aigateway/model-route-selectors/chainsaw-test.yaml
+++ b/test/e2e/chainsaw/aigateway/model-route-selectors/chainsaw-test.yaml
@@ -1,0 +1,171 @@
+apiVersion: chainsaw.kyverno.io/v1alpha1
+kind: Test
+metadata:
+  name: aigateway-model-route-selectors
+spec:
+  description: |
+    AIGatewayModel route selectors. Stands up a Konnect AIGateway
+    control plane, a shared OpenAI-compatible provider, and three AIGatewayModel
+    resources - one per `config.route.model` selector type (headers, body, path) -
+    then drives real chat-completions traffic to verify each selector type routes
+    requests to the correct target based on its configured alias `values`, and
+    that a non-matching alias is consistently rejected.
+  bindings:
+    - name: test_name
+      value: ($namespace)
+    - name: konnect_token
+      value: (env('KONNECT_TOKEN'))
+    - name: konnect_url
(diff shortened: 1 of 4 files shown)


programmer04 (dismissed):

test/e2e/chainsaw/aigateway/model-route-selectors/chainsaw-test.yaml
```diff
@@ -0,0 +1,171 @@
+apiVersion: chainsaw.kyverno.io/v1alpha1
+kind: Test
+metadata:
+  name: aigateway-model-route-selectors
+spec:
+  description: |
+    AI GW 1.1 — AIGatewayModel route selectors. Stands up a Konnect AIGateway
```
randmonkey: Should we update the version of AI GW here?

randmonkey (dismissed):

test/e2e/chainsaw/aigateway/model-route-selectors/chainsaw-test.yaml
alacuku: Yep, we should.

alacuku: @randmonkey PTAL

programmer04 (approved):

PR #5237, opened 2026-08-11
fix: fix routes become unaccepted and removed from dataplane when listener is transient unprogrammed

### LINKED ISSUE #4507: Routes incorrectly marked  Accepted=False/NoMatchingParent and removed from dataplane during transient listener state change

WeichengWang1:

In #4521 we tried to solve #4507, but just solved it partially.
As described in #4507, gateway controller will wipe listeners' status, and write back later, which makes listeners' status transient not Programmed.
Attachable listeners existed but none of them ready caused by such case, the *Route will become not Accepted and be removed from dataplane, which is unexpected.

This PR reordered the per-listener matching loop so the Programmed check runs last, and moved the `attachedListenerNames` append to happen right before it, once every structural check has passed. 
So a listener that's the exact right match but transiently not-yet-Programmed still shows up in `attachedListenerNames`, instead of vanishing from it and tripping the len==0 "no listener could ever attach" escape hatch in `isGatewayProgrammedForRoute`.

**Which issue this PR fixes**

Fixes #4507

- This PR split transient not-yet-Programmed with terminal not Programmed by additional checks towards `ResolvedRefs` condition in `isGatewayProgrammedForRoute` to avoid endless requeue.

- This PR reordered the per-listener matching loop so the Programmed check runs last, corresponding move the case `!listenerReady` to the last in:
```
reason := gatewayapi.RouteReasonNoMatchingParent
			switch {
			case matchingHostname != nil && *matchingHostname == metav1.ConditionFalse:
				// If there is no matchingHostname, the gateway Status Condition Accepted
				// must be set to False with reason NoMatchingListenerHostname
				reason = gatewayapi.RouteReasonNoMatchingListenerHostname
			case !allowedByAllowedRoutes || !allowedBySupportedKinds:
				reason = gatewayapi.RouteReasonNotAllowedByListeners
			case parentRef.SectionName != nil && !allowedByListenerName:
				// If ParentRef specified listener names but none of the listeners matches the name,
				// the gateway Status Condition Accepted must be set to False with reason RouteReasonNoMatchingParent.
				reason = gatewayapi.RouteReasonNoMatchingParent
			case parentRef.Port != nil && !portMatched:
				// If ParentRef specified a Port but none of the listeners matched, the gateway Status
				// Condition Accepted must be set to False with reason NoMatchingListenerPort
				reason = gatewayapi.RouteReasonNoMatchingParent
			case len(attachedListenerNames) > 0 && !listenerReady:
				reason = gatewayapi.RouteReasonNotAllowedByListeners
			}
```

**PR Readiness Checklist**:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -107,6 +107,9 @@
   Previously a single transient failure (e.g. a rejected admission webhook
   call) left the Service without its `konghq.com/plugins` annotation
   indefinitely, since nothing else would trigger another reconcile.
+- fix routes become unaccepted and removed from dataplane when listener is
+  transient unprogrammed
+  [#5237](https://github.com/Kong/kong-operator/pull/5237)
 
 ## [v2.3.0-rc.3]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -107,8 +107,8 @@
   Previously a single transient failure (e.g. a rejected admission webhook
   call) left the Service without its `konghq.com/plugins` annotation
   indefinitely, since nothing else would trigger another reconcile.
-- fix routes become unaccepted and removed from dataplane when listener is
-  transient unprogrammed
+- Fix Gateway API routes becoming transiently unaccepted and removed from dataplane
+  when listener is intermittently marked as Programmed=False.
   [#5237](https://github.com/Kong/kong-operator/pull/5237)
 
 ## [v2.3.0-rc.3]
--- a/ingress-controller/internal/controllers/gateway/route_utils.go
+++ b/ingress-controller/internal/controllers/gateway/route_utils.go
@@ -259,7 +259,7 @@ func getSupportedGatewayForRoute[T gatewayapi.RouteT](
 			// route would attach to, regardless of whether it's Programmed yet. Record it
 			// here, before the Programmed check, so a transiently-not-yet-Programmed
 			// listener (e.g. while the Gateway controller re-renders listener status) still
-			// shows up in attachedListenerNames. isGatewayProgrammedForRoute relies on this
+			// shows up in attachedListenerNames. isGatewaySettledForRoute relies on this
 			// to tell "no listener could ever attach this route" (terminal) apart from "the
 			// right listener exists but isn't Programmed yet" (transient, should wait).
 			attachedListenerNames = append(attachedListenerNames, listener.Name)
@@ -615,28 +615,23 @@ func isGatewaySettledForRoute(gateway supportedGatewayWithCondition) bool {
 	}
 
 	generation := gateway.gateway.Generation
+	sawTransient := false
 	for _, listenerName := range gateway.attachedListenerNames {
-		// Any one attached listener being Programmed is enough to proceed - mirrors `isRouteAccepted.
+		// Any one attached listener being Programmed is enough to proceed - mirrors `isRouteAccepted`.
 		if err := listenerProgrammedInStatus(listenerName, gateway.gateway.Status.Listeners); err == nil {
 			return true
 		}
-
 		// A listener whose ResolvedRefs condition is explicitly False will never
-		// become Programmed until that reference is fixed.
-		if gatewayapi.ListenerResolvedRefsFalse(listenerName, generation, gateway.gateway.Status.Listeners) {
+		// become Programmed until that reference is fixed. The same holds for other
+		// listener validation failures with dedicated reasons.
+		if gatewayapi.ListenerResolvedRefsFalse(listenerName, generation, gateway.gateway.Status.Listeners) ||
+			gatewayapi.ListenerProgrammedFalseForSettledReason(listenerName, generation, gateway.gateway.Status.Listeners) {
 			continue
 		}
-
-		// Other listener validation failures with dedicated reasons.
-		if gatewayapi.ListenerProgrammedFalseForSettledReason(listenerName, generation, gateway.gateway.Status.Listeners) {
-			continue
-		}
-
-		// Transient not Programmed.
-		return false
+		// Transient not Programmed - keep looking for a Programmed sibling first.
+		sawTransient = true
 	}
-
-	return true
+	return !sawTransient
 }
 
 // isHTTPReferenceGranted checks that the backendRef referenced by the HTTPRoute is granted by a ReferenceGrant.
--- a/ingress-controller/internal/controllers/gateway/route_utils.go
+++ b/ingress-controller/internal/controllers/gateway/route_utils.go
@@ -619,6 +619,13 @@ func isGatewayProgrammedForRoute(gateway supportedGatewayWithCondition) bool {
 		if err := listenerProgrammedInStatus(listenerName, gateway.gateway.Status.Listeners); err == nil {
 			return true
 		}
+
+		// A listener whose ResolvedRefs condition is explicitly False will never become Programmed
+		// until that reference is fixed. That's not a transient NotProgrammed, so treat it as Programmed
+		// to avoid waiting forever.
+		if gatewayapi.ListenerResolvedRefsFalse(listenerName, gateway.gateway.Status.Listeners) {
+			return true
+		}
 	}
 
 	return false
--- a/ingress-controller/internal/controllers/gateway/route_utils.go
+++ b/ingress-controller/internal/controllers/gateway/route_utils.go
@@ -182,7 +182,6 @@ func getSupportedGatewayForRoute[T gatewayapi.RouteT](
 			allowedByAllowedRoutes  = false
 			allowedBySupportedKinds = false
 			allowedByListenerName   = false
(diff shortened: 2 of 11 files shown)


ingress-controller/internal/gatewayapi/listener_attach.go
```diff
@@ -144,6 +144,26 @@ func ListenerSupportsRouteInStatus[T RouteT](route T, listenerName SectionName,
 	return nil
 }
 
+// ListenerResolvedRefsFalse reports whether listenerName's ResolvedRefs condition
+// in lss is explicitly False.
+func ListenerResolvedRefsFalse(listenerName SectionName, lss []ListenerStatus) bool {
```
pmalek: Can we add a UT for this?

pmalek (commented):
The change looks good to me. Let's make sure to add a changelog entry for this.

Left 1 minor comment

ingress-controller/internal/controllers/gateway/route_utils.go
```diff
@@ -607,6 +619,13 @@ func isGatewayProgrammedForRoute(gateway supportedGatewayWithCondition) bool {
 		if err := listenerProgrammedInStatus(listenerName, gateway.gateway.Status.Listeners); err == nil {
 			return true
 		}
+
+		// A listener whose ResolvedRefs condition is explicitly False will never become Programmed
+		// until that reference is fixed. That's not a transient NotProgrammed, so treat it as Programmed
+		// to avoid waiting forever.
+		if gatewayapi.ListenerResolvedRefsFalse(listenerName, gateway.gateway.Status.Listeners) {
```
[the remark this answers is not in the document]
WeichengWang1: nice catch!

ingress-controller/internal/controllers/gateway/route_utils.go
```diff
@@ -607,6 +619,13 @@ func isGatewayProgrammedForRoute(gateway supportedGatewayWithCondition) bool {
...
+			return true
+		}
```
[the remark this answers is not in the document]
WeichengWang1: nice catch, but if the listener is Programmed, we should return true instead of continue

WeichengWang1:
/muthur review this PR again, all review comments resolved and also rename isGatewayProgrammedForRoute to isGatewaySettledForRoute.

CHANGELOG.md
```diff
@@ -107,6 +107,9 @@
   Previously a single transient failure (e.g. a rejected admission webhook
   call) left the Service without its `konghq.com/plugins` annotation
   indefinitely, since nothing else would trigger another reconcile.
+- fix routes become unaccepted and removed from dataplane when listener is
+  transient unprogrammed
```
pmalek: How about:


```suggestion
- Fix Gateway API routes becoming transiently unaccepted and removed from dataplane when listener is
  intermittently marked as Programmed=False.
```

ingress-controller/internal/controllers/gateway/route_utils.go
```diff
@@ -603,13 +614,24 @@ func isGatewayProgrammedForRoute(gateway supportedGatewayWithCondition) bool {
 		return true
 	}
 
+	generation := gateway.gateway.Generation
+	sawTransient := false
```
pmalek: This is a bit unclear: saw transient what? Transient programmed, not programmed, something else?

ingress-controller/internal/controllers/gateway/route_utils.go
```diff
@@ -590,7 +601,7 @@ func isRouteAccepted(gateways []supportedGatewayWithCondition) bool {
 	return false
 }
 
-func isGatewayProgrammedForRoute(gateway supportedGatewayWithCondition) bool {
+func isGatewaySettledForRoute(gateway supportedGatewayWithCondition) bool {
```
pmalek: Nit: perhaps something like:

```suggestion
func isGatewayProgrammedSettledForRoute(gateway supportedGatewayWithCondition) bool {
```

with a comment explaining would help in readability? (I'm fine with other names as well but plain `isGatewaySettledForRoute` could suggest that we're awaiting for a condition `Settled`, which is not the case.

ingress-controller/internal/controllers/gateway/route_utils.go
WeichengWang1: good point, renamed it to transientNotProgrammed, and added some comments towards isGatewayProgrammedSettledForRoute

ingress-controller/internal/controllers/gateway/route_utils.go
WeichengWang1: updated in 64fd58f4bafb5dc538156cf5306e0b0e0eff1d22

randmonkey (approved):

PR #5263, opened 2026-08-12
feat(crd-from-oas)Add SupportCrossNamespaceReference in ReferenceConfig

### LINKED ISSUE #5258: Support cross namesapce reference in generated CRDs in `crd-from-oas`
Currently crd-from-oas does not support cross namespace reference for the fields referencing other resources in CRDs. Although the field contains a `namespace` field, The generated `resolve*Ref` functions directly ignores the `namespace` of the reference.

Now we need to update the `crd-from-oas` to support configuring CRDs to refer resources in other namespaces. It should include:

- [ ] Adding fields to configure whether cross namespace reference is supported in `ReferenceConfig`
- [ ] Generating the code for checking reference grants and resolving cross namespace references
- [ ] Generating the watch/index functions on `KongReferenceGrant` in the controllers for CRDs that supports cross namespace reference

randmonkey:

Add `SupportCrossNamespaceReference` to support referencing resources in other namespaces.

When it is set to true, it generates code for the CRDs to check and resolve references to resources in other namespaces and watching `KongReferenceGrant`s in the controllers.

**Which issue this PR fixes**

Fixes #5258

**PR Readiness Checklist**:

### DIFF
--- a/crd-from-oas/pkg/generator/templates.go
+++ b/crd-from-oas/pkg/generator/templates.go
@@ -545,6 +545,7 @@ import (
 {{- if .NeedsClient}}
 
 {{if .NeedsSecretFetchImport}}	corev1 "k8s.io/api/core/v1"
+{{end}}{{if .NeedsCrossNamespaceCheck}}	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 {{end}}{{if .References}}	apierrors "k8s.io/apimachinery/pkg/api/errors"
 {{end}}	"sigs.k8s.io/controller-runtime/pkg/client"
 {{- end}}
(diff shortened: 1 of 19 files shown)


crd-from-oas/pkg/generator/templates.go
```diff
@@ -1121,6 +1124,39 @@ func (obj *{{$.EntityName}}) ResolveKonnectReferences(ctx context.Context, cl cl
...
+			kind = "{{.DefaultKind}}"
+		}
+		checks = append(checks, CrossNamespaceReferenceCheck{
+			FromGVK:       metav1.GroupVersionKind{Group: GroupVersion.Group, Version: GroupVersion.Version, Kind: "{{$.EntityName}}"},
+			ToGVK:         metav1.GroupVersionKind{Group: GroupVersion.Group, Version: GroupVersion.Version, Kind: kind},
+			FromNamespace: obj.GetNamespace(),
+			ToNamespace:   ns,
+			ToName:        ref.Name,
+		})
+	}
```
WeichengWang1, line 1155: ToGVK always built from same package GroupVersion as FromGVK — assumes referenced kind lives in same API group as referrer.
Will it be possible that  a future SupportCrossNamespaceReference ref points cross-group?

randmonkey: Currently `RefernceConfig` does not support referrring to resources in other groups. So I decide not to change this since this is for an unrelated feature. If we do want to generate CRDs referring to resources in other groups/versions here, we should also change this when we implement it.

WeichengWang1 (approved):

PR #5265, opened 2026-08-13
feat: enable AIGatewayAgent and AIGatewayModel to reference IdentityProvider

### LINKED ISSUE #5264: Support AIGatewayModel/AIGatewayAgent reference AIGatewayIdentityProvider
## Problem Statement

`AIGatewayIdentityProvider` is added into `spec.apiSpec.access` of `AIGatewayAgent`, which should be a cross object reference, same in `spec.apiSpec.api.access` of `AIGatewayModel`.
Currently it's a plain text requires Konnect names, we need to make it able to reference the CRD `AIGatewayIdentityProvider`.

## Proposed Solution

...

## Additional Information

...

## Acceptance Criteria

- [ ] As a user, I can ...

WeichengWang1:

This PR added reference entries in `config.yaml` turning `IdentityProviders` into CRD refs (`AIGatewayIdentityProviderRef`, kind `AIGatewayIdentityProvider`) for `AIGatewayAgent.spec.apiSpec.access.identityProviders` and both `AIGatewayModel.spec.apiSpec.{api,model}.access.identityProviders`, Also updated samples yaml and chainsaw tests.

**Which issue this PR fixes**

Fixes #5264 

**PR Readiness Checklist**:


crd-from-oas/pkg/generator/templates.go
```diff
@@ -1147,7 +1147,9 @@ func (obj *{{$.EntityName}}) ResolveKonnectReferences(ctx context.Context, cl cl
 	// ancestor pointer means that part of the config wasn't set, so the payload
 	// is left untouched.
 {{- end}}
+{{- if .Cond}}
```
randmonkey, line 1150: Why did we add a `if .Cond` in the template layer? Would it be redandunt? (In other words, is there any scenario that when the `.Cond` is false, the `if .Cond` block in the generated go source will have syntax error or runtime error?)

WeichengWang1: Before the fix, the template unconditionally wrapped the injection body:

      if {{.Cond}} {
      ... inject resolved refs ...
      }

This assumed Cond is never empty. That assumption held for every ref shape that existed before — because they all had at least one pointer ancestor: the ACL union wrapper, a root-union variant selector (API/Model on AIGatewayModel), or an array-of-structs hop.

spec.apiSpec.access.identityProviders on AIGatewayAgent, breaks it: Access is a plain value-typed struct field (Access AIGatewayAgentAccess, not a pointer, not part of any union), and IdentityProviders is a leaf array (arrays are never pointers). Walking that chain, appendRefInjection finds zero pointer segments → Cond = "".

Rendered output:

      if  {
              access, _ := payload["access"].(map[string]any)
              ...
      }

if  { — empty condition, invalid Go.

randmonkey (approved):

PR #5280, opened 2026-08-13
fix(hybridgateway): propagate konghq.com/tags to generated KongPlugin…

### LINKED ISSUE #5277: konghq.com/tags annotation not propagated to Konnect Plugin entity in Hybrid Gateway mode
## Current Behavior

For Hybrid Gateway (Gateway API) mode, `konghq.com/tags` set on a `KongPlugin` does not reach the resulting Plugin entity in Konnect, whether the plugin is generated from built-in `HTTPRoute` filters (e.g. `RequestHeaderModifier`) or referenced via an `ExtensionRef` filter pointing at a user-managed `KongPlugin`.

This is a distinct gap from #4891 (fixed by #4924): that fix propagated `konghq.com/tags` into `spec.tags` for `KongRoute`, `KongService`, `KongUpstream`, `KongTarget`, `KongCertificate`, and `KongSNI`. `KongPlugin` and `KongPluginBinding` were explicitly out of scope for that fix (per the PR description) because neither type has a `spec.tags` field to populate — confirmed in `api/configuration/v1/kongplugin_types.go` and `api/configuration/v1alpha1/kongpluginbinding_types.go`.

Checked against the merge commit for #4924 (`d0732aa68e5d3f388ea19a6aad23fac843d9fb79`): it touched only `controller/hybridgateway/builder/{kongcertificate,kongroute,kongservice,kongsni,kongtarget,kongupstream}.go` and related converters/tests — nothing under `controller/hybridgateway/plugin/` or `controller/hybridgateway/pluginbinding/`.

## Root cause

In `controller/hybridgateway/plugin/plugin.go`, `PluginsForRule`:
- For the `ExtensionRef` case, the doc comment states the referenced `KongPlugin` is "returned as-is without rebuild, annotation modification, or existence checks" — but the code below it does exactly that: it fetches the referenced plugin, then calls `builder.NewKongPlugin()...WithAnnotations(httpRoute, pRef)...Build()` to produce a new, differently-named copy carrying over only `PluginName` and `Config`, not the original's own annotations.
- The same `.WithAnnotations(httpRoute, pRef)` call is used for the built-in-filter-generated plugins earlier in the same function.

In `controller/hybridgateway/pluginbinding/pluginbinding.go`, both `BindingForPluginAndRoute` and `BindingForPluginAndService` call the same `.WithAnnotations(httpRoute, pRef)` on the generated `KongPluginBinding`.

All of these route through `metadata.BuildAnnotations(obj, parentRef)` (`controller/hybridgateway/metadata/annotation.go`), which only ever sets two internal tracking annotations (`GatewayOperatorHybridGatewaysAnnotation` and a `hybrid-routes-<kind>` key) — it never reads or forwards a `konghq.com/tags` annotation from either the `HTTPRoute` or the source `KongPlugin`.

Contrast with the plain (non-Hybrid-Gateway) `KongPluginBinding` reconciliation path in `controller/konnect/ops/ops_kongpluginbinding.go`, which does work correctly: `GenerateTagsForObject(pluginBinding, metadata.ExtractTags(plugin)...)` reads the `konghq.com/tags` annotation directly off the referenced `KongPlugin`'s own metadata (`pkg/metadata/tags.go`, `ExtractTags`) and forwards it to Konnect. This code path predates #4891 (present since Sept 2024, commit `4491b846`) and is unaffected by it — it's simply not used by the Hybrid Gateway (`HTTPRoute`) reconciler.

**Net effect:** for Hybrid Gateway-generated Plugins and PluginBindings, there is currently no path for a `konghq.com/tags` annotation to reach the resulting Konnect Plugin entity — whether set on the `HTTPRoute` or on a user-managed `KongPlugin` referenced via `ExtensionRef`.

## Steps to reproduce

1. Deploy Kong Operator in Hybrid Gateway mode (Konnect), Kong Gateway 3.15.0.2.
2. Create a `KongPlugin` with `metadata.annotations["konghq.com/tags"]` set, e.g.:
   ```yaml
   kind: KongPlugin
   metadata:
     name: my-plugin
     annotations:
       konghq.com/tags: "my-tag"
   plugin: some-plugin
   config: {}
   ```
3. Reference it from an `HTTPRoute` via an `ExtensionRef` filter (or use any built-in filter that generates a plugin, e.g. `RequestHeaderModifier`).
4. After reconciliation, inspect the operator-generated `KongPlugin` copy and the Konnect Plugin entity it produces.

**Expected:** the Konnect Plugin entity carries `tags: ["my-tag"]`.
**Actual:** the generated `KongPlugin` copy (and the resulting Konnect Plugin entity) has no `konghq.com/tags` value — only internal Gateway API tracking annotations are present.

## Environment

- Kong Gateway: 3.15.0.2
- Kong Operator: 2.3.0-rc.3
- Confirmed still present, unfixed, on `origin/main` as of this filing (no changes to `controller/hybridgateway/plugin/plugin.go`, `controller/hybridgateway/pluginbinding/pluginbinding.go`, or `controller/hybridgateway/metadata/annotation.go` since `v2.3.0-rc.3`).

## Related

- #4891 — original `konghq.com/tags` propagation report (Service/Route/Certificate scope)
- #4924 — fix for #4891; explicitly excludes `KongPlugin`/`KongPluginBinding` since neither CRD has a `spec.tags` field

## Suggested fix shape

Since `KongPlugin`/`KongPluginBindingSpec` have no `spec.tags` field, this can't be fixed with a `WithSpecTags`-style change like #4924. Two options:
1. Add a `spec.tags` field to `KongPlugin`/`KongPluginBindingSpec` and wire it into the Hybrid Gateway builders the same way #4924 did for the other types, or
2. Fix `metadata.BuildAnnotations` (or add a variant) to merge in the `konghq.com/tags` annotation from the source `HTTPRoute` and/or the source `KongPlugin` (for the `ExtensionRef` case) when generating the copy, rather than only setting internal tracking annotations.

Happy to provide a minimal repro repo if useful.

bowale01:
## Summary

Fixes https://github.com/Kong/kong-operator/issues/5277

When the hybrid gateway controller creates operator-managed KongPlugin copies (for both built-in
filter plugins and ExtensionRef-referenced user plugins), it only set internal tracking annotations
via `metadata.BuildAnnotations()`. The `konghq.com/tags` annotation from the source
HTTPRoute/GRPCRoute or user-managed KongPlugin was never forwarded to the generated copy.

This caused the Konnect ops layer to produce Plugin entities without the user-specified tags,
since `metadata.ExtractTags()` on the generated copy returned nothing.

## Changes

- Added `WithTagsAnnotation()` to `KongPluginBuilder` that merges `konghq.com/tags` from source
  objects (route and/or original plugin) onto the generated copy's annotations. Tags from multiple
  sources are deduplicated.
- Applied it in all plugin generation paths:
  - Built-in filter plugins (tags from HTTPRoute/GRPCRoute)
  - ExtensionRef plugins (tags from both route and original KongPlugin)
  - Request-termination plugins (tags from route)

## How it works

The Konnect ops layer already handles tag extraction correctly in `kongPluginBindingToSDKPluginInput`:

tags := GenerateTagsForObject(pluginBinding, metadata.ExtractTags(plugin)...)

It fetches the referenced KongPlugin and extracts tags from its annotations. The problem was that
the generated copy never had the annotation set. This fix ensures it does.

## Testing

- All existing unit tests pass for `controller/hybridgateway/builder/` and `controller/hybridgateway/plugin/`
- Follows the same pattern used for KongRoute/KongService/KongUpstream tag propagation (from #4554)

### DIFF
--- a/controller/hybridgateway/builder/kongplugin.go
+++ b/controller/hybridgateway/builder/kongplugin.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"maps"
+	"sort"
 	"strings"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -130,6 +131,7 @@ func (b *KongPluginBuilder) WithTagsAnnotation(sources ...pkgmetadata.ObjectWith
 	if b.plugin.Annotations == nil {
 		b.plugin.Annotations = make(map[string]string)
 	}
+	sort.Strings(deduped)
 	b.plugin.Annotations[pkgmetadata.AnnotationKeyTags] = strings.Join(deduped, ",")
 	return b
 }
--- a/controller/hybridgateway/builder/kongplugin.go
+++ b/controller/hybridgateway/builder/kongplugin.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"maps"
+	"strings"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
@@ -13,6 +14,7 @@ import (
 	"github.com/kong/kong-operator/v2/controller/hybridgateway/metadata"
 	gwtypes "github.com/kong/kong-operator/v2/internal/types"
 	"github.com/kong/kong-operator/v2/modules/manager/scheme"
+	pkgmetadata "github.com/kong/kong-operator/v2/pkg/metadata"
 )
 
 // KongPluginBuilder is a builder for configurationv1.KongPlugin resources.
@@ -99,6 +101,37 @@ func (b *KongPluginBuilder) MustBuild() configurationv1.KongPlugin {
 		panic(fmt.Errorf("failed to build KongPlugin: %w", err))
 	}
 	return plugin
+}
+
+// WithTagsAnnotation merges the konghq.com/tags annotation value from the given
+// sources into the KongPlugin being built. This ensures that when the generated
+// KongPlugin copy is later read by the Konnect ops layer (via metadata.ExtractTags),
+// the user-supplied tags are present. Multiple sources are merged and deduplicated.
+func (b *KongPluginBuilder) WithTagsAnnotation(sources ...pkgmetadata.ObjectWithAnnotations) *KongPluginBuilder {
+	var allTags []string
+	for _, src := range sources {
+		if src == nil {
+			continue
+		}
+		allTags = append(allTags, pkgmetadata.ExtractTags(src)...)
+	}
+	if len(allTags) == 0 {
+		return b
+	}
+	// Deduplicate while preserving order.
+	seen := make(map[string]struct{}, len(allTags))
+	deduped := make([]string, 0, len(allTags))
+	for _, t := range allTags {
+		if _, ok := seen[t]; !ok {
+			seen[t] = struct{}{}
+			deduped = append(deduped, t)
+		}
+	}
+	if b.plugin.Annotations == nil {
+		b.plugin.Annotations = make(map[string]string)
+	}
+	b.plugin.Annotations[pkgmetadata.AnnotationKeyTags] = strings.Join(deduped, ",")
+	return b
 }
 
 // WithPluginName sets the plugin name for the KongPlugin being built.
--- a/controller/hybridgateway/builder/kongplugin_test.go
+++ b/controller/hybridgateway/builder/kongplugin_test.go
@@ -204,3 +204,113 @@ func TestKongPluginBuilder_ChainedCalls(t *testing.T) {
 	assert.NotNil(t, plugin.Annotations)
 	assert.JSONEq(t, string(config), string(plugin.Config.Raw))
 }
+
+func TestKongPluginBuilder_WithTagsAnnotation(t *testing.T) {
+	t.Run("tags from single source", func(t *testing.T) {
+		route := &gwtypes.HTTPRoute{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "test-route",
+				Namespace: "default",
+				Annotations: map[string]string{
+					"konghq.com/tags": "team-payments,env-prod",
+				},
+			},
+		}
+
+		plugin, err := NewKongPlugin().
+			WithName("test-plugin").
+			WithTagsAnnotation(route).
+			Build()
+		require.NoError(t, err)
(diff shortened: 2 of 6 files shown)


randmonkey:
@bowale01 Thanks for contributing. Currently our security policies does not allow some of the required checks to run from forks, so we will submit another PR to cherry-pick your commits and merge it. Sorry for the inconvenience.

controller/hybridgateway/builder/kongplugin.go
```diff
@@ -101,6 +103,37 @@ func (b *KongPluginBuilder) MustBuild() configurationv1.KongPlugin {
...
+	for _, t := range allTags {
+		if _, ok := seen[t]; !ok {
+			seen[t] = struct{}{}
+			deduped = append(deduped, t)
+		}
+	}
+	if b.plugin.Annotations == nil {
+		b.plugin.Annotations = make(map[string]string)
+	}
+	b.plugin.Annotations[pkgmetadata.AnnotationKeyTags] = strings.Join(deduped, ",")
```
randmonkey, line 135: To keep the annotation stable when not changed, I think the `deduped` should be sorted here:
```suggestion
    sort.Strings(deduped)
	b.plugin.Annotations[pkgmetadata.AnnotationKeyTags] = strings.Join(deduped, ",")
```

randmonkey (commented):
Could we add some tests to ensure that the tags are correctly added?

bowale01:
> Could we add some tests to ensure that the tags are correctly added?

Thanks! Applied the sort fix and added tests. Pushing now.

controller/hybridgateway/builder/kongplugin.go
bowale01: > To keep the annotation stable when not changed, I think the `deduped` should be sorted here:

Applied the sort and added tests. Pushed in 9e1c6c8.

randmonkey (approved):

randmonkey:
@bowale01 Approved. Thanks for quick response. We are sorry that the required checks cannot run from forks, we have to submit another PR to cherry-pick your commits. You will be still be honored as the contributor. 
Thanks again for contribution!

bowale01:
> @bowale01 Approved. Thanks for quick response. We are sorry that the required checks cannot run from forks, we have to submit another PR to cherry-pick your commits. You will be still be honored as the contributor. Thanks again for contribution!

You are welcome

randmonkey:
Since #5284 is merged, close this.

PR #5306, opened 2026-08-18
fix(konnect): Truncate tags and label values by unicode runes

### LINKED ISSUE #5305: The `truncate` to truncate tags is possible to break unicode characters and produce malformed tags
## Current Behavior

The `truncate` function in to truncate tags truncates the tags by bytes. This is possilbe to cut at the point in the middle of a unicode character, and produce a malformed tag in the created Konnect entity. This causes the Konnect entity fails to apply with the following error:

```json
{"level":"error","ts":"2026-08-18T07:23:38Z","logger":"KongRoute","msg":"operation in Konnect API failed","controller":"KongRoute","controllerGroup":"configuration.konghq.com","controllerKind":"KongRoute","KongRoute":{"name":"http.default-httproute-echo-2.cpe85a9357.f6700184.m00","namespace":"default"},"namespace":"default","name":"http.default-httproute-echo-2.cpe85a9357.f6700184.m00","reconcileID":"6eaeb5d2-66d1-4396-bd3a-468533a86128","konnect_id":"977c64c5-ec90-4bac-88de-33af9dd508cb","op":"update","namespacedName":"default/http.default-httproute-echo-2.cpe85a9357.f6700184.m00","duration":"258.668812ms","error":"failed to update KongRoute default/http.default-httproute-echo-2.cpe85a9357.f6700184.m00: API error occurred: Status 400\n{\"code\":3,\"details\":[{\"@type\":\"type.googleapis.com/kong.admin.model.v1.ErrorDetail\",\"field\":\"tags[0]\",\"messages\":[\"must match pattern '^(?:[\\\\x21-\\\\x2B\\\\x2D\\\\x2E\\\\x30-\\\\x7E\\\\p{N}\\\\p{L}]+(?: *[\\\\x21-\\\\x2B\\\\x2D\\\\x2E\\\\x30-\\\\x7E\\\\p{N}\\\\p{L}])*)?$'\"],\"type\":\"ERROR_TYPE_FIELD\"}],\"message\":\"validation error: must match pattern '^(?:[\\\\x21-\\\\x2B\\\\x2D\\\\x2E\\\\x30-\\\\x7E\\\\p{N}\\\\p{L}]+(?: *[\\\\x21-\\\\x2B\\\\x2D\\\\x2E\\\\x30-\\\\x7E\\\\p{N}\\\\p{L}])*)?$'\"}"}
```

In k8s CELs, `MaxLength` for a `string` or equivalant typed field means the maximum unicode points in the string. Also, Konnect supports tags with no more than 128 unicode characters (actually it also supports > 128 chars but accurate upper limit not clear yet). So we should unify the behavior to support maximum of 128 unicode characters (code points).

## Expected Behavior

- tags with no more than 128 unicode characters are not truncated
- tags with more than 128 unicode characters has exactly the first 128 unicode chars in the tags in Konnect

## Steps To Reproduce

Create an `HTTPRoute` like the following manifest in hybrid gateway:

```yaml
kind: HTTPRoute
metadata:
  name: httproute-echo-2
  namespace: default
  annotations:
    konghq.com/strip-path: "true"
    # 44 chinese characters which encodes to 44*3 = 132 > 128 bytes, and the truncate cuts in the middle of a character
    konghq.com/tags: "测试中文一二三四五六七八九十一二三四五六七八九十一二三四五六七八九十一二三四五六七八九十" 
spec:
  parentRefs:
  - name: hybrid-kong
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /echo-2
    backendRefs:
      - name: echo
        kind: Service
        port: 80
```

and it will fail to apply in Konnect with the error mentioned above.

## Operator Version

current `main`

randmonkey:

Truncate tags and label values by unicode runes (code points) instead of by bytes to avoid cutting in the middle that causes tags containing non-ASCII characters fails to apply in Konnect.

**Which issue this PR fixes**

Fixes #5305

**PR Readiness Checklist**:

### DIFF
--- a/controller/konnect/ops/objects_metadata.go
+++ b/controller/konnect/ops/objects_metadata.go
@@ -5,6 +5,7 @@ import (
 	"maps"
 	"slices"
 	"sort"
+	"unicode/utf8"
 
 	"github.com/samber/lo"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -158,9 +159,11 @@ func WithKubernetesMetadataLabelsPtr(obj ObjectWithMetadata, userSetLabels map[s
 	return out
 }
 
+// truncate truncates a string to the specified limit in runes, since tags supports unicode runes.
+// If the string is shorter than or equal to the limit, it returns the original string.
 func truncate(s string, limit int) string {
-	if len(s) <= limit {
+	if utf8.RuneCountInString(s) <= limit {
 		return s
 	}
-	return s[:limit]
+	return string([]rune(s)[:limit])
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -120,6 +120,9 @@
 - Fix Gateway API routes becoming transiently unaccepted and removed from dataplane
   when listener is intermittently marked as Programmed=False.
   [#5237](https://github.com/Kong/kong-operator/pull/5237)
+- Konnect entities: Fix truncating of tags to cut at 128 unicode runes
+  (UTF8 code points).
+  [#5306](https://github.com/Kong/kong-operator/pull/5306)
 
 ## [v2.3.0-rc.3]
 


controller/konnect/ops/objects_metadata.go
```diff
@@ -158,9 +159,11 @@ func WithKubernetesMetadataLabelsPtr(obj ObjectWithMetadata, userSetLabels map[s
 	return out
 }
 
+// truncate truncates a string to the specified limit in runes, since tags supports unicode runes.
+// If the string is shorter than or equal to the limit, it returns the original string.
 func truncate(s string, limit int) string {
-	if len(s) <= limit {
+	if utf8.RuneCountInString(s) <= limit {
```
WeichengWang1, line 165: `truncate` is a shared helper, do all caller of truncate() currently recognize limit as len of runes?

randmonkey: Yes. It is used for truncate label values and tags, and both of them recognize limit as len of runes. (label values only support a subset of ASCII so effectively the same)

WeichengWang1 (approved):

PR #5318, opened 2026-08-19
Add support for Kong Secret storage in Konnect Config Store

justindavies:
What this PR does / why we need it:

Adds support for keeping a KongCertificate's private key out of Konnect entirely by routing it through a Konnect Vault (Config Store), automatically, without requiring the key's source Secret to know anything about vaults.

Today, a KongCertificate backed by secretRef sends whatever is in the Secret's tls.key straight to Konnect as a literal value. Konnect's permissions are coarse (org/control-plane level, not per-field), so anyone with certificate read access there can read the private key in plaintext. Customers managing certificates with cert-manager can't work around this today, since cert-manager always writes real key material into its Secret and has no concept of Kong's vault:// reference syntax.

This PR closes that gap with two complementary changes:

1. Gateway API vault-ref support: a Gateway Listener's tls.certificateRefs can now point at a Secret whose tls.key/tls.crt holds a literal {vault://prefix/key} reference instead of PEM data (controller/pkg/secrets/cert.go's new IsTLSSecretValidOrVaultRef, wired into both controller/gateway/controller_reconciler_utils.go and controller/hybridgateway/converter/gateway.go). Previously this only worked for a hand-authored KongCertificate+KongSNI; Gateway API-managed listeners rejected it as an invalid TLS secret. This is for keys that are already vault-resident by some other means (a different backend, or a Config Store secret written out-of-band).
2. Automatic push-to-vault (controller/konnect/ops/vaultsecret.go): a Secret referenced by a KongCertificate's secretRef can now be marked with a new annotation, konghq.com/vault-secret: <KongVault name>. When present, the operator pushes the Secret's tls.key value into that KongVault's backing Konnect Config Store (creating or updating a stable, deterministically-named entry — see vaultSecretKeyName) and sends Konnect only a {vault://...} reference, never the literal key. This is the piece that makes cert-manager compatible: cert-manager keeps writing real PEM key material into its Secret exactly as it always does, completely unaware that anything downstream is happening. Rotation is transparent — the same Config Store entry gets updated in place, and the KongCertificate's Konnect object never needs to change.

  - The target Config Store entry and its owning control plane/store IDs are recorded in a new status.vaultSecret field on KongCertificate (api/configuration/v1alpha1/kongcertificate_types.go), so a later reconcile knows what to update and deletion knows what to remove — even if the source Secret is already gone by then.
  - deleteCertificate now also deletes the Config Store secret it owns, closing the "orphaned secret on delete" gap.
  - The key name is length-prefix-encoded (k8s-<len(ns)>-<ns>-<len(name)>-<name>-tls-key) specifically to avoid two different Secrets colliding on the same Config Store entry (e.g. ns=foo-bar,name=baz vs ns=foo,name=bar-baz would otherwise both produce foo-bar-baz).
  - Scope: only the primary secretRef is routed this way; secretRefAlt (and cert/cert_alt generally) are unaffected for now.

Live-verified end-to-end against two real Konnect environments, including: cert-manager-issued certs with real rotation (secret delete → cert-manager reissues → Config Store entry updates in place → Konnect cert stays on the same vault reference throughout, confirmed via TLS handshake modulus matching), the naming-collision fix, and delete cleanup.

Which issue this PR fixes

Fixes #

Special notes for your reviewer:

- New CRD field: KongCertificate.status.vaultSecret (optional, additive-only — make verify.crd-breaking-changes should be clean, though I wasn't able to run it locally due to an unrelated mise tooling limitation and would appreciate someone confirming in CI). Rollout ordering matters: the updated CRD needs to be applied before/alongside this operator version, or the new status field is silently pruned by the API server (no crash, just the field never persists, and deletion cleanup then has nothing to work from).
- Rollback caveat worth calling out in release notes: once any Secret is marked with konghq.com/vault-secret, rolling the operator back to a version predating this PR is not safe without first un-marking those Secrets — the old code has no knowledge of the annotation and will happily write the literal key to Konnect on its next reconcile of that certificate. This isn't a new class of bug, just a real interaction worth documenting since it's easy to hit via a bad rollback or a stale replica.
- ConfigStoreSecretsSDK is added to the hand-maintained SDKWrapper interface (controller/konnect/ops/sdk/sdkfactory.go), not the generated one — Config Store secrets aren't a CRD-backed entity (per design, there's deliberately no KonnectConfigStoreSecret CRD; secret values are explicitly kept out of the declarative surface), so there's nothing for the codegen pipeline to key off. The mock (test/mocks/sdkmocks/sdkfactory_mock.go) was updated by hand to match.
- Known, deliberate limitations (documented, not silently missing):
  - No local change-detection hash — the operator always pushes/updates on reconcile rather than skipping when the value is unchanged. Simpler and correct, just not optimal on write volume.
  - No protection against a coincidental pre-existing Config Store entry at the same computed name — Config Store secrets carry no ownership metadata, so an existing entry is adopted (treated as ours) rather than rejected. The alternative (refuse to touch it) would turn an ordinary "controller restarted between create and status-persist" case into a stuck reconcile, which seemed like the worse trade-off.

PR Readiness Checklist:

Complete these before marking the PR as ready to review:

- [ ] the CHANGELOG.md release notes have been updated to reflect significant changes


tao12345666333:
Thank you, Justin. I will close this PR and replace it with https://github.com/Kong/kong-operator/issues/5355

PR #5366, opened 2026-08-25
feat: enable KONG_USE_STANDARD_GRPC_STATUS_CODES and GRPCRoute conformance tests

### LINKED ISSUE #2345: conformance: `GRPCRouteHeaderMatching`, `GRPCExactMethodMatching`, `GRPCRouteListenerHostnameMatching`

WeichengWang1:

set KONG_USE_STANDARD_GRPC_STATUS_CODES="on" so that Kong will response with standard grpc status code, and then all GRPCRoute conformance tests can pass.

**Which issue this PR fixes**

Fixes #2345

**PR Readiness Checklist**:


CHANGELOG.md
```diff
@@ -114,6 +114,8 @@
 - CRD-from-OAS: enable resources to reference other resources from different
   namespaces by setting `supportCrossNamespaceReference` in `ReferenceConfig`.
   [#5263](https://github.com/Kong/kong-operator/pull/5263)
+- Enable KONG_USE_STANDARD_GRPC_STATUS_CODES and GRPCRoute conformance tests
```
randmonkey:
```suggestion
- Enable KONG_USE_STANDARD_GRPC_STATUS_CODES and GRPCRoute conformance tests.
```
Should we mention that this changed the default behavior of dataplane and only supports Kong 3.15(?) and later?

test/conformance/skipped_tests_test.go
```diff
@@ -6,12 +6,7 @@ import (
 	"github.com/kong/kong-operator/v2/test"
 )
 
-var skippedTestsShared = []string{
```
randmonkey, line 9: 👍
Should we remove the `skippedTestsShared` or add a comment?

randmonkey (commented):
LGTM, but we can explain the change clearer.

CHANGELOG.md
WeichengWang1: good point, it indeed changed the default behavior of dataplane, but I don't think it's necessary since GRPCRoute is supported from this version 2.3, so there is no behavior changes, and instead, it's a new feature and we announce that it's gateway api conformant.

test/conformance/skipped_tests_test.go
WeichengWang1: I prefer not removing it, just in case we need add something again in the future, and keep it align the format of `skippedTestsForHybrid` & `skippedTestsForStandard`, WDYT?
Will add a comment in next commits

CHANGELOG.md
randmonkey: But I still think the changelog entry is not very clear. How about

```suggestion
- Enable KONG_USE_STANDARD_GRPC_STATUS_CODES on dataplanes by default and run all core level
 GRPCRoute conformance tests.
```
?

WeichengWang1: sounds good, will modify it in the next commit.

randmonkey (approved):

PR #5425, opened 2026-08-27
fix(eventgateway): reflect full Deployment rollout in KegDataPlane Ready condition

### LINKED ISSUE #5340: Fix Ready condition to reflect full Deployment rollout status for AIGatewayDataPlane and KegDataPlane
## Problem Statement

Both `AIGatewayDataPlane` and `KegDataPlane` can report `Ready=True` too early (as soon as any pod is ready), even while a rollout of a spec change is still in progress, stuck, or effectively only serving pods from a previous version. This gives users false confidence that a recent change has taken effect when it may not have.

## Proposed Solution

`Ready=True` should only be reported once the current version of the resource's configuration has been fully and successfully rolled out to the workload, not merely once some pods happen to be ready. Until that point, `Ready` should remain `False` (or otherwise clearly indicate the rollout is still in progress), and it should continue to also depend on all other prerequisites (certificate readiness, control-plane connectivity, and service readiness) being satisfied, as it does today.

## Additional Information

This applies the same way to both `AIGatewayDataPlane` and `KegDataPlane`: both compute readiness the same, overly permissive way today, so the fix should behave consistently across the two.

## Acceptance Criteria

- [ ] ~~Documentation is updated with this new feature / enhancement~~: corrects the behavior of an existing condition, no new user-facing field or API surface.
- [ ] ~~Helm chart is updated to work with this feature / enhancement~~: no new resources or permissions involved.
- [ ] As a user, I can trust that `Ready=True` on an `AIGatewayDataPlane` or `KegDataPlane` means the latest configuration change has actually fully rolled out, not that it's merely in progress or stuck.

alacuku:

Ready no longer flips True until the current generation is fully rolled out. Adds a shared DeploymentRolloutComplete helper and a kegdataplane-lifecycle chainsaw test.

**Which issue this PR fixes**

Fixes #5340

**PR Readiness Checklist**:


alacuku:
@muthur review this pr

randmonkey (approved):

PR #5470, opened 2026-09-01
feat(aigateway): make AIGatewayDataPlane.spec.controlPlaneRef optional

### LINKED ISSUE #5458: Make AIGatewayDataPlane.spec.controlPlaneRef optional to support manually-managed control planes
## Problem Statement

`AIGatewayDataPlane.spec.controlPlaneRef` is a required field. It only supports one variant, `konnectNamespacedRef`, which forces every AI Gateway data plane to be paired with a `KonnectAIGateway` that Kong Operator itself creates, registers with Konnect, and manages end to end, including auto-generating and registering the mTLS client certificate used to connect to it.

This makes it impossible to deploy an `AIGatewayDataPlane` standalone against a control plane or certificate that already exist by other means (created manually, provisioned by another tool, or managed outside Kong Operator's lifecycle). Users in this situation have no supported path: they cannot omit `controlPlaneRef`, and there is no alternative reference type for a manually-managed control plane.

There's already a partial escape hatch: `spec.deployment.podTemplateSpec` lets a user overlay raw env vars onto the Deployment, and those take precedence over the operator's strongly-typed fields. But it can't be used at all today because `controlPlaneRef` being required blocks object creation before that escape hatch is ever reached.

## Proposed Solution

Make `spec.controlPlaneRef` optional. When it's unset, the operator performs no `KonnectAIGateway` lookup and no certificate registration automation for that `AIGatewayDataPlane`, leaving the user free to wire up Konnect (or any other) connectivity manually through `spec.deployment.podTemplateSpec`, the same escape hatch that already exists today.

## Additional Information

The Deployment env vars driven by `controlPlaneRef` today (`KONG_CLUSTER_CONTROL_PLANE`, `KONG_CLUSTER_SERVER_NAME`, `KONG_CLUSTER_TELEMETRY_ENDPOINT`, `KONG_CLUSTER_TELEMETRY_SERVER_NAME`) and the certificate registration flow (self-signed cert, then an `AIGatewayDataPlaneCertificate`, then Konnect) are internal implementation details, not documented as a manual-configuration contract. Nothing else about the referenced `KonnectAIGateway` object is consumed by the controller: not its Konnect API auth, not `Spec.APISpec`, not its Konnect entity ID.

## Acceptance Criteria
- [x] As a user, I can deploy an `AIGatewayDataPlane` without a `controlPlaneRef` and manually configure its Konnect connectivity via `spec.deployment.podTemplateSpec` env vars and volumes

alacuku:

Allows deploying an AIGatewayDataPlane against a manually-managed control plane and certificate, configured via spec.deployment.podTemplateSpec, without Kong Operator owning a KonnectAIGateway.

**Which issue this PR fixes**

Fixes #5458 

**PR Readiness Checklist**:


api/aigateway/v1alpha1/aigatewaydataplane_types.go
```diff
@@ -35,6 +37,7 @@ import (
 // +kubebuilder:resource:shortName=aigwdp,categories=kong
 // +kubebuilder:printcolumn:name="Ready",description="The Resource is ready",type=string,JSONPath=`.status.conditions[?(@.type=='Ready')].status`
 // +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
+// +kubebuilder:validation:XValidation:rule="(!has(oldSelf.spec.controlPlaneRef)) ? true : (!has(self.status) || !self.status.conditions.exists(c, c.type == 'Ready' && c.status == 'True')) ? true : has(self.spec.controlPlaneRef)", message="spec.controlPlaneRef cannot be removed once the AIGatewayDataPlane is Ready"
```
pmalek: `Ready` condition flaps so I don't think it's the best candidate for this.

For example, when user changes the image, `Ready` is set to false and `controlPlaneRef` can be changed then.

api/aigateway/v1alpha1/aigatewaydataplane_types.go
```diff
@@ -43,7 +46,7 @@ type AIGatewayDataPlane struct {
 	// Spec defines the desired state of AIGatewayDataPlane.
 	//
 	// +required
-	Spec AIGatewayDataPlaneSpec `json:"spec,omitzero"`
+	Spec AIGatewayDataPlaneSpec `json:"spec,omitempty"`
```
pmalek, line 49: Why this change?

pmalek (commented):
Are we going to change the `ControlPlaneRef` type as discussed offline to contain the `kind`, e.g. 

```
spec:
  controlPlaneRef:
    type: namespacedRef
    namespacedRef:
      kind: OnpremAIGateway | KonnectAIGateway
      name: ai-gw-cp
```

?

alacuku:
> Are we going to change the `ControlPlaneRef` type as discussed offline to contain the `kind`, e.g.
> 
> ```
> spec:
>   controlPlaneRef:
>     type: namespacedRef
>     namespacedRef:
>       kind: OnpremAIGateway | KonnectAIGateway
>       name: ai-gw-cp
> ```
> 
> ?

It's out of scope for this PR.

api/aigateway/v1alpha1/aigatewaydataplane_types.go
alacuku: What do you suggest?

pmalek: We can make this field immutable when set.

api/aigateway/v1alpha1/aigatewaydataplane_types.go
alacuku: Making `controlPlaneRef` optional makes `AIGatewayDataPlaneSpec{}` (all fields zero) a valid config for the first time, since `ControlPlaneRef` nil is its zero value.
`omitzero` on `Spec` then drops `spec` from the JSON entirely for that now-valid case, so `omitempty` is needed to keep it on the wire.

pmalek (approved):

PR #5474, opened 2026-09-01
ci(muthur): pin ksai scripts to action tag

bartsmykla:

`/muthur` has failed on every run since Aug 31 with `TypeError: ownsVerb is not a function`.

Kong/ksai is internal and this repo is public, so the reviewer action is checked out and run by local path. A local-path action gets no `github.action_ref`, and that is exactly what the action falls back on when `ksai_ref` is unset. It therefore resolved an empty ref, which `actions/checkout` reads as Kong/ksai's default branch. So since #5361 dropped `ksai_ref`, the action has been pinned at `v1.17.5` while the trusted scripts it loads came from `main`. A function rename on `main` on Aug 31 broke the pair 45 minutes later.

Naming the ref explicitly puts the scripts and the action back on one tag. That is what #5361 intended - it just needs the reviewer's own tag rather than the kreview plugin tag that caused the original `MODULE_NOT_FOUND`.

`issue_comment` workflows always run from the default branch, so this cannot be exercised from the PR branch. It takes effect on merge.

Separately, I am adding a guard on the Kong/ksai side so an empty ref fails loudly instead of quietly reading `main`.

**PR Readiness Checklist**:

### DIFF
--- a/.github/workflows/muthur.yaml
+++ b/.github/workflows/muthur.yaml
@@ -9,9 +9,9 @@ name: Muthur
 #
 # kong-operator is public and Kong/ksai is internal, so ksai actions are checked out locally
 # via sparse-checkout and referenced by path rather than by a remote `uses:`, which GitHub
-# won't resolve from a public repo into a private one. codeowners-authz, ai-reviewer-federated
-# and the kreview plugin each version independently in Kong/ksai, so each is pinned to its own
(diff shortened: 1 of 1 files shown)


pmalek (approved):

pmalek:
Overriding CI checks to unblock muthur reviewer.

Issue #4046, opened 2026-04-28
Gateway API conformance: Implement ListenerSets GEP-1713


pmalek:
### Problem statement

`ListenerSet` API has been graduated to v1 in Gateway API 1.5.

This issue tracks the implementation of that CRD in Kong Operator.

Gateway API issue: https://github.com/kubernetes-sigs/gateway-api/issues/1713
GEP: https://gateway-api.sigs.k8s.io/geps/gep-1713/
Spec: https://github.com/kubernetes-sigs/gateway-api/blob/v1.5.0/apis/v1/listenerset_types.go

### Anything else

This issue tracks the self hosted implementation of `ListenerSet`. Hybrid gateway support will be tracked in a dedicated issue.

### Acceptance criteria

- [ ] Kong Operator users can use `ListenerSet` CRD for self hosted environments
- [ ] `ListenerSet` conformance tests are enabled

PR #5489, opened 2026-09-01
fix(ingress-controller): skip parentRefs of unsupported kinds instead of erroring

SLoeuillet:
Same fix as Kong/kubernetes-ingress-controller#8079, applied here because this is where new development lands per KIC's README. Reported as Kong/kubernetes-ingress-controller#8078.

### What this changes

`parentRefsForRoute` rejects a whole route as soon as one `parentRef` names a kind other than a core Gateway API `Gateway`:

```go
if group != gatewayv1.GroupName || kind != "Gateway" {
    return nil, fmt.Errorf("unsupported parent kind %s/%s", group, kind)
}
```

That runs *before* `getSupportedGatewayForRoute` compares the parent's `GatewayClass.spec.controllerName` against `GetControllerName()`, so a route owned by a different Gateway API implementation never reaches the "not mine, skip it" path — even though that path exists and is used for every other not-mine case (missing Gateway, missing GatewayClass, foreign `controllerName`, all `continue`).

Because such a route is never going to change, the reconcile error repeats on controller-runtime's backoff forever. We hit this on a cluster where Envoy Gateway owns an `HTTPRoute` parented to a `ListenerSet`, in a namespace the Kong controller also watches:

```
error controllers.HTTPRoute Reconciler error {"error": "unsupported parent kind gateway.networking.k8s.io/ListenerSet"}
```

This filters unsupported parentRefs out instead. A route left with no supported parentRef resolves to `ErrNoSupportedGateway`, which every route controller already handles — it clears any stale status and drops the object from the dataplane cache, returning no error. That is exactly the right outcome for a route owned by another controller, and it needs no new error handling.

It also fixes a smaller thing: the old code `return`ed on the first bad ref, so a route combining an unsupported parent with a legitimate `Gateway` parent was discarded whole. Filtering keeps the `Gateway` one.

The filter allocates a new slice rather than removing in place (e.g. `slices.DeleteFunc`) on purpose: `refs` aliases the route's own `Spec.ParentRefs`, and the route comes from the shared informer cache, so in-place removal would mutate a cached object.

### Relationship to #4046

This is **not** ListenerSet support and does not implement any part of #4046. The controller does not need to understand `ListenerSet` here — it only needs to not error out on routes that reference one. The fix is generic to any parent kind the controller cannot resolve.

It is worth noting that `ListenerSet` reaches many more clusters than it used to: it graduated to the **standard** channel as `gateway.networking.k8s.io/v1` in Gateway API v1.5.0 and is still standard in v1.6.1, so users no longer have to opt into the experimental bundle to have the CRD present — and therefore to have routes that reference it.

### Testing

- The existing `invalid parentRef kind rejected` case now asserts `ErrNoSupportedGateway` rather than the raw error, and is renamed to match the new semantics.
- New `TestParentRefsForRoute` covers: a `Gateway` ref kept; omitted Group/Kind defaulted and kept; a `ListenerSet` ref dropped; an unrelated group dropped; and an unsupported ref not discarding a supported one alongside it.

`go test ./ingress-controller/internal/controllers/gateway/...` passes.

### Note

I left `CHANGELOG.md` alone — there is no `Unreleased` section to append to and it looks maintainer-curated at release time. Happy to add an entry wherever you'd like it.

### DIFF
--- a/ingress-controller/internal/controllers/gateway/route_utils.go
+++ b/ingress-controller/internal/controllers/gateway/route_utils.go
@@ -62,6 +62,14 @@ func (g supportedGatewayWithCondition) GetSectionName() mo.Option[string] {
 
 // parentRefsForRoute provides a list of the parentRefs given a Gateway APIs route object
 // (e.g. HTTPRoute, TCPRoute, e.t.c.) which refer to the Gateway resource(s) which manage it.
+//
+// parentRefs pointing at a kind this controller cannot resolve (e.g. a ListenerSet, or any
+// Gateway API kind introduced after the version this controller is built against) are filtered
+// out rather than rejected: such a reference means the route is managed by some other
+// controller. Routes left with no supported parentRef are handled by getSupportedGatewayForRoute
+// returning ErrNoSupportedGateway, which callers already treat as "not ours". Returning an error
+// here instead makes reconciliation of a route this controller does not own fail permanently,
+// retrying forever without ever converging.
 func parentRefsForRoute[T gatewayapi.RouteT](route T) ([]gatewayapi.ParentReference, error) {
 	var refs []gatewayapi.ParentReference
 	switch r := (any)(route).(type) {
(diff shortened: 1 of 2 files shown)


pmalek:
Hi @SLoeuillet,

Thanks for the PR. It looks good. Due to limitations in our CI I've cherry picked your commit to https://github.com/Kong/kong-operator/pull/5554 ( so that the full test suite can run ).

Let us know if you have any comments on this. I'm going to close this one here, let's keep the discussion in 5554.

Issue #5342, opened 2026-08-21
Support referencing an existing TLS Secret for mTLS certificates on AIGatewayDataPlane and KegDataPlane (manual certificate injection)


alacuku:
## Problem Statement

`AIGatewayDataPlane` and `KegDataPlane` currently only support **automatic** mTLS certificate provisioning: the operator generates the certificate and registers it with Konnect on the user's behalf. There is no way for a user to supply their own pre-existing certificate (e.g. issued by cert-manager or an external PKI) for the operator to use instead. This blocks use cases where certificate issuance and rotation must stay under the user's or an external system's control, for example a dedicated internal CA, compliance requirements, or existing PKI tooling that the organization already standardizes on.

## Proposed Solution

Allow a user to reference an existing Kubernetes TLS Secret as an alternative to automatic provisioning. When such a reference is supplied:

- The operator must not generate, register, rotate, or delete a certificate on the user's behalf, and must not modify the referenced Secret in any way: it remains entirely user-owned.
- The operator validates that the referenced Secret exists and contains a usable certificate and key. If it is missing or invalid, the resource is reported as not ready, with a clear reason explaining why.
- The certificate is put into use by the running workload just as an automatically-provisioned one would be.
- Any change to the Secret's contents, or to the reference itself, results in the workload picking up the new certificate (i.e. triggers a rollout) without manual intervention.

## Additional Information

This applies the same way to both `AIGatewayDataPlane` and `KegDataPlane`: both follow the same automatic-provisioning-only model today, so the manual option should behave consistently across the two.

## Acceptance Criteria

- [ ] Documentation is updated with this new feature / enhancement
- [ ] Tests are added for this feature / enhancement
- [ ] Helm chart is updated to work with this feature / enhancement
- [ ] As a user, I can reference an existing TLS Secret on an `AIGatewayDataPlane` or `KegDataPlane` instead of relying on operator-generated certificates, and have changes to that Secret picked up automatically, without the operator ever creating, modifying, or deleting the Secret or any certificate resource on my behalf.

PR #5548, opened 2026-09-04
feat(aigateway): support manual mTLS certificate Secret for AIGatewayDataPlane

alacuku:

Adds spec.certificateSecret.{provisioning,secretRef} to AIGatewayDataPlane. Users can now supply their own same-namespace TLS Secret for the mTLS client certificate instead of relying on the operator to auto-provision and sign one. provisioning is Manual or Automatic (default Automatic, matching prior behavior); secretRef names the Secret and is required when Manual, forbidden when Automatic. This is enforced via CEL validation at the API server, so an invalid combination is rejected on write.

What it brings for users
- Bring your own certificate support (e.g. issued by an external PKI or cert-manager) instead of being locked into the operator's auto-issued cert.
- Ability to switch between Manual and Automatic provisioning on an existing resource.
- Stale auto-provisioned cert Secrets get cleaned up when no longer needed.

Breaking / behavior changes
1. Bug fix with a side effect: the operator no longer provisions or mounts an mTLS client certificate for an AIGatewayDataPlane with no spec.controlPlaneRef. Previously it always did, signed by the operator's shared cluster CA, even when nothing needed it. Any resource that relied on that (e.g. hand wiring the cert into spec.deployment.podTemplateSpec to reach a same cluster ControlPlane) will lose the auto issued cert on upgrade and needs to supply its own. Note that spec.certificateSecret is not a workaround here: without controlPlaneRef it just sets CertificateProvisioned=False/ControlPlaneRefMissing and the operator stops reconciling the Deployment, HPA, and ingress Service entirely.
2. One time rolling restart: the Pod template now carries a certificate checksum annotation, so every existing AIGatewayDataPlane Deployment with a spec.controlPlaneRef will roll once on upgrade.

**Which issue this PR fixes**

Fixes #

Part of #5342 

**PR Readiness Checklist**:


controller/aigateway/dataplane/owned_secret.go
```diff
@@ -28,11 +28,96 @@ import (
...
+		if cs != nil {
+			apimeta.SetStatusCondition(&aigwdp.Status.Conditions, metav1.Condition{
+				Type:               string(aigatewayv1alpha1.CertificateProvisionedType),
+				Status:             metav1.ConditionFalse,
+				Reason:             string(aigatewayv1alpha1.CertificateControlPlaneRefMissingReason),
+				Message:            aigatewayv1alpha1.CertificateControlPlaneRefMissingMessage,
+				ObservedGeneration: aigwdp.Generation,
+			})
+		}
+		return op.Noop, nil, nil
```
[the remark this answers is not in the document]
alacuku: It was a bug, the operator was not supposed to provision and mount the mtls secret when the `controlPlaneRef` was not set.

alacuku:
/ksai review this PR and check that everything has been addressed from the previous review.

tao12345666333 (dismissed):

controller/aigateway/dataplane/owned_secret.go
```diff
@@ -20,19 +20,116 @@ import (
...
+	if aigatewaycp == nil {
+		if cs != nil {
+			apimeta.SetStatusCondition(&aigwdp.Status.Conditions, metav1.Condition{
+				Type:               string(aigatewayv1alpha1.CertificateProvisionedType),
+				Status:             metav1.ConditionFalse,
+				Reason:             string(aigatewayv1alpha1.CertificateControlPlaneRefMissingReason),
+				Message:            aigatewayv1alpha1.CertificateControlPlaneRefMissingMessage,
+				ObservedGeneration: aigwdp.Generation,
+			})
+		}
```
WeichengWang1, line 73: CertificateProvisioned=False/ControlPlaneRefMissing won't be cleared.

alacuku: @WeichengWang1 PTAL: 757c3ebdd0865eb05585829a5b416f9060b3d328

WeichengWang1 (approved):

PR #5597, opened 2026-09-08
chore(ci): update the KSAI workflows for ksai/v3.40.7

bartsmykla:
## Motivation

The KSAI workflows this repository calls have changed shape since v3.40.5. Moving the pin alone is not enough: an input a caller passes that the workflow no longer declares, or a permission it does not grant, is refused by GitHub before any job starts and reports nothing a reader can act on.

This repository holds its four federation identifiers in secrets. GitHub allows `secrets` on a step's `with:` and refuses it on a reusable-workflow call's, so the automated migration stops on them rather than writing a caller that fails at startup. They are not secret — the shipped consumer template carries them in plain text and `federations/team-k8s.yaml` in `Kong/ksai` is where they come from — so this pull request writes the values, and everything below it is what the migration produced from there.

This carries the change and the pins together, which is why it is one pull request rather than a version bump.

## Implementation

- The job is replaced by calls to the published workflows rather than running the action directly, so the runner, the timeout and the concurrency lane are the called workflow's from now on. Any `concurrency:` block this file had is removed for that reason, and the workflow is renamed to `KSAI`
- `authorization_github_token`, `kong_org_github_token`, `source_org_github_token` are not carried over. The published workflow computes those values itself and passes them to the action, so setting them on the caller is no longer possible or needed
- `ksai_ref` is not carried over: it names which Kong/ksai commit a local-path caller checks out, and a caller of the published workflow is pinned by its own uses: line, so the adopted shape has nothing to point it at
- There is no `ksai-actions` preset this repository can extend, so none was added to your Renovate config. Until there is one, Renovate cannot tell one `Kong/ksai` release target from another, and this will not be the last pull request of its kind
- Everything else this repository had set is kept: its runner, its inputs, its secrets and its comments. Only the lines above change

> [!WARNING]
> **The `if:` on `gate` and `review` is replaced by the published workflow's own.**
>
> If yours narrowed what starts a run, this widens it, and every pull request the new one admits spends your Anthropic budget. Yours was:
>
> ```yaml
> if: >- github.event.issue.pull_request && contains(github.event.comment.body, '/ksai')
> if: ${{ needs.gate.outputs.authorized == 'true' }}
> ```
>
> `require_label`, `skip_label`, `review_drafts` and `review_bots` are inputs on that call and say the same thing without an expression to maintain.

`/ksai` works exactly as it does today.

### DIFF
--- a/.github/workflows/ksai.yaml
+++ b/.github/workflows/ksai.yaml
@@ -1,149 +1,136 @@
 name: KSAI
 
-# Federated code review driven by the shared kreview skills in Kong/ksai. Triggered by a
-# `/ksai <ask>` comment on a pull request.
-#
-# Only protected-branch contexts can authenticate (the federation rule requires
-# ref_protected). An issue_comment workflow always executes from the default branch, so
-# this file has to be on master before `/ksai` does anything.
-#
-# kong-operator is public and Kong/ksai is internal, so ksai actions are checked out locally
-# via sparse-checkout and referenced by path rather than by a remote `uses:`, which GitHub
-# won't resolve from a public repo into a private one. codeowners-authz and ai-reviewer-federated
-# version independently in Kong/ksai, so each is pinned to its own release tag, bumped by
-# Renovate. The kreview plugin and the trusted scripts are read from the reviewer's own tag.
+permissions: {}
 
 on:
   issue_comment:
     types: [created]
-
-env:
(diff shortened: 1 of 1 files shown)


bartsmykla:
Superseded by #5598, opened from a branch in this repository now that I have write access — a fork PR cannot mint the federated token, since a fork run is handed no secrets.

PR #5598, opened 2026-09-08
chore(ci): update the KSAI workflows for ksai/v3.40.9

bartsmykla:
## Motivation

The KSAI workflows this repository calls have changed shape since v3.40.5. Moving the pin alone is not enough: an input a caller passes that the workflow no longer declares, or a permission it does not grant, is refused by GitHub before any job starts and reports nothing a reader can act on.

`Kong/ksai` is internal and this repository is public, so the calls resolve to `Kong/ksai-public`, the generated snapshot published for exactly this case. That also fixes the model selection that failed here on `v3.40.5`: the local-path caller fell back to the old action, which does not support the OSS models in `allowed_models`.

## Implementation

- The two hand-rolled jobs are replaced by calls to the published workflows, pinned at `Kong/ksai-public@v3.40.9` by commit digest. The runner, the timeout and the concurrency lane are the called workflow's from now on, and `allowed_models` is carried across unchanged.
- The four federation identifiers move out of secrets into plain values. GitHub allows the `secrets` context on a step's `with:` and refuses it on a reusable-workflow call's, so they could not be carried as they were. They are not secret — the shipped consumer template carries them in plain text and `federations/team-k8s.yaml` in `Kong/ksai` is where they come from.
- `push_as_app: false`, so every write goes through `GITHUB_TOKEN` and posts as `github-actions[bot]`. That is what the previous caller did: it granted the write scopes to `GITHUB_TOKEN` on the job and minted `KONG_READONLY` only for the CODEOWNERS check. **The App needs no new permission.**
- `client_id` passes `GH_APP__KONG_READONLY__APP_ID` and the private key comes from the matching secret, so the run keeps minting with this repository's own App rather than the shared KSAI one.
- The automatic review request job is **not** included. It asks GitHub for a *team* reviewer, which needs `pull-requests: write` on the App and cannot be done with `GITHUB_TOKEN` at all. This workflow answered `/ksai` comments only before, so nothing is lost; add it once the App has that permission if automatic reviews are wanted. `clear_request` goes with it.
- `authorization_github_token`, `kong_org_github_token`, `source_org_github_token` and `ksai_ref` are not carried over — the published workflow computes each of them itself.
- Validated against the published files at `v3.40.9`: every input the two jobs pass is declared, the required secret is supplied, and no job is short a permission the called workflow needs.

> [!NOTE]
> `/ksai` works as it does today. The `ksai` job additionally answers a review-thread comment and a review submitted on a draft this flow opened, where the old file answered issue comments only.

### DIFF
--- a/.github/workflows/ksai.yaml
+++ b/.github/workflows/ksai.yaml
@@ -1,149 +1,136 @@
 name: KSAI
 
-# Federated code review driven by the shared kreview skills in Kong/ksai. Triggered by a
-# `/ksai <ask>` comment on a pull request.
-#
-# Only protected-branch contexts can authenticate (the federation rule requires
-# ref_protected). An issue_comment workflow always executes from the default branch, so
-# this file has to be on master before `/ksai` does anything.
-#
-# kong-operator is public and Kong/ksai is internal, so ksai actions are checked out locally
-# via sparse-checkout and referenced by path rather than by a remote `uses:`, which GitHub
-# won't resolve from a public repo into a private one. codeowners-authz and ai-reviewer-federated
-# version independently in Kong/ksai, so each is pinned to its own release tag, bumped by
-# Renovate. The kreview plugin and the trusted scripts are read from the reviewer's own tag.
+permissions: {}
 
 on:
   issue_comment:
     types: [created]
-
-env:
-  # renovate: datasource=github-tags depName=Kong/ksai extractVersion=^ksai/v(?<version>.+)$
-  KSAI_REF: ksai/v3.40.5
+  pull_request_review_comment:
+    types: [created]
+  pull_request:
+    types: [review_requested, opened, synchronize, reopened, ready_for_review, edited]
+  pull_request_review:
+    types: [submitted, dismissed]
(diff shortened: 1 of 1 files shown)


pmalek:
@bartsmykla 

```
  Error: Kong/ksai-public/.github/workflows/ksai.yml@v3.40.7 is not pinned to a full length commit SHA.
  Error: Kong/ksai-public/.github/workflows/ksai-request.yml@v3.40.7 is not pinned to a full length commit SHA.
  Error: Kong/ksai-public/.github/workflows/ksai-hold.yml@v3.40.7 is not pinned to a full length commit SHA.
```

bartsmykla:
@pmalek on it

pmalek:
@bartsmykla I guess we're missing permissions on the app:

```
Error: The permissions requested are not granted to this installation. - https://docs.github.com/rest/reference/apps#create-an-installation-access-token-for-an-app
```

bartsmykla:
@pmalek checking

pmalek (approved):

PR #5600, opened 2026-09-08
fix(hybridgateway): resolve KongPlugin configFrom and configPatches for ExtensionRef filters

sandromodarelli:

A `KongPlugin` attached to an `HTTPRoute` or `GRPCRoute` through an `ExtensionRef` filter was mirrored to the control plane from `spec.config` alone. `spec.configFrom` and `spec.configPatches` were never read, so a plugin sourcing its configuration from a `Secret` was pushed without it, silently — the route still reported `Programmed: True`.

Both fields are now resolved before the mirrored `KongPlugin` is built, the referenced `Secret`s are watched so a change triggers a reconcile, and a failure to resolve them surfaces as an error instead of an empty config.

**Which issue this PR fixes**

Fixes FTI-7908

- The resolution logic is ported from the KIC path (`ingress-controller/internal/dataplane/kongstate`), which `controller/` cannot import because it lives under `internal/`.
- e2e coverage lands in the existing `httproute-filters` chainsaw suite: `configFrom`, `configPatches` and a `Secret` rotation, each asserting both the mirrored `KongPlugin` config and the header served by the data plane. The suite passes against Konnect with this change and fails without it.
- `assert-kongPluginConfig.yaml` is a new step template because the existing `assert-kongPlugin.yaml` matches every `KongPlugin` in the namespace, and a source plugin using `configFrom` carries no `spec.config` for it to evaluate.

**PR Readiness Checklist**:


controller/hybridgateway/watch/mapfuncs_secret.go
```diff
@@ -109,3 +116,80 @@ func routesForClientCertSecret(ctx context.Context, cl client.Client, secretName
...
+			}
+		}
+		return lo.Uniq(requests)
+	}
+}
+
+// kongPluginsForConfigSecret returns the "namespace/name" keys of the KongPlugins in the given
+// namespace that source their configuration from the given Secret, in the format expected by the
+// KongPluginsOn*Route field indexes.
+func kongPluginsForConfigSecret(ctx context.Context, cl client.Client, secretNamespace, secretName string) []string {
```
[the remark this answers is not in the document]
sandromodarelli: I wouldn't create a new index to handle this. Secrets are cache-filtered by `konghq.com/secret=true`, so these events are infrequent and the per-namespace KongPlugin count is small

sandromodarelli:
/ksai review the pr and check that everything has been addressed form the previous round

sandromodarelli:
/ksai review the pr and check that everything has been addressed form the previous round

controller/hybridgateway/plugin/config.go
```diff
@@ -0,0 +1,188 @@
...
+// empty raw configuration is patched as if it were an empty JSON object.
+func applyConfigPatches(
+	ctx context.Context,
+	cl client.Client,
+	namespace string,
+	rawConfig json.RawMessage,
+	patches []configurationv1.ConfigPatch,
+) (json.RawMessage, error) {
+	raw := rawConfig
+	if len(raw) == 0 {
```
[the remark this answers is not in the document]
sandromodarelli: i won't apply this change. `apiextensionsv1.JSON.UnmarshalJSON` discards the null literal, so `Raw` comes back empty and the existing len(raw) == 0 guard already converts it to {}

controller/hybridgateway/watch/mapfuncs_secret.go
```diff
@@ -109,3 +118,90 @@ func routesForClientCertSecret(ctx context.Context, cl client.Client, secretName
...
+		}
+		return lo.Uniq(requests)
+	}
+}
+
+// MapGRPCRouteForPluginConfigSecret returns a handler.MapFunc that, given a Secret, lists all
+// KongPlugins in the same namespace sourcing their configuration from it via spec.configFrom or
+// spec.configPatches, then returns reconcile.Requests for all GRPCRoutes that reference those
+// KongPlugins in an ExtensionRef filter.
+func MapGRPCRouteForPluginConfigSecret(cl client.Client) handler.MapFunc {
```
tao12345666333: nit: It is very similar to the `MapHTTPRouteForPluginConfigSecret` above. Should we consider extracting it into a helper function?

sandromodarelli: Nice catch. helper implemented

controller/hybridgateway/plugin/config.go
```diff
@@ -0,0 +1,174 @@
...
+
+	return patched, nil
+}
+
+// secretValue returns the bytes held by the given key of the given Secret.
+func secretValue(ctx context.Context, cl client.Client, namespace, name, key string) ([]byte, error) {
+	secret := &corev1.Secret{}
+	if err := cl.Get(ctx, types.NamespacedName{Namespace: namespace, Name: name}, secret); err != nil {
+		return nil, fmt.Errorf("failed to fetch plugin configuration secret %s/%s: %w", namespace, name, err)
+	}
```
WeichengWang1, line 175: KO cached Secrets with label `konghq.com/secret: "true"`, is it a well-known pre-condition? If not, how about adding a hint in the error msg?

WeichengWang1 (approved):

PR #3024, opened 2026-01-08
feat: `Gateway`'s underlying resources static naming

mlavacca:
**What this PR does / why we need it**:

**Which issue this PR fixes**

Fixes #3015 

**Special notes for your reviewer**:

**PR Readiness Checklist**:

Complete these before marking the PR as `ready to review`:

- [ ] the `CHANGELOG.md` release notes have been updated to reflect significant changes

PR #5659, opened 2026-09-11
fix(gateway): namespace-qualify the Konnect Control Plane name under static naming

### LINKED ISSUE #4079: `static-naming` annotation ignores Gateway namespace when generating Konnect control plane names
When using the `gateway-operator.konghq.com/static-naming: "true"` annotation on a Gateway resource, the resulting Konnect control plane name only reflects the Gateway's name, rather than a qualified name including the namespace. This causes issues when the Operator and Gateway are in different namespaces or when trying to maintain a `namespace-gatewayname` convention.

A preliminary investigation suggests the reconciler logic only references `gateway.Name` here:
[https://github.com/Kong/kong-operator/blob/8d11b2d66a7ab206752b9e6e3caf62842b9bd057/controller/gateway/controller_reconciler_utils.go#L1651](https://github.com/Kong/kong-operator/blob/8d11b2d66a7ab206752b9e6e3caf62842b9bd057/controller/gateway/controller_reconciler_utils.go#L1651)

## Current Behavior
The Konnect control plane name results in `gateway.name`. 
* **Example:** A Gateway named `development-external` in namespace `development-gateways` results in a control plane named `development-external`.

## Expected Behavior
The control plane name should follow a qualified naming convention, specifically `{gateway.namespace}-{gateway.name}`.
* **Example:** A Gateway named `development-external` in namespace `development-gateways` should result in `development-gateways-development-external`.

## Steps To Reproduce
1. Deploy the Kong Gateway Operator.
2. Create a Gateway in a non-default namespace (e.g., `development-gateways`) with the following annotation:
   ```yaml
   metadata:
     annotations:
       gateway-operator.konghq.com/static-naming: "true"
     name: development-external
     namespace: development-gateways
   ```
3. Check the resulting Control Plane name in the Konnect UI/API.
4. Observe that the namespace prefix is missing.

## Operator Version
2.1.2

## `kubectl version`
*(N/A)*

---

> **Note for Maintainers:** As pointed out in the discussion, fixing this will be a **breaking change** for users currently relying on the single-name mapping. Coordination or a transition strategy may be required.

matteogastaldello:

> **TL;DR** — Under static naming, the Konnect Control Plane name is namespace-qualified again (`<namespace>_<gateway-name>`), so two same-named Gateways in different namespaces stop colliding in Konnect. The dynamic path is untouched.

**The bug.** A `KonnectGatewayControlPlane` has two names: `metadata.name`, unique per namespace, and `spec.createControlPlaneRequest.name`, which must be unique across the whole Konnect org. #3357 replaced the old deterministic `<namespace>-<gateway-name>` Konnect name with the Kubernetes name plus a client-side random suffix, and that suffix is the anti-collision mechanism. Static naming (#3024) has no suffix, by design, so on that path the namespace qualification was removed with nothing in its place.

**The symptom.** Two Gateways named `edge-gw` in `staging` and `production`, both statically named, are legal in Kubernetes but collide in Konnect. The second create returns HTTP 409 and that Gateway is stuck at `Programmed=False`.

**The fix and what it costs.** Static naming qualifies the Konnect name with the namespace again, joined with an underscore; the dynamic path keeps #3357's behaviour, locked by `gateway/konnect-dynamic-naming`. An underscore is valid in neither a namespace nor a Gateway name, so `<namespace>_<gateway-name>` splits unambiguously, where `<namespace>-<gateway-name>` does not: both sides may contain a hyphen. The cost is that the two names no longer match on the static path (`edge-gw` in Kubernetes, `staging_edge_gw` in Konnect). Nothing reads the Konnect name to find an entity — lookups go through the `k8s-uid` label or the Konnect ID — and the namespace is still exposed in Konnect as the `k8s-namespace` label. A bare predictable Kubernetes name, matching Kubernetes and Konnect names, and org-wide uniqueness can't all hold at once under static naming. This drops the second.

**No renames on upgrade.** `createKonnectGatewayControlPlane` runs only when the Gateway owns no Control Plane yet, and `enforceKonnectGatewayControlPlaneSpec` never touches the name, so existing Control Planes are left alone. To move an existing one onto the new scheme, edit `spec.createControlPlaneRequest.name`: Konnect renames in place and keeps the ID.

**Which issue this PR fixes**

Fixes #4079

> **TL;DR** — The fix lives at the call site, not in `setObjectNamespaceName`. Static naming had no test coverage anywhere in the repo before this PR; it now has unit, envtest, and Konnect-backed e2e coverage.

**Why not `setObjectNamespaceName`.** It sets `metadata.name` *and* returns the string that feeds the Konnect name, so changing it changes both — and `metadata.name` isn't where the bug is. A KGCP named `default_hybrid` breaks every manifest that references it by name, including the #3024 sample. The helper is also shared with `DataPlane` and `ControlPlane`, whose names never reach Konnect and play no part in the `(org_id, name)` collision. It stays functionally as is; only the duplicated annotation check moved out, into `hasStaticNaming()`.

**Tests.** `TestEnforceKonnectGatewayControlPlaneSpec` now seeds the pre-fix unqualified Konnect name in every case and asserts it survives. That name is write-once at creation and nothing enforced it, so without this test a future change there would silently rename every existing Konnect Control Plane on upgrade. An envtest covers the same on the reconciler path. `gateway/konnect-static-naming-cross-namespace` was confirmed to fail without the fix, on the qualified-name assertion for the first Gateway; its `Programmed=True` and non-null Konnect ID checks also verify that Konnect accepts both Control Planes, which envtest can't do without a Konnect backend. It's the first Konnect-backed test in the `gateway/` suite and needs no workflow change.

**Manual verification.** Against a real Konnect org, the two Gateways produce `staging_edge_gw` and `production_edge_gw` with distinct Konnect IDs, both `Programmed=True`.

**Out of scope.** The same namespace and name on two clusters sharing one Konnect org still conflict; this is noted in the release note. #3357 closes that axis with randomness, and static naming exists to make names deterministic, so closing it on the static path needs an explicit cluster identifier rather than a suffix.

**Docs.** [Kong/developer.konghq.com#7321](https://github.com/Kong/developer.konghq.com/pull/7321) updates the static naming how-to for the new name. It stays in draft until this ships, because the operator docs are not versioned: merging it earlier would put a page in front of readers that contradicts the release they are running.

**PR Readiness Checklist**:

### DIFF
--- a/test/e2e/chainsaw/gateway/konnect-dynamic-naming/chainsaw-test.yaml
+++ b/test/e2e/chainsaw/gateway/konnect-dynamic-naming/chainsaw-test.yaml
@@ -1,0 +1,170 @@
+apiVersion: chainsaw.kyverno.io/v1alpha1
+kind: Test
+metadata:
+  name: gateway-konnect-dynamic-naming
+spec:
+  description: |
+    Non-regression lock for the default (dynamic) naming path of Gateway-owned
+    KonnectGatewayControlPlanes.
+
+    Without the static-naming annotation the operator generates
+    "<gateway-name>-<random suffix>" for the Kubernetes resource and uses that same string
+    as the Konnect Control Plane name, which is the invariant introduced in #3357. The fix
+    for #4079 qualifies the Konnect name with the namespace only under static naming; this
+    test makes sure the dynamic path is not changed by accident.
+  bindings:
+    - name: test_name
+      value: ($namespace)
+    - name: konnect_token
+      value: (env('KONNECT_TOKEN'))
+    - name: konnect_url
+      value: (env('KONNECT_SERVER_URL') || 'eu.api.konghq.tech')
+    - name: gateway_image
+      value: (env('GATEWAY_IMAGE'))
+    # A short literal rather than the usual "<test_name>-gateway": the assertions below
+    # compare the generated name against the namespace, and a name that already embeds the
+    # namespace would make those comparisons meaningless.
+    - name: gateway_name
+      value: edge-gw
+    - name: listener_http_port
+      value: 80
+    - name: auth_secret_name
+      value: (join('-', [($test_name), 'konnect-auth-secret']))
+    - name: auth_secret_namespace
+      value: ($namespace)
+    - name: konnect_auth_name
+      value: (join('-', [($test_name), 'konnect-api-auth']))
+    - name: konnect_auth_namespace
+      value: ($namespace)
+
+  steps:
+    - name: 01-create-auth-secret
+      description: Create the Secret holding the Konnect token.
+      use:
+        template: ../../common/_step_templates/apply-assert-konnectAuthSecret.yaml
+
+    - name: 02-create-konnect-api-auth
+      description: Create the KonnectAPIAuthConfiguration.
+      use:
+        template: ../../common/_step_templates/apply-assert-konnectAPIAuthConfiguration-secretref.yaml
+
+    - name: 03-create-gateway-configuration
+      description: |
+        Create the Konnect-backed GatewayConfiguration and GatewayClass. These are local
+        files rather than the shared templates because the shared
+        apply-assert-gatewayConfiguration.yaml always emits an explicit authRef.namespace,
+        which would require a KongReferenceGrant this test does not need.
+      try:
+        - apply:
+            file: 01-gatewayconfiguration.yaml
+        - assert:
+            file: 02-assert-gatewayclass.yaml
+
+    - name: 04-create-gateway
+      description: Create a Gateway without the static-naming annotation.
+      try:
+        - apply:
+            file: 03-gateway.yaml
+
+    - name: 05-assert-konnect-control-plane-programmed
+      description: The Gateway-owned KonnectGatewayControlPlane reaches Programmed in Konnect.
+      try:
+        - assert:
+            file: 04-assert-kgcp.yaml
+
+    - name: 06-assert-dynamic-naming-invariants
+      description: |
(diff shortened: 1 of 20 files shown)


test/e2e/chainsaw/gateway/konnect-dynamic-naming/chainsaw-test.yaml
```diff
@@ -0,0 +1,123 @@
+apiVersion: chainsaw.kyverno.io/v1alpha1
+kind: Test
+metadata:
+  name: gateway-konnect-dynamic-naming
+spec:
+  description: |
+    Non-regression lock for the default (dynamic) naming path. Without the static-naming
+    annotation the operator generates "<gateway-name>-<random suffix>" and uses that same
+    string as the Konnect name (#3357). The #4079 fix qualifies the Konnect name only under
```
WeichengWang1, line 9: How about replacing these #NNNN with links? 
Ditto in other files.

pmalek (commented):
> Out of scope. The same namespace and name on two clusters sharing one Konnect org still conflict; this is noted in the release note

This is fine. That's the case for all other resources.

---


LGTM, let's just address @WeichengWang1 's comment and this can be merged.

test/e2e/chainsaw/gateway/konnect-dynamic-naming/chainsaw-test.yaml
matteogastaldello: For sure!

WeichengWang1 (approved):

PR #5723, opened 2026-09-16
feat(konnect): block KonnectConfigStore deletion while store holds entries

### LINKED ISSUE #5708: KonnectConfigStore: block deletion while store holds entries (remove Force:true)
Sub-issue of #5355.

## Problem Statement

Today, deleting a `KonnectConfigStore` CR force-deletes the Konnect Config Store **and all secret entries it holds**: the generated delete op passes `Force: true` (introduced in #5193, configured in `crd-from-oas/config.yaml`).

One accidental store deletion is an instant TLS outage for every SNI referencing those entries — measured against a live data plane: a missing entry means the affected SNI serves **no certificate at all** (no fallback), until the entry is restored. There is no warning and no chance to clean up references first.

## Proposed Solution

Remove `Force: true` from the `KonnectConfigStore` delete op in `crd-from-oas/config.yaml` and regenerate. Konnect rejects deletion of a non-empty store (400 "can not delete config store with secrets"), so the CR will stay terminating instead of silently cascading.

Surface this state clearly:
- A status condition / event on the `KonnectConfigStore` CR reporting that deletion is blocked by existing entries, so the operator knows exactly what to clean up.
- User-facing docs updated: the current behavior note in `crd-from-oas/config.yaml` ("deleting a KonnectConfigStore deletes every secret in that store") must be revised.

This is a deliberate behavior change to an already-shipped feature, reviewed and approved in the KonnectConfigStoreSync design review (Q2 decision, 2026-09-16). Context: the upcoming `KonnectConfigStoreSync` controller (#5355) writes store entries declaratively, which makes the silent cascade significantly more dangerous — entries it manages must never be destroyed as a side effect of deleting the store CR.

## Additional Information

- Design doc (ERD): KonnectConfigStoreSync — Declarative Secret → Konnect Config Store Synchronization, §5.5 / L11 / R11, decision Q2
- `Force: true` origin: #5193
- Measured gateway behavior for missing entries: no fallback certificate, recovery ≤ 15s after re-creation

## Acceptance Criteria

- [ ] Deleting a `KonnectConfigStore` CR whose store still holds entries does NOT delete the entries; the CR reports a blocked-deletion condition/event naming the reason
- [ ] Deleting a `KonnectConfigStore` CR whose store is empty succeeds as before
- [ ] Tests are added/updated for both paths (blocked and empty)
- [ ] Documentation is updated to describe the new deletion behavior
- [ ] CHANGELOG entry calls out the behavior change
- [ ] ~~Helm chart is updated~~ (no chart changes needed — CRD/regeneration only)

tao12345666333:

Deleting a `KonnectConfigStore` CR previously force-deleted the Konnect config store **and every secret entry it held** (behavior shipped in #5193 via `Force: true` on the delete op) — one accidental CR deletion instantly removed all stored certificates/keys and broke every SNI referencing them.

This PR changes the delete path to report-and-block:

- Removes `Force: true` from the KonnectConfigStore delete operation in `crd-from-oas/config.yaml` and regenerates `zz_generated_ops_konnect_configstore.go`.
- Adds a guarded delete (`controller/konnect/ops/ops_konnectconfigstore_manual.go`) that wraps the generated delete op: when Konnect rejects the deletion with its typed 400 (store still holds entries), it lists the blocking secret entry keys via the ConfigStoreSecrets SDK and returns a typed `KonnectConfigStoreNotEmptyError` carrying those keys.
- The reconciler keeps the CR's cleanup finalizer, sets the `Programmed` condition to `False` with the new `DeletionBlocked` reason and a capped, sorted list of blocking entry keys in the message, and retries with backoff. Once the entries are removed from the store in Konnect, deletion proceeds automatically.
- Deleting a store that is already empty works exactly as before.

**Which issue this PR fixes**

Fixes #5708

- The new hand-written file is named `ops_konnectconfigstore_manual.go` following the house `_manual.go` helper convention: the crd-from-oas generator skips *all* ops generation for an entity when a hand-written `ops_<entity>.go` file exists, so the plain per-entity name would delete the generated ops.
- `ClearInstanceFromError` in `ops.go` now preserves the `KonnectConfigStoreNotEmptyError` wrapper (it previously unwrapped to the bare innermost SDK error, which would drop the blocking keys); the `Instance` field is still cleared in place either way.
- The `DeletionBlocked` reason constant is added in `api/konnect/v1alpha1/konnect_conditions.go` next to the existing `Programmed` reasons.
- This changes already-shipped behavior (#5193) and is called out under "Breaking changes" in the CHANGELOG.

**PR Readiness Checklist**:


tao12345666333:
Note for reviewers: the `controller/mcpserver/mcpserverfetcher.go` pagination cursor fix that was briefly part of this PR was removed in c629bcb87 and now lives in its own standalone PR: #5727.

controller/konnect/reconciler_generic.go
```diff
@@ -609,6 +610,30 @@ func (r *KonnectEntityReconciler[T, TEnt]) Reconcile(ctx context.Context, ent TE
...
+						errNotEmpty.DeletionBlockedMessage(),
+					); errStatus != nil || !res.IsZero() {
+						return res, errStatus
+					}
+					// Do not return the error: the blockage is resolved out of
+					// band in Konnect, which produces no watch event, so the
+					// default error backoff would degrade to ~16min between
+					// retries while logging an error and recording a failed
+					// operation metric on every attempt.
+					return ctrl.Result{RequeueAfter: ctrlconsts.RequeueWithBackoff}, nil
```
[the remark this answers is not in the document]
tao12345666333: addressed https://github.com/Kong/kong-operator/pull/5723/commits/87ef1b1247f14f8de2eb0555d69fafd5876cf1c8

controller/konnect/ops/ops_konnectconfigstore_manual.go
```diff
@@ -0,0 +1,230 @@
...
+func listConfigStoreSecretKeys(
+	ctx context.Context,
+	configStoreSecretsSDK sdkkonnectgo.ConfigStoreSecretsSDK,
+	obj *konnectv1alpha1.KonnectConfigStore,
+) ([]string, error) {
+	var (
+		keys      []string
+		pageAfter *string
+	)
+	for range configStoreSecretsListMaxPages {
```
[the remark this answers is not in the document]
tao12345666333: addressed https://github.com/Kong/kong-operator/pull/5723/commits/87ef1b1247f14f8de2eb0555d69fafd5876cf1c8

controller/konnect/ops/ops_konnectconfigstore_manual.go
```diff
@@ -0,0 +1,257 @@
...
+
+	konnectv1alpha1 "github.com/kong/kong-operator/v2/api/konnect/v1alpha1"
+)
+
+// configStoreNotEmptyMessage is the (case-insensitive) substring of the error
+// detail that Konnect returns when rejecting the deletion of a config store
+// that still holds secret entries. The Konnect API exposes no machine-readable
+// error code for this case, so detection relies on the typed 400 error plus
+// this message, centralized here.
+const configStoreNotEmptyMessage = "can not delete config store with secrets"
```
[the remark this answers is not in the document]
tao12345666333: Addressed https://github.com/Kong/kong-operator/pull/5723/commits/84a21bae1a393c5276b92ecb5721daf4196b1ff0

controller/konnect/ops/ops_konnectconfigstore_manual.go
```diff
@@ -0,0 +1,257 @@
...
+func listConfigStoreSecretKeys(
+	ctx context.Context,
+	configStoreSecretsSDK sdkkonnectgo.ConfigStoreSecretsSDK,
+	obj *konnectv1alpha1.KonnectConfigStore,
+) ([]string, bool, error) {
+	var (
+		keys      []string
+		pageAfter *string
+		truncated bool
+	)
```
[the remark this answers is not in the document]
tao12345666333: addressed https://github.com/Kong/kong-operator/pull/5723/commits/84a21bae1a393c5276b92ecb5721daf4196b1ff0

controller/konnect/ops/ops_konnectconfigstore_manual.go
```diff
@@ -0,0 +1,275 @@
...
+// expects only the item cursor, so the URI must be parsed.
+func pageAfterCursorFromNextPageURL(next *string) (string, error) {
+	if next == nil {
+		return "", nil
+	}
+	u, err := url.Parse(*next)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse next page URI %q: %w", *next, err)
+	}
+	return u.Query().Get("page[after]"), nil
```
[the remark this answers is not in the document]
tao12345666333: fixed https://github.com/Kong/kong-operator/pull/5723/commits/55fd624ba6a0b68f25624311abbdbb00ee869077

controller/konnect/ops/ops_konnectconfigstore_manual.go
```diff
@@ -0,0 +1,283 @@
...
+					"id", obj.GetKonnectStatus().GetKonnectID(),
+					"error", err.Error(),
+				)
+			truncated = true
+			break
+		}
+		if next == "" {
+			break
+		}
+		if pageAfter != nil && next == *pageAfter {
```
randmonkey: Is it useful to list so many configStore secrets in Konnect to just provide the number of referenced secret in the user facing message? I think only telling that there are still remaining secrets is enough.

controller/konnect/ops/ops_konnectconfigstore_manual.go
```diff
@@ -0,0 +1,283 @@
...
+
+// ErrorIsConfigStoreNotEmpty returns true if the provided error is a Konnect
+// 400 rejecting the deletion of a config store that still holds secret entries.
+func ErrorIsConfigStoreNotEmpty(err error) bool {
+	if errBadRequest, ok := errors.AsType[*sdkkonnecterrs.BadRequestError](err); ok {
+		return strings.Contains(
+			strings.ToLower(errBadRequest.Detail),
+			configStoreNotEmptyMessage,
+		)
+	}
```
WeichengWang1, line 54: Is this string matching reliable? can we add integration tests towards it?

controller/konnect/ops/ops_konnectconfigstore_manual.go
```diff
@@ -0,0 +1,283 @@
...
+	if errSDK, ok := errors.AsType[*sdkkonnecterrs.SDKError](err); ok {
+		return errSDK.StatusCode == http.StatusBadRequest &&
+			strings.Contains(strings.ToLower(errSDK.Body), configStoreNotEmptyMessage)
+	}
```
WeichengWang1: Ditto

controller/konnect/ops/ops_konnectconfigstore_manual.go
randmonkey: +1 for this. This looks like the error message returned from Konnect so the value is not guaranteed constant. (Or maybe we should ask Koko team to confirm how to ditinguish the config store not empty error?)

controller/konnect/ops/ops_konnectconfigstore_manual.go
tao12345666333: make sense, let me change the logic

controller/konnect/ops/ops_konnectconfigstore_manual.go
tao12345666333: Good catch! Let me see how to modify it more appropriately.

controller/konnect/ops/ops_konnectconfigstore_manual.go
```diff
@@ -0,0 +1,283 @@
...
+}
+
+// maxKeysInDeletionBlockedMessage caps the number of entry keys included in
+// the user-facing DeletionBlocked condition message.
+const maxKeysInDeletionBlockedMessage = 10
+
+// DeletionBlockedMessage returns a user-facing message describing that the
+// config store deletion is blocked by the secret entries it still holds, so
+// the user knows exactly what to clean up before deletion can proceed.
+func (e KonnectConfigStoreNotEmptyError) DeletionBlockedMessage() string {
```
randmonkey, line 46: This looks too long and not very clear. How about adding some tool funtions to generate the `deletion blocked ...` like 
```go
func (e KonnectConfigStoreNotEmptyError) deletionBlockedMessageStillHoldsEntries()  string {
    if len(e.Keys) == 0 {
        return "..."
    }
   if e.KeysTruncated {
        return fmt.Sprintf("...",len(e.Keys))
    }
    return fmt.Sprintf("...",len(e.Keys))
}
```
and `remove the entries ....` by: 
```go
func (e KonnectConfigStoreNotEmptyError) deletionBlockedMessageRemoveEntriesInstruction() string {
    keysToReport := min(len(e.Keys), maxKeysInDeletionBlockedMessage)
   ...
}
 ```
Then construct the message here by if `e.ListErr` is not nil?

tao12345666333: Thank you, The logic in this function has been cleaned up.

randmonkey (approved):

PR #5727, opened 2026-09-16
fix(mcpserver): use page[after] cursor from next-page URI when listing MCP servers

tao12345666333:

`MCPServersFetcher.fetchAll` passed the full `meta.page.next` URI from Konnect list responses as the `PageAfter` request parameter, which expects only the `page[after]` item cursor it contains. This malformated every page 2+ request: the fetch loop either retried the rejected request forever (never returning any MCP server) or, if Konnect ignored the invalid cursor, re-fetched the first page indefinitely.

This PR extracts the item cursor from the next-page URI via a new `pageAfterCursorFromNextPageURL` helper, and makes pagination fail closed: an unfollowable page (non-200 response, unparseable next URI, missing cursor, or a repeated/cycling cursor) now fails the fetch, which is retried with a backoff, instead of returning a truncated list — such a list is treated as authoritative by `syncMCPServers` and would have deleted the in-cluster `MCPServer`s missing from it.

**Which issue this PR fixes**

N/A - pre-existing bug found during review of #5723

The same next-page-URI cursor-extraction helper pattern is used in the KonnectConfigStore ops package in #5723, where this fix was originally bundled before being extracted into this standalone PR.

**PR Readiness Checklist**:


controller/mcpserver/mcpserverfetcher_test.go
```diff
@@ -426,3 +431,251 @@ func TestSyncMCPServers(t *testing.T) {
...
+			}
+			return http.StatusOK, pageJSON(new(baseURL+"/v1/mcp-cp/"+cpID+"/mcp-servers"), "srv-1")
+		})
+
+		servers, err := f.fetchAll(t.Context())
+		require.ErrorContains(t, err, "carries no page[after] cursor")
+		assert.Nil(t, servers)
+		assert.Equal(t, []string{""}, cursors())
+	})
+}
```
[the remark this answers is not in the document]
tao12345666333: The comments here are reasonable, but I think the test case is not very reasonable, so I made some other modifications.

https://github.com/Kong/kong-operator/pull/5727/commits/4de8ac3249aedb81add0dd63de0ae78f8a547493

controller/mcpserver/mcpserverfetcher.go
```diff
@@ -121,7 +123,14 @@ func (f *MCPServersFetcher) run(ctx context.Context) {
...
 				if err != nil {
-					log.Error(logger, err, "failed to fetch MCP servers", "controlPlaneID", cpID)
+					if ctx.Err() != nil {
+						return
+					}
+					// Requeue: a failed fetch yields an incomplete list, and
+					// syncing one would delete in-cluster MCPServers whose
+					// Konnect counterparts simply were not listed.
+					log.Error(logger, err, "failed to fetch MCP servers, retrying", "controlPlaneID", cpID)
+					time.AfterFunc(b.Duration(), f.wake)
```
pmalek: Rather than sleeping here, we can add a select on 

```
				select {
				case <-time.NewTimer(b.Duration()).C:
					f.wake()
				case <-ctx.Done():
					return
				}
```

tao12345666333: Good catch! Addressed https://github.com/Kong/kong-operator/pull/5727/commits/8906ca3b8a92ecb829ac9d63a7bf8d77b87711cc

pmalek (approved):

PR #5799, opened 2026-09-21
ci: revert reusable workflow for github app token

pmalek:

GitHub's runner drops job outputs crossing a reusable-workflow (workflow_call) boundary when the value matches a masked secret — the app token is auto-masked, so outputs.token from __github_app_token.yaml arrived empty everywhere. Backport failed hard (Input required and not supplied: github_token); nightly hit the same warning but passed only because the empty gh-pat didn't break that path.

- `backport.yaml`, `nightly.yaml`, `__release-workflow.yaml` — reverted to inline github-app-token jobs minting the token within each workflow (pre-#5776 pattern, not subject to the guard). Backport now uses the app token instead of `PAT_GITHUB`, and regained its lost job-level permissions: contents: write.
- Deleted `__github_app_token.yaml`; no references remain.

Example failure: https://github.com/Kong/kong-operator/actions/runs/35599720503/job/106332759029

> Error: Error: Input required and not supplied: github_token

**PR Readiness Checklist**:


pmalek:
/ksai review this pr, check if backport's backport job still requires contents write permissions given that it's now using the github app token

programmer04 (approved):

PR #5804, opened 2026-09-21
fix(aigwconfiguration)!: AIGatewayMCPServer route matcher type change (breaking)

alacuku:

`AIGatewayMCPServer`'s route matcher (`spec.apiSpec.<variant>.config.route`, present in all 5 variants: `conversion-only`, `conversion-listener`, `listener`, `passthrough-listener`, `upstream-server`) was generated by `crd-from-oas` as `map[string]string` instead of a structured object. Root cause: the matcher's OpenAPI schema is composed via `allOf` (a base object schema plus a validation-only `anyOf` sibling requiring "at least one of hosts/paths/methods/headers"), and the generator had no handling for that shape. A schema or property with no properties of its own fell back to `map[string]string` (or `any`).

Because `map[string]string` only allows scalar string values, `paths`/`hosts`/`methods` could never be expressed as the arrays Kong's route matching actually requires. `kubectl apply` accepted it happily, since a bare string is a valid map value. The operator's own request-building code then failed to convert it into the shape the Konnect SDK expects, for every variant, every time:

```
failed creating AIGatewayMCPServer SDK request: failed to unmarshal into
AIGatewayMCPServerConversionOnly: could not unmarshal `{"paths":"/mcp/flights-conversion-only"}`
into any supported union types for AIGatewayMCPServerRouteWithMatcher
```

**Before, and after**

```yaml
# Before: route accepted only scalar strings, silently invalid for
# anything except a single path/host/method, and always failed against
# the real Konnect SDK regardless.
spec:
  apiSpec:
    conversion-only:
      config:
        route:
          paths: /mcp/flights-conversion-only
```

```yaml
# After: route.paths/hosts/methods are proper lists, headers is a
# free-form object (list of values per header name), matching the real
# Konnect schema.
spec:
  apiSpec:
    conversion-only:
      config:
        route:
          paths:
            - /mcp/flights-conversion-only
            - /mcp/flights-conversion-only/v2
          methods:
            - GET
            - POST
          hosts:
            - flights.mcp.example.com
          headers:
            x-mcp-client:
              - flights-e2e
```

**Fix, in two commits**:

1. `fix(crd-from-oas)`: teach the generator to merge an `allOf`'s member properties into the schema/property instead of falling back to `map[string]string`/`any`.
2. `fix(aiconfiguration)`: regenerate the API types, CRDs, Helm chart, docs, and chart golden snapshots against the existing, unchanged OpenAPI spec. No SDK version bump needed, the spec already had the correct schema. Adds a chainsaw e2e test (`test/e2e/chainsaw/aigateway/mcpserver-types`) covering all 5 variants. It spreads `paths` (multi-value), `hosts`, `methods`, and `headers` across them so every field the bug affected gets exercised between them, with `conversion-listener` combining all four at once, matching the exact shape from the bug report.

**Which issue this PR fixes**

Fixes #

- This is a breaking CRD schema change (`route`'s type changes from `map[string]string` to a structured object). It's unavoidable: the old shape could never actually express an array valued matcher, so nothing had a working config depending on it. The only thing anyone could have had is a config that was already permanently failing with the error above. Under the new schema, a bad manifest (e.g. `paths: /mcp/foo` instead of a list) is now rejected at `kubectl apply` time with a clear validation error, instead of failing silently deep inside the SDK.
- Verified end to end against a real Konnect account: applied all 5 variants via the new chainsaw test, confirmed `Programmed: True` with real Konnect IDs, and cross-checked the raw `GET /v1/ai-gateways/{id}/mcp-servers/{id}` response. `paths`, `hosts`, `methods`, and `headers` are all correctly stored server-side.

**PR Readiness Checklist**:


pmalek (commented):
Nice find! The change looks good to me but 2 comments:
- one is that we need to handle the chart 1MiB issue
- two: let's add manifest samples for these changes

alacuku:
Waiting for https://github.com/Kong/kong-operator/pull/5819 to be merged.

sandromodarelli (approved):

pmalek (approved):

PR #5826, opened 2026-09-22
ci: mint the release-bot app token in the jobs that use it

matteogastaldello:

Kong caps classic PATs at 7 days, so `secrets.PAT_GITHUB` (team-k8s-bot) is no longer a renewable credential. The release bot now mints a short-lived GitHub App token in each job that uses one, the shape #5814 introduced.

The echo-push guard changes with it. A ref pushed with an App token reports `github.actor` as `<app-slug>[bot]`, so `look_for_release` compares the actor against the App's own login, read from the mint it runs when the push created a ref.

**Which issue this PR fixes**

Fixes #

- Org-side changes on this repository, none of which the workflow can set or verify from here:
  - the App must be a bypass actor on the `release-branches` ruleset (id 3169974). That ruleset has a `creation` rule on `refs/heads/release/*`, and `create-release-branch` now creates that ref with the App token: without the bypass it fails with `Reference update failed`.
  - the App needs `contents: write` (create the branch, push the cherry-pick), `pull_requests: write` (open both PRs) and `workflows: write` (one of the two PRs edits `.github/workflows/charts-sync.yaml`).
- The mint itself is not in question here: `nightly.yaml` mints a token for this repository every night, and the latest run (2026-09-22) is green.
- This takes effect for releases from `main` as soon as it merges. Patch releases off `release/2.3.x` keep using `PAT_GITHUB` until this is backported, so the secret cannot be deleted yet.
- The first release after the merge is the canary; `create-release-branch` is where a missing ruleset bypass would surface.
- CI-only change, no CHANGELOG entry.

**PR Readiness Checklist**:


pmalek (approved):
LGTM. We need to make sure to backport this to 2.3, 2.2 and 2.1

